PR #7, opened 2020-08-20
Remove discprov fallbacks from libs init

raymondjacobson:


audius-infra:
Preview this change https://frontend-demo.audius.co/rj-remove-fallbacks

PR #9, opened 2020-08-24
Add LICENSE

raymondjacobson:


jowlee:
You could also add a license field to the package.json file

PR #16, opened 2020-09-06
Use music-metadata-browser to extract metadata from audio files

Borewit:
Would you be interested using [music-metadata-browser](https://github.com/Borewit/music-metadata-browser) to extract metadata from audio files?

It supports TypeScript & pomises out of the box.
Works with virtual any audio file type.

Fixed actual mime-type assigned to `Blob.type` representing artwork.


raymondjacobson:
Hey @Borewit thanks for opening this PR and sorry for the radio silence. Taking a look at both packages, it seems as if 
https://bundlephobia.com/result?p=music-metadata-browser@2.1.7
is quite a lot large than jsmediatags
https://bundlephobia.com/result?p=jsmediatags@3.9.3
which we've found to be suitable for our needs.

Is there any case study you can point to that demonstrates that this supports more types for us out of the box or similar?

Borewit:
> Is there any case study you can point to that demonstrates that this supports more types for us out of the box or similar?

This is the [list of file types support by music-metadata](https://github.com/Borewit/music-metadata#support-for-audio-file-type), versus MP3, MP4 & FLAC in jsmediatags. If you find something which is not supported, [let me know](https://github.com/Borewit/music-metadata/issues).

To me jsmediatags seems pretty much abandonded as well.

raymondjacobson:
Hey @Borewit, that's certainly cool how many types you've got and how well documented it is! I don't think there's anything short-term we can do here though because we don't explicitly need the extra types and it doesn't justify the 4-5x package size increase. If there's any way we can consume a subset of the package here to get size parity with jsmediatags, please lmk and re-open! Thanks!

Borewit:
Thanks for you reply @raymondjacobson. I the past I tried a kind of plug-gable audio type sub-modules in the past, did not reduce things drastically as the core is quite big and people had issues with linking the right things in. That was not worth it. Living with a slightly bigger package is indeed the price which one has to pay.

PR #23, opened 2020-09-25
Move client to API Trending

piazzatron:
### Trello Card Link
https://trello.com/c/h5UW0oXf

### Description
Use `v1/full/tracks/trending` for API.

This needs https://github.com/AudiusProject/audius-protocol/pulls to merge. 

**Current Known Issues**
- [ ] Trending keeps trying to load more tiles at 100 entries (possible lineup bug?)
- [X] FIXED Switching to a different time-range and then selecting a new genre disables the previous time range tab buttons

### Dragons

🚨 **Trending**

### How Has This Been Tested?

Ran local discprov pointed to prod discovery snapshot, passed that in as override endpoint for APIClient


audius-infra:
Preview this change https://frontend-demo.audius.co/piazz-api-trending

src/containers/track-page/TrackPageProvider.tsx
```diff
@@ -395,11 +395,10 @@ class TrackPageProvider extends Component<
     })
 
     const releaseDate = track ? track.release_date || track.created_at : ''
-    const mood = track ? track.mood : ''
     const description = getTrackPageDescription({
       releaseDate: releaseDate ? formatDate(releaseDate) : '',
-      description: track ? track.description : '',
-      mood,
+      description: track?.description ?? '',
+      mood: track?.description ?? '',
```
jowlee, line 401: mood?

src/containers/track-page/store/retrieveTrending.ts
```diff
@@ -0,0 +1,69 @@
...
+    offset,
+    limit,
+    currentUserId: encodedUserId,
+    timeRange
+  })
+
+  const currentGenre = yield select(getTrendingGenre)
+
+  // If we changed genres, do nothing
+  if (currentGenre !== genre) return []
```
jowlee, line 65: Should this matter if the generator code is in a takeLatest saga, then it would be canceled which is effectively the same thing? I assume this is for the case when you switch genres quickly before the first saga got to this point?

src/services/AudiusBackend.js
```diff
@@ -1041,25 +1055,6 @@ class AudiusBackend {
     }
   }
 
-  static async getTrendingTracks({ genre, offset, limit, timeRange }) {
```
jowlee, line 1044: @raymondjacobson we are getting rid of the optimistic load with this change, just wanted to flag. Should be fine if dp selection isn't slow.

src/services/audius-api-client/AudiusAPIClient.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  _constructUrl(
+    path: string,
+    queryParams: { [key: string]: string | number | undefined | null }
+  ) {
+    if (this.initializationState.state !== 'initialized')
+      throw new Error('_constructURL called uninitialized')
+    const params = Object.entries(queryParams)
+      .filter(p => p[1] !== undefined && p[1] !== null)
+      .map(p => `${p[0]}=${p[1]}`)
```
jowlee, line 120: might want to run `encodeURIComponent` here.

src/services/AudiusBackend.js
jowlee: nvm there is eager selection in the new api 😍  thx @piazzatron

jowlee (approved):
Nice work @piazzatron ! Looks fantastic.

src/containers/track-page/store/retrieveTrending.ts
raymondjacobson: good point, let's merge now and table this. doesn't hurt to have

PR #26, opened 2020-10-01
Add DP v1 api user follows/following

jowlee:
### Trello Card Link
https://www.notion.so/audiusproject/Porting-Client-to-API-9beefcc007ff4242be96124874606227
`/users/<id>/following` & `/users/<id>/followers`

### Description
Updates the get user following and get user follows endpoint to use the new api DP v1 api

### Dragons

This shouldn't be merged until DP is merged

### How Has This Been Tested?
I ran this locally against a local DP pointed at a prod DB snapshot. and loaded the user followers and following modals from the user page on the desktop.

### DIFF
--- a/src/containers/discover-page/store/lineups/trending/reducer.js
+++ b/src/containers/discover-page/store/lineups/trending/reducer.js
@@ -11,6 +11,7 @@ const initialState = {
   ...initialLineupState,
   antiBot: true,
   dedupe: true,
+  containsDeleted: false,
   prefix: PREFIX
 }
 
--- a/src/containers/favorites-page/store/sagas.ts
+++ b/src/containers/favorites-page/store/sagas.ts
@@ -12,24 +12,18 @@ import { createUserListProvider } from 'containers/user-list/utils'
 import { FavoriteType } from 'models/Favorite'
 import { trackFavoriteError, playlistFavoriteError } from './actions'
 import { watchFavoriteError } from './errorSagas'
-import { encodeHashId } from 'utils/route/hashIds'
 
 const getPlaylistFavorites = createUserListProvider<Collection>({
   getExistingEntity: getCollection,
   extractUserIDSubsetFromEntity: (collection: Collection) =>
     collection.followee_saves.map(r => r.user_id),
-  fetchAllUsersForEntity: ({ limit, offset, entityId, currentUserId }) => {
-    const playlistId = encodeHashId(entityId)!
-    const encodedUserId = currentUserId
-      ? encodeHashId(currentUserId) ?? undefined
-      : undefined
-    return apiClient.getPlaylistFavoriteUsers({
+  fetchAllUsersForEntity: ({ limit, offset, entityId, currentUserId }) =>
+    apiClient.getPlaylistFavoriteUsers({
       limit,
       offset,
-      playlistId,
-      currentUserId: encodedUserId
-    })
-  },
+      playlistId: entityId,
+      currentUserId
+    }),
   selectCurrentUserIDsInList: getUserIds,
   canFetchMoreUsers: (collection: Collection, combinedUserIDs: ID[]) =>
     combinedUserIDs.length < collection.save_count,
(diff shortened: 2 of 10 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/jowlee-user-follow

jowlee:
Added endpoints for tracks/playlist favorites and reposts

src/containers/favorites-page/store/sagas.ts
```diff
@@ -33,8 +40,18 @@ const getTrackFavorites = createUserListProvider<Track>({
...
   extractUserIDSubsetFromEntity: (track: Track) =>
     track.followee_saves.map(r => r.user_id),
-  fetchAllUsersForEntity: ({ limit, offset, entityId }) =>
-    AudiusBackend.getFavoritersForTrack({ limit, offset, trackId: entityId }),
+  fetchAllUsersForEntity: ({ limit, offset, entityId, currentUserId }) => {
+    const trackId = encodeHashId(entityId)!
+    const encodedUserId = currentUserId
+      ? encodeHashId(currentUserId) ?? undefined
+      : undefined
+    return apiClient.getTrackFavoriteUsers({
```
raymondjacobson, line 48: nit, I think we should put the encoded hash id calls inside apiClient.getTrackFavoriteUsers. I think michael did this

src/services/audius-api-client/AudiusAPIClient.ts
```diff
@@ -60,6 +110,156 @@ class AudiusAPIClient {
     return adapted
   }
```
raymondjacobson, line 112: Not for this pr, but we're gonna need to break this file up sooner rather than later before it's too late

raymondjacobson (approved):

src/containers/discover-page/store/lineups/trending/reducer.js
```diff
@@ -11,6 +11,7 @@ const initialState = {
   ...initialLineupState,
   antiBot: true,
   dedupe: true,
+  containsDeleted: false,
```
raymondjacobson, line 14: nice.

src/services/audius-api-client/AudiusAPIClient.ts
piazzatron: Yep +1

src/containers/reposts-page/store/sagas.ts
```diff
@@ -36,13 +43,25 @@ const getTrackReposts = createUserListProvider<Track>({
...
+    currentUserId
   }: {
     limit: number
     offset: number
     entityId: ID
-  }) =>
-    AudiusBackend.getRepostersForTrack({ limit, offset, trackId: entityId }),
+    currentUserId: ID | null
+  }) => {
+    const trackId = encodeHashId(entityId)!
```
piazzatron, line 54: let's avoid using force unwraps - we should throw an error if the ID is unencodable for some reason

piazzatron: same comment for other one

src/containers/following-page/store/sagas.ts
```diff
@@ -3,25 +3,39 @@ import { USER_LIST_TAG } from '../FollowingPage'
...
+    entityId,
+    currentUserId
   }: {
     limit: number
     offset: number
     entityId: ID
-  }) => AudiusBackend.getFollowees(entityId, limit, offset),
+    currentUserId: ID | null
+  }) => {
+    const encodedUserId = currentUserId
```
piazzatron, line 28: maybe we should have a wrapper function around encodeHashID or encodeHashId should take a nullable value so we can avoid this song and dance everywhere

src/services/audius-api-client/AudiusAPIClient.ts
```diff
@@ -19,6 +27,48 @@ type GetTrendingArgs = {
   genre?: string
 }
 
+type GetFollowingArgs = {
+  profileUserId: string
+  currentUserId?: string
```
piazzatron, line 32: I forgot this on trending, but I have an 'OpaqueID' type that we could use for this

src/containers/favorites-page/store/sagas.ts
piazzatron: I don't think I did this actually, but I agree w you. Less leaky abstraction if we can keep all the opaque IDs contained in this client

piazzatron (approved):
Looking slick 🎉

PR #37, opened 2020-10-09
hotfix 10 9 bugfix

raymondjacobson:
### Description
This fixes two bugs we've actually had for a while:
* direct linking into /history fails because we don't wait for libs init
* eager load util straight up had a bug where it was not using a full URL if url params were passed in. this led to visiting /profile and looking at their reposts, you'd actually just see YOUR feed. 

### Dragons

nah

### How Has This Been Tested?

Manually by running the dapp


audius-infra:
Preview this change https://frontend-demo.audius.co/rj-hotfix-10-9-bugfix

piazzatron (approved):

PR #38, opened 2020-10-09
Fix client bugs from API work

piazzatron:
### Description
- Fix track upload polling bug
- Fix playlist creation bug

### Dragons

None

### How Has This Been Tested?

Tested locally

### DIFF
--- a/src/store/cache/collections/sagas.js
+++ b/src/store/cache/collections/sagas.js
@@ -147,7 +147,8 @@ function* confirmCreatePlaylist(uid, userId, formFields, source) {
         // If the user is trying to upload artwork, check that an artwork gets set.
         if (formFields.artwork && formFields.artwork.file) {
           check = playlist =>
-            some([playlist], toConfirm) && playlist.cover_art_sizes
+            some([playlist], toConfirm) &&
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/piazz-playlist-fix

src/store/cache/collections/sagas.js
```diff
@@ -147,7 +147,8 @@ function* confirmCreatePlaylist(uid, userId, formFields, source) {
         // If the user is trying to upload artwork, check that an artwork gets set.
         if (formFields.artwork && formFields.artwork.file) {
           check = playlist =>
-            some([playlist], toConfirm) && playlist.cover_art_sizes
+            some([playlist], toConfirm) &&
+            playlist.playlist_image_sizes_multihash
```
raymondjacobson, line 151: don't we massage out this field into cover_art_size? maybe add a comment if im missing something

raymondjacobson (approved):

src/store/cache/collections/sagas.js
piazzatron: discussed on slack

PR #42, opened 2020-10-13
Fix legacy playlist cover art fields

jowlee:
### Trello Card Link
na

### Description
The old DP api returns the playlist fields `playlist_image_multihash` and `playlist_image_sizes_multihash` instead of `cover_art` and `cover_art_sizes` respectively returned from the v1 API. 

Since the collections saga func watchFetchCoverArt was updated to only look for the new fields, I added a mapping in the audiusBackend instead of updating the `watchFetchCoverArt` function to handle both fields for redux store consistency, but I could see the argument to also add the check in watchFetchCoverArt. lmk what you guys think.  

Also fix the confirmer comparing against the correct image field. 

### Dragons

no

### How Has This Been Tested?

I ran it against staging!

### DIFF
--- a/src/services/AudiusBackend.js
+++ b/src/services/AudiusBackend.js
@@ -335,8 +335,21 @@ class AudiusBackend {
       coverArtSizes[DefaultSizes.OVERRIDE] = placeholderCoverArt
     }
 
+    const legacyFieldMapping = {}
+    if (
+      collection.playlist_image_sizes_multihash &&
+      !collection.cover_art_sizes
+    ) {
+      legacyFieldMapping.cover_art_sizes =
+        collection.playlist_image_sizes_multihash
+    }
+    if (collection.playlist_image_multihash && !collection.cover_art) {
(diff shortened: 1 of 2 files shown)


src/services/AudiusBackend.js
```diff
@@ -718,7 +731,10 @@ class AudiusBackend {
     } catch (err) {
       console.error(err)
     }
-    return feedItems
+    return feedItems.map(item => {
+      if (item.playlist_id) return AudiusBackend.getCollectionImages(item)
```
raymondjacobson, line 735: nice -- so moving feed to v1 should help deal w/ this right?

raymondjacobson (approved):
good w/ me until code is on new API

src/services/AudiusBackend.js
jowlee: Yes, if we move feed and getPlaylists over the v1 then we shouldn't need this anymore!

PR #43, opened 2020-10-13
Fix loading on notifications when empty

jowlee:
### Trello Card Link
na

### Description
If a user has no notification, it's left in a spinning state forever b/c the status is never set to success. 

### Dragons

no

### How Has This Been Tested?

ran it locally against a new account w/ no notifs and an old account w/ notifs

### DIFF
--- a/src/containers/notification/store/sagas.ts
+++ b/src/containers/notification/store/sagas.ts
@@ -433,6 +433,14 @@ function* getNotifications(isFirstFetch: boolean) {
             )
           )
         }
+      } else if (status !== Status.SUCCESS) {
+        yield put(
+          notificationActions.fetchNotificationSucceeded(
+            [], // notifications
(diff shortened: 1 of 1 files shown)


src/containers/notification/store/sagas.ts
```diff
@@ -433,6 +433,14 @@ function* getNotifications(isFirstFetch: boolean) {
             )
           )
         }
+      } else if (status !== Status.SUCCESS) {
+        yield put(
+          notificationActions.fetchNotificationSucceeded(
+            [], // notifications
+            0, // totalUnread
```
raymondjacobson, line 440: this seems like a bad code smell that we would report "success" in the case of non-success. Can we make a new action?

audius-infra:
Preview this change https://frontend-demo.audius.co/jowlee-no-notifs-loading

PR #45, opened 2020-10-14
Redux store for wallet + dashboard

piazzatron:
### Description
Lays down redux groundwork for new dashboard functionality. 

The `wallet slice and sagas` are responsible for tracking claims + balance, and calling `WalletClient` to transfer + claim.

The `token-dashboard slice and sagas` manage flows through the dashboard, and fire`wallet slice` actions to transfer + claim.

Feel free to bikeshed on names :) 

### Dragons

None

### How Has This Been Tested?

Fired actions manually in Redux Debugger, made sure balances changed appropriately, flow through the dashboard was as expected.


audius-infra:
Preview this change https://frontend-demo.audius.co/piazz-wallet-store

package.json
```diff
@@ -17,6 +17,7 @@
     "antd": "3.20.4",
     "array-pack-2d": "^0.1.1",
     "array-range": "^1.0.1",
+    "bn.js": "^5.1.3",
```
raymondjacobson, line 20: Let's add bn.js as a peer dependency rather a strict one so we don't end up getting even more bloat. And let's make it super super permissive (i think >= 4.0.0)

```
"peerDependencies": {
    "bn.js": ">= 4.0.0"
  }
```

web3 itself uses ^4.11.9
https://github.com/ethereum/web3.js/blob/1.x/package.json#L104

src/containers/token-dashboard/sagas.ts
```diff
@@ -0,0 +1,104 @@
+import { select } from 'redux-saga-test-plan/matchers'
```
raymondjacobson, line 1: just following our patterns (which we should def consolidate), we should probs at least put this in containers/token-dashboard/store/sagas.ts

src/store/backend/sagas.js
```diff
@@ -75,8 +76,12 @@ export function* setupBackend() {
     console.info('Reconnected')
   }
 
+  // Init APICLient
   yield call(() => apiClient.init())
 
+  // Fetch wallet claim + balance
+  yield all([put(getClaim()), put(getBalance())])
```
raymondjacobson, line 83: can we do this on page load rather than on all sagas?

jowlee (approved):
bn.js can't handle decimals, so when we get the wei (10^-18) from libs, I'm assuming that you want to store it as that and just display it, so I think that's something we should keep in mind / comment at ever conversion later on.

package.json
piazzatron: good idea

src/containers/token-dashboard/sagas.ts
piazzatron: will move

src/store/backend/sagas.js
piazzatron: discussed on call - makes sense to do it before page load. but this doesn't delay load or anything since it's just firing some actions

src/containers/token-dashboard/sagas.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  const claimingState: ModalState = {
+    stage: 'CLAIM',
+    claimState: {
+      stage: 'CLAIMING'
+    }
+  }
+
+  // Set loading state
+  yield all([
```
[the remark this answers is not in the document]
piazzatron: yeah it is, just in the habit of using `all`

src/containers/token-dashboard/sagas.ts
```diff
@@ -0,0 +1,104 @@
...
+  if (!sendData) return
+  const { recipientWallet, amount } = sendData
+  yield put(walletSend({ recipientWallet, amount: amount.toString() }))
+}
+
+function* claim() {
+  const claimableBalance: ReturnType<typeof getClaimableBalance> = yield select(
+    getClaimableBalance
+  )
+  if (!claimableBalance) return
```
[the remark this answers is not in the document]
piazzatron: Oh good callout

piazzatron:
> bn.js can't handle decimals, so when we get the wei (10^-18) from libs, I'm assuming that you want to store it as that and just display it, so I think that's something we should keep in mind / comment at ever conversion later on.

good callout 👍

PR #61, opened 2020-10-17
Fix user missing images

raymondjacobson:
Since we don't call processAndCache in this flow, we don't reformat the user into having the proper image fields


piazzatron (approved):
If we want to keep it consistent with how we do tracks, we could put in `onAfterAddToCache` in retrieve

PR #63, opened 2020-10-17
Fix handle case sensitivity

raymondjacobson:
### Description
https://audius.co/DJRROS3
Links with non-case-sensitivity in handles are broken. We always cache handles as lowercase
```
handleStatuses[status.handle.toLowerCase()] = {
            id: status.id,
            status: status.status
          }
```

So we've probably had some bugs from this for a while

### Dragons

### How Has This Been Tested?

verified fetching both http://localhost:3002/DJRROS3 and http://localhost:3002/djrros3 work


audius-infra:
Preview this change https://frontend-demo.audius.co/hotfix-handle-sensitivity

jowlee (approved):

PR #69, opened 2020-10-21
UI Touchups

raymondjacobson:
### Description
* Stops optimizely info logs on dev mode (I don't think we actively use them, but it's easy enough to re-enable)
* Switches around /signup lazy loading so we only lazy load the metamask component (which requires libs content). This makes first load on audius.co/signup a bit nicer
* Disables the send button if there's an error on the "send" wallet client
* Fixes dashboard on lower res displays

### Dragons

### How Has This Been Tested?

Pushed to joey.audius.co and played around sign up & dashboard

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -91,9 +91,9 @@
       }
     },
     "@audius/stems": {
-      "version": "0.2.10",
-      "resolved": "https://registry.npmjs.org/@audius/stems/-/stems-0.2.10.tgz",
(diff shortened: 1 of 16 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/rj-touchups

piazzatron (approved):
Looks awesome! Ty for doing all this

PR #76, opened 2020-10-22
Fix web3 race

piazzatron:
### Description

Attempts to address `this.web3.eth.Contract is not a constructor` error by having internal web3 stick a `web3Loaded` flag on the window, and testing for that in backend setup, rather than the existence of window.Web3.

### Dragons

Nah

### How Has This Been Tested?

Tested locally

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -34,7 +34,7 @@
 
     <script
       defer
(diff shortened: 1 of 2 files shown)


raymondjacobson (approved):
DUH! Great solution :)

audius-infra:
Preview this change https://frontend-demo.audius.co/piazz-web3-race

PR #93, opened 2020-10-29
Fix input bg

raymondjacobson:
### Description
I believe this bug has been around for forever (or browsers changed how they do things, which I doubt).
https://github.com/AudiusProject/audius-client-old/commit/8f57bd28309630b6410f87fd95269d60db69834d#diff-081e9cd26dc46773db656642905fae33c5851bdb37c4b4ec38ea416c0b7c41ecR38
This PR added (I think erroneously) a background color to disabled placeholders, which actually show on top of handles (placeholder being the elevated "handle" here)
Screen Shot 2020-10-29 at 2 27 23 PM

This manifests as a "bug" where artists coming in who are verified on twitter, prepopulate their sign on form, and aren't allowed to change their handle (input disabled). So it displays weirdly and they back out and try again (sometimes even giving up) which does write to identity & generates a lot of noise

### Dragons

low-level input change

### How Has This Been Tested?

* Made it so that sign up thought I was verified from twitter and made sure the input looked correct
* Manually disabled all the inputs and walked through the app (didn't see anything weird)
* Searched through codebase of all <Input with disabled and didn't catch anything obvious

### DIFF
--- a/src/components/data-entry/Input.module.css
+++ b/src/components/data-entry/Input.module.css
@@ -40,8 +40,7 @@
   font-size: var(--font-xl);
 }
 
-.input.disabled,
-.input.disabled .placeholder {
+.input.disabled {
   cursor: default !important;
   background-color: var(--neutral-light-9);
 }
@@ -87,6 +86,7 @@
 }
 
 :global(.placeholder.focus).placeholder {
+  /* Z-index of 2 here to be on top of password managers */
   z-index: 2;
   pointer-events: none;
   color: var(--neutral-light-4);


audius-infra:
Preview this change https://frontend-demo.audius.co/rj-fix-input

piazzatron (approved):
Looks fine to me - only thing I can think of is that `.placeholder` no longer has `cursor: default !important;`. Any idea if that was important?

raymondjacobson:
> Looks fine to me - only thing I can think of is that `.placeholder` no longer has `cursor: default !important;`. Any idea if that was important?

I was looking thru it and I couldn't find any reason. I think the error in the original PR I linked was just adding placeholder to everything to get the text color to change and didn't realize that disabled has a BG color change instead. The placeholder isn't cursor:pointer ever anyway

PR #102, opened 2020-11-04
Allow track null in polling confirmation

raymondjacobson:
### Description
Makes it so track confirmer can tolerate initial 404s

### Dragons

Touches upload flow

### How Has This Been Tested?

No reselection!

Screen Shot 2020-11-04 at 10 28 03 AM

### DIFF
--- a/src/services/audius-api-client/AudiusAPIClient.ts
+++ b/src/services/audius-api-client/AudiusAPIClient.ts
@@ -431,7 +431,10 @@ class AudiusAPIClient {
     return adapted
   }
 
-  async getTrack({ id, currentUserId, unlistedArgs }: GetTrackArgs) {
+  async getTrack(
+    { id, currentUserId, unlistedArgs }: GetTrackArgs,
+    retry = true
+  ) {
     const encodedTrackId = this._encodeOrThrow(id)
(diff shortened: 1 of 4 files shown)


src/services/audius-api-client/AudiusAPIClient.ts
```diff
@@ -431,7 +431,10 @@ class AudiusAPIClient {
     return adapted
   }
 
-  async getTrack({ id, currentUserId, unlistedArgs }: GetTrackArgs) {
+  async getTrack(
+    { id, currentUserId, unlistedArgs }: GetTrackArgs,
+    retry = true
```
piazzatron, line 436: nit: any reason not to put retry in the GetTrackArgs?

piazzatron (approved):

src/services/audius-api-client/AudiusAPIClient.ts
raymondjacobson: I figured it made more sense to leave out because it's not really part of the "args" themselves. maybe it could be part of a "config" object or something like that, but i don't think we'll use this all that much

audius-infra:
Preview this change https://frontend-demo.audius.co/rj-allow-null

PR #107, opened 2020-11-05
Add Track Upload Metrics

piazzatron:
### Description

- Adds success + error events for track upload
- Fixes collection uploads not following collection upload code path

### Dragons

Upload is scary, but this is pretty tame for upload.

### How Has This Been Tested?

Tested every upload type in both failure + success cases, logged out the events we'll be sending up.

### DIFF
--- a/src/containers/upload-page/store/sagas.js
+++ b/src/containers/upload-page/store/sagas.js
@@ -341,7 +341,7 @@ function* uploadWorker(requestChan, respChan, progressChan) {
  *
  * tracks is of type [{ track: ..., metadata: ... }]
  */
-export function* handleUploads(tracks, isCollection, isStem = false) {
+export function* handleUploads({ tracks, isCollection, isStem = false }) {
   const numWorkers = getNumWorkers(tracks.map(t => t.track.file))
 
   // Map of shape {[trackId]: { track: track, metadata: object, artwork?: file, index: number }}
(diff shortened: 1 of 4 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/piazz-track-upload-fix

src/containers/upload-page/store/sagas.js
```diff
@@ -488,6 +494,21 @@ export function* handleUploads(tracks, isCollection, isStem = false) {
...
+
+  // Report success + failure events
+  const uploadType = isCollection
+    ? isAlbum
+      ? 'album'
+      : 'playlist'
+    : 'multi_track'
+  yield reportSuccessAndFailureEvents({
+    // Don't report non-uploaded tracks due to playlist upload abort
+    numSuccess: tracks.length - failedRequests.length - numOutstandingRequests,
```
raymondjacobson, line 506: seems a little dangerous to do this math rather than tracking a number through. i could see us intro-ing a bug in the tracking of failures or something and then losing successes. i can't think of anything better though.

raymondjacobson (approved):

src/containers/upload-page/store/sagas.js
piazzatron: Yeah, this worried me a little bit too. It's definitely fine as it is now but I'm always down w defensive programming; it'd be easy to just add a count here. I'll do it

PR #112, opened 2020-11-07
Local Dev - Remove 3 follower requirement

hareeshnagaraj:
### Description
Minor change to min follower count, just to make local dev smoother - I always use dapp and end up refreshing randomly so I figured this would be nicer.

### Dragons

None that i can think of

### How Has This Been Tested?

Created users locally and verified that the gate does not block on 3 follows - will verify on stage later as well (it should still be 3 followers required)

### DIFF
--- a/src/containers/sign-on/components/desktop/FollowPage.tsx
+++ b/src/containers/sign-on/components/desktop/FollowPage.tsx
@@ -19,7 +19,10 @@ const messages = {
   following: 'Following'
 }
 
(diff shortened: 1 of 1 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/hn_localdev_follow_req

raymondjacobson (approved):
☮️

raymondjacobson:
@hareeshnagaraj let's merge!!

PR #131, opened 2020-12-02
Fix broken scroll on edit playlist mobile

piazzatron:
### Trello Card Link

https://trello.com/c/s04A5Jac

### Description
- Scrolling on edit playlist page was broken (would get stuck). Moved to using our native scroll lock solution to handle this instead of body-scroll-lock npm package.
- Improved useScrollLock ergonomics

### Dragons

None

### How Has This Been Tested?

Tested in iOS simulator safari

### DIFF
--- a/src/containers/nav/mobile/TopLevelPage.tsx
+++ b/src/containers/nav/mobile/TopLevelPage.tsx
@@ -1,8 +1,7 @@
-import React, { useEffect } from 'react'
+import React from 'react'
 import cn from 'classnames'
 import { connect } from 'react-redux'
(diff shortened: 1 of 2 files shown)


raymondjacobson (approved):
Maybe we should use the one from stems and add this there?

audius-infra:
Preview this change https://frontend-demo.audius.co/piazz-bugs-dec-2

piazzatron:
> Maybe we should use the one from stems and add this there?

Actually realized we can't do that because the correct actions aren't in stems - instead we should wrap the stems version in the client. Will change it

PR #132, opened 2020-12-03
Fix deeplink into collection on mobile

piazzatron:
### Trello Card Link

https://trello.com/c/6wa2YJ7y

### Description

Reset race condition that would break deeplinking into collections on mobile. Solution is to reset on Collection page unmount, rather than during mount.

### Dragons

Probably not

### How Has This Been Tested?

Tested in simulator and browser, still trying to test on device...


audius-infra:
Preview this change https://frontend-demo.audius.co/piazz-playlist-deeplink-fix

raymondjacobson (approved):

PR #133, opened 2020-12-04
Forward logs to mobile

piazzatron:
### Description

Fixes the pain of using Safari for debugging mobile app.

### Dragons

None. Curious if you think this is the right place to init the logging service, though?

### How Has This Been Tested?

Tested in mobile, we see logs!


raymondjacobson (approved):
amazing

PR #137, opened 2020-12-09
Fix trending badge

jowlee:
### Trello Card Link
na

### Description
FIxes the trending badge on the track page.

The error was that this [line in retrieveTrending](https://github.com/AudiusProject/audius-client/blob/079c4d34a2475f55a8bc8ffc850be679f3a16b9d/src/containers/track-page/store/retrieveTrending.ts#L60) was comparing null and undefined and thought that they were different genres. 

It now only expects a string or null so it can strictly compare the values. 

Screen Shot 2020-12-09 at 12 06 24 PM

### Dragons

no

### How Has This Been Tested?
I ran the dapp locally against staging.

### DIFF
--- a/src/containers/track-page/store/retrieveTrending.ts
+++ b/src/containers/track-page/store/retrieveTrending.ts
@@ -15,7 +15,7 @@ import { AppState } from 'store/types'
 
 type RetrieveTrendingArgs = {
   timeRange: TimeRange
-  genre?: string
+  genre: string | null
   offset: number
   limit: number
   currentUserId: ID | null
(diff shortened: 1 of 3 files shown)


src/containers/track-page/store/retrieveTrending.ts
```diff
@@ -15,7 +15,7 @@ import { AppState } from 'store/types'
 
 type RetrieveTrendingArgs = {
   timeRange: TimeRange
-  genre?: string
+  genre: string | null
```
piazzatron, line 18: Can we use nullable here?

src/services/audius-api-client/AudiusAPIClient.ts
```diff
@@ -71,7 +71,7 @@ type GetTrendingArgs = {
   offset?: number
   limit?: number
   currentUserId: Nullable<ID>
-  genre?: string
+  genre: string | null
```
piazzatron, line 74: Same here

piazzatron (approved):
Good god. Nice find

audius-infra:
Preview this change https://frontend-demo.audius.co/jowle-fix-trending-badge

PR #141, opened 2020-12-15
[AUD-13] Fix profile completion meter

piazzatron:
### Trello Card Link

https://trello.com/c/HFeIWwAf

### Description
Profile completion meter broken for two reasons:
- Updates to profile weren't synced to local storage, so we could hit 5 followees but when we load up again, the old profile with 4th followees is cached
- Visiting your Profile Page loads the cached account user embedded into your tracks + collections into the redux store. This profile is potentially out of date from your local redux state, so you have a bug where you go follow the 5th person, confirm, go the profile page, add user with 4 follows into the store, write that out to local storage, and then reload and still see 4

### Dragons

- Cache changes

### How Has This Been Tested?

Tested locally with profile completion meter

### DIFF
--- a/src/store/cache/collections/utils/addUsersFromCollections.ts
+++ b/src/store/cache/collections/utils/addUsersFromCollections.ts
@@ -2,15 +2,30 @@ import { makeUid } from 'utils/uid'
 import { UserCollection } from 'models/Collection'
 import { Kind } from 'store/types'
 import { reformat as reformatUser } from 'store/cache/users/utils'
-import { put } from 'redux-saga/effects'
+import { put, select } from 'redux-saga/effects'
 import * as cacheActions from 'store/cache/actions'
+import { getAccountUser } from 'store/account/selectors'
+import { uniqBy } from 'lodash'
 
+/**
+ * Adds users from collection metadata to cache.
+ * Dedupes users and removes self.
+ * @param metadataArray
+ */
 export function* addUsersFromCollections(metadataArray: Array<UserCollection>) {
-  const users = metadataArray.map(m => ({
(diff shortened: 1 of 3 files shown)


src/store/cache/collections/utils/addUsersFromCollections.ts
```diff
@@ -2,16 +2,31 @@ import { makeUid } from 'utils/uid'
...
+    getAccountUser
+  )
+  const currentUserId = accountUser?.user_id
+  let users = metadataArray.map(m => ({
     id: m.user.user_id,
     uid: makeUid(Kind.USERS, m.user.user_id),
     metadata: reformatUser(m.user)
   }))
 
+  // Removes duplicates and self
```
raymondjacobson, line 26: perhaps if currentUserId { } around the filter so for non-logged in users we can short circuit the logic

raymondjacobson (approved):
lgtm! do we want to replace (or just remove) the similar instance of this that we found in profile page sagas?

piazzatron:
> lgtm! do we want to replace (or just remove) the similar instance of this that we found in profile page sagas?

ahh ty for reminding, going to remove it

src/store/cache/collections/utils/addUsersFromCollections.ts
piazzatron: yeah that's better

piazzatron:
> > lgtm! do we want to replace (or just remove) the similar instance of this that we found in profile page sagas?
> 
> ahh ty for reminding, going to remove it

actually removing it broke this. sad

Issue #92, opened 2020-10-29
Media buttons not working


koloz193:
Currently the media buttons (play/pause/prev/next) don't seem to work in controlling music playback. Would love to be able to control the music when in another desktop window.

zkolodny:
@piazzatron yea, so maybe its different but im on the macbook pro with a touchbar and the controls don't seem to be tracking

kerkkoh:
I just tested this on Windows 10, with media keys, on both Chrome 87.0.4280.66 (64 bit) & Firefox 83.0 (64-bit). "Pause" and "Play" seem to be working just fine (I believe these are working by default for all audio from the browser), but I can confirm that both "Previous" and "Next" are not working.

sliptype:
Closing, see https://github.com/AudiusProject/audius-client/pull/161

PR #161, opened 2020-12-23
Add media session action handlers

jowlee:
### Description
Adds [Media Session Action Handlers](https://developer.mozilla.org/en-US/docs/Web/API/MediaSession/setActionHandler)
NOTE: because of the hack in AudioStream play/pause does not work correctly in chrome. 

Closes AUD-31
### Dragons

none

### How Has This Been Tested?

I used the next, previous, play, and pause on the keyboard to trigger these actions.

### DIFF
--- a/src/store/queue/sagas.ts
+++ b/src/store/queue/sagas.ts
@@ -3,9 +3,11 @@ import {
   call,
   put,
   select,
+  take,
   takeLatest,
   takeEvery
 } from 'redux-saga/effects'
+import { eventChannel } from 'redux-saga'
 
 import {
   play,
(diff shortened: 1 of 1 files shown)


src/store/queue/sagas.ts
```diff
@@ -281,14 +283,53 @@ export function* watchRemove() {
   })
 }
 
+/**
+ * Set media session action handlers
+ * Reference: https://developer.mozilla.org/en-US/docs/Web/API/MediaSession/setActionHandler
+ * NOTE: in chrome there'a a hack in AudioStream where we do not call pause and thus play/pause
```
raymondjacobson, line 289: this seems kinda unfortunate. i imagine you're talking about this https://github.com/AudiusProject/audius-client/blob/ace8ea441f8f8c3e4b4a934ead769dcef7869c1d/src/audio/AudioStream.ts#L334

is there any way around that or does play/pause not work entirely?

jowlee: If you check the html audio object, it never calls pause, so checking `audius.paused` always returns false.
So on when hitting the play/pause btn, it will always think it's playing even when it's paused, but when it's playing you can pause it.

raymondjacobson: can we pull in the paused state from somewhere else? like a selector? i dont think we should ship if this is still borked

piazzatron:
Let's remember to close out: https://github.com/AudiusProject/audius-client/issues/92 when we finish

There was an interesting comment on this issue:
https://github.com/AudiusProject/audius-client/issues/92#issuecomment-731746594

@jowlee any if this would address the issue? Could we just listen for the key press events?

jowlee:
I don't think we should ship this feature, b/c of the hack we have to use. Even if we pulled play/pause state from somewhere  else and made it work correcly, the browser might still think it's playing when it's paused and display the wrong toggle button (On the touchbar for the macbook, it always shows the pause btn and never play).

PR #169, opened 2021-01-05
Fix edit playlist overflow

jowlee:
### Description
* Update stems from 0.3.2 -> 0.3.3
* Adds an Edit Playlist Modal component - it reuses the ui of the create playlist component, but is added at the root level to the Modals container. 
* Adds an edit playlist option to the dropdown of the playlist card.
* Updates the edit playlist modal in the playlist page to use the new component.
* Added a zIndex file to collect all z-indices used in the app. 
* Removed toggle key in edit track modal

Closes AUD-84

### Dragons
**Yes**: I updated stems from 0.3.2 -> 0.3.3 which has the modal style changes.
I played around with the modals and they seemed to be working fine. 

### How Has This Been Tested?
I ran the dapp against staging. I clicked edit playlist and tried all the options from the overflow menu as well as from the playlist/album page and it all worked as intended.

### DIFF
--- a/src/components/create-playlist/CreatePlaylistModal.js
+++ b/src/components/create-playlist/CreatePlaylistModal.js
@@ -193,7 +193,6 @@ CreatePlaylistModal.propTypes = {
   onSave: PropTypes.func,
   // When the delete button is clicked
   onDelete: PropTypes.func,
-  // When the delete button is clicked
   zIndex: PropTypes.number
 }
 
--- a/src/components/create-playlist/CreatePlaylistModal.js
+++ b/src/components/create-playlist/CreatePlaylistModal.js
@@ -25,7 +25,7 @@ const CreatePlaylistModal = props => {
     artwork: false
   })
 
-  const { visible, metadata, editMode } = props
+  const { zIndex, visible, metadata, editMode } = props
 
   // On modal open (similar to componentDidMount, but this component is mounted before being displayed)
   useEffect(() => {
@@ -119,6 +119,7 @@ const CreatePlaylistModal = props => {
       titleClassName={styles.modalTitle}
       isOpen={props.visible}
       onClose={onCancel}
+      zIndex={zIndex}
     >
       <div className={styles.createPlaylist}>
         <UploadArtwork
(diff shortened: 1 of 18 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/jowlee-fix-edit-playlist-overflow

src/components/create-playlist/CreatePlaylistModal.js
```diff
@@ -191,7 +192,9 @@ CreatePlaylistModal.propTypes = {
   // When the save button is clicked
   onSave: PropTypes.func,
   // When the delete button is clicked
-  onDelete: PropTypes.func
+  onDelete: PropTypes.func,
+  // When the delete button is clicked
```
raymondjacobson, line 196: remove comment :)

src/store/application/ui/editPlaylistModal/selectors.ts
```diff
@@ -0,0 +1,20 @@
...
+const getBase = (state: AppState) => state.application.ui.editPlaylistModal
+export const getIsOpen = (state: AppState) => {
+  return getBase(state).isOpen
+}
+
+export const getCollectionId = (state: AppState) => {
+  return getBase(state).collectionId
+}
+
+export const getCollectionAndUser = (state: AppState) => {
```
raymondjacobson, line 14: i'd love us to standardize this selector in the cache or something like that -- we already have some things defined like `getTracksFromCollection` etc.

src/utils/zIndex.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Standardize the use of zIndex across the application
+ *
+ * Reference: https://www.notion.so/audiusproject/Client-Modal-Design-cc9d722f2aed4a0abf14fb44f97c246e
```
raymondjacobson, line 4: I'd remove this reference since it's for modals and not exposed in notion. i think we should just get better about documentation inside code -- like let's specify around these two zIndexes why we have them. If you think some of the notes there are super useful, let's just copy the text into this doc-comment.

// Set to 1000 to account for nested modals inside, which take a higher z-index
EDIT_TRACK_MODAL = 1000

raymondjacobson (approved):

jowlee:
AUD-84

PR #173, opened 2021-01-05
[AUD-54] Fix flickering drop shadow in nav column

piazzatron:
### Description

Switching between tracks with the same artwork would cause the dropshadow to do an ugly flicker (highly visible here: https://audius.co/mrcarmack/album/blu-ep-154) because when we load the new track, we haven't calculated the averageRGB yet, so we briefly display the placeholder while calculating the new one.

This fix pulls the averageRGB info out of the track model and into a new redux slice that maps the image CID -> averageRGB to allow us to reuse the average color info for tracks with the same artwork.

### Dragons

Nah

### How Has This Been Tested?

Tested on desktop + mobile browser

### DIFF
--- a/src/store/application/ui/average-color/slice.ts
+++ b/src/store/application/ui/average-color/slice.ts
@@ -1,0 +1,35 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit'
+import Color from 'models/common/Color'
+import { CID } from 'models/common/Identifiers'
+import { AppState } from 'store/types'
+
+const initialState: { colors: { [multihash: string]: Color } } = {
+  colors: {}
+}
+
+/**
+ * This slice tracks computed average colors for a given track art CID.
+ * Colors is a map of art cid -> Color
+ */
+const slice = createSlice({
+  name: 'application/ui/averageColor',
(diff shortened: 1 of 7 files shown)


src/store/application/ui/average-color/slice.ts
```diff
@@ -0,0 +1,37 @@
...
+      state.colors[multihash] = color
+    }
+  }
+})
+
+export const { setColor } = slice.actions
+
+export const getAverageColor = (
+  state: AppState,
+  { multihash }: { multihash: Nullable<CID> }
```
raymondjacobson, line 33: maybe nitty, but might be nice to export a second selector here that takes state and track and does the logic you did in the mapstateto's

raymondjacobson (approved):
lgtm! sweet! any noticeable perf changes?

src/store/application/ui/average-color/slice.ts
piazzatron: yeah dis is a good idea, will do

piazzatron:
> lgtm! sweet! any noticeable perf changes?

none that I saw!

PR #182, opened 2021-01-07
[AUD-110] Fix Unfollow from Desktop User List

piazzatron:
### Description

- Changes follow button to immediately show 'follow' after unfollowing, as per discussion w product and design
- Fixes unfollow from userlist broken on desktop
- Looks like we were missing unsubscribe when unfollow on a few buttons, following the behavior on mobile here. Not 100% sure this is necessary?

### Dragons

None

### How Has This Been Tested?

Locally

### DIFF
--- a/src/components/artist/ArtistPopover.tsx
+++ b/src/components/artist/ArtistPopover.tsx
@@ -18,6 +18,7 @@ import { FollowSource } from 'services/analytics'
 import { useUserCoverPhoto, useUserProfilePicture } from 'hooks/useImageSize'
 import { WidthSizes, SquareSizes } from 'models/common/ImageSizes'
 import { MountPlacement } from 'components/types'
+import { setNotificationSubscription } from 'containers/profile-page/store/actions'
 
 enum Placement {
   Top = 'top',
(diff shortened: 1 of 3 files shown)


raymondjacobson:
Seems reasonable, can you double check w/ forrest on this copy change tho? I think the follow/unfollow stuff was intentional ( but maybe confusing )

audius-infra:
Preview this change https://frontend-demo.audius.co/aud-110

piazzatron:
> Seems reasonable, can you double check w/ forrest on this copy change tho? I think the follow/unfollow stuff was intentional ( but maybe confusing )

This ended up being really funny. 

A) This wasn't even the bug QA reported
but
B) Forrest agreed that the current button state was real wonky.

Ended up going back and forth w him and Julian a bit, ended up with this.

src/components/artist/ArtistPopover.tsx
```diff
@@ -154,8 +155,10 @@ const mapDispatchToProps = (dispatch: Dispatch) => ({
   goToRoute: (route: string) => dispatch(pushRoute(route)),
   onFollow: (userId: ID) =>
     dispatch(socialActions.followUser(userId, FollowSource.HOVER_TILE)),
-  onUnfollow: (userId: ID) =>
+  onUnfollow: (userId: ID) => {
     dispatch(socialActions.unfollowUser(userId, FollowSource.HOVER_TILE))
+    dispatch(setNotificationSubscription(userId, false, true))
```
raymondjacobson, line 160: +1 ty for catching these

raymondjacobson (approved):

PR #184, opened 2021-01-08
Fix tracklist missing prop on mobile history pg

jowlee:
### Description
Closes AUD-56
Bug: On the mobile listening history page, when you reposted a song, it would not show up as reposted. This was caused by a missing prop. 

Solution: add the isReposted prop!

### Dragons

no

### How Has This Been Tested?

Ran the dapp locally against staging w/ the desktop browser set to mobile width. Then went to the mobile listening history page and reposted a track and checked that it was reposted and the option said unrepost after.


audius-infra:
Preview this change https://frontend-demo.audius.co/jowlee-fix-mobile-history-pg-tacks

raymondjacobson (approved):

PR #190, opened 2021-01-12
[AUD-118] Fix web3 race on mobile init

raymondjacobson:
### Description
Fixes race condition on window.Web3 on initial load.

The reason this was happening is because the metamask sign in component was importing Utils directly from libs. Utils, in turn, imports libs/src/Web3.js which checks to see if Web3 is provided by the window or whether requiring it is necessary. In order to do that check in libs/src/Web3.js, Web3 must be already present on the window object. Since Web3 itself is lazy loaded in the client, directly importing Utils from libs is not safe.

Doing so through AudiusBackend as done in this PR is though because the correct ordering enforced --
1. Web3 is lazy loaded
2. Libs is lazy loaded
3. Metamask can use libs utils

### Dragons

n/a

### How Has This Been Tested?

1. Created an account (bleh) using metamask
2. Verified that race condition for web3 doesn't occur on mobile local dev any longer

### DIFF
--- a/src/containers/sign-on/components/desktop/MetaMaskModal.js
+++ b/src/containers/sign-on/components/desktop/MetaMaskModal.js
@@ -1,11 +1,12 @@
 import React, { Component } from 'react'
 import cn from 'classnames'
 import PropTypes from 'prop-types'
-import { Utils } from '@audius/libs'
 import { Button, ButtonType } from '@audius/stems'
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/rj-aud-118

jowlee (approved):

PR #191, opened 2021-01-12
[AUD-111] Prevent touch actions on nav and bottom bar

raymondjacobson:
### Description
A bug exists where you can lose the ability to scroll on the edit playlist page.

Really what this boils down to is you creating a scroll focus on the nav/play bar and scrolling accidentally (which triggers inertial scrolling on the window/document). If you then try to immediately scroll the edit playlist page, since the window/document hasn't lost scroll focus, you continue to scroll it. The user then gets frustrated and scrolls much more aggressively which creates more inertia and it takes a while to regain scroll focus where you want it.

Touch action none is supported pretty widely and from my testing I believe this prevents the issue:
https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action

### Dragons

n/a

### How Has This Been Tested?

iOS simulator

### DIFF
--- a/src/components/bottom-bar/BottomBar.module.css
+++ b/src/components/bottom-bar/BottomBar.module.css
@@ -15,4 +15,5 @@
   display: flex;
   align-items: center;
   justify-content: space-evenly;
+  touch-action: none;
 }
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/rj-aud-111

piazzatron (approved):
Nice find

PR #194, opened 2021-01-13
Send theme change messages to native layer

piazzatron:
### Description

This & native layer code fixes two bugs:
- Manually setting the theme in the app wouldn't change the StatusBar color - so if your phone was in dark mode and you set the theme to light, your status bar is invisible. 
- Our react-native-webview package has an evil default where it manages the StatusBar color unless you tell it otherwise, and it has weird behavior (randomly changing the color to be white when it shouldn't be, etc). https://github.com/react-native-webview/react-native-webview/issues/735

### Dragons

Native change did update the Webview package. But seems fine.

### How Has This Been Tested?

On physical iOS device

### DIFF
--- a/src/store/application/ui/theme/sagas.ts
+++ b/src/store/application/ui/theme/sagas.ts
@@ -7,6 +7,7 @@ import { SET_THEME, ThemeActions, setTheme as setThemeAction } from './actions'
 import Theme from 'models/Theme'
 import { PrefersColorSchemeMessage } from 'services/native-mobile-interface/android/theme'
 import { getIsIOS } from 'utils/browser'
+import { ThemeChangeMessage } from 'services/native-mobile-interface/theme'
 
 const NATIVE_MOBILE = process.env.REACT_APP_NATIVE_MOBILE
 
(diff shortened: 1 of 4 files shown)


src/store/application/ui/theme/sagas.ts
```diff
@@ -38,12 +39,15 @@ function* watchNativeTheme() {
 
 export function* watchSetTheme() {
   yield takeEvery(SET_THEME, function* (action: ThemeActions) {
-    setTheme(action.theme)
+    const { theme } = action
+    setTheme(theme)
+    const message = new ThemeChangeMessage(theme)
```
raymondjacobson, line 44: i dont think this would break if not native, but maybe check if NATIVE_MOBILE first?

raymondjacobson (approved):

audius-infra:
Preview this change https://frontend-demo.audius.co/aud-63

src/store/application/ui/theme/sagas.ts
piazzatron: 👍

PR #195, opened 2021-01-13
Bump libs to 1.1.9

piazzatron:


piazzatron:
I think keccak may be a more common error than we realized.

https://sentry.io/organizations/audius/issues/1976205053/?project=1457231&query=is%3Aunresolved&statsPeriod=14d

raymondjacobson (approved):

PR #196, opened 2021-01-13
Prevent a route switch if already on login/signup

jowlee:
### Description
If you load in the signin page or the signup page and quickly switch to the signup page, it goes back to the signup page. 
This is because a saga re-routes to the signup page if fetch account fails.
The fix is to check if the user is already on either and not push route if the user is on signup or signin. 

Closes AUD-121

### Dragons

nope

### How Has This Been Tested?

I loaded the signup page and quickly switched to the signin page and it didn't switch back. 
I loaded the signin page and it didn't switch to the signup page

### DIFF
--- a/src/store/account/mobileSagas.ts
+++ b/src/store/account/mobileSagas.ts
@@ -11,6 +11,8 @@ export const ENTROPY_KEY = 'hedgehog-entropy-key'
 
 function* watchFetchAccountFailed() {
   yield takeEvery(accountActions.fetchAccountFailed.type, function* () {
+    // Do not push route if the user is already on the signup or signin page
+    // or else it will toggle the UI page.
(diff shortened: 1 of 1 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/jowlee-mobile-signup-toggle

src/store/account/mobileSagas.ts
```diff
@@ -11,7 +11,9 @@ export const ENTROPY_KEY = 'hedgehog-entropy-key'
 
 function* watchFetchAccountFailed() {
   yield takeEvery(accountActions.fetchAccountFailed.type, function* () {
-    yield put(pushRoute(SIGN_UP_PAGE))
+    if (!doesMatchRoute(SIGN_IN_PAGE) && !doesMatchRoute(SIGN_UP_PAGE)) {
```
piazzatron, line 14: nit: can we add a brief comment explaining why this is necessary?

piazzatron (approved):

PR #199, opened 2021-01-15
[AUD-129] Fix profile pic not loading

raymondjacobson:
### Description
Sometimes the user's profile pic does not load in the nav column. This is a race condition that is manifested by us caching _profilePictureSizes on the user object in local storage and that user not getting updated with a fresh image before the image is rendered by the useImageSize hook.

The reason it's bad to cache _profilePictureSizes in local storage is that our image references are sometimes blob references (e.g. blob:audius.co/xxx-yyy-zzz), a side effect of how fetchCID needs to work. These references are no longer valid on a new session, but there's a chance we could read the (outdated) local storage ones before having refetched.

### Dragons

n/a

### How Has This Been Tested?

Able to reproduce the bug by visiting my own profile repeatedly and delaying image fetch requests so the race condition manifests more easily (the image has to be pulled from the object in local storage *before* being fetched by the nav column component).

### DIFF
--- a/src/store/cache/users/sagas.js
+++ b/src/store/cache/users/sagas.js
@@ -129,6 +129,33 @@ function* watchAdd() {
   })
 }
 
+/**
+ * Prunes blob url values off of a user.
+ * @param {User} user
+ */
+const pruneBlobValues = user => {
+  const returned = {
+    ...user,
+    _profile_picture_sizes: { ...user._profile_picture_sizes },
+    _cover_photo_sizes: { ...user._cover_photo_sizes }
+  }
+  if (returned._profile_picture_sizes) {
+    Object.keys(returned._profile_picture_sizes).forEach(size => {
+      if (returned._profile_picture_sizes[size].startsWith('blob')) {
+        delete returned._profile_picture_sizes[size]
+      }
+    })
+  }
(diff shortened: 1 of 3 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/rj-aud-129

src/store/cache/users/sagas.js
```diff
@@ -129,6 +129,33 @@ function* watchAdd() {
...
+    Object.keys(returned._profile_picture_sizes).forEach(size => {
+      if (returned._profile_picture_sizes[size].startsWith('blob')) {
+        delete returned._profile_picture_sizes[size]
+      }
+    })
+  }
+  if (returned._cover_photo_sizes) {
+    Object.keys(returned._cover_photo_sizes).forEach(size => {
+      if (returned._cover_photo_sizes[size].startsWith('blob')) {
+        returned._cover_photo_sizes[size] = 'pokemon'
```
raymondjacobson, line 152: oops lol

piazzatron (approved):
LGTM. How would you feel about making a helpers.ts file for this or something & importing that, so we can have the new code in TS rather than contributing to the legacy JS?

PR #214, opened 2021-02-08
Remove optimizely trending bool

jowlee:
### Description
Removes the optimizely boolean value `OPTIMIZED_TRENDING_BADGE_ENDPOINT`
We introduced this remote config value to wait for all discprovs to have a new endpoint before consuming it on the client.
The nodes running the current version of discprov 0.3.28 (check [dashboard](https://dashboard.audius.org/services/discovery-node) - 17/18 Discovery nodes) have this endpoint. 

Closes AUD-256

### Dragons

none

### How Has This Been Tested?  
The remote config value was already flipped from false to true and the client was working as intended. 
This change removes the now, unused code path.
I ran the client against prod locally after the changes and still saw the trending badges

### DIFF
--- a/src/containers/track-page/store/sagas.js
+++ b/src/containers/track-page/store/sagas.js
@@ -1,5 +1,5 @@
 import moment from 'moment'
-import { all, fork, call, put, select, takeEvery } from 'redux-saga/effects'
+import { fork, call, put, select, takeEvery } from 'redux-saga/effects'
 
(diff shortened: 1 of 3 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/jowlee-remove-omtimizely-trending

raymondjacobson (approved):
yay!

PR #219, opened 2021-02-09
Prevent pull to refresh if confirming

jowlee:
### Description
On mobile, if the user edits their profile and pulls down to refresh, the old state is displayed. 
This change prevents the user from pulling to refresh if the user's profile is confirming. 

Closes AUD-213

### Dragons

none

### How Has This Been Tested?

I ran this locally, ngroked and on my phone edited my profile, and made sure that I could not pull to refresh while it was confirming.


raymondjacobson (approved):
fine with me i suppose, but if feel like the better fix is to allow the pull to refresh, but make it so it does nothing. users will be confused as to why they can't pull probably

PR #229, opened 2021-02-18
Pulling property off undefined object throws error in creator node autoselect

dmanjunath:
### Description
The autoselect function expects an object and since nothing is passed in, the function is trying to pull properties off undefined and it errors. This change just passes an empty object so the default properties get pulled off in the autoselect and fixes the error.

Upload_•_Audius

This is only a problem in the old user metadata -> creator flow. New users created via libs don't face this issue because libs correctly passes in an object
### Dragons

### How Has This Been Tested?

Ran the client pointed against prod and verified replica set is assigned

Upload_•_Audius

### DIFF
--- a/src/services/AudiusBackend.js
+++ b/src/services/AudiusBackend.js
@@ -452,7 +452,7 @@ class AudiusBackend {
   }
 
   static async autoSelectCreatorNodes() {
-    return audiusLibs.ServiceProvider.autoSelectCreatorNodes()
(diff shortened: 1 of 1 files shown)


raymondjacobson (approved):

audius-infra:
Preview this change https://frontend-demo.audius.co/dm-autoselect-undef-hotfix

soulfulAdam (approved):

soulfulAdam (approved):

soulfulAdam (approved):

PR #263, opened 2021-03-10
Hide recaptcha always with custom render

raymondjacobson:
### Description
High level of confidence that this will make the captcha error message never show.
I believe the outer div is where the text got added -- effectively it replaced the contents of the outer div in this screenshot:
Screen Shot 2021-03-10 at 3 34 40 PM

Screen Shot 2021-03-10 at 3 52 15 PM

By choosing where to render recaptcha (inside of the ghost div, that should take care of things)

Screen Shot 2021-03-10 at 3 53 29 PM

### Dragons

touches index.html 

### How Has This Been Tested?

Against staging

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -83,7 +83,15 @@
     <!-- end Google Analytics -->
 
     <!-- start reCAPTCHA -->
-    <!-- <script src="https://www.google.com/recaptcha/api.js?render=6LfVR-0ZAAAAADFcqNM1P1IafKwQwN0E_l-gxQ9q"></script> -->
+    <script>
+      onRecaptchaLoad = () => grecaptcha.render('recaptcha-hidden', {
+        'sitekey' : '6LfVR-0ZAAAAADFcqNM1P1IafKwQwN0E_l-gxQ9q',
+        'size': 'invisible',
+        'expired-callback': () => null,
+        'expired-callback': () => null
+      })
+    </script>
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/rj-recaptcha-dom-fix

vicky-g:
If im understanding this correctly, you
1. load the recaptcha render into the div with `id=recaptcha-hidden`
2. all the recaptcha stuff gets loaded into that element
3. that element is always hidden
4. so, effectively, anything recaptcha tries to append via their script gets hidden anyway

raymondjacobson:
> If im understanding this correctly, you
> 
> 1. load the recaptcha render into the div with `id=recaptcha-hidden`
> 2. all the recaptcha stuff gets loaded into that element
> 3. that element is always hidden
> 4. so, effectively, anything recaptcha tries to append via their script gets hidden anyway

5. profit?

vicky-g:
> > If im understanding this correctly, you
> > 
> > 1. load the recaptcha render into the div with `id=recaptcha-hidden`
> > 2. all the recaptcha stuff gets loaded into that element
> > 3. that element is always hidden
> > 4. so, effectively, anything recaptcha tries to append via their script gets hidden anyway
> 
> 1. profit?

### PROFITTTTTTTTTTTTTTTT 🤑

vicky-g (approved):
### 💰 💰 💰 💰 💰 💰 💰 💰 💰 💰 💰

PR #269, opened 2021-03-16
Disable electron captcha

raymondjacobson:
### Description
Electron assets are served from file:// and thus have no domain for captcha to verify against.

### Dragons

n/a

### How Has This Been Tested?

* Reproduced error locally w/ electron on sign up
* Disabling mitigates that error


audius-infra:
Preview this change https://frontend-demo.audius.co/rj-electron-captcha-fix

vicky-g (approved):
I'm assuming electron is used to dev native apps. With that said, do we have any followup plans to enable recaptcha with electron? seems like it'd be valuable

raymondjacobson:
> I'm assuming electron is used to dev native apps. With that said, do we have any followup plans to enable recaptcha with electron? seems like it'd be valuable

Yes - w.r.t the comment, we need to serve the assets from a server rather than the FS

PR #271, opened 2021-03-17
Trending Playlists 

piazzatron:
### Description

- Adds trending rewards pages for desktop + mobile
- Fix misc other things, esp playlist skeleton tiles

Should be basically reviewable by commit.

### Dragons

Changed some playlist tile guts

### How Has This Been Tested?

Tested locally against canary prod GCP GP

### DIFF
--- a/src/components/track/helpers.ts
+++ b/src/components/track/helpers.ts
@@ -45,6 +45,7 @@ export const getCollectionWithFallback = (collection: Collection | null) => {
       is_delete: false,
       activity_timestamp: '',
       _co_sign: undefined,
+      playlist_owner_id: -1,
       _cover_art_sizes: {
         '150x150': '',
         '480x480': '',
--- a/src/containers/lineup/hooks.ts
+++ b/src/containers/lineup/hooks.ts
@@ -1,0 +1,74 @@
+import { useMemo } from 'react'
+import { useDispatch } from 'react-redux'
+import { makeGetLineupMetadatas } from 'store/lineup/selectors'
+import { getBuffering, getPlaying } from 'store/player/selectors'
+import { useSelector } from 'utils/reducer'
+import { LineupVariant } from 'containers/lineup/types'
+import { AppState } from 'store/types'
+import { LineupState } from 'models/common/Lineup'
+import { makeGetCurrent } from 'store/queue/selectors'
+import { isMobile } from 'utils/clientUtil'
+
+type LineupActions = any
+
+type UseLineupProps = {
+  getLineupSelector: (state: AppState) => LineupState<{}>
+  actions: LineupActions
+  variant?: LineupVariant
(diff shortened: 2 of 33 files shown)


src/components/track/helpers.ts
```diff
@@ -45,6 +45,7 @@ export const getCollectionWithFallback = (collection: Collection | null) => {
       is_delete: false,
       activity_timestamp: '',
       _co_sign: undefined,
+      playlist_owner_id: -1,
```
raymondjacobson, line 48: could this break anything? maybe some selector assuming if (playlist_owner_id) or something

src/containers/lineup/hooks.ts
```diff
@@ -0,0 +1,74 @@
...
+  numPlaylistSkeletonRows?: number
+  scrollParent?: HTMLElement
+}
+
+/**
+ * Returns props for a Lineup component.
+ * Requires at least a selector and actions
+ * See example usage in `TrendingPlaylistPage`
+ * */
+export const useLineup = ({
```
raymondjacobson, line 27: maybe we could call this useLineupProps since you still need to render separately? wdyt?

src/containers/trending-playlists/store/lineups/collections/sagas.ts
```diff
@@ -0,0 +1,38 @@
...
+import { getUserId } from 'store/account/selectors'
+import { LineupSagas } from 'store/lineup/sagas'
+import { PREFIX, trendingPlaylistLineupActions } from './actions'
+import { getLineup } from './selectors'
+import apiClient from 'services/audius-api-client/AudiusAPIClient'
+import { processAndCacheCollections } from 'store/cache/collections/utils'
+import Collection, { UserCollectionMetadata } from 'models/Collection'
+
+function* getPlaylists({ limit, offset }: { limit: number; offset: number }) {
+  const time = 'week' as 'week'
```
raymondjacobson, line 11: wow this type is necessary?

raymondjacobson (approved):
code looks great! will do a pass at the UI side after we get stuff merged onto rewards branch, etc.

thx for fixing typos

src/containers/lineup/hooks.ts
piazzatron: Yeah I dig it

src/containers/trending-playlists/store/lineups/collections/sagas.ts
piazzatron: yeah, in cases like this TS doesn't know to narrow a string type to a string literal

src/components/track/helpers.ts
piazzatron: I don't think so - this should basically just return a dummy collection, but we forgot to include owner_id bc we didn't use it anywhere on the dummy collection

PR #273, opened 2021-03-18
Fix for extra requests from DynamicImage

meeped:
### Description
This fixes an issue where the app was trying to request the current path as an image whenever a user hovered over a component using Dynamic image.
Screen Shot 2021-03-18 at 9 25 48 pm

I've removed the url('') from the CSS background change. I can see that the component supports BG image gradients but I think if we are changing the style as opposed to adding an additional css style it should work.

The above change removed all of the requests except for one which was caused by the empty image in the bottom player bar artwork section. This was fixed by adding the transparentPlaceholderImg.

### Dragons

I don't know the app well enough to know if this will have unattended consequences so happy to revisit this if you need me to.

### How Has This Been Tested?
Tested image loading in Firefox / Chrome.


src/components/dynamic-image/DynamicImage.tsx
```diff
@@ -52,7 +54,7 @@ const fadeIn = (
     if (image === placeholder) {
       ref.current.style.backgroundColor = 'unset'
       // Apply an empty url here so we can use the background image prop to set a gradient
```
raymondjacobson, line 56: mind dropping this comment since it no longer applies?

raymondjacobson (approved):
Really nice subtle catch. Thank you!

src/components/dynamic-image/DynamicImage.tsx
meeped: Good catch, done!

PR #289, opened 2021-03-23
Remove visualizer component from app.js

meeped:
### Description
**This is just a speculative pull request and assumes that the visualizer isn't in use at the moment.** I was unable to figure out how to trigger it on web or mobile but noticed that it had a decent affect on our first boot time.

On my device (2.4 GHz 8-Core Intel Core i9) it added around ~400ms of parsing time. I also suspect but haven't proved that the webgl libs it uses might cause us to switch to the discrete GPU (where available) for web acceleration.

visualizer-timing

### Dragons

If we are still using the visualizer I can rework this to use react suspense or something similar and only load when triggered.

### How Has This Been Tested?
Tried triggering the visualizer / reading through the code to find where it was used. Removed the code paths and tested the app without finding any issues.

### DIFF
--- a/src/containers/App.js
+++ b/src/containers/App.js
@@ -81,7 +81,6 @@ import SomethingWrong from 'containers/something-wrong/SomethingWrong'
 import RequiresUpdate from 'containers/requires-update/RequiresUpdate'
 import PlayBarProvider from 'containers/play-bar/PlayBarProvider'
 import UploadType from 'containers/upload-page/components/uploadType'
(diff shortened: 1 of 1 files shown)


raymondjacobson:
Visualizer is used (web not mobile) and can be summoned with `v`

Would be down to look into better codesplitting options

meeped:
> Visualizer is used (web not mobile) and can be summoned with `v`
> 
> Would be down to look into better codesplitting options

Awesome thank you, I'll close this for now and dive into codesplitting!

PR #294, opened 2021-03-23
Audio Page UI

piazzatron:
### Description
- UI for Audio page rewards.
- Adds new withMobileStyle hook
- Excludes: navigation to audio page, modals

**Tagged multiple ppl for exposure to withMobileStyle**

### Dragons

No

### How Has This Been Tested?

Locally


src/containers/audio-rewards-page/AudioRewardsPage.tsx
```diff
@@ -6,15 +6,38 @@ import Page from 'components/general/Page'
...
-export const AudioRewardsPage = () => {
+export const RewardsContent = () => {
+  const wm = useWithMobileStyle(styles.mobile)
+
+  return (
+    <>
+      <ExplainerTile className={wm(styles.explainerTile)} />
+      <WalletModal />
+      <div className={wm(styles.cryptoContentContainer)}>
+        <BalanceTile className={wm(styles.balanceTile)} />
```
jowlee, line 31: I see how the func `useWithMobileStyle` works, so the `.mobile` classname is used with every `wm`  instance if it is mobile. 
What are your thoughts on using just a conditional `.mobile` only at the top level and each className instead of having .mobile in it, just uses the .mobile selector from the parent in the css file.

src/containers/audio-rewards-page/RewardsTile.module.css
```diff
@@ -0,0 +1,102 @@
+.rewardsTile {
+  background-color: var(--white);
+  margin-bottom: 32px;
+  padding: 42px 28px 32px;
+}
+.title {
+  font-weight: 800;
```
jowlee, line 7: 700

src/containers/audio-rewards-page/Tiers.tsx
```diff
@@ -127,9 +120,15 @@ export const Tier = ({
...
-      className={cn(styles.tierContainerWrapper, {
-        [styles.tierContainerActive]: isActive
-      })}
+      className={cn(
+        styles.tierContainerWrapper,
+        {
+          [styles.tierContainerActive]: isActive
+        },
+        {
+          [styles.compact]: isCompact
```
jowlee, line 129: You can make this one object instead of two.

src/containers/audio-rewards-page/Tiles.tsx
```diff
@@ -49,41 +32,16 @@ export const Tile: React.FC<TileProps> = ({ className, children }) => {
...
 
   return (
-    <Tile
-      className={cn([
-        styles.claimTile,
-        { [styles.claimCollapsed]: hasNoClaim },
-        className
-      ])}
-    >
+    <Tile className={cn([styles.balanceTile, className])}>
```
jowlee, line 39: nbd, but should we be consistent w/ how we pass in styles for classnames - I know that their api allows this though.

jowlee (approved):
Looks good! Thanks for cleaning a lot of this up.
Am getting 
`Attempted import error: 'IconCrown' is not exported from '@audius/stems'.` when trying to run though. might need stems publish and update

src/containers/audio-rewards-page/AudioRewardsPage.tsx
piazzatron: Discussed on slack - I like the top level .mobile, will try to do that going forward when possible

PR #298, opened 2021-03-24
Improve icon popup

raymondjacobson:
### Description
- Fix some weird styling on svg in NavBanner (purple bleeding over edges, etc.)
- Prevent popup from scrolling with content
- Add computed position so that position doesn't overflow window

### Dragons

### How Has This Been Tested?

Locally against staging -- image selection popup as well as navbanner sorter popup


src/components/general/IconPopup.tsx
```diff
@@ -86,7 +86,7 @@ const IconPopup: React.FC<IconPopupProps> = ({
               className={styles.item}
               onClick={handleMenuItemClick(item)}
             >
-              <span className={styles.icon}>{item.icon}</span>
+              {item.icon && <span className={styles.icon}>{item.icon}</span>}
```
sddioulde, line 89: nice catch

src/components/general/Popup.js
```diff
@@ -25,68 +85,102 @@ const Popup = ({
...
+    wrapper,
+    placeholder,
+    triggerRef,
+    computedPosition,
+    setOriginalTopPosition
+  ])
+
+  // Callback invoked on each scroll. Uses original top position to scroll with content.
+  const watchScroll = useCallback(
+    (scrollParent, initialScrollPosition) => {
```
sddioulde, line 163: nbd, but I think this function can just take the scrollParent as argument, given that initialScrollPosition is just the scrollParent.scrollTop?

sddioulde (approved):

src/components/general/Popup.js
raymondjacobson: it's not actually -- initial position refers to the position when the popup first opened, whereas scrollParent.scrollTop gets the "current" scroll pos. Adding a comment

PR #299, opened 2021-03-25
[AUD-462] Throttle upload relays for collections

raymondjacobson:
### Description
Throttles the maximum number of concurrent /relay requests on collection uploads.
Screen Shot 2021-03-24 at 6 45 14 PM

As it stands today (in the above screen shot) we wait until all uploads for a collection happen and then we spam the relay endpoint. For 200 concurrent tracks, we would send 200x concurrency leading to too many tracks in the same block and discovery struggles to index.

After this change, we bundle the relays & track posts into groups and collate any errors.
Screen Shot 2021-03-24 at 7 05 52 PM

### Dragons

Touches core playlist/album upload flow

### How Has This Been Tested?

- Uploaded playlist of 28 tracks on staging
- Multi upload of 8 tracks (not collection -- should be unimpacted)
- Single track upload (not collection -- should be unimpacted)

### DIFF
--- a/src/containers/upload-page/store/sagas.js
+++ b/src/containers/upload-page/store/sagas.js
@@ -45,6 +45,7 @@ import { trackNewRemixEvent } from 'store/cache/tracks/sagas'
 import { reportSuccessAndFailureEvents } from './utils/sagaHelpers'
 
 const MAX_CONCURRENT_UPLOADS = 4
+const MAX_CONCURRENT_REGISTRATIONS = 4
 const MAX_CONCURRENT_TRACK_SIZE_BYTES = 40 /* MB */ * 1024 * 1024
 const UPLOAD_TIMEOUT_MILLIS = 60 /* min */ * 60 /* sec */ * 1000 /* ms */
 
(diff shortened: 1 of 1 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/rj-aud-462

src/containers/upload-page/store/sagas.js
```diff
@@ -526,10 +527,43 @@ export function* handleUploads({
...
+      for (
+        let i = 0;
+        i < sortedMetadata.length;
+        i += MAX_CONCURRENT_REGISTRATIONS
+      ) {
+        const concurrentMetadata = sortedMetadata.slice(
+          i,
+          i + MAX_CONCURRENT_REGISTRATIONS
+        )
+        const {
```
piazzatron, line 551: I think there's a small bug here - if we return an error in the current batch, we break the for loop before we append the returned track ids to the outer trackIds slice, which IIRC we need to have for reverting associations later on.

piazzatron (approved):
Looks good besides the one issue I commented 👍

dmanjunath (approved):
Incredible!

PR #304, opened 2021-03-26
Add collectibles tab to profile page

sddioulde:
### Description
Add collectibles tab to profile page

### Dragons

Nothing too dangerous. Maybe look out for useTabs and confirm that changes in some of the useEffect and useCallback dependencies are reasonable

### How Has This Been Tested?

tested locally on browsers and mobile:
- only users with tier silver and up can see collectibles tab
- on a profile page, if the wallet and associated wallets have related and visible image or video collectibles, there should be a collectibles tab, otherwise there is no collectibles tab unless the user has an eligible tier and has collectibles so that they can update their preferences.
- user can only edit collectibles preferences when on their own profile
- modal for editing of the preferences is only available in desktop
- saved preferences determine the order in which collectibles are displayed
- playing of videos only in modals when in desktop, and drawers when in mobile
- muting and unmuting videos


src/containers/collectibles/components/CollectiblesPage.module.css
```diff
@@ -0,0 +1,528 @@
...
+.modalMediaWrapper {
+  position: relative;
+  margin: 24px;
+  width: 342px;
+  height: 342px;
+  max-width: 512px;
+  max-height: 512px;
+}
+
+video {
```
jowlee, line 312: Since video is a native html element, I would with this within a selector ie. `.container video` just in case.

src/containers/collectibles/components/CollectiblesPage.tsx
```diff
@@ -0,0 +1,629 @@
...
+
+const VISIBLE_COLLECTIBLES_DROPPABLE_ID = 'visible-collectibles-droppable'
+
+const messages = {
+  title: 'COLLECTIBLES',
+  subtitlePrefix: 'A collection of NFT collectibles owned and created by '
+}
+
+// @ts-ignore
+const VisibleCollectible = props => {
```
jowlee, line 38: Would try to type unless there's something really hard about adding types for this 🙃

src/containers/collectibles/components/CollectiblesPage.tsx
```diff
@@ -0,0 +1,629 @@
...
+        <img
+          className={styles.editMedia}
+          src={imageUrl}
+          alt='Visible collectible thumbnail'
+        />
+      </div>
+      <div className={styles.editRowTitle}>{name}</div>
+      <div>
+        {isOwned ? (
+          <span className={cn(styles.owned, styles.editStamp)}>OWNED</span>
```
jowlee, line 65: Would put this text in `messages` as well as created

src/containers/collectibles/components/CollectiblesPage.tsx
```diff
@@ -0,0 +1,629 @@
...
+
+      setCollectiblesMetadata({
+        ...collectiblesMetadata,
+        ...newCollectiblesMap,
+        order: collectiblesMetadata.order.concat(
+          Object.keys(newCollectiblesMap)
+        )
+      })
+    }
+    // eslint-disable-next-line
```
jowlee, line 390: why disable?

src/containers/collectibles/components/CollectiblesPage.tsx
```diff
@@ -0,0 +1,629 @@
...
+            Visit audius.co from a desktop browser to hide and sort your NFT
+            collectibles.
+          </div>
+        </div>
+      </Drawer>
+    </div>
+  )
+}
+
+export default CollectiblesPage
```
jowlee, line 629: somewhat of a large file, could break up into multiple files - if you wanted - no need to change though

src/containers/collectibles/components/CollectiblesPage.tsx
```diff
@@ -0,0 +1,629 @@
...
+        [id]: collectiblesMetadata?.id ?? {},
+        order: (collectiblesMetadata?.order ?? []).filter(
+          tokenId => tokenId !== id
+        )
+      })
+    },
+    [setCollectiblesMetadata, collectiblesMetadata]
+  )
+
+  const onDragEnd = (result: any) => {
```
jowlee, line 435: Could you type if not too hard 🙃

src/containers/collectibles/components/CollectiblesPage.tsx
```diff
@@ -0,0 +1,629 @@
...
+    const visibleCollectibleKeySet = new Set(
+      getVisibleCollectibles().map(c => c.id)
+    )
+    return collectibleList.filter(c => !visibleCollectibleKeySet.has(c.id))
+  }, [getVisibleCollectibles, collectibleList])
+
+  return (
+    <div className={styles.collectiblesWrapper}>
+      <div className={styles.wrapper}>
+        <div className={cn(styles.header, { [styles.mobileHeader]: isMobile })}>
```
jowlee, line 482: @piazzatron made a helper for mobile styles that could be used. 
My personal preference would be a top level `.mobile` class next to `collectiblesWrapper` and then for these you won't need to add isMobile every time, but instead use a more specified selector ie. `.mobile .header`

src/containers/collectibles/components/CollectiblesPage.tsx
```diff
@@ -0,0 +1,629 @@
...
+
+      <Drawer
+        isOpen={showUseDesktopDrawer}
+        onClose={() => setShowUseDesktopDrawer(false)}
+      >
+        <div className={styles.editDrawer}>
+          <IconGradientCollectibles className={styles.editDrawerIcon} />
+          <div className={styles.editDrawerTitle}>Edit Collectibles</div>
+          <div className={styles.editDrawerContent}>
+            Visit audius.co from a desktop browser to hide and sort your NFT
```
jowlee, line 620: could you put this text in the messages var

src/containers/collectibles/components/types.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export enum CollectibleType {
+  IMAGE = 'IMAGE',
+  VIDEO = 'VIDEO'
+}
+
+export type Collectible = {
+  id: string
+  name: string | null
```
jowlee, line 13: Michael made a nice `Nullable<TYPE>` wrapper that we try to use instead of `TYPE | null`

src/containers/profile-page/components/desktop/ProfilePage.tsx
```diff
@@ -159,6 +165,8 @@ const ProfilePage = ({
   loadMoreUserFeed,
   loadMoreArtistTracks,
   openCreatePlaylistModal,
+  // Collectibles
```
jowlee, line 168: what is this comment for?

src/containers/profile-page/components/desktop/ProfilePage.tsx
```diff
@@ -239,6 +247,20 @@ const ProfilePage = ({
     record(make(Name.TRACK_UPLOAD_OPEN, { source: 'profile' }))
   }, [goToRoute, record])
 
+  const { tierNumber } = useSelectTierInfo(userId ?? 0)
```
jowlee, line 250: I feel like this shouldn't default to 0 when not defined?

jowlee (approved):

src/containers/collectibles/components/CollectiblesPage.tsx
sddioulde: there was something annoying about it, then I didn't circle back. good catch

src/containers/collectibles/components/CollectiblesPage.tsx
sddioulde: let's chat about it :)

src/containers/collectibles/components/CollectiblesPage.tsx
sddioulde: yeah that's fair. it has like 5 components, could just move those

src/containers/collectibles/components/CollectiblesPage.tsx
sddioulde: yes sir!

src/containers/collectibles/components/types.ts
sddioulde: cool

src/containers/profile-page/components/desktop/ProfilePage.tsx
sddioulde: lol. nothing. part of my original thinking to avoid confusing myself. will remove

src/containers/profile-page/components/desktop/ProfilePage.tsx
sddioulde: followed an example from the codebase `Tiers.tsx lines 199 and 200`. I believe 0 will just end up giving you '0' for balance anyway. but yeah I agree it's not ideal, can circle back on it

audius-infra:
Preview this change https://frontend-demo.audius.co/sd-feature-collectibles

PR #319, opened 2021-03-30
Tiles connect wallets

jowlee:
### Description
Adds the Tiles.tsx and Tiles.module.css back to the audius-rewards-page from what is on master
Adds the new styles from multi-wallet support to the balance and wallet actions tiles in the Tiles file.
Adds a drawer for mobile clicking of connect wallet.

### Dragons
none

### How Has This Been Tested?
Ran locally


audius-infra:
Preview this change https://frontend-demo.audius.co/jowlee-tiles-connect-wallets

src/containers/mobile-connect-wallets-drawer/store/slice.ts
```diff
@@ -0,0 +1,26 @@
+import { createSlice } from '@reduxjs/toolkit'
```
piazzatron, line 1: We can use our modals slice & hooks for this rather than creating an entirely new slice :)

src/containers/mobile-connect-wallets-drawer/MobileConnectWalletsDrawer.tsx
```diff
@@ -0,0 +1,35 @@
...
+import styles from './MobileConnectWalletsDrawer.module.css'
+
+const messages = {
+  title: 'Connect Additional Wallets With Your Account',
+  visit: 'Visit audius.co from a desktop browser',
+  reason:
+    'Show off your NFT Collectibles and flaunt your $AUDIO with a VIP badge on your profile.'
+}
+
+const MobileConnectWalletsDrawer = ({ onClose }: { onClose: () => void }) => {
```
piazzatron, line 17: just a stylistic thing, for these I've been creating the onClose action creator inside the component rather than passing it by props, since we're already connected to the store to get the visibility state. how do you feel about that vs passing it in?

piazzatron (approved):

piazzatron:
LGTM - just sanity checking the tiles code + CSS was pulled from latest master, yea?

src/containers/mobile-connect-wallets-drawer/MobileConnectWalletsDrawer.tsx
jowlee: I could go either way. In this case the parent has to open the modal which is pretty similar to the close function so I don't necessary think it's cleaner either way.

PR #326, opened 2021-03-31
Route - authenticated 

jowlee:
### Description
Created a hoot to redirect the user if not logged in

### Dragons
none

### How Has This Been Tested?
Ran it locally - in incognito when not signed in load the /audio route and watch it redirect to the login page

### DIFF
--- a/src/containers/audio-rewards-page/AudioRewardsPage.tsx
+++ b/src/containers/audio-rewards-page/AudioRewardsPage.tsx
@@ -12,12 +12,13 @@ import RewardsTile from './RewardsTile'
 import { isMobile } from 'utils/clientUtil'
 import { useMobileHeader } from 'components/general/header/mobile/hooks'
 import MobilePageContainer from 'components/general/MobilePageContainer'
-import { AUDIO_PAGE, BASE_URL } from 'utils/route'
+import { AUDIO_PAGE, BASE_URL, TRENDING_PAGE } from 'utils/route'
 import { useWithMobileStyle } from 'hooks/useWithMobileStyle'
 import NavContext, {
   LeftPreset,
   RightPreset
 } from 'containers/nav/store/context'
+import { useRequiresAccount } from 'hooks/useRequiresAccount'
 
 export const messages = {
   title: '$AUDIO & Rewards',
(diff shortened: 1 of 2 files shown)


src/containers/audio-rewards-page/AudioRewardsPage.tsx
```diff
@@ -26,7 +27,7 @@ export const messages = {
 
 export const RewardsContent = () => {
   const wm = useWithMobileStyle(styles.mobile)
-
+  useRequiresAccount(TRENDING_PAGE)
```
piazzatron, line 30: Pretty sure this page doesn't require an account? It's live right now without this...

src/hooks/useRequiresAccount.ts
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Checks that a user is signed in, else redirects to the sign in
+ * page with a banner saying - you need an account to do that
+ */
+export const useRequiresAccount = (route?: string) => {
+  const account = useSelector(getAccountUser)
+  const dispatch = useDispatch()
+  useEffect(() => {
+    if (!account) {
+      if (route) dispatch(updateRouteOnExit(route))
```
piazzatron, line 21: should this be an if / else?

src/containers/audio-rewards-page/AudioRewardsPage.tsx
jowlee: If will not error, but this page does show account specific information like your balance which may be confusing to see if you're not logged in. Thoughts @raymondjacobson - maybe we just trigger redirect on click send/receive/manage wallets similar to when you click favorite track when not logged in.

src/hooks/useRequiresAccount.ts
jowlee: This should remain an if statement. This just updates the route to go to when exiting the signup page. 
This function mimics the sagaHelper. https://github.com/AudiusProject/audius-client/blob/a175e0f87d962f7e1d07483339bf6caad1171cb0/src/utils/sagaHelpers.js#L111-L121

src/containers/audio-rewards-page/AudioRewardsPage.tsx
piazzatron: since we're already live with the rewards page for unauthed users, this seems like a product question to me. let's rope in the team

jowlee:
Closed in favor of #327 which is on top of master

PR #333, opened 2021-04-02
Preload wallet web3 providers

jowlee:
### Description
Lazy load in the wallet web3 providers & preload when the manage wallet modal is opened

### Dragons
none

### How Has This Been Tested?
Ran locally and saw the separate chunk.


src/store/token-dashboard/sagas.ts
```diff
@@ -417,6 +421,11 @@ function* watchForDiscordCode() {
   yield put(setDiscordCode({ code: appended }))
 }
 
+function* preloadWalletProviders() {
+  yield loadWalletConnect()
```
raymondjacobson, line 425: do you think this would be better to just do on the /audio page instead of in the modal?

raymondjacobson (approved):

audius-infra:
Preview this change https://frontend-demo.audius.co/jowlee-preload-web3-providers

src/store/token-dashboard/sagas.ts
jowlee: I don't think it makes too much of a difference - might be slightly better on the /audio page to help reduce potential lag (depending on internet speed) w/ the web3modal overlay appearing. Will move though.

PR #339, opened 2021-04-06
Fix repeat and shuffle

jowlee:
### Description
On native mobile shuffle and repeat were broken 🙃 
This should fix both. 
* Shuffle was never implemented
* Repeat broken when we switched to using redux toolkit and forgot to destruct the action's content from the new `payload` body of the action vs getting the value directly 

### Dragons
Yes, this is dangerous as it touched playing on native mobile. 
Please take some time to play around and test it 🙏 

### How Has This Been Tested?
I ran the client locally against a simulator and tested w/ a playlist to ensure it was shuffling and repeating correctly.

### DIFF
--- a/src/services/native-mobile-interface/queue.ts
+++ b/src/services/native-mobile-interface/queue.ts
@@ -17,13 +17,31 @@ type TrackInfo = {
 export type Tracks = TrackInfo[]
 
 export class PersistQueueMessage extends NativeMobileMessage {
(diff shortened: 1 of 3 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/jowlee-mobile-fix-repeat-shuffle

raymondjacobson (approved):

raymondjacobson:
let's maybe wait on this until after our thurs release

PR #341, opened 2021-04-06
Fix associated wallet override

jowlee:
### Description
The user metadata w/ associated wallet was being overriden with null when not updating the associated wallets field in the metadata directly. This was due to an oversite in fetching the user's ipfs metadata before re-writing to it. The associated wallet metadata (ie. signature) is not passed as part of the DP response, so we must request the user's ipfs metadata to update it before posting a new version.  

### Dragons
none

### How Has This Been Tested?

Ran locally w/ users that have associated wallets and w/out associated wallet to ensure the flow works are expected. 
Also ran updates to the associated wallets metadata to ensure that those were also updated correctly and not over-riden by later updates.

### DIFF
--- a/src/services/AudiusBackend.js
+++ b/src/services/AudiusBackend.js
@@ -34,6 +34,7 @@ import {
 } from './remote-config/Provider'
 import { monitoringCallbacks } from './serviceMonitoring'
 import { isElectron } from 'utils/clientUtil'
+import { getCreatorNodeIPFSGateways } from 'utils/gatewayUtil'
 
 export const IDENTITY_SERVICE = process.env.REACT_APP_IDENTITY_SERVICE
 export const USER_NODE = process.env.REACT_APP_USER_NODE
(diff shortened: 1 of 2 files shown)


src/services/AudiusBackend.js
```diff
@@ -936,8 +937,31 @@ class AudiusBackend {
     }
   }
 
+  static async fetchUserAssociatedWallets(user) {
```
raymondjacobson, line 940: do you want token dashboard sagas to re-use this function too?

raymondjacobson (approved):

src/services/AudiusBackend.js
jowlee: yeah can do

audius-infra:
Preview this change https://frontend-demo.audius.co/jowlee-fix-associated-wallet-metadata

raymondjacobson (approved):

PR #344, opened 2021-04-07
Collectibles touchups

raymondjacobson:
### Description

Commits should be self-descriptive & reviewed 1 by 1. Requires somewhat thorough review/testing. The 2nd commit (profile page deep links) is a bit hefty, but allows us to link to things like
audius.co/rayjacobson/collectibles
but not have that take up a full history state (so the back button will not switch tabs).

Demo on ray.audius.co (http://ray.audius.co/rayjacobson/collectibles)

### Dragons

Touches useTabs (minorly, but touches it)

### How Has This Been Tested?

Locally against prod, web + mobile

### DIFF
--- a/src/containers/profile-page/components/desktop/ProfilePage.tsx
+++ b/src/containers/profile-page/components/desktop/ProfilePage.tsx
@@ -442,9 +442,11 @@ const ProfilePage = ({
     ]
 
     if (
-      profileHasCollectiblesTierRequirement &&
-      (profileHasVisibleImageOrVideoCollectibles ||
-        (profileHasCollectibles && isUserOnTheirProfile))
+      // `has_collectibles` is a shortcut that is only true iff the user has a modified collectibles state
+      profile?.has_collectibles ||
+      (profileHasCollectiblesTierRequirement &&
+        (profileHasVisibleImageOrVideoCollectibles ||
+          (profileHasCollectibles && isUserOnTheirProfile)))
     ) {
       headers.push({
         icon: <IconCollectibles />,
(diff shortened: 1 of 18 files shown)


raymondjacobson:
@piazzatron could use your eye on commit 5ec4269 (Add profile page deep links). Had to touch useTabs to address some difference in desktop vs. mobile behavior.

audius-infra:
Preview this change https://frontend-demo.audius.co/rj-collectibles-touchups

src/containers/profile-page/components/desktop/ProfilePage.tsx
```diff
@@ -442,9 +442,11 @@ const ProfilePage = ({
     ]
 
     if (
-      profileHasCollectiblesTierRequirement &&
-      (profileHasVisibleImageOrVideoCollectibles ||
-        (profileHasCollectibles && isUserOnTheirProfile))
+      // `has_collectibles` is a shortcut that is only true iff the user has a modified collectibles state
+      profile?.has_collectibles ||
```
sddioulde, line 446: Not sure if this is correct. This seems like the collectibles tab would be displayed in the following scenarios even though it shouldn't:
- profile used to have a tier silver or higher, but not has downgraded tiers
OR
- profile has collectibles but all of them are hidden and user is not on their profile

piazzatron (commented):
Found two bugs, one not so bad one pretty bad:

- (Bad) Try switching from an artist page to a non-artist page, when you're on tracks/albums on the artist page. Crash :(
- (Not so bad) If you're on a profile page and switch to another page, you lose the URL but we preserve the tab. So michael/playlists -> forrest still shows us on the playlist tab, but the URL just says /forrest instead of /forrest/playlists.

Both of these seem to stem from useTabs state persisting through multiple profile pages. I wonder if we could key this smartly we nuke internal state when we switch pages?

raymondjacobson:
> Found two bugs, one not so bad one pretty bad:
> 
> * (Bad) Try switching from an artist page to a non-artist page, when you're on tracks/albums on the artist page. Crash :(
> * (Not so bad) If you're on a profile page and switch to another profile page, you lose the URL but we preserve the tab. So michael/playlists -> forrest still shows us on the playlist tab, but the URL just says /forrest instead of /forrest/playlists.
> 
> Both of these seem to stem from useTabs state persisting through multiple profile pages. I wonder if we could key this smartly to nuke internal state when we switch pages?

Fix was much more straight forward -- just unset activeTab within profile page provider. A key would be good, but i started doing it and breaking tons of stuff

sddioulde (approved):

PR #345, opened 2021-04-07
Hotfix diplo

raymondjacobson:
### Description
Adds a mechanism to stream by mp3 instead of HLS and exposes that option via optimizely.

### Dragons

### How Has This Been Tested?

### DIFF
--- a/src/audio/AudioStream.ts
+++ b/src/audio/AudioStream.ts
@@ -132,9 +132,9 @@ class AudioStream {
     this.errorRateLimiter = new Set()
   }
 
-  _initContext = () => {
+  _initContext = (shouldSkipAudioContext = false) => {
     this.audio.addEventListener('canplay', () => {
-      if (!this.audioCtx) {
+      if (!this.audioCtx && !shouldSkipAudioContext) {
         // Set up WebAudio API handles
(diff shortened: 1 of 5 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/hotfix-diplo

src/services/audius-api-client/AudiusAPIClient.ts
```diff
@@ -915,6 +915,17 @@ class AudiusAPIClient {
     console.debug('APIClient: Initialized')
   }
 
+  makeUrl = async (
```
jowlee, line 918: don't think this needs to be async

src/audio/AudioStream.ts
```diff
@@ -184,79 +184,91 @@ class AudioStream {
     onEnd: () => void,
     prefetchedSegments = [],
     gateways = [],
-    info = { title: '', artist: '' }
+    info = { id: '', title: '', artist: '' },
```
jowlee, line 187: You added id here, but I don't think it's used.

jowlee (approved):
Worked w/ diplo's track locally against prod for me.

PR #347, opened 2021-04-07
Close modal for top level mobile pages

jowlee:
### Description
Closes the top level pages on mobile if the user routes to another page. 

Fixes AUD-509

### Dragons

none

### How Has This Been Tested?

locally in the chrome browser w/ mobile view

### DIFF
--- a/src/hooks/useHasChangedRoute.ts
+++ b/src/hooks/useHasChangedRoute.ts
@@ -1,5 +1,6 @@
-import { useEffect, useState } from 'react'
+import { useEffect } from 'react'
 import { useLocation } from 'react-router-dom'
+import useInstanceVar from 'hooks/useInstanceVar'
 
 /**
  * Custom hook that fires a callback when the route changes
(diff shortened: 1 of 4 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/jowlee-edit-mobile-playlist

src/hooks/useHasChangedRoute.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Custom hook that fires a callback when the route changes
+ * Normally, you can use `useEffect` on the pathname from useLocation, but this
+ * hook ignores the first call of useEffect
+ *
+ * @param {function} onRouteChange the callback fired when the route changes
+ */
+const useHasChangedRoute = (onRouteChange: () => void) => {
+  const { pathname } = useLocation()
+  const [currentRoute, setCurrentRoute] = useState(pathname)
```
piazzatron, line 13: Wonder if this could be an instance var rather than state to prevent a re-render?

piazzatron (approved):
Love the new hook 😍
wanna add it to our hooks readme?

src/hooks/useHasChangedRoute.ts
jowlee: Yeah, shouldn't make a difference - might be slightly better w/ instance var in that it would re-run unnecessarily. 
I initially wrote it w/ useRef - but changed it. Can use instanceVar.

PR #357, opened 2021-04-08
GIF fixes

raymondjacobson:
### Description
1. try preview / thumbnails first just in case they are there--they will be smaller
2. try Range request uber hack (works most of the time) 
3. write faster gif worker to be used instead of resize worker (we just need to read into a jpeg which JIMP will do). also use same model as average rgb worker since that keeps a persistent worker up & running

4*. additional change to `getCollectibleImage` to instead of preloading, use fetch and use Content-Type to determine gif. There are currently some gifs that will not be treated as such because there is no extension in the URL. This stops them from autoplaying in the collectible detail -- but it also doesn't show the play button. We should do this check earlier in the stack, per our other convo today about restructuring when we determine gif.

### Dragons

### How Has This Been Tested?

ray.audius.co
/matthewchaim/collectibles
/coopahtroopa/collectibles
/rayjacobson/collectibles
/rac/collectibles

### DIFF
--- a/src/containers/collectibles/helpers.ts
+++ b/src/containers/collectibles/helpers.ts
@@ -3,8 +3,7 @@ import {
   Collectible,
   CollectibleType
 } from 'containers/collectibles/components/types'
-import { resizeImage } from 'utils/imageProcessingUtil'
-import { preload } from 'utils/image'
+import { gifPreview } from 'utils/imageProcessingUtil'
 
 const OPENSEA_AUDIO_EXTENSIONS = ['mp3', 'wav', 'oga']
 const NULL_ADDRESS = '0x0000000000000000000000000000000000000000'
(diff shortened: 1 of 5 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/rj-gif-fix

samid737:
the `CollectiblesPage_detailsDescription__jtdtC` div class does not allow text selection when trying to copy the user links in the description of the collectible.

src/containers/collectibles/helpers.ts
```diff
@@ -111,20 +123,23 @@ export const getCollectibleImage = async (collectible: Collectible) => {
...
   if (imageOriginalUrl?.endsWith('.gif')) {
     return await getFrameFromGif(imageOriginalUrl, name || '')
   }
-  if (imagePreviewUrl?.endsWith('.gif')) {
-    return await getFrameFromGif(imagePreviewUrl, name || '')
-  }
-  if (imageThumbnailUrl?.endsWith('.gif')) {
-    return await getFrameFromGif(imageThumbnailUrl, name || '')
-  }
   if (imageUrl) {
```
sddioulde, line 138: maybe we should try other images in case there is no image_url i.e. thumbnail, preview, original?

sddioulde (approved):

PR #374, opened 2021-04-12
Fix iOS safari video thumbnail and limit video support

sddioulde:
### Description
Fix iOS safari video thumbnail and limit video support

### Dragons

### How Has This Been Tested?

Tested locally

### DIFF
--- a/src/containers/collectibles/helpers.ts
+++ b/src/containers/collectibles/helpers.ts
@@ -9,6 +9,7 @@ import { gifPreview } from 'utils/imageProcessingUtil'
  * extensions based on OpenSea metadata standards
  * https://docs.opensea.io/docs/metadata-standards
  */
+const OPENSEA_AUDIO_EXTENSIONS = ['mp3', 'wav', 'oga']
 const OPENSEA_VIDEO_EXTENSIONS = [
   'gltf',
   'glb',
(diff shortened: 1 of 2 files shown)


src/containers/collectibles/helpers.ts
```diff
@@ -9,16 +9,19 @@ import { gifPreview } from 'utils/imageProcessingUtil'
...
   'mp4',
   'm4v',
   'ogv',
-  'ogg'
+  'ogg',
+  'mov'
 ]
-const OPENSEA_AUDIO_EXTENSIONS = ['mp3', 'wav', 'oga']
+
+const SUPPORTED_VIDEO_EXTENSIONS = ['webm', 'mp4', 'ogv', 'ogg', 'mov']
```
raymondjacobson, line 24: Just read a bit more into this
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video

it sounds like not all browsers support the same things and you can use multiple <source> elements to fallback between formats. Since the API results are so hairy, i think what we have here is good

src/containers/collectibles/helpers.ts
```diff
@@ -88,13 +82,7 @@ export const assetToCollectible = async (
   let gifUrl = null
```
raymondjacobson, line 82: Could use a larger doc-string style comment on `assetToCollectible` that explains our algo here, basically one that explains in a bit more detail why we check extensions and then check mime type (for the case where there is no extension)

raymondjacobson (approved):

src/containers/collectibles/helpers.ts
sddioulde: yes!

PR #382, opened 2021-04-16
Display user badges in search and close modal on click outside

sddioulde:
### Description
Display user badges in search and close modal on click outside

### Dragons

userId no longer being used in .../desktop/Card.tsx

### How Has This Been Tested?

Tested locally


raymondjacobson (approved):

src/components/card/desktop/Card.tsx
```diff
@@ -47,7 +47,6 @@ const cardSizeStyles = {
 type CardProps = {
   className?: string
   id: ID
-  userId: ID
```
jowlee, line 50: If we drop `userId` then we have to be sure that the `id` prop passed into card is always the userId because this passed into UserBadges and it expects an `userId`

src/containers/explore-page/components/desktop/CollectionsPage.tsx
```diff
@@ -83,7 +83,6 @@ const CollectionsPage = ({
         setDidLoad={setDidLoadPlaylist}
         key={playlist.playlist_id}
         id={playlist.playlist_id}
-        userId={playlist.playlist_owner_id}
```
jowlee, line 86: The card component passes the id prop to the UserBadges component now, so it won't retrieve the correct userId now

src/containers/saved-page/components/desktop/SavedPage.tsx
```diff
@@ -156,7 +156,6 @@ const SavedPage = ({
             setDidLoad={setDidLoadAlbums}
             key={album.playlist_id}
             id={album.playlist_id}
-            userId={album.playlist_owner_id}
```
jowlee, line 159: Same as prev comment

src/containers/profile-page/components/desktop/ProfilePage.tsx
```diff
@@ -493,7 +491,6 @@ const ProfilePage = ({
         key={index}
         size='medium'
         id={playlist.playlist_id}
-        userId={playlist.playlist_owner_id}
```
jowlee, line 496: If you look on the playlist tab of a user's profile with badges, it's now missing badges

src/components/card/desktop/Card.tsx
sddioulde: gotchu gotchu

PR #386, opened 2021-04-19
Include promise allsettled polyfill where it is missing

sddioulde:
Some clients do not have the Promise.allSettled function which causes some errors e.g. [here](https://sentry.io/organizations/audius/discover/audius-dapp:296df7dc3f2849bab953905b8d078681/?field=title&field=event.type&field=project&field=user.display&field=timestamp&name=All+Events&query=&sort=-timestamp&statsPeriod=24h&widths=-1&widths=-1&widths=-1&widths=-1&widths=-1) This is to mitigate that issue.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5831,15 +5831,93 @@
       }
     },
     "array.prototype.map": {
-      "version": "1.0.2",
-      "resolved": "https://registry.npmjs.org/array.prototype.map/-/array.prototype.map-1.0.2.tgz",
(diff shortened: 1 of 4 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/sd-all-settled

raymondjacobson (approved):
nice! change seems reasonable. let's make sure to link the sentry issue in the PR somewhere

PR #392, opened 2021-04-20
Fix verified not showing in autocomplete

raymondjacobson:
### Description
Adds a controlled variant to UserBadges since autocomplete results are not cached

### Dragons

### How Has This Been Tested?

Screen Shot 2021-04-20 at 11 12 45 AM
Screen Shot 2021-04-20 at 11 12 57 AM

### DIFF
--- a/src/components/search/SearchBarResult.js
+++ b/src/components/search/SearchBarResult.js
@@ -59,7 +59,8 @@ const SearchBarResult = memo(props => {
     imageMultihash,
     creatorNodeEndpoint,
     size,
-    defaultImage
+    defaultImage,
+    isVerifiedUser
   } = props
   const isUser = kind === Kind.USERS
 
@@ -85,6 +86,7 @@ const SearchBarResult = memo(props => {
               className={styles.verified}
               userId={userId}
               badgeSize={10}
+              isVerifiedOverride={isVerifiedUser}
             />
           )}
         </span>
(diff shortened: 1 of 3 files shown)


piazzatron (approved):
Slightly nitty but maybe we want `isVerified ?? isVerifiedOverride` instead of `||`. Reason being, with the way we have things rn if isVerified is true and we want to override it to false, it won't work. But if isVerified is false and we want to override it to true, it works, which is inconsistent

raymondjacobson:
> Slightly nitty but maybe we want `isVerified ?? isVerifiedOverride` instead of `||`. Reason being, with the way we have things rn if isVerified is true and we want to override it to false, it won't work. But if isVerified is false and we want to override it to true, it works, which is inconsistent

niceee

piazzatron:
> > Slightly nitty but maybe we want `isVerified ?? isVerifiedOverride` instead of `||`. Reason being, with the way we have things rn if isVerified is true and we want to override it to false, it won't work. But if isVerified is false and we want to override it to true, it works, which is inconsistent
> 
> niceee

or even `isVerifiedOverride ?? isVerified` if we always want the override to take precedence

raymondjacobson:
> > > Slightly nitty but maybe we want `isVerified ?? isVerifiedOverride` instead of `||`. Reason being, with the way we have things rn if isVerified is true and we want to override it to false, it won't work. But if isVerified is false and we want to override it to true, it works, which is inconsistent
> > 
> > 
> > niceee
> 
> or even `isVerifiedOverride ?? isVerified` if we always want the override to take precedence

Yeah it needs to be that and default needs to be undefined not false, sounds good tho 👍

PR #404, opened 2021-05-03
Shim Object.entries

jowlee:
### Description
Shim the function `Object.entries`
Note that the function is widely supported - https://caniuse.com/object-entries
But there are a lot of error in sentry for `Object.entries is not a function` which seemingly come from react spring, but specifically this function. This is my best guess for the fix.

### Dragons
none

### How Has This Been Tested?
I ran the function and it has no affect even if Object.entries is defined.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17816,14 +17816,92 @@
       }
     },
     "object.entries": {
-      "version": "1.1.1",
-      "resolved": "https://registry.npmjs.org/object.entries/-/object.entries-1.1.1.tgz",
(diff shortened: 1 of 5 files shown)


raymondjacobson (approved):

raymondjacobson:
did react spring get upgraded by a patch version or something?

jowlee:
> did react spring get upgraded by a patch version or something?

We are using react-spring version 8.0.27 which is that latest patch for major version 8 & is 2 yrs old (The current major versions w/ updates is v9). So I don't think it's because of the package. 🤷‍♂️

PR #407, opened 2021-05-05
Fix various search issues

piazzatron:
### Description
Fixes a bunch of things:
- Adds badges in autocomplete search
- Removes %20 in search bar if you search for multiple words (pressing enter, full search)
- Stops sorting exact handle matches to top on client, since DP is smart now
- Fix bug where if you hover over an autocomplete option and then press arrow key down, two opts show selected (go see this one on prod)
- Fix bug where a 5px dropshadow from the header would steal focus as you scrolled through autocomplete results (a @forrestb find)

Had to break up the selectors for badge stuff a bit, but no logic changes.

### Dragons

Nah

### How Has This Been Tested?

Locally against DP1

### DIFF
--- a/src/components/general/Page.module.css
+++ b/src/components/general/Page.module.css
@@ -39,6 +39,8 @@
 
 .frosted {
   backdrop-filter: blur(10px);
+  position: relative;
+  z-index: 10;
 }
 
 .headerBoxShadow {
--- a/src/containers/search-bar/SearchBar.js
+++ b/src/containers/search-bar/SearchBar.js
@@ -60,11 +60,12 @@ class SearchBar extends Component {
       this.setState({ value: '' })
       return
     }
+    const decodedValue = decodeURIComponent(value)
 
     if (!this.isTagSearch() && fetch) {
-      this.props.fetchSearch(value)
+      this.props.fetchSearch(decodedValue)
     }
-    this.setState({ value })
+    this.setState({ value: decodedValue })
   }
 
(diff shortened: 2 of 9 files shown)


src/components/general/Page.module.css
```diff
@@ -39,6 +39,8 @@
 
 .frosted {
   backdrop-filter: blur(10px);
+  position: relative;
+  z-index: 10;
```
raymondjacobson, line 43: why did we need this?

audius-infra:
Preview this change https://frontend-demo.audius.co/piazz-client-search-improvements

src/containers/search-bar/SearchBar.js
```diff
@@ -60,11 +61,12 @@ class SearchBar extends Component {
       this.setState({ value: '' })
       return
     }
+    const decodedValue = decodeURIComponent(value)
```
raymondjacobson, line 64: we should make sure to test searches that intentionally include a % sign, etc. as well as tag search

raymondjacobson (approved):
let's merge tmw after release, but generally looks great!! would love to play around on staging w/ it

nice putting up with antd

src/components/general/Page.module.css
piazzatron: should have been called out in the commit message for this one. It was this crazy thing where we had a 5 pixel bar that stole focus as you hovered down autocomplete results

PR #409, opened 2021-05-05
Reduce sentry notification polling errors

jowlee:
### Description
Only report the errors where the server doesn't return a 200 status code
- This should stop error when the client in unable to make a request (ie. background)
Also check for an account before continuing the notification flow. 
Seems like there are cases where the user is not signed in but tries to fetch notifications?

### Dragons
none

### How Has This Been Tested?
ran locally and simulated no internet and a bad response to test that it hit the correct code path on each

### DIFF
--- a/src/containers/notification/store/sagas.ts
+++ b/src/containers/notification/store/sagas.ts
@@ -392,23 +392,38 @@ const checkIfNotifcationsChanged = (
 
 export function* getNotifications(isFirstFetch: boolean) {
   try {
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/jowlee-notif

audius-infra:
Preview this change https://frontend-demo.audius.co/jowlee-notif

raymondjacobson (approved):

PR #414, opened 2021-05-07
Add podcast skip buttons

raymondjacobson:
### Description
Move next/prev buttons to providers and add podcast skip buttons

### Dragons

Nah

### How Has This Been Tested?

* Locally against prod/staging
* Local mobile web + apps


src/containers/now-playing/NowPlaying.tsx
```diff
@@ -292,6 +294,15 @@ const NowPlaying = g(
     ])
 
     const onPrevious = () => {
+      if (track.genre === Genre.PODCASTS) {
+        const position = timing.position
+        const newPosition = position - SKIP_DURATION_SEC
+        if (newPosition > 0) {
```
sddioulde, line 300: should we use seek(Math.mix(0, newPosition)) and not do this check

src/containers/now-playing/NowPlaying.tsx
```diff
@@ -300,6 +311,17 @@ const NowPlaying = g(
       }
     }
 
+    const onNext = () => {
+      if (track.genre === Genre.PODCASTS) {
+        const newPosition = timing.position + SKIP_DURATION_SEC
+        if (newPosition < timing.duration) {
```
sddioulde, line 317: here too, should we use seek(Math.min(newPosition, timing.duration)) and not do this check

src/containers/play-bar/desktop/PlayBar.js
```diff
@@ -224,12 +226,46 @@ class PlayBar extends Component {
...
+      audio,
+      seek,
+      previous,
+      reset,
+      currentQueueItem: { track }
+    } = this.props
+    if (track.genre === Genre.PODCASTS) {
+      const position = audio.getPosition()
+      const newPosition = position - SKIP_DURATION_SEC
+      if (newPosition > 0) {
```
sddioulde, line 239: here too

src/containers/play-bar/desktop/PlayBar.js
```diff
@@ -224,12 +226,46 @@ class PlayBar extends Component {
...
+      audio,
+      seek,
+      next,
+      currentQueueItem: { track }
+    } = this.props
+    if (track.genre === Genre.PODCASTS) {
+      const duration = audio.getDuration()
+      const position = audio.getPosition()
+      const newPosition = position + SKIP_DURATION_SEC
+      if (newPosition < duration) {
```
sddioulde, line 263: and here

src/containers/now-playing/NowPlaying.tsx
raymondjacobson: subtle difference -- if newPosition is <= 0, instead, we want to not return and fall through, so i think this is best. thoughts?

raymondjacobson: nvm u right

audius-infra:
Preview this change https://frontend-demo.audius.co/rj-aud-588

sddioulde (approved):

PR #417, opened 2021-05-10
Confirm transactions using block has and block number

sddioulde:
### Description
This will make the confirmation of different actions such as repost, save, upload, signup, follow, etc. to use the same confirmation paradigm. That is, by using the transaction receipt block hash and block number and verifying that they have been indexed.

### Dragons

This touches many flows in the dapp and we want to make sure those still work as intended.

### How Has This Been Tested?

Spun up local libs with changes, and spun up local dapp and tested all the impacted flows.

### DIFF
--- a/src/store/cache/collections/sagas.js
+++ b/src/store/cache/collections/sagas.js
@@ -1,36 +1,31 @@
 import {
-  call,
-  put,
   all,
+  call,
+  fork,
+  put,
   select,
-  takeLatest,
   takeEvery,
-  fork
+  takeLatest
 } from 'redux-saga/effects'
-import { pick, some } from 'lodash'
 
 import * as collectionActions from 'store/cache/collections/actions'
 import * as confirmerActions from 'store/confirmer/actions'
 import * as cacheActions from 'store/cache/actions'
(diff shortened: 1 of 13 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/sd-confirmer

src/store/cache/collections/sagas.js
```diff
@@ -308,27 +304,26 @@ function* confirmEditPlaylist(playlistId, userId, formFields) {
...
+
+        const confirmed = yield call(confirmTransaction, blockHash, blockNumber)
+        if (!confirmed) {
+          throw new Error(
+            `Could not confirm playlist edition for playlist id ${playlistId}`
+          )
         }
-        return yield call(pollPlaylist, confirmedPlaylistId, userId, check)
+
+        return (yield call(AudiusBackend.getPlaylists, userId, [
```
raymondjacobson, line 324: this probably should use the API client like poll user was switched out for

raymondjacobson (approved):
lgtm! minus making sure libs is updated in this PR

PR #419, opened 2021-05-10
Make sure scrubber tracks podcast seeker

raymondjacobson:
### Description
Fixes bug in 15s podcast seeker where scrubber does not stay in sync (needs a new mediakey)

### Dragons

### How Has This Been Tested?

Web
Mobile web
Mobile
Mobile & skipping via lock screen

### DIFF
--- a/src/containers/now-playing/NowPlaying.tsx
+++ b/src/containers/now-playing/NowPlaying.tsx
@@ -298,6 +298,9 @@ const NowPlaying = g(
         const position = timing.position
         const newPosition = position - SKIP_DURATION_SEC
         seek(Math.max(0, newPosition))
+        // Update mediakey so scrubber updates
+        setTiming({ position: newPosition, duration: timing.duration })
+        setMediaKey(mediaKey => mediaKey + 1)
       } else {
         const shouldGoToPrevious = timing.position < RESTART_THRESHOLD_SEC
         if (shouldGoToPrevious) {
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/rj-fix-podcast-scrub

jowlee:
> > > Would ideally like something better than incrementing an int to force an update, but this should work.
> > 
> > 
> > thoughts on what would be a better pattern? we use this media key thing for similar purposes on mobile, but i'm not attached to it
> 
> We could use the same thing we use for tabs and have a callback fn that you invoke called scrubberRecalculator or something

This is for updating the scrubber location right? 
Shouldn't that just be updated by some audio duration state that's updated by clicking on podcast seek and not need the media key int?

raymondjacobson:
> > > > Would ideally like something better than incrementing an int to force an update, but this should work.
> > > 
> > > 
> > > thoughts on what would be a better pattern? we use this media key thing for similar purposes on mobile, but i'm not attached to it
> > 
> > 
> > We could use the same thing we use for tabs and have a callback fn that you invoke called scrubberRecalculator or something
> 
> This is for updating the scrubber location right?
> Shouldn't that just be updated by some audio duration state that's updated by clicking on podcast seek and not need the media key int?

Media key is the way in which the scrubber forces recalculation. It's not based on the current audio time since that doesn't update in state--the animation runs on its own accord

jowlee:
> > > > > Would ideally like something better than incrementing an int to force an update, but this should work.
> > > > 
> > > > 
> > > > thoughts on what would be a better pattern? we use this media key thing for similar purposes on mobile, but i'm not attached to it
> > > 
> > > 
> > > We could use the same thing we use for tabs and have a callback fn that you invoke called scrubberRecalculator or something
> > 
> > 
> > This is for updating the scrubber location right?
> > Shouldn't that just be updated by some audio duration state that's updated by clicking on podcast seek and not need the media key int?
> 
> Media key is the way in which the scrubber forces recalculation. It's not based on the current audio time since that doesn't update in state--the animation runs on its own accord

But isn't the timing State updated everywhere the media key state is updated, so that alone should update state?
Except in the onPrevious and onNext in desktop, but that could be changed?

PR #430, opened 2021-05-27
Add feature flag for playlist updates

sddioulde:
### Dragons

### How Has This Been Tested?

### DIFF
--- a/src/containers/saved-page/components/mobile/SavedPage.tsx
+++ b/src/containers/saved-page/components/mobile/SavedPage.tsx
@@ -1,10 +1,4 @@
-import React, {
-  ReactNode,
-  useCallback,
-  useEffect,
-  useContext,
-  useState
-} from 'react'
+import React, { ReactNode, useCallback, useEffect, useContext } from 'react'
 import { Button, ButtonType } from '@audius/stems'
 
 import useTabs from 'hooks/useTabs/useTabs'
@@ -37,8 +31,7 @@ import {
 import { HeaderContext } from 'components/general/header/mobile/HeaderContextProvider'
 import MobilePageContainer from 'components/general/MobilePageContainer'
 import Spin from 'antd/lib/spin'
-import { useFlag } from 'containers/remote-config/hooks'
-import { FeatureFlags, flagDefaults } from 'services/remote-config/FeatureFlags'
+import { useArePlaylistUpdatesEnabled } from 'containers/remote-config/hooks'
 
 const emptyTabMessages = {
   afterSaved: "Once you have, this is where you'll find them!",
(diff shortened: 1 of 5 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/sd-playlist-updates-ff

src/containers/saved-page/components/mobile/SavedPage.tsx
```diff
@@ -265,6 +273,16 @@ const PlaylistCardLineup = ({
   playlistUpdates: number[]
   updatePlaylistLastViewedAt: (playlistId: number) => void
 }) => {
+  const { isLoaded, isEnabled } = useFlag(FeatureFlags.PLAYLIST_UPDATES_ENABLED)
```
piazzatron, line 276: Don't think we need this whole song and dance - isEnabled *should* default to the default value if we're not loaded (double check me here), and I don't think we need a useEffect and useState, we should be able to just directly use isEnabled or possibly some combination like `const shouldShow = (isLoaded && isEnabled)`. But, test it. Sorry for the flag code still being a little rough, @raymondjacobson was just pointing out today that we may have some issues around default values with flags.

piazzatron (approved):
I think we'll have a much easier job pulling this out & cleaning things up if we can wrap the PLAYLIST_UPDATES_ENABLED bit in it's own hook somewhere, which we can later delete. I did something similar here if you want to check it out for reference:

https://github.com/AudiusProject/audius-client/commit/95d7c3c20451de4c3e0cad757111a136e4bfb6b2#diff-6d5832d310c0e5850472a8b133180fd5d9ca5041855844c6fd83c9e73cb1d390R100

Otherwise looks good. Approved pending comments.

src/containers/saved-page/components/mobile/SavedPage.tsx
raymondjacobson: yeah, i think @piazzatron is correct here -- all we need to do is check for isLoaded && isEnabled.

The one caveat around defaults is interesting. Just important to note that default isn't what's used, basically ever. Default only is what's used if there's an error loading the optimizely config, regardless of whether the flag is even defined in optimizely.

raymondjacobson:
> I think we'll have a much easier job pulling this out & cleaning things up if we can wrap the PLAYLIST_UPDATES_ENABLED bit in it's own hook somewhere, which we can later delete. I did something similar here if you want to check it out for reference:
> 
> [95d7c3c#diff-6d5832d310c0e5850472a8b133180fd5d9ca5041855844c6fd83c9e73cb1d390R100](https://github.com/AudiusProject/audius-client/commit/95d7c3c20451de4c3e0cad757111a136e4bfb6b2#diff-6d5832d310c0e5850472a8b133180fd5d9ca5041855844c6fd83c9e73cb1d390R100)
> 
> Otherwise looks good. Approved pending comments.

I don't feel strongly here. One of the use cases is not in a component, so that would have to deviate from the hook. Searching for the PLAYLIST_UPDATES_ENABLED key would find everything w/ this change as is

raymondjacobson (approved):

src/containers/saved-page/components/mobile/SavedPage.tsx
sddioulde: Yep you are correct sir! Ah so much simpler thanks

sddioulde: We don't even need to check for isLoaded. Just isEnabled

sddioulde: Hmm just saw your comment about defaults. I think we're still good here since we have the `flagDefaults` that we set that is used in `getFeatureEnabled`

raymondjacobson: Yeah -- i think we're good here. the only thing to note is that defaults should probably always be false and really dont have much meaning when they are true, we should probably change the others that are set to true to false, and then just make sure to delete the flag entirely after the feature is launched

piazzatron (approved):

PR #437, opened 2021-06-02
Update "accept" field on image uploads to allowable images

rickyrombo:
### Description

Certain image formats are not supported by the back end, so we should filter them out on the client.

Adds an explicit `accept` prop to React Dropzone (docs: https://react-dropzone.js.org/#!/Accepting%20specific%20file%20types ) so that invalid types aren't mistakenly uploaded.

### Dragons

- Higher coupling with the utils file (the cost of a single source of truth for file support)
- Defaulted the `accept` prop to `audio/*` when `type` is not `'image'` as the only types for the `ImageSelectionPopup` are `'track'`, `'stem'`, and `'image'` (and both track and image should be audio files). Previously, no such filter existed.

### How Has This Been Tested?

Tested manually locally by uploading an individual track.

### DIFF
--- a/src/components/image-selection/ImageSelectionButton.js
+++ b/src/components/image-selection/ImageSelectionButton.js
@@ -77,7 +77,7 @@ const ImageSelectionButton = ({
           <ReactDropzone
             onDrop={onSelect}
             className={styles.dropzone}
-            accept='image/*'
+            accept='image/png, image/jpeg'
           >
             <Button
               className={cn(styles.button, styles.noPopup, {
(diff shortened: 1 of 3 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/mp-bday-party-no-gifs

src/components/image-selection/ImageSelectionButton.js
```diff
@@ -77,7 +78,7 @@ const ImageSelectionButton = ({
           <ReactDropzone
             onDrop={onSelect}
             className={styles.dropzone}
-            accept='image/*'
+            accept={ALLOWED_IMAGE_FILE_TYPES.join(', ')}
```
raymondjacobson, line 81: you *could* export ALLOWED_IMAGE_FILE_TYPES.join(', ') from the imageProcessingUtil and that might be less error prone since this requires a comma and a space?

raymondjacobson (approved):
lgtm! nice solution

src/components/image-selection/ImageSelectionButton.js
rickyrombo: I like keeping it in array form and formatting where necessary - easier to understand why it's formatted that way

PR #439, opened 2021-06-04
Save shuffle and repeat states to local storage

rickyrombo:
### Description

Save the user's shuffle/repeat states to local storage to persist it on page reload.

### Dragons

### How Has This Been Tested?

Manually tested on Microsoft Edge Desktop:
1) Toggle Shuffle/Repeat settings
2) Refresh page
3) Observe settings persist

1) Open local storage in devtools
2) Set relevant keys to 1 (for the "ANIMATE_OFF_ALL" etc keys)
3) Observe the settings change to 2

### DIFF
--- a/src/components/play-bar/repeat-button/RepeatButton.js
+++ b/src/components/play-bar/repeat-button/RepeatButton.js
@@ -14,11 +14,23 @@ const RepeatStates = Object.freeze({
   ANIMATE_SINGLE_OFF: 5
 })
 
+const REPEAT_STATE_LS_KEY = 'repeatState'
+const getRepeatState = defaultState => {
+  const localStorageRepeatState = window.localStorage.getItem(
+    REPEAT_STATE_LS_KEY
+  )
+  if (localStorageRepeatState === null) {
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-persist-player-settings

piazzatron (approved):
Looks great! Only thing that jumps out to me is that `getRepeatState` and `getShuffleState` are essentially the same function, with different LS keys. I know we didn't want to do a whole local storage refactor, but maybe we could pull this out into something like a `getLocalStorageWithDefault` util which accepts both LS key + default val?

rickyrombo:
> Looks great! Only thing that jumps out to me is that `getRepeatState` and `getShuffleState` are essentially the same function, with different LS keys. I know we didn't want to do a whole local storage refactor, but maybe we could pull this out into something like a `getLocalStorageWithDefault` util which accepts both LS key + default val?

Yeah this definitely would be more DRY but I might save that for the full refactor, if for nothing else than colocation of the function definition to its usage

PR #444, opened 2021-06-08
Fix reposts tab content ordering/pagination

raymondjacobson:
### Description
There were a couple issues with the chunk of code in the reposts tab that added "unconfirmed" reposted content in (https://github.com/AudiusProject/audius-client/commit/368abbc67453083e245fbd42b09c592bcf2d65e4).

1. The logic checks anything that's in the cache and only dedupes against what is currently coming back in the lineup pagination. If you visited some track pages/trending/etc. where you had tracks that were reposted, the existing code basically ends up shoving all of that content into your reposts tab (even if you had reposted it months ago)
2. The logic runs on every pagination so you'd get the same content repeated many times at the end of each page
3. The logic only accounted for tracks, not collections

### Dragons

### How Has This Been Tested?

- Loaded my prod reposts tab in a clean state (no confirming actions) and paginated through. Made a lot more sense.
- Ran against staging and made sure that having unconfirmed track & collection reposts do indeed show up in the tab

### DIFF
--- a/src/containers/profile-page/store/lineups/feed/sagas.js
+++ b/src/containers/profile-page/store/lineups/feed/sagas.js
@@ -13,12 +13,16 @@ import { LineupSagas } from 'store/lineup/sagas'
 import { getUserId } from 'store/account/selectors'
 import { retrieveUserReposts } from './retrieveUserReposts'
 import { getTracks } from 'store/cache/tracks/selectors'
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/rj-something-wrong-with-reposts

piazzatron (approved):
Nice find, easy to follow with the comments

PR #447, opened 2021-06-11
Add log to check whether we hit catch block of createPlaylist

sddioulde:
### Description

This log should help us see why we get COLLECTION_CREATE_PLAYLIST_NO_ID errors in sentry. We currently assume that we never hit this code path. This log is to double-check this assumption.

### Dragons

### How Has This Been Tested?


piazzatron (approved):

audius-infra:
Preview this change https://frontend-demo.audius.co/sd-log-create-playlist-catch

PR #451, opened 2021-06-15
Fix playlist library bug when explore playlist is not found

raymondjacobson:
### Description
For some reason, while dog-fooding the playlist library feature, I ended up with an explore playlist "Under the Radar" which is not the canonical string. "Under The Radar" is. I looked at identity and I'm the only user with this (i must have been playing with this in prod before we launched explore).

https://discoveryprovider.audius.co/v1/full/users/7eP5n?user_id=7eP5n

I'm going to manually issue the write to chain to clear this from my subscribed playlists, but in general, I think this safety check would be good on the off chance that the playlist literally cannot be found given the id reference

### Dragons

### How Has This Been Tested?

Local against prod


audius-infra:
Preview this change https://frontend-demo.audius.co/rj-fix-playlist-library-render-bug

sddioulde (approved):

PR #455, opened 2021-06-17
Handle duplicates in playlist library

raymondjacobson:
### Description
Currently on prod, I do not believe you could end up in a situation with duplicates, but I did for my account because I was playing around with it on release.audius.co before using audius.co where it tried to re-"migrate" me over to the new library from the identity stored library. Since there are not db constraints around how playlist library is handled, I think it's best to add this safeguard.

On every reorder, with this change, we will just remove duplicates if somehow they happen to exist.

### Dragons

### How Has This Been Tested?

- Unit
- Locally against prod, repairing my duplicates

### DIFF
--- a/src/store/playlist-library/helpers.test.js
+++ b/src/store/playlist-library/helpers.test.js
@@ -3,6 +3,7 @@ import {
   findIndexInPlaylistLibrary,
   findInPlaylistLibrary,
   removeFromPlaylistLibrary,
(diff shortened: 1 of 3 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/rj-handle-duplicates-in-playlist-library

sddioulde (approved):

PR #456, opened 2021-06-18
Switch back to audius RPC endpoint for stage

raymondjacobson:
### Description
https://poa-gateway.staging.audius.co/ is healthy again and much more performant than the public sokol RPC endpoint.

### Dragons

n/a just staging env

### How Has This Been Tested?

Locally against stage
1. Reposted track while pointed at a discovery node & identity also on our RPC
2. Uploaded a track while pointed at creator node, discovery node & identity also on our RPC


hareeshnagaraj:
looks great! will let @sliptype approve

sliptype (approved):
In prod we have a fallback RPC endpoint, but I guess we don't really need one for stage

Looks good!

raymondjacobson:
> In prod we have a fallback RPC endpoint, but I guess we don't really need one for stage
> 
> Looks good!

No harm in having it & probably good to be consistent w/ prod. I'll add the fallback!

audius-infra:
Preview this change https://frontend-demo.audius.co/rj-repoint-poa-gateway

PR #460, opened 2021-06-24
AUD-651 Update styles for UserListModal to allow seeing the bottom of the list

Kyle-Shanks:
### Description

There was an error with styling where the bottom of the UserList in the UserListModal would not be visible. This updates the styles of the UserListModal to use flexbox to force the content to the the correct size within the modal. 

Should fix AUD-651

### Dragons

I think this will only affect these two modals (Reposts and Favorites), but there may be other modals that are affected

### How Has This Been Tested?

I opened the modal and it looked good. :chef-kiss:

### DIFF
--- a/src/containers/user-list-modal/components/UserListModal.module.css
+++ b/src/containers/user-list-modal/components/UserListModal.module.css
@@ -1,6 +1,8 @@
 .modalBody {
   max-width: 360px;
   width: 100%;
+  display: flex;
+  flex-direction: column;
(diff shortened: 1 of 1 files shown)


raymondjacobson (approved):
Amazing! I did a quick sanity check on https://frontend-demo.audius.co/kj-aud-651/ray and it looks good.

hopefully there aren't any unintended consequences. modals are annoying :)

PR #462, opened 2021-06-25
Retrigger useFlag when userId gets set

raymondjacobson:
### Description
userId gets set after the useFlag hook may be called. The useMemo prevents recomputation, so add the account selector in to the mix for USER_ID experiments

### Dragons

### How Has This Been Tested?

- Locally against prod: verified user id experiment triggers (playlist library update dots). Verified mobile still works as well.
- Locally against stage: verified that track polling session id exp is working as intended


sddioulde (approved):

audius-infra:
Preview this change https://frontend-demo.audius.co/rj-fix-user-id-exp

PR #463, opened 2021-06-25
sk-aud-646 Overflow Menus & AddToPlaylistModal

sliptype:
### Description

This PR:

* Removes `CascadingMenu` in favor of using `Popup`
* Adds `PopupMenu` which accepts a render function for the popup trigger

image

* Renames `IconPopup` to `PopupMenuIconButton` and utilizes `PopupMenu` here
* Removes `hooks/useClickOutside.js` because it is not referenced (we are using the Stems version everywhere)
* Updates the styling of the "mobile" Toasts rendered via `ToastContext`, the idea is to migrate all toasts to use this going forward
* Adds an `AddToPlaylistModal` for the desktop add to playlist flow, since we no longer have nested menus

image

Notes:
* Didn't end up needing an protocol changes after all, after reviewing how the mobile AddToPlayist fetches playlist images

#### Future Scope
* Move `Popup`, `PopupMenu`, `PopupMenuIconButton`, and `IconButton` into stems
* Convert `Popup.js` to TypeScript
* Update menu item generation logic to be more declarative

### Dragons

Because `Popup` is still in JS, adding the `renderTrigger` prop could be breaking if not all usages of `Popup` are updated. To my knowledge they are all updated.

Most of this is styling and view logic so nothing _too_ dangerous.

### How Has This Been Tested?

Manually against staging. Found all instances of the components that were rendering `CascadingMenu` before (via `Menu` mostly) and verified that the styling and behavior were correct.


audius-infra:
Preview this change https://frontend-demo.audius.co/sk-aud-646

src/containers/menu/TrackMenu.tsx
```diff
@@ -249,11 +223,7 @@ const TrackMenu = (props: TrackMenuProps) => {
 
   const menu = getMenu()
 
-  return (
-    <CascadingMenu menu={menu} mount={props.mount} className={props.className}>
-      {props.children}
-    </CascadingMenu>
-  )
+  return props.children(menu.items)
```
raymondjacobson, line 226: I like the FaCC pattern choice here, but we should chat about how we feel as a team. afaik, this may be the first use?

raymondjacobson (commented):
Couple small UI things i noticed, but overall, this is super clean. Impressed with the surgery.

Screen Shot 2021-06-25 at 3 37 07 PM

* the popup stays fixed as the window scrolls here. it should be doable (like the sorter on the profile page tabs) to make it scroll w. the window.

Screen Shot 2021-06-25 at 2 47 44 PM

* I've got a fat playlist and this layout got a little funky. Probably need a min-width or something like that
* I think the underline at the top of the modal shouldn't extend the entire width (Doesn't in most of our other modals). I think some intentional modal-level padding would do the job

raymondjacobson:
> Couple small UI things i noticed, but overall, this is super clean. Impressed with the surgery.
> 
> Screen Shot 2021-06-25 at 3 37 07 PM
> 
> * the popup stays fixed as the window scrolls here. it should be doable (like the sorter on the profile page tabs) to make it scroll w. the window.
> 
> Screen Shot 2021-06-25 at 2 47 44 PM
> 
> * I've got a fat playlist and this layout got a little funky. Probably need a min-width or something like that
> * I think the underline at the top of the modal shouldn't extend the entire width (Doesn't in most of our other modals). I think some intentional modal-level padding would do the job

we do the "text-overflow ellipsis" a lot for this purpose too

src/components/actions-tab/ActionsTab.module.css
```diff
@@ -94,13 +104,13 @@
 .actionButton svg,
 .iconContainer svg,
 .iconKebabHorizontalWrapper svg {
-  transition: all .07s ease-in-out !important;
+  transition: all 0.07s ease-in-out !important;
```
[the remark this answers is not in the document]
sliptype: Prettier made this change so I think we can use that as our convention for the precision. Would be nice to do a css style guide though

src/components/general/PopupMenuIconButton.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
+
+import cn from 'classnames'
+
+import { PopupMenu, PopupMenuProps } from 'components/general/PopupMenu'
+
```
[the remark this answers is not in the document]
sliptype: Working on it currently!

src/containers/menu/TrackMenu.tsx
sliptype: There isn't much reason to have these as components, my plan is to change them into functions that accept `dispatch` in my next PR. I can't decide where in the directory structure functions like that should live though. `containers` doesn't really make sense, but neither does `utils` or `selectors`. Any thoughts?

src/components/toast/mobile/ToastLinkContent.module.css
```diff
@@ -0,0 +1,13 @@
+.text {
+  margin-right: 8px;
+}
+
+.link {
+  color: var(--static-white);
+  text-decoration: underline !important;
```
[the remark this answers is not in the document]
sliptype: I'm asking him about this, but just to confirm you are talking about the text-underline fading out on hover right?

sliptype:
> Couple small UI things i noticed, but overall, this is super clean. Impressed with the surgery.
> 
> Screen Shot 2021-06-25 at 3 37 07 PM
> 
> * the popup stays fixed as the window scrolls here. it should be doable (like the sorter on the profile page tabs) to make it scroll w. the window.
> 
> Screen Shot 2021-06-25 at 2 47 44 PM
> 
> * I've got a fat playlist and this layout got a little funky. Probably need a min-width or something like that
> * I think the underline at the top of the modal shouldn't extend the entire width (Doesn't in most of our other modals). I think some intentional modal-level padding would do the job

The scrolling issue was because we have some logic to find the closest scrolling parent that checks for `overflow-y` being auto/scroll. But setting `overflow-x: hidden` implicitly sets `overflow-y: auto`, so anywhere the popup had a parent with `overflow-x: hidden` closer than the actual scroll container it was broken.

The solution here was to set `overflow-y: hidden` explicitly (and check that this didn't break anything). This does update the `useTabs` containers so it could be fairly dangerous/far-reaching. But it looks like we are not relying on the `overflow-y` being auto to do the scrolling in those containers.

The text overflow is fixed on the playlists:

image

I matched the modal styling from the GenreSelectionModal which was also missing that header margin. But I've added it in both places now. This styling needs to get moved into stems also
image

src/components/toast/mobile/ToastLinkContent.module.css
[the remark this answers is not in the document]
sliptype: Done!

src/containers/menu/TrackMenu.tsx
raymondjacobson: maybe a `helpers` file or something like that near the component that uses it w/ a sub-folder containing both?

i think we should probably get rid of the containers vs. components structure over time, esp. with useDispatch, etc.

src/hooks/useTabs/TabStyles.module.css
```diff
@@ -99,9 +99,9 @@
   overflow: hidden;
 }
 
-.clippedOverflowX {
+.clippedOverflow {
   overflow-x: hidden;
-  overflow-y: visible;
+  overflow-y: hidden;
```
raymondjacobson, line 104: @piazzatron just in case you can think of anything this breaks. let's soak this change on stage for a while & go through weekend QA.
@sliptype might be worth testing mobile layout using the devtools thing

raymondjacobson (approved):

src/containers/menu/TrackMenu.tsx
sliptype: Ok I like that. And I agree, the container/component distinction doesn't provide much value when using the redux hooks

src/hooks/useTabs/TabStyles.module.css
sliptype: Yeah everything looks good on mobile too, I'll go ahead and merge and we can keep our eyes peeled in staging

PR #468, opened 2021-06-28
Add import/order eslint rule and fix import order project-wide

sliptype:
### Description

This PR adds the `import/order` eslint rule.

It uses the default order defined [here](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md#importorder-enforce-a-convention-in-module-import-order) with the exception of `react` which is placed at the top of the file.

The import order can be automatically fixed by running `npm run lint:fix`

### Dragons

Touches practically every file, but the app compiles and works properly.

Updates the package-lock.json `lockFileVersion` to `2`, which should be backwards compatible with npm 6 and below.

Adds a list of aliases in `.eslintrc` due to [this issue](https://github.com/benmosher/eslint-plugin-import/issues/807). These aliases will have to be kept updated with the actual directory structure, or we could generate them dynamically.

### How Has This Been Tested?

Manually, a breaking change here would most likely prevent app from compiling


audius-infra:
Preview this change https://frontend-demo.audius.co/sk-eslint-import-order

raymondjacobson (approved):
ama

PR #473, opened 2021-07-02
Fix zindex issue with notification menu

sliptype:
### Description

Fixes a small bug I noticed with the z index being too low for the notification overflow menu.

I originally lowered the z-index of the Menu (specific to tracks, users, collections, and notifications) to account for the frosted header on Feed and Trending pages, but this was breaking because the Notification menu is opened from another popup.

### Dragons

Are there any other places popup menus happen from modals/dialogs? 

### How Has This Been Tested?

### DIFF
--- a/src/containers/menu/Menu.tsx
+++ b/src/containers/menu/Menu.tsx
@@ -38,7 +38,7 @@ const Menu = (props: MenuProps) => {
       position='bottomRight'
       renderTrigger={props.children}
       popupClassName={className}
-      zIndex={12}
+      zIndex={menu.type === 'notification' ? 10000 : 12}
     />
   )
 


audius-infra:
Preview this change https://frontend-demo.audius.co/sk-fix-zindex-bug-popup-menu

raymondjacobson (approved):

src/containers/menu/Menu.tsx
```diff
@@ -38,7 +38,7 @@ const Menu = (props: MenuProps) => {
       position='bottomRight'
       renderTrigger={props.children}
       popupClassName={className}
-      zIndex={12}
+      zIndex={menu.type === 'notification' ? 10000 : 12}
```
raymondjacobson, line 41: i dont think there are any other spots, but maybe worth looking into making sure modals show up on top of this

sliptype: They do! The popup up menu closes whenever an item is clicked. But even in the case that it's open when a modal appears the modal shows over it

PR #480, opened 2021-07-08
Fix old accounts

raymondjacobson:
### Description
It seems like users no longer have profile_picture, but have a single non sizes CID in the _sizes prop.
https://discoveryprovider.audius.co/users?handle=emilyhou
I couldn't think of a better way to handle this unfortunately.

### Dragons

### How Has This Been Tested?

localhost:3002/emilyhou

### DIFF
--- a/src/services/AudiusBackend.js
+++ b/src/services/AudiusBackend.js
@@ -210,11 +210,26 @@ const fetchImageCID = async (cid, creatorNodeGateways = [], cache = true) => {
   await waitForLibsInit()
   // Else, race fetching of the image from all gateways & return the image url blob
   try {
-    const image = await audiusLibs.File.fetchCID(
-      cid,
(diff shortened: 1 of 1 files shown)


jowlee (approved):
This makes sense. In the case that both promises do not resolve (Or the single case w/out the '/' in the cid) it would be stuck at the promise.race forever - not sure if we want to catch that.

raymondjacobson:
> This makes sense. In the case that both promises do not resolve (Or the single case w/out the '/' in the cid) it would be stuck at the promise.race forever - not sure if we want to catch that.

Yeah... I thought of that. I figure it's probably ok since it results w/ the same behavior of just no image

audius-infra:
Preview this change https://frontend-demo.audius.co/rj-aud-773

PR #481, opened 2021-07-08
Fix playlist library staleness bug

raymondjacobson:
### Description
It's possible while fetching another profile to refetch your own account as part of the followers/followees/mutual followers and these endpoints don't respect (or support) the ?user_id route param fo the current user.
In those cases, we don't really need to refetch the current user, so filter it out.

### Dragons

### How Has This Been Tested?

Repro was from my account => profile /emilyhou and verified the bug no longer shows up with this change


audius-infra:
Preview this change https://frontend-demo.audius.co/rj-aud-660

sddioulde (approved):

PR #485, opened 2021-07-12
Add support for referrer and is_mobile_user events

raymondjacobson:
### Description

Add referrer and is_mobile_user events tracking

_referrer_ gets added on sign up and parsed off of the URL, e.g. audius.co/signup?ref=ray
_is_mobile_user_ gets added along w/ the existing flow that writes hasUserSignedInOnMobile to the identity service for notification email processing

### Dragons

Nope other than there should be a point of notification refactor where we remove the identity field and use the user metadata one. That would also include a discovery database migration for backward compat.

### How Has This Been Tested?

Ran against local backend services & tested setting `referrer` and `is_mobile_user`

### DIFF
--- a/src/containers/sign-on/SignOnProvider.tsx
+++ b/src/containers/sign-on/SignOnProvider.tsx
@@ -121,7 +121,13 @@ export class SignOnProvider extends Component<SignOnProps, SignOnState> {
       }
     })
 
-    this.props.fetchReferrer(this.props.location)
+    const referrerHandle = new URLSearchParams(this.props.location.search).get(
+      'ref'
+    )
+    if (referrerHandle) {
+      this.props.fetchReferrer(referrerHandle)
+    }
+
     const closeModalHotkey = setupHotkeys({
       27 /* Escape */: () => {
         if (
@@ -476,8 +482,8 @@ function mapDispatchToProps(dispatch: Dispatch) {
     onSignIn: (email: string, password: string) =>
       dispatch(signOnAction.signIn(email, password)),
     fetchFollowArtists: () => dispatch(signOnAction.fetchAllFollowArtists()),
-    fetchReferrer: (location: Location) =>
-      dispatch(signOnAction.fetchReferrer(location)),
+    fetchReferrer: (handle: string) =>
+      dispatch(signOnAction.fetchReferrer(handle)),
     signUp: (email: string, password: string, handle: string) =>
       dispatch(signOnAction.signUp(email, password, handle)),
     setTwitterProfile: (
--- a/src/containers/sign-on/SignOnProvider.tsx
+++ b/src/containers/sign-on/SignOnProvider.tsx
@@ -5,7 +5,7 @@ import {
   replace as replaceRoute,
   goBack
 } from 'connected-react-router'
-import { UnregisterCallback } from 'history'
+import { Location, UnregisterCallback } from 'history'
 import { sampleSize } from 'lodash'
 import { connect } from 'react-redux'
 import { withRouter, RouteComponentProps } from 'react-router-dom'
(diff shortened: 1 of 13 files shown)


sliptype (approved):
👍

audius-infra:
Preview this change https://frontend-demo.audius.co/rj-aud-707

src/containers/sign-on/SignOnProvider.tsx
```diff
@@ -121,6 +121,7 @@ export class SignOnProvider extends Component<SignOnProps, SignOnState> {
       }
     })
 
+    this.props.fetchReferrer(this.props.location)
```
jowlee, line 124: Does this mean that the `ref` url query will only work for the /signin or /signup route? - just making sure that this is what we want?

src/containers/sign-on/store/sagas.js
```diff
@@ -112,6 +114,20 @@ function* fetchFollowArtistGenre(followArtistCategory) {
   }
 }
 
+function* fetchReferrer(action) {
+  const { location } = action
+  const handle = new URLSearchParams(location.search).get('ref')
```
jowlee, line 119: Not sure if you would need a url decode here?

jowlee (approved):

src/containers/sign-on/SignOnProvider.tsx
raymondjacobson: yep. it's just for referral to start a new account. what were you thinking?

src/containers/sign-on/store/sagas.js
raymondjacobson: the other place we do handle param work is with matchPath and that doens't call a uri decode https://github.com/ReactTraining/react-router/blob/42933fe141819e4662113ab2c320bf86be3490fb/packages/react-router/modules/matchPath.js#L28

so i think we're fine?

src/containers/sign-on/SignOnProvider.tsx
```diff
@@ -475,6 +476,8 @@ function mapDispatchToProps(dispatch: Dispatch) {
     onSignIn: (email: string, password: string) =>
       dispatch(signOnAction.signIn(email, password)),
     fetchFollowArtists: () => dispatch(signOnAction.fetchAllFollowArtists()),
+    fetchReferrer: (location: Location) =>
```
piazzatron, line 479: Maybe this action could take in the referrer handle, rather than taking in location and parsing it out in the sagas - since actions are supposed to be serializable, and it looks like location has some functions attached on it

src/containers/sign-on/store/sagas.js
```diff
@@ -227,17 +243,21 @@ function* signUp(action) {
...
-          password,
-          createUserMetadata,
-          alreadyExisted
+          {
+            email,
+            password,
+            formFields: createUserMetadata,
+            hasWallet: alreadyExisted,
+            referrer
+          }
```
piazzatron, line 260: <3

src/store/account/mobileSagas.ts
```diff
@@ -52,6 +56,29 @@ function* watchAccountRecovery() {
...
+  if (!isMobileUser || isMobileUser !== 'true') {
+    try {
+      // Legacy method to update whether a user has signed in on
+      // native mobile. Used in identity service for notification indexing
+      yield call(AudiusBackend.updateUserEvent, {
+        hasSignedInNativeMobile: true
+      })
+      // Updates the user metadata with an event `is_mobile_user` set to true
+      // if the account is being fetched from a mobile context
+      yield call(updateProfileAsync, {
```
piazzatron, line 70: If sign out then in, I think we'll try to write out the is_mobile_user flag again? Maybe we could check for it here (i.e. if it's already true, do nothing)

piazzatron (approved):

src/store/account/mobileSagas.ts
raymondjacobson: we need to clear it unfortunately b/c they could sign in as a different user. the alternative solution here would be to go through a lot of trouble of piping out is_mobile_user from the discovery node side and serve it on the current user request, but i think that's heavyweight

PR #487, opened 2021-07-14
Fix linked module import order lint

sliptype:
### Description

Adding the `import/order` eslint rule caused an issue when a local module was `npm link`ed. The module was no longer considered "external" and therefore the rule required it to be in a different import group. This PR fixes that by specifying the directories from which modules should be considered "external"

### Dragons

Are there any other modules we commonly link?

If we were to add a `stems` or `audius-protocol` directory in the client, this would consider files imported from those directories external. But I don't think there would ever be a reason to have directories of those names in the client

### How Has This Been Tested?

Manually, by linking/unlinking and running the linter

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -25,6 +25,14 @@ module.exports = {
   },
   plugins: ['react', 'react-hooks', '@typescript-eslint', 'jest', 'import'],
   settings: {
+    // Only modules resolved from these folders will be considered "external".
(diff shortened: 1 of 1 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/sk-fix-linked-module-import-order-lint

raymondjacobson (approved):
this guy!!!!!

PR #491, opened 2021-07-15
TikTok share sound fixes

sliptype:
### Description

This PR:

* Fixes a bug where the `creator_node_endpoint` was being used incorrectly, causing the request to the users CN to fail and always falling back to the public ipfs gateways
* Sets the default value for the TikTok sharing feature flag  to `false`
* Adopts the pattern of using the `modal` slice to handle open/close state while using a custom slice for the business logic

### Dragons

No

### How Has This Been Tested?

Manually

### DIFF
--- a/src/containers/share-sound-to-tiktok-modal/ShareSoundToTikTokModal.tsx
+++ b/src/containers/share-sound-to-tiktok-modal/ShareSoundToTikTokModal.tsx
@@ -42,7 +42,7 @@ const fileRequirementErrorMessages = {
 }
 
 const ShareSoundToTikTokModal = () => {
-  const [isOpen, setIsOpen] = useModalState('ShareSoundToTikTokModal')
+  const [isOpen, setIsOpen] = useModalState('ShareSoundToTikTok')
   const dispatch = useDispatch()
 
   const track = useSelector(getTrack)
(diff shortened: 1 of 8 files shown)


src/containers/share-sound-to-tiktok-modal/ShareSoundToTikTokModal.tsx
```diff
@@ -41,9 +42,9 @@ const fileRequirementErrorMessages = {
 }
 
 const ShareSoundToTikTokModal = () => {
+  const [isOpen, setIsOpen] = useModalState('ShareSoundToTikTok')
```
raymondjacobson, line 45: should this string be an exported const somewhere?

raymondjacobson (approved):

audius-infra:
Preview this change https://frontend-demo.audius.co/sk-tiktok-share-sound-fixes

src/containers/share-sound-to-tiktok-modal/ShareSoundToTikTokModal.tsx
sliptype: `useModalState` accepts a union type so typescript checks that this is valid. This is how we are doing it in all the other instances

 It also autocompletes the string!

It might be valuable to discuss using union types in Stems too, because then you don't need an extra import to do `ButtonType.PRIMARY` for example. But it only makes sense if we expect all consumers of Stems to be using typescript

PR #492, opened 2021-07-15
Add conditional hcaptcha display on rewards page

sddioulde:
### Description

Display hcaptcha for users to verify themselves

### Dragons

### How Has This Been Tested?

Spun up client locally and tested that modal shows up and that submission of the captcha generates token and makes api call. Api counterpart has also been tested.


audius-infra:
Preview this change https://frontend-demo.audius.co/sd-hcaptcha

src/containers/audio-rewards-page/components/modals/RewardsModals.tsx
```diff
@@ -7,12 +7,20 @@ import TransferAudioMobileDrawer from './TransferAudioMobileDrawer'
 import TrendingRewardsModal from './TrendingRewards'
 import VerifiedUpload from './VerifiedUpload'
 
+const HCaptchaModal = React.lazy(() => import('./HCaptchaModal'))
```
raymondjacobson, line 10: this should be fine, but you may want to consider lazyWithPreload if the user hits the parent component here to start the fetching of the modal/hcaptcha code and then have a separate condition to actually do the modal display logic

raymondjacobson (approved):

src/containers/audio-rewards-page/components/modals/RewardsModals.tsx
```diff
@@ -1,18 +1,29 @@
...
 
 import { isMobile } from 'utils/clientUtil'
+import lazyWithPreload from 'utils/lazyWithPreload'
 
 import TopAPIModal from './TopAPI'
 import TransferAudioMobileDrawer from './TransferAudioMobileDrawer'
 import TrendingRewardsModal from './TrendingRewards'
 import VerifiedUpload from './VerifiedUpload'
 
+const HCaptchaModal = lazyWithPreload(() => import('./HCaptchaModal'))
```
raymondjacobson, line 11: i don't actually think we need to preload this (yet) since we're not going to turn it on until there's evidence to? but should be fine either way

raymondjacobson: > i don't actually think we need to preload this (yet) since we're not going to turn it on until there's evidence to? but should be fine either way

sorry somewhat whiplash from my part

sddioulde: all good, will make the change

PR #496, opened 2021-07-19
Update NotificationPanel to use Popup from stems

sliptype:
### Description

This PR:
* Updates `NotificationPanel` to use `Popup` from stems, removing it's dependency on ant-d

image

* Removes `Popup` and `PopupMenu` in favor of using the components from stems

### Dragons

Once again touching all overflow menus,  all the various cases everything look to be working though

### How Has This Been Tested?

Manually, against staging data


src/index.css
```diff
@@ -1,5 +1,4 @@
 @import '~antd/dist/antd.css';
-@import '~@audius/stems/dist/stems.css';
```
sliptype, line 2: When linking stems into the client, this import was causing React to complain that files were being imported from outside of `src`. Moving the import into `app.tsx` resolved this issue. Because stems is all css modules, this move to earlier in the list of imports shouldn't have any negative effects

src/containers/notification/components/desktop/NotificationBlock.tsx
```diff
@@ -231,7 +232,7 @@ const NotificationBlock = (props: NotificationBlockProps) => {
         <div className={styles.body}>{body}</div>
         {!header && (
           <div className={styles.menuContainer}>
-            <Menu menu={props.menuProps}>
+            <Menu menu={props.menuProps} zIndex={zIndex.NAVIGATOR_POPUP + 1}>
```
jowlee, line 235: Thoughts on making this another zIndex field instead of adding to the popup so that we know don't do this somewhere else and break this?

src/containers/notification/NotificationPanel.tsx
```diff
@@ -94,95 +97,102 @@ const NotificationPanel = ({
...
-      destroyOnClose
+    <Popup
+      anchorRef={anchorRef}
+      className={styles.popup}
+      isVisible={panelIsOpen}
+      checkIfClickInside={(target: EventTarget) => {
+        if (target instanceof Element && anchorRef) {
+          return (
+            anchorRef.current.contains(target) ||
+            [...target.classList].some(c => c.includes('PopupMenu'))
```
jowlee, line 108: Is the `PopupMenu` a class hardcoded in the @audius/stems project?
Thoughts on something like `import navStyles from './NavHeader.module.css'` and then check against `navStyles.iconNotification` here instead?

jowlee (approved):
🙏

src/containers/notification/components/desktop/NotificationBlock.tsx
sliptype: Yeah I was going back and forth on this, I think having a separate key for the `NAVIGATOR_POPUP_OVERFLOW_POPUP` is better. Added!

src/containers/notification/NotificationPanel.tsx
sliptype: The `anchorRef` is used to check if the click is inside of the button that triggers the NotificationPanel. The second check is for the overflow menu on an individual notification itself. But I agree that using the classname is pretty hacky. 

I've been meaning to add `forwardRefs` to the stems components anyway and I think this is a perfect use case:
https://github.com/AudiusProject/audius-client/pull/496/commits/0d588281d739cbaf71aea483e2a455382e893cb6#diff-774e572244f6064b7880927b357385d2c46ace5b14dd2a0dba55c32c3e25c861R108

Any thoughts?

audius-infra:
Preview this change https://frontend-demo.audius.co/sk-use-popup-from-stems

raymondjacobson (approved):
I think the drop-shadow on the notification panel is less with this popup component compared to what it was. I think the shadow is probably intentionally pretty dark for this one, but maybe we should just update across the board. Probably a julian Q.

Screen Shot 2021-07-21 at 4 55 14 PM
Screen Shot 2021-07-21 at 4 55 11 PM

src/containers/notification/NotificationPanel.tsx
```diff
@@ -94,95 +99,102 @@ const NotificationPanel = ({
   }, [hasMore, status, fetchNotifications])
 
   return (
-    <AntModal
-      wrapClassName={styles.modalContainerWrapper}
```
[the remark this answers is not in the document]
sliptype: Somehow I missed these comments, yeah I think you are right. Let me check if those styles can be deleted and I'll ask Julian about the drop shadow

PR #497, opened 2021-07-19
Fix subtle bug in explore layout

raymondjacobson:
### Description

Because the layout for the explore page is now controlled by a feature flag, the first few renders happen when the feature flag is off and then renders happen when it's on. This causes two issues:
1. The layout jumps when the Remixables tile shows up
2. It's possible that the array of explore tiles gets instantiated before the feature flag loads at all, making remixables never show

Both of these issues are fixed by
1. Moving the decision to render the remixables tab into the conmponent lifecycle rather than as a global
2. Wait to render explore until remote config loads in

### Dragons

### How Has This Been Tested?

1. Feature flag on
2. Feature flag off

### DIFF
--- a/src/containers/explore-page/ExplorePage.tsx
+++ b/src/containers/explore-page/ExplorePage.tsx
@@ -1,7 +1,8 @@
 import React from 'react'
 
-import { connect } from 'react-redux'
+import { connect, useSelector } from 'react-redux'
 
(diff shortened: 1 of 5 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/rj-fix-subtle-explore-remote-config-bug

Kyle-Shanks (approved):

PR #500, opened 2021-07-22
AUD-837 Update client to call new route API

rickyrombo:
### Description

AUD-837 Update client to call new route api

Updates the client to remove IDs from the URLs by querying the API via slug/handle combo.

### Dragons

- **ANYWHERE I MISSED** that uses IDs in the route (not including the `/tracks/<hashid>`)
- Need help testing!!!

### How Has This Been Tested?

Everything was manual 😨 
Still ongoing

TODOS:
- [x] Clean up unnecessary changes and any lingering `console.logs`
- [x] ~~Update routes in links to new `permalink`~~ #502 
- [x] Follow up on questions on existing routing logic
- [x] Verify unlisted tracks still work
- [x] Update full remixes page to use new routing scheme
- [x] Fix permalink.toLowerCase() not defined error and remove console.log()


audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-track-route-api

rickyrombo:
Putting up for an early review - still going to do a pass to make sure everything is good but wanted to get early feedback on general approach etc

src/containers/track-page/store/sagas.js
```diff
@@ -140,39 +141,43 @@ function* getMoreByThisArtist(trackId, ownerHandle) {
...
+        track = tracks && tracks.length === 1 ? tracks[0] : null
+      }
+      if (!track) {
         const isReachable = yield select(getIsReachable)
         if (isReachable) {
           yield put(pushRoute(NOT_FOUND_PAGE))
           return
         }
+      } else {
+        yield fork(getMoreByThisArtist, track.track_id, handle)
```
rickyrombo, line 172: Note that before this change we could start to get tracks by the artist because we had the track ID ahead of time.

raymondjacobson (commented):
Some things that I noticed while playing around w/ it


1. upload track => visit track page (shows a weird redirect from -id to the no -id URL). this seems to be the case across most new "slug" tracks. i imagine this is b/c we haven't changed internal routing yet, but that seems important to do

2. upload hidden track => visit track page immediately (404)

3. visit artist dashoard => visit hidden track page (404)

4. share link from a track page copies the old style URL (which does a proper redirect when linked to)

5. Setting the "Remix of" track using a URL shows "track not found" because `getTrackId` method in remix-settings-modal/store/sagas.ts calls into the trackRouteParser and then calls retrieve with no id

6. embed player share copies old style link (track page => embed => click share btn on track). probably known. fortunately general admission is the sole responsible entity for handling that: https://github.com/AudiusProject/general-admission/blob/634e9fdee4d55ae49f0c11fc76f9bf964157ead1/src/servlets/bedtime/index.ts#L21

rickyrombo:
> Some things that I noticed while playing around w/ it
> 
> 1. upload track => visit track page (shows a weird redirect from -id to the no -id URL). this seems to be the case across most new "slug" tracks. i imagine this is b/c we haven't changed internal routing yet, but that seems important to do

Yep, this is because of TODO #2, I haven't updated the links yet to use the new URLs

> 2. upload hidden track => visit track page immediately (404)
> 3. visit artist dashoard => visit hidden track page (404)

Somewhat expected at the moment - I haven't touched unlisted yet (was working on that today)

> 4. share link from a track page copies the old style URL (which does a proper redirect when linked to)

Same as #1

> 5. Setting the "Remix of" track using a URL shows "track not found" because `getTrackId` method in remix-settings-modal/store/sagas.ts calls into the trackRouteParser and then calls retrieve with no id

Ooh let me look into this, I'll double check all calls into trackRouteParser too.

> 6. embed player share copies old style link (track page => embed => click share btn on track). probably known. fortunately general admission is the sole responsible entity for handling that: https://github.com/AudiusProject/general-admission/blob/634e9fdee4d55ae49f0c11fc76f9bf964157ead1/src/servlets/bedtime/index.ts#L21

Ah, yeah, I have another linear item to update GA. Good reminder to be sure to put a PR up for that too before this goes in.

rickyrombo:
@raymondjacobson unlisted tracks should work now - and for 1 and 4 see #502

raymondjacobson:
> @raymondjacobson unlisted tracks should work now - and for 1 and 4 see #502

Awesome!!

raymondjacobson:
Looks good! only 1 thing I found--

- renaming track doesn't immediately update the URL (after the confirmer updates the URL does). share button doesn't copy the updated URL as well after rename. I think this is because when you edit the track, we don't update the permalink in the cache

rickyrombo:
> Looks good! only 1 thing I found--
> 
> * renaming track doesn't immediately update the URL (after the confirmer updates the URL does). share button doesn't copy the updated URL as well after rename. I think this is because when you edit the track, we don't update the permalink in the cache

I think it's impossible to update the URL until the DN has indexed, as we won't know the conflict/collision ID. Share button is weird though, will look!

raymondjacobson:
LGTM minus this one comment, which I don't think we got to yet https://github.com/AudiusProject/audius-client/pull/500#discussion_r675216113

src/store/cache/tracks/selectors.ts
```diff
@@ -7,6 +7,9 @@ export const getTrack = (
   state: AppState,
   props: { id?: ID | null; uid?: UID | null; permalink?: string | null }
 ) => {
+  if (props.permalink && typeof props.permalink !== 'string') {
+    console.log(typeof props.permalink, props.permalink)
```
raymondjacobson: log

raymondjacobson (approved):

src/containers/remixes-page/RemixesPageProvider.tsx
```diff
@@ -96,7 +94,7 @@ const RemixesPageProvider = ({
       actions: tracksActions,
       scrollParent: containerRef as any,
       loadMore: (offset: number, limit: number) => {
-        loadMore(offset, limit, { trackId })
+        loadMore(offset, limit, { trackId: originalTrack?.track_id || null })
```
raymondjacobson, line 97: i think this is probably the same behavior as before, but definitely a little odd.

we do have this withNullGuard construct that we've used in a full places to prevent from getting into this annoying situation https://github.com/AudiusProject/audius-client-old/blob/ad02f295906122c234ae5a0bd593c684129dce17/src/containers/remixes-page/components/desktop/RemixesPage.tsx#L33 (used here but a level down in the mobile/desktop rather than at the provider level)

raymondjacobson (approved):

src/containers/remixes-page/RemixesPageProvider.tsx
raymondjacobson: also nit, if we keep this

originalTrack?.track_id ?? null probs better

PR #504, opened 2021-07-28
Link solana wallet

jowlee:
### Description
Adds the ability to link solana wallets with [sollet](https://www.sollet.io/) and [phantom](https://docs.phantom.app/)

* Updates to the user metadata schema to add field `associated_spl_wallets`
* Adds libs dependency for sol-wallet-adapter needed for sollet integration
* Updates connect connect wallet saga flow to add logic for sollet/phantom connection w/ solana web3
* Updates UI of connect wallet modal w/ chain icon (eth/sol) collection count and balance
* Adds new modal for Audio Breakdown (On mobile, it's a page)
* Adds remote config variables:
  * `DISPLAY_SOLANA_WEB3_PROVIDER_PHANTOM`: for connecting solana wallets with phantom
  * `DISPLAY_SOLANA_WEB3_PROVIDER_SOLLET`: solana wallets w/ sollet

### Dragons

### How Has This Been Tested?
This was run locally against a remote setup to test that the wallets were updated correctly. 

Closes AUD-795


src/services/opensea-client/OpenSeaClient.ts
```diff
@@ -213,6 +213,36 @@ class OpenSeaClient {
...
+
+  async getSPLCollectibleCount(splWallet: string) {
+    // TODO: figure out how to get spl wallet assets
+    return 0
+  }
+
+  async getSPLCollectibleCountForMultipleWallets(wallets: string[]) {
+    // TODO: figure out how to get spl wallet assets
+    return wallets.map(() => 0)
+  }
```
jowlee, line 245: @sddioulde Have you looked into the solana NFTs? 
Also does the first couple methods for collectible counts seem reasonable. hopefully won't take too long.

src/services/AudiusBackend.js
```diff
@@ -2476,6 +2505,18 @@ class AudiusBackend {
...
     )
   }
+
+  /**
+   * Transfers the user's ERC20 AUDIO into SPL WAUDIO to their solana user bank account
+   * @param {BN} balance The amount of AUDIO to be transferred
+   */
+  static async getAddressWAudioBalance(address) {
+    await waitForLibsInit()
+    const waudioBalance = await audiusLibs.solanaWeb3Manager.getWAudioBalance(
```
jowlee, line 2515: NOTE: For local dev, if this is not init, the client will break

src/services/opensea-client/OpenSeaClient.ts
jowlee: We are not longer supporting collectible counts in this feature for now

src/containers/audio-rewards-page/components/modals/AudioBreakdownModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+const messages = {
+  modalTitle: '$AUDIO BREAKDOWN',
+  total: 'TOTAL $AUDIO',
+  audiusWallet: 'AUDIUS WALLET',
+  audiusWalletDescription: 'You can use this $AUDIO throughout the app',
+  linkedWallets: 'LINKED WALLETS',
+  linkedWalletsDescription:
+    'Linked wallets are more secure but not all features are supported',
+  linkedWalletsTooltip: 'Something here'
```
jowlee, line 28: Waiting on copy from julian and forrest here

audius-infra:
Preview this change https://frontend-demo.audius.co/jowlee-link-solana-wallet

src/store/token-dashboard/sagas.ts
```diff
@@ -335,28 +657,55 @@ function* connectWallet() {
...
+    if (removeChain === 'eth') {
+      const currentAssociatedWallets: Record<string, any> = yield call(
+        AudiusBackend.fetchUserAssociatedEthWallets,
+        updatedMetadata
+      )
+      if (
+        currentAssociatedWallets &&
+        !(removeWallet in currentAssociatedWallets)
+      ) {
+        // The wallet already exists in the assocaited wallets set
```
rickyrombo, line 676: associated

rickyrombo: also, wouldn't `!(removeWallet in currentAssociatedWallets)` mean it's _not_ in the associated wallets set? Should this comment say it _doesn't_ exist in the associated wallet set?

src/store/token-dashboard/slice.ts
```diff
@@ -150,11 +172,20 @@ const slice = createSlice({
...
       {
-        payload: { associatedWallets }
-      }: PayloadAction<{ associatedWallets: AssociatedWallets }>
+        payload: { associatedWallets, chain }
+      }: PayloadAction<{ associatedWallets: AssociatedWallets; chain: Chain }>
     ) => {
-      state.associatedWallets.connectedWallets = associatedWallets
-      state.associatedWallets.confirmingWallet = null
+      if (chain === 'sol') {
+        state.associatedWallets.connectedSolWallets = associatedWallets
```
rickyrombo, line 179: Why have separate collections of wallets rather than one collection of different wallet types (eg a "chain" prop on the wallet themselves)? It seems like a lot of the work here is in splitting the logic between the two collections.

package.json
```diff
@@ -7,9 +7,10 @@
   "private": true,
   "dependencies": {
     "@audius/libs": "1.2.13",
-    "@audius/stems": "0.3.12",
+    "@audius/stems": "^0.3.13",
```
rickyrombo, line 10: intentional?

rickyrombo (commented):
Note: I didn't really dive in to the css or verify that the subcomponents you moved out of the existing components into their own components kept the same functionality.

src/store/token-dashboard/slice.ts
jowlee: Yeah, I think it'd be the same amount of work either way.
The Solana wallets and Eth wallets are treated differently: Fetching their balances, signing messages, and formatting/displaying the wallets in the UI differentiates the two.
I could go either way for how it's stored though, just thought it was more explicit to split them up this way so we don't mix methods between the two.

package.json
jowlee: Ideally a patch version update wouldn't break us, but I'll change it, thx

src/containers/audio-rewards-page/components/WalletsTable.tsx
```diff
@@ -0,0 +1,241 @@
...
+const COPIED_TOAST_TIMEOUT = 2000
+
+const messages = {
+  copied: 'Copied To Clipboard!',
+  linkedWallets: 'LINKED WALLETS',
+  collectibles: 'COLLECTIBLES',
+  audio: '$AUDIO'
+}
+
+const shortedSPLAddress = (addr: string) => {
```
raymondjacobson, line 35: shorten

src/containers/audio-rewards-page/components/WalletsTable.tsx
```diff
@@ -0,0 +1,241 @@
...
+  return `${addr.substring(0, 4)}...${addr.substr(addr.length - 5)}`
+}
+
+const shortenEthAddress = (addr: string) => {
+  return `0x...${addr.substr(addr.length - 7)}`
```
raymondjacobson, line 40: probably worth having 0,4 at the beginning too? like we do on protocl dash?

src/services/AudiusBackend.js
```diff
@@ -1022,13 +1024,40 @@ class AudiusBackend {
     return null
   }
 
+  /**
+   * Retrieves the user's associated wallets from IPFS using the user's metadata CID and creator node endpoints
+   * @param {Object} user The user metadata which contains the CID for the metadata multihash
```
raymondjacobson, line 1029: @param {User}

src/services/AudiusBackend.js
```diff
@@ -1022,13 +1024,40 @@ class AudiusBackend {
...
+   * @returns Object The associated wallets mapping of address to nested signature
+   */
+  static async fetchUserAssociatedSolWallets(user) {
+    const gateways = getCreatorNodeIPFSGateways(user.creator_node_endpoint)
+    const cid = user?.metadata_multihash ?? null
+    if (cid) {
+      const metadata = await fetchCID(
```
raymondjacobson, line 1036: shouldn't we only need to fetch this CID once? maybe we should follow the pattern on the profile page for collectibles and do this somewhere else:
https://github.com/AudiusProject/audius-client/blob/b8b852bd4a819694eb60ae27b82bad79bed4e79b/src/containers/profile-page/store/sagas.js#L56

src/services/web3-modal/index.ts
```diff
@@ -84,6 +109,62 @@ export const createSession = async (config: Config): Promise<any> => {
...
+            const wallet = new SolWallet(providerUrl, network)
+            const walletConnection: Promise<string> = new Promise(
+              (resolve, reject) => {
+                wallet.on('connect', publicKey => {
+                  resolve(publicKey.toBase58())
+                })
+              }
+            )
+            await wallet.connect()
+            console.log('finihsed')
```
raymondjacobson, line 156: logs

src/store/token-dashboard/slice.ts
```diff
@@ -40,17 +40,32 @@ export type ModalState = Nullable<
...
+  collectibleCount: number
+}[]
 
-export type ConfirmRemoveWalletAction = PayloadAction<{ wallet: WalletAddress }>
+export type ConfirmRemoveWalletAction = PayloadAction<{
+  wallet: WalletAddress
+  chain: Chain
+}>
 
+export type Chain = 'sol' | 'eth'
```
raymondjacobson, line 54: i know mike likes the string options here, but an enum would let you switch statement the reducers a bit more cleanly 🤷

raymondjacobson (approved):

rickyrombo (approved):

src/services/AudiusBackend.js
jowlee: interesting, yeah it's odd bc fetching the metadata and storing it redux means that the format of the current user's metadata is different from all other users in the cache.
Specifically for the linked wallets - usually we only fetch just user's linked wallets from a discovery endpoint, bc we don't need the extra metadata for signature.
I personally like the idea of fetching before setting the new wallets which is where this is used vs on init fetching of user because it leads to less of a chance of messing up the user's metadata in between. but there're a couple places I could reduce the network calls for this. 
Thoughts?

PR #507, opened 2021-07-30
Aud-791 tiktok overflow sharing

sliptype:
### Description

This PR:
* Adds __Share To TikTok__ to track overflow menus for the owner of the track
Screenshot_20210730-132204_Audius Music
Screenshot_20210730-132211_Audius Music
* Uses the `tracks/stream` endpoint to fetch the track instead of using the cid and fetchCID
* Adds mobile styling for ShareToTikTokModal
* Fixes an issue with overflow width of the ShareBanner buttons (https://linear.app/audius/issue/AUD-798/add-min-width-or-wrap-on-tik-tok-track-is-live-buttons)
* Adds the message types to interact with the mobile client for TikTok auth

### Dragons

Nothing really, only touching TikTok stuff which is behind a feature flag. Very minor tweaks to the upload sagas

### How Has This Been Tested?
 
Manually, on desktop/android device/ios simulator


src/containers/share-sound-to-tiktok-modal/ShareSoundToTikTokModal.tsx
```diff
@@ -42,6 +47,9 @@ const fileRequirementErrorMessages = {
 }
 
 const ShareSoundToTikTokModal = () => {
+  const mobile = isMobile()
```
raymondjacobson, line 50: this works, we also have a useIsMobile() hook that does the same thing. i don't really have a strong preference, but probably we should pick a pattern 🤷

src/containers/share-sound-to-tiktok-modal/store/sagas.ts
```diff
@@ -20,22 +24,43 @@ const TIKTOK_SHARE_SOUND_ENDPOINT =
...
-    const { data } = yield call(
-      window.audiusLibs.File.fetchCID,
-      action.payload.cid,
-      getCreatorNodeIPFSGateways(creator_node_endpoint),
-      () => {}
+    // Fetch the track blob
+    const encodedTrackId = encodeHashId(id)
+
+    const response = yield call(
+      window.fetch,
```
raymondjacobson, line 55: do we need window here? maybe this code should live in apiClient itself

raymondjacobson (approved):
Looks great! couple small things, but I'd be cool merging this beforehand to get on mobile + then addressing some of the tiny stuff afterwards

PR #510, opened 2021-08-03
Update fetchRemixables query to limit the number of times an artist can appear

Kyle-Shanks:
### Description

Limiting artists popping up too many times in the remixables playlist

### DIFF
--- a/src/containers/smart-collection/store/sagas.ts
+++ b/src/containers/smart-collection/store/sagas.ts
@@ -117,10 +117,24 @@ function* fetchRemixables() {
   const tracks: TrackMetadata[] = yield call(
     Explore.getRemixables,
     currentUserId,
-    COLLECTIONS_LIMIT
+    75 // limit
   )
 
-  const processedTracks: Track[] = yield call(processAndCacheTracks, tracks)
+  // Limit the number of times an artist can appear
+  const artistLimit = 3
+  const artistCount: Record<number, number> = {}
+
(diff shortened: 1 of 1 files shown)


src/containers/smart-collection/store/sagas.ts
```diff
@@ -117,10 +117,24 @@ function* fetchRemixables() {
...
-    COLLECTIONS_LIMIT
+    75 // limit
   )
 
-  const processedTracks: Track[] = yield call(processAndCacheTracks, tracks)
+  // Limit the number of times an artist can appear
+  const artistLimit = 3
+  const artistCount: Record<number, number> = {}
+
+  const filteredTracks = tracks.filter(trackMetadata => {
```
raymondjacobson, line 127: this looks like it would work, but we should probably take the most popular (first) 3 not just any 3 since the filter will go in any order

probably just a for of loop would be better

Kyle-Shanks: Does `Array.filter` not go in the order? I always assumed that it did?

audius-infra:
Preview this change https://frontend-demo.audius.co/kj-Add-artist-limit-to-remixables-query

src/containers/smart-collection/store/sagas.ts
raymondjacobson: im wrong

raymondjacobson (approved):

PR #521, opened 2021-08-12
Fix TikTok CSRF cookie on mobile

sliptype:
### Description

This PR:

* Adds support for the new way of fetching the TikTok access token on mobile. Previously, the authorization code was fetched in the native layer spawned oauth webview and then passed to the web layer which performed the second request for the access token. The required csrf cookie persisted on android but not on ios. The new way performs the access token request in the oauth webview so that the csrf cookie is available

* Adds a `auth-redirect` route to the public site to show a spinner before the auth polling catches the redirect (instead of flashing the landing page)

### Dragons

Need to make sure the mobile client includes these changes because they are not backwards compatible

### How Has This Been Tested?


audius-infra:
Preview this change https://frontend-demo.audius.co/sk-fix-csrf-cookie-on-mobile

raymondjacobson (approved):

PR #527, opened 2021-08-14
Fix permalink selector to lowercase

raymondjacobson:
### Description

We lowercase all permalinks in storage:
https://github.com/AudiusProject/audius-client/blob/daf20f01b3737d8ca9a975b71f7f662eaa2c9aeb/src/store/cache/tracks/selectors.ts#L14

The selector does not lowercase though, so tracks w/ capitalization are not fetched properly by the selector

^^^^^^^
After this first change, realizing that there's already a permalink selector that does the .toLowerCase and some of the logic can be consolidated. I believe this second fix is safe as well.

### Dragons

Touches track retrieval

### How Has This Been Tested?

Locally against prod (mobile + web)

http://localhost:3002/Galazy_Music
-> click on first track
-> 404
(prior to change)

No 404 after

* Uploaded track
* Navigated between tracks
* Linked directly to track

### DIFF
--- a/src/store/cache/tracks/selectors.ts
+++ b/src/store/cache/tracks/selectors.ts
@@ -19,11 +19,6 @@ export const getTrack = (
   })
 }
 
-export const getTrackByPermalink = (
-  state: AppState,
-  props: { permalink: string }
-) => state.tracks.permalinks[props.permalink.toLowerCase()] || null
-
 export const getStatus = (state: AppState, props: { id?: ID | null }) =>
   (props.id && state.tracks.statuses[props.id]) || null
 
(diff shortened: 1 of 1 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/rj-permalinks-selector-fix

src/store/cache/tracks/selectors.ts
```diff
@@ -18,10 +18,7 @@ export const getTrack = (
     kind: Kind.TRACKS
   })
 }
-export const getTrackByPermalink = (
-  state: AppState,
-  props: { permalink: string }
-) => state.tracks.permalinks[props.permalink] || null
```
raymondjacobson, line 24: this was missing .toLowerCase

piazzatron (approved):

rickyrombo:
This looks good to me - however it makes me want to ensure that we do collision detection on the backend case-insensitive then too (or remove the toLowerCase call on the client)

PR #534, opened 2021-08-18
Add Solana collectibles

sddioulde:
### Description

Add Solana collectibles to collectibles page.

### Dragons

### How Has This Been Tested?

- Tested manually with hardcoded solana address.
- Going to test with connected phantom solana wallet.


raymondjacobson (commented):
will take a closer look today. some function-level comments would be helpful tho :)

audius-infra:
Preview this change https://frontend-demo.audius.co/sd-solana-nft

src/containers/collectibles/solCollectibleHelpers.ts
```diff
@@ -0,0 +1,188 @@
...
+  const files = nft.properties.files
+  // should we restrict video file extensions here?
+  // MP4, MOV, GLB
+  // GLTF??
+  // https://github.com/metaplex-foundation/metaplex/blob/81023eb3e52c31b605e1dcf2eb1e7425153600cd/js/packages/web/src/views/artCreate/index.tsx#L318
+  // DO WE CARE ABOUT VR NFTs??
+  const videoFile = files?.find(
+    file => typeof file === 'object' && file.type.includes('video')
+  ) as SolanaNFTPropertiesFile
+  // https://github.com/metaplex-foundation/metaplex/blob/397ceff70b3524aa0543540584c7200c79b198a0/js/packages/web/src/components/ArtContent/index.tsx#L107
```
rickyrombo, line 42: this link already broke I think. Try: https://github.com/metaplex-foundation/metaplex/blob/194a82f35ecc5178c9595ea1dbbdc0bdf4ec42b4/js/packages/web/src/components/ArtContent/index.tsx#L107

src/containers/collectibles/solCollectibleHelpers.ts
```diff
@@ -0,0 +1,188 @@
...
+      url,
+      frameUrl: nft.image || null
+    }
+  }
+  return null
+}
+
+const nftImage = async (nft: SolanaNFT): Promise<Nullable<SolanaNFTMedia>> => {
+  const files = nft.properties.files
+  // should we restrict image file extensions here?
```
rickyrombo, line 83: What are the reasons to restrict/not to restrict? Does it work with other file types out of the box? Do we not want it to? Etc. (Same with video).

I think restricting could put us in better control, say if someone tries to put a mkv or something huge on there and we have a sluggish feeling user experience. On the flip-side, the request isn't going against our server, so I'm not sure it matters? Are there security implications?

src/containers/collectibles/solCollectibleHelpers.ts
```diff
@@ -0,0 +1,188 @@
...
+    externalLink: nft.external_url,
+    isOwned: true,
+    chain: Chain.Sol
+  } as Collectible
+
+  if (nft.properties.creators.some(creator => creator.address === address)) {
+    collectible.isOwned = false
+  }
+
+  const { url, frameUrl, collectibleMediaType } = ((await nftGif(nft)) ||
```
rickyrombo, line 173: Are NFTs => Collectibles really 1:1? It looks like each NFT can have multiple files, but we only get one URL/Collectible from each NFT?

src/services/solana-client/SolanaClient.ts
```diff
@@ -0,0 +1,137 @@
...
+      .map((value, i) => {
+        const ownerAddress = wallets[i]
+        const mintAddresses = value
+          .map(v => ({
+            mint: v.account.data.parsed.info.mint,
+            tokenAmount: v.account.data.parsed.info.tokenAmount
+          }))
+          .filter(({ tokenAmount }) => {
+            // nfts generally have a supply of 1 and they have 0 decimal places
+            // return tokenAmount.uiAmount === 1 && tokenAmount.decimals === 0
```
rickyrombo, line 42: This looks like a TODO

src/services/solana-client/SolanaClient.ts
```diff
@@ -0,0 +1,137 @@
...
+            return true
+          })
+          .map(({ mint }) => mint)
+        /**
+         * {
+         *   mintAddresses: ['mintAddress1', 'mintAddress2', ...],
+         *   ownerAddress: 'solanaAddress'
+         * }
+         */
+        // console.log({ mintAddresses, ownerAddress })
```
rickyrombo, line 52: nit: remove comment when done?

rickyrombo (commented):
Agree with Ray that some top-level comments would be helpful, and there's some lingering comments I think, but other than that I think it looks good!

src/containers/collectibles/solCollectibleHelpers.ts
sddioulde: did it? im still able to access

src/containers/collectibles/solCollectibleHelpers.ts
sddioulde: They're supposed to be 1:1 from my understanding. Each NFT can have multiple files in the metadata, but what I saw was that it's because it may be a video nft where 1 file is the video, another is the audio, another may be a thumbnail for the video...

src/services/solana-client/SolanaClient.ts
sddioulde: for sure. thanks. need to clean up

src/services/solana-client/SolanaClient.ts
sddioulde: that's right, need to decide to do or not to do

src/containers/collectibles/solCollectibleHelpers.ts
sddioulde: These are all great questions. Maybe Metaplex already handles not including weird file types as they do specify what types they want [here](https://github.com/metaplex-foundation/metaplex/blob/81023eb3e52c31b605e1dcf2eb1e7425153600cd/js/packages/web/src/views/artCreate/index.tsx#L318).
For now will not restrict and we'll watch how it plays out. If need to restrict, we can base off above link

PR #555, opened 2021-09-07
AUD-916 Delay NotificationPanel image load

sliptype:
### Description

This PR:
* Adds a 500ms delay to the loading of profile pictures in the notification panel. This fixes an issue with the animation lagging when there are a lot of notifications with profile pictures
* Adds types and tests to the `useImageSize` hook

Would recommend testing on https://sebastian.audius.co/ to see the behavior with prod quantity of notifications 

### Dragons

### How Has This Been Tested?

Unit tests & against production


src/containers/notification/components/desktop/UserHeader.tsx
```diff
@@ -38,19 +40,34 @@ export const UserImage = ({
...
+    SquareSizes.SIZE_150_BY_150,
+    undefined,
+    undefined,
+    loadImage
   )
 
+  // Loading the images immediately causes lag in the NotificationPanel animation
+  useEffect(() => {
+    if (!loadImage) {
+      const t = setTimeout(() => {
```
sliptype, line 56: Considered hooking into the Popup animation end to trigger this. But that would require changes to the Popup so I opted to keep it simple with a timeout for now

src/hooks/useImageSize.ts
```diff
@@ -19,138 +19,140 @@ import {
...
+  ImageSizes extends ImageSizesObject<ImageSize>
+> = {
+  // The action to dispatch to fetch the desired size
+  action: (id: number, size: ImageSize) => void
+  // The fallback if no sizes are available
   defaultImage?: string
+  // A unique id (used to prevent duplicate fetches)
+  id?: number | null
+  // A flag (default = true) that will trigger the image load. Can be used to delay the load
+  load?: boolean
```
sliptype, line 65: This is a new option because `onDemand` doesn't allow for using an already cached image before triggering the fetch for the desired size

src/hooks/useImageSize.ts
```diff
@@ -19,138 +19,140 @@ import {
...
+    if (smaller) {
+      return fallbackImage(smaller)
+    }
+
+    return defaultImage
   }
 
-  if (!onDemand) return getImageSize()
-  return getImageSize
+  // TODO: sk - disambiguate the return value so it can be typed
```
sliptype, line 153: Future scope because it would touch every usage of `useImageSize`

audius-infra:
Preview this change https://frontend-demo.audius.co/sk-aud-916-notif-panel-image-load

raymondjacobson:
Playing around w/ it a bit, I see a delay in loading my own profile pic
https://www.loom.com/share/5182d1087bb3439188ee41e5a8c353aa

Aso, if we do want to commit to this delayed load, we should fade the profile pictures in from an empty/loading gradient state rather than the "no profile pic" default image like we do w/ artwork
Screen Shot 2021-09-07 at 4 19 26 PM

sliptype:
> Playing around w/ it a bit, I see a delay in loading my own profile pic
> https://www.loom.com/share/5182d1087bb3439188ee41e5a8c353aa
> 
> Also, if we do want to commit to this delayed load, we should fade the profile pictures in from an empty/loading gradient state rather than the "no profile pic" default image like we do w/ artwork
> Screen Shot 2021-09-07 at 4 19 26 PM

Good catch! That was due to a slight change in how `useImageSize` was returning default values. I've fixed it now to only return the `defaultImage` if the requests/overridden size exists in the `sizes` object but its value is falsey. This is how it was working before

jowlee (approved):
Looks good, nice test!

PR #556, opened 2021-09-08
Add wait between validating email

jowlee:
### Description
Creates a new action/saga for validating email, throttling the email check to 2 seconds
The throttle method was used in place of a delay in the existing saga so that it does not have to wait on the very first pass.

Closes AUD-930

### Dragons
none

### How Has This Been Tested?
Running locally

manually


audius-infra:
Preview this change https://frontend-demo.audius.co/jowlee-wait-signon

raymondjacobson (approved):

raymondjacobson:
I think 2s is reasonable since leading is allowed

PR #567, opened 2021-09-13
Fix 3d object nft frame in case of gif

sddioulde:
### Description

3D nfts on opensea with a gif frame were display the gif instead of a still image on the collectibles page. This PR makes an image out of the gif so the card shows a still image.

### Dragons

### How Has This Been Tested?

Spun app locally and tested account with 3d nft that had a gif as a frame. The frame now shows a still image instead of a gif.

### DIFF
--- a/src/containers/collectibles/ethCollectibleHelpers.ts
+++ b/src/containers/collectibles/ethCollectibleHelpers.ts
@@ -146,12 +146,23 @@ export const assetToCollectible = async (
       gifUrl = imageUrls.find(url => url?.endsWith('.gif'))!
     } else if (isAssetThreeDAndIncludesImage(asset)) {
       mediaType = CollectibleMediaType.THREE_D
-      frameUrl = imageUrls.find(
-        url => url && NON_IMAGE_EXTENSIONS.every(ext => !url.endsWith(ext))
-      )!
       threeDUrl = [animation_url, animation_original_url, ...imageUrls].find(
         url => url && SUPPORTED_3D_EXTENSIONS.some(ext => url.endsWith(ext))
       )!
+      frameUrl = imageUrls.find(
+        url => url && NON_IMAGE_EXTENSIONS.every(ext => !url.endsWith(ext))
+      )!
+      // image urls may not end in known extensions
+      // just because the don't end with the NON_IMAGE_EXTENSIONS above does not mean they are images
+      // they may be gifs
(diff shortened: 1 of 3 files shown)


src/containers/collectibles/ethCollectibleHelpers.ts
```diff
@@ -146,12 +146,23 @@ export const assetToCollectible = async (
...
       )!
+      frameUrl = imageUrls.find(
+        url => url && NON_IMAGE_EXTENSIONS.every(ext => !url.endsWith(ext))
+      )!
+      // image urls may not end in known extensions
+      // just because the don't end with the NON_IMAGE_EXTENSIONS above does not mean they are images
+      // they may be gifs
+      // example: https://lh3.googleusercontent.com/rOopRU-wH9mqMurfvJ2INLIGBKTtF8BN_XC7KZxTh8PPHt5STSNJ-i8EQit8ZTwE3Mi8LK4on_4YazdC3Cl-HdaxbnKJ23P8kocvJHQ
+      const res = await fetch(frameUrl, { method: 'HEAD' })
+      const hasGifFrame = res.headers.get('Content-Type')?.includes('gif')
```
raymondjacobson, line 160: is this latency / additional req. volume at all something to worry about?

src/services/WebWorker.js
```diff
@@ -43,6 +43,7 @@ export default class WebWorker {
             resolve(event.data.result)
           } else {
           }
```
raymondjacobson, line 45: can we just remove? i think equally as confusing now that there's a todo. it seems like it does nothing?

raymondjacobson (approved):

audius-infra:
Preview this change https://frontend-demo.audius.co/sd-nft

src/services/WebWorker.js
sddioulde: yeah added todo to get a comment from you :). will remove

src/containers/collectibles/ethCollectibleHelpers.ts
sddioulde: It will depend on what page you're looking at and how many 3d nfts they have whose images are gifs. I assume that's going to be more of an edge case. We could do the check on CollectibleDetails but that would make to already non-trivial logic in there more complex

audius-infra:
Preview this change https://frontend-demo.audius.co/sd-nft

PR #573, opened 2021-09-16
AUD-671 Sign Up/Sign In Redesign

rickyrombo:
### Description
AUD-671
- Make the sidebar have a "Sign In" link instead of "Sign Up" when not logged in.
- Submitting an email that already exists in the "Sign Up" flow no longer errors but sends you to the "Sign In" page with the email pre-filled and the password field focused.
- Adjusted the "Sign Up" page to include a button instead of a link for "Have an account? Sign In"

### Dragons

- ~I dislike the way I handled the issues with the Saga for the `EmailPage`, would love to brainstorm other ideas here.~ Updated: Feedback still welcome but looks better now.

### How Has This Been Tested?

- Manually/locally.

- None.

### DIFF
--- a/src/containers/sign-on/SignOnProvider.tsx
+++ b/src/containers/sign-on/SignOnProvider.tsx
@@ -194,6 +194,13 @@ export class SignOnProvider extends Component<SignOnProps, SignOnState> {
 
   onNextPage = () => {
     const { page, isMobile } = this.props
+    if (page === Pages.PASSWORD) {
+      const {
+        fields: { email },
+        recordCompletePassword
+      } = this.props
+      recordCompletePassword(email.value)
+    }
     if (page === Pages.PROFILE) {
       const {
         signUp,
@@ -231,20 +238,6 @@ export class SignOnProvider extends Component<SignOnProps, SignOnState> {
     }
     this.addRouteHash(page)
     this.props.nextPage(isMobile)
-  }
-
-  handleOnContinue = (page: Pages) => {
-    return () => {
-      const { email } = this.props.fields
-      if (page === Pages.PASSWORD) {
-        this.props.onEmailSubmitted(email.value)
-        return
-      } else if (page === Pages.PROFILE) {
-        this.props.recordCompletePassword(email.value)
-      }
-      this.addRouteHash(page)
-      this.props.goToPage(page)
-    }
   }
 
   onPrevPage = () => {
@@ -427,10 +420,10 @@ export class SignOnProvider extends Component<SignOnProps, SignOnState> {
       onClickReadMetaMaskConfig: this.onClickReadMetaMaskConfig,
       closeModal: this.closeModal,
       onNextPage: this.onNextPage,
-      handleOnContinue: this.handleOnContinue,
       onPrevPage: this.onPrevPage,
       onConfigureWithMetaMask: this.onConfigureWithMetaMask,
       onEmailChange: this.onEmailChange,
+      onEmailSubmitted: this.props.onEmailSubmitted,
       onPasswordChange: this.onPasswordChange,
       onNameChange: this.onNameChange,
       onAddFollows: this.props.addFollows,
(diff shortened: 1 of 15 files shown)


src/containers/sign-on/components/desktop/EmailPage.tsx
```diff
@@ -30,53 +30,78 @@ const errorMessages = {
...
 
   onBlur = () => {
-    this.setState({ attempted: true })
+    this.setState({ showValidation: true })
   }
 
-  onKeyDown = e => {
-    if (e.keyCode === 13 /** enter */) {
+  onKeyDown = (e: KeyboardEvent) => {
+    if (e.key === 'Enter') {
```
rickyrombo, line 70: FYI: `keyCode` is deprecated.

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

not sure if I should be using `key` or `code` here - mdn recommends `code` but other sources recommend using `key` - not sure if it matters.

src/containers/sign-on/SignOnProvider.tsx
```diff
@@ -237,7 +237,8 @@ export class SignOnProvider extends Component<SignOnProps, SignOnState> {
     return () => {
       const { email } = this.props.fields
       if (page === Pages.PASSWORD) {
-        this.props.recordCompleteEmail(email.value)
+        this.props.onEmailSubmitted(email.value)
```
rickyrombo, line 240: This breaks the existing pattern and requires the saga to handle recording the event, which I don't like. But we do need to ensure we check the email prior to attempting to move pages...

src/containers/sign-on/store/sagas.js
```diff
@@ -213,13 +212,35 @@ function* validateHandle(action) {
...
+}
+
 function* validateEmail(action) {
   try {
     if (!isValidEmailString(action.email)) {
       yield put(signOnActions.validateEmailFailed('characters'))
       return
     }
-    yield put(signOnActions.validateEmailInUse(action.email))
+    yield delay(2000)
```
rickyrombo, line 242: TODO: Will change back to the constant

audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-signin

src/containers/sign-on/store/sagas.js
```diff
@@ -213,25 +212,30 @@ function* validateHandle(action) {
...
+  if (!isValidEmailString(action.email)) {
+    yield put(signOnActions.validateEmailFailed('characters'))
+    return
+  }
   try {
-    if (!isValidEmailString(action.email)) {
-      yield put(signOnActions.validateEmailFailed('characters'))
-      return
+    const inUse = yield call(AudiusBackend.emailInUse, action.email)
+    if (inUse) {
```
raymondjacobson, line 222: maybe worth commenting here on the expected UX around what happens if the email is taken (goes to sign in)

raymondjacobson (commented):
i think this strategy is good so long as we can confirm with forrest & julian that doing the mobile sign on the same way here is also cool (since originally out of scope). I imagine at the very least we need to turn the copy of that first button from "Sign Up" to "Continue" but we may want to make the buttons wider, etc.

also,
Screen Shot 2021-09-16 at 6 14 53 PM

one funky styling thing i saw (you need to have the extension installed to see this)

raymondjacobson:
there's also some weird difference in the animation in the new version, perhaps b/c the Audius logo doesn't overlap where it was before. not sure what we should do there. thoughts? i think the animations could be cleaned up a bit overall

audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-signin

audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-signin

audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-signin

rickyrombo:
Ran probers locally with https://github.com/AudiusProject/probers/pull/34 and they passed.

audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-signin

raymondjacobson (approved):
LGTM minus white color on spinner maybe needed

PR #583, opened 2021-09-22
Fallback to parsing the ID from URL on 404

rickyrombo:
### Description

Updates the `getTrackByHandleAndSlug` method to fallback to trying to parse the ID from the route if the initial request 404s. This fixes the links sent out by mobile users using outdated app versions.

Also removes one-off since that link should now also work.

### Dragons

### How Has This Been Tested?

Tested manually on -instrumental-prod.-by-mattrickbeats-530271 and -instrumental-prod.-by-mattrickbeats-100000 and -instrumental-prod.-by-mattrickbeats-1 to ensure the proper logic. (First one redirects to the proper track, second 404s, third 404s)

Tempted to add some logs here so that we can see more clearly when we can remove the logic. Thoughts @raymondjacobson ?

### DIFF
--- a/src/services/audius-api-client/AudiusAPIClient.ts
+++ b/src/services/audius-api-client/AudiusAPIClient.ts
@@ -712,7 +712,18 @@ class AudiusAPIClient {
       args,
       true
     )
-    if (!trackResponse) return null
+    // Try the old route method, ensuring that the track once found has the same owner handle
+    if (!trackResponse) {
+      const matches = args.slug.match(/[0-9]{6,}$/)
+      if (!matches) return null
+      const oldTrackResponse = await this.getTrack({
+        id: parseInt(matches[0])
+      })
+      if (!oldTrackResponse || oldTrackResponse.user.handle !== args.handle) {
+        return null
+      }
+      return oldTrackResponse
+    }
     return adapter.makeTrack(trackResponse.data)
   }
 
--- a/src/services/audius-api-client/AudiusAPIClient.ts
+++ b/src/services/audius-api-client/AudiusAPIClient.ts
@@ -713,8 +713,9 @@ class AudiusAPIClient {
       true
     )
     // Try the old route method, ensuring that the track once found has the same owner handle
+    // Ensure at least 5 digits (anything lower has old route in the DB)
     if (!trackResponse) {
-      const matches = args.slug.match(/[0-9]{6,}$/)
+      const matches = args.slug.match(/[0-9]{5,}$/)
       if (!matches) return null
       const oldTrackResponse = await this.getTrack({
         id: parseInt(matches[0])
--- a/scripts/workers-site/index.js
+++ b/scripts/workers-site/index.js
@@ -29,21 +29,6 @@ function checkIsBot(val) {
 }
 
 async function handleEvent(event) {
-  // One-off fix for old-style route shared on twitter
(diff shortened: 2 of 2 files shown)


src/services/audius-api-client/AudiusAPIClient.ts
```diff
@@ -712,7 +712,18 @@ class AudiusAPIClient {
       args,
       true
     )
-    if (!trackResponse) return null
+    // Try the old route method, ensuring that the track once found has the same owner handle
+    if (!trackResponse) {
+      const matches = args.slug.match(/[0-9]{6,}$/)
```
raymondjacobson, line 717: worthy a try catch for this?

src/services/audius-api-client/AudiusAPIClient.ts
```diff
@@ -712,7 +712,18 @@ class AudiusAPIClient {
       args,
       true
     )
-    if (!trackResponse) return null
+    // Try the old route method, ensuring that the track once found has the same owner handle
+    if (!trackResponse) {
```
raymondjacobson, line 716: do we do 6 here b/c we know that all the tracks that would have issues were over 100,000? can we comment on that if that's the case?

is it safer to do split(-)[-1] instead?

raymondjacobson (approved):
We could add an analytics event. We won't see the console.log or console.error if we add something here unless sentry throws down the line. Probably not worth it (?)

src/services/audius-api-client/AudiusAPIClient.ts
rickyrombo: I could query to make sure but let's change to 5 to be certain. I don't like split(-) because that may not return a number and will always match anything with spaces in the title

src/services/audius-api-client/AudiusAPIClient.ts
rickyrombo: Does `match` throw?

raymondjacobson: meant the whole section. i dont think anything could throw here because parseInt returns NaN on failure. just was thinking aloud

src/services/audius-api-client/AudiusAPIClient.ts
raymondjacobson: but it would already be in the case of a 404 so something was already pretty wonky w/ the url? a try/catch/NaN check and going to 404 would suffice there as well

src/services/audius-api-client/AudiusAPIClient.ts
rickyrombo: (the regex shouldn't match any non-int, unless someone uses a number larger than an int can store I guess)

src/services/audius-api-client/AudiusAPIClient.ts
rickyrombo: I think matching digits is the safer route? What's the benefit of doing multiple checks/requests?

raymondjacobson: i think this is probably fine -- but i think potentially had some unthought of edge cases vs. split('-') being super straightforward (to me). im down to ship this the way it is

audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-old-routes-still-work

rickyrombo:
We can add analytics in a fast follow/next release if we find it needed I think.

PR #598, opened 2021-09-29
Use dev tools smart prod mode

raymondjacobson:
### Description

Per https://medium.com/@zalmoxis/using-redux-devtools-in-production-4c5b56c5600f, we do a bit less work if we use this version of the dep
```
The difference is that it will not include the instrumentation enhancer, so will not affect the performance and will not store anything if not required.
```

### Dragons

### How Has This Been Tested?

ray.audius.co


audius-infra:
Preview this change https://frontend-demo.audius.co/rj-dev-tools-prod

sliptype (approved):

PR #614, opened 2021-10-18
Libs wormhole

jowlee:
### Description

* Updates/adds prod env vars for solana and bridge
* Adds the libs updated wormhole config to libs constructor
* Updates the libs transfer to solana call

**NOTE**: Need to wait on libs update which is waiting on wormhole update

Closes AUD-997

### Dragons
Switches to using `craco` instead of react react app scripts - this is so that we can use the wasm loader

### How Has This Been Tested?
Ran it locally against prod and transferred tokens


raymondjacobson (approved):

audius-infra:
Preview this change https://demo.audius.co/jowlee-wormhole

PR #615, opened 2021-10-18
[AUD-666] Update sign up flow to auto-follow Hot & New Playlist

Kyle-Shanks:
### Description
[AUD-666]
Adding a new saga and updating sign up saga to follow the Hot & New playlist

### How Has This Been Tested?

Manually tested

### DIFF
--- a/src/containers/sign-on/store/sagas.js
+++ b/src/containers/sign-on/store/sagas.js
@@ -13,12 +13,13 @@ import {
 } from 'redux-saga/effects'
 
 import * as accountActions from 'common/store/account/reducer'
+import { retrieveCollections } from 'common/store/cache/collections/utils'
 import { fetchUserByHandle, fetchUsers } from 'common/store/cache/users/sagas'
 import { processAndCacheUsers } from 'common/store/cache/users/utils'
 import { getIGUserUrl } from 'components/general/InstagramAuth'
 import AudiusBackend from 'services/AudiusBackend'
 import { getCityAndRegion } from 'services/Location'
-import { Name } from 'services/analytics'
+import { FavoriteSource, Name } from 'services/analytics'
 import apiClient from 'services/audius-api-client/AudiusAPIClient'
 import { getRemoteVar, IntKeys, StringKeys } from 'services/remote-config'
 import { fetchAccountAsync } from 'store/account/sagas'
@@ -27,6 +28,7 @@ import * as backendActions from 'store/backend/actions'
 import { waitForBackendSetup } from 'store/backend/sagas'
 import * as confirmerActions from 'store/confirmer/actions'
 import { confirmTransaction } from 'store/confirmer/sagas'
+import { saveCollection } from 'store/social/collections/actions'
 import * as socialActions from 'store/social/users/actions'
 import { setHasRequestedBrowserPermission } from 'utils/browserNotifications'
 import { isValidEmailString } from 'utils/email'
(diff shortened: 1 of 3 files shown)


src/containers/collection-page/components/desktop/CollectionPage.tsx
```diff
@@ -126,6 +126,7 @@ const CollectionPage = ({
   onClickFavorites,
   onClickDescriptionExternalLink
 }: CollectionPageProps) => {
+  console.log({ playlistId })
```
raymondjacobson, line 129: log

src/containers/sign-on/store/sagas.js
```diff
@@ -60,6 +62,8 @@ const messages = {
 // Users ID to filter out of the suggested artists to follow list and to follow by default
 /* user id 51: official audius account */
 const defaultFollowUserIds = new Set([51])
+/* user id 12372: official audius account */
+const stagingDefaultFollowUserIds = new Set([12372])
```
raymondjacobson, line 66: can we instead define defaultFollowUserIds based on env?

Otherwise, we're polluting the variable space w/ a staging variable on production builds too

let defaultFollowUserIds = new Set([])
if (process.env.ENV === 'production') defaultFollowUserIds = new Set([51])
if (process.env.ENV === 'staging') defaultFollowUserIds = new Set([12372])

src/containers/sign-on/store/sagas.js
```diff
@@ -484,6 +524,17 @@ function* watchConfigureMetaMask() {
   yield takeLatest(signOnActions.CONFIGURE_META_MASK, configureMetaMask)
 }
 
+function* watchFollowHotAndNew() {
+  while (
+    yield all([
```
raymondjacobson, line 529: I think I would prefer that we update the `followArtists` saga and use the existing flow there rather than having a different paradigm for playlist auto follows

I also this should be more generic -- some sort of `followCollections` saga that can be agnostic to the specific playlist/album it is (like how we do artists)

src/containers/sign-on/store/sagas.js
```diff
@@ -409,6 +422,25 @@ function* signIn(action) {
...
 
+function* followHotAndNew() {
+  if (process.env.REACT_APP_ENVIRONMENT !== 'production') return
+  yield call(waitForBackendSetup)
+  try {
+    // Hot & New playlist id
+    const hotNewId = 4281
+    const result = yield retrieveCollections(null, [hotNewId])
+
+    if (
```
raymondjacobson, line 433: why this check instead of just a truthy/existence check on result?

Kyle-Shanks: No real reason, just for safety to make sure that the playlist for that id is the one we're expecting

src/containers/sign-on/store/sagas.js
```diff
@@ -432,6 +457,12 @@ function* followArtists() {
       const userIndex = userIdsToFollow.findIndex(fId => fId === userId)
       if (userIndex > -1) hasFollowConfirmed[userIndex] = true
     }
+
+    // Auto-follow Hot & New Playlist
+    if (IS_PRODUCTION) {
```
raymondjacobson, line 462: almost there, but we should do this in parallel w/ the previous follow users. that would be a 

yield fork(...) above when we do that work. or alternatively pulling that work into a forked saga

PR #619, opened 2021-10-20
Change Password UI

rickyrombo:
### Description

**Note:** it's not quite ready for an in-depth review but a quick review for design structure etc would be appreciated!

Latest at https://marcus.audius.co

Creates a button in the Settings page that allows a user to change their password. Note that unlike a password reset, changing a password marks the former password as deleted.

See also:

Hedgehog changes to support sending the old password on reset and a new password confirmation method:
- https://github.com/AudiusProject/hedgehog/pull/36 
- https://github.com/AudiusProject/hedgehog/pull/37
- https://github.com/AudiusProject/hedgehog/pull/40

Identity and libs changes to support deleting old lookup key when sent:
- https://github.com/AudiusProject/audius-protocol/pull/1968

Stems change to add the lock icon
- https://github.com/AudiusProject/stems/pull/60

**TODO:**
- [ ] Animations between pages
- [x] Fix bug where user is logged out if changing password fails
- [x] Test with Identity changes https://github.com/AudiusProject/audius-protocol/pull/1968/files
- [x] Add metrics for usage, success, failure
- [x] Discuss whether we need to check if the identity service handles the deletion properly?
- [x] Change OnCompleted to separate OnFailure/OnSuccess
- [x] Split tiles on the settings screen into two columns and make buttons the same width

### Dragons

- ❗ Users can get locked out if this change isn't done correctly! ❗ 
- Updates libs for the hedgehog changes
- Updates stems for the lock icon

### How Has This Been Tested?

- Manually tested against staging accounts

- Added two amplitude metrics for usage and success/failure:
  - Settings: Start Change Password
  - Settings: Complete Change Password

### DIFF
--- a/src/containers/change-password/ChangePassword.tsx
+++ b/src/containers/change-password/ChangePassword.tsx
@@ -1,6 +1,6 @@
 import React, { useCallback, useEffect, useState } from 'react'
 
-import { Button, IconLock } from '@audius/stems'
+import { Button, ButtonType, IconLock } from '@audius/stems'
 import cn from 'classnames'
 import { useDispatch, useSelector } from 'react-redux'
 
@@ -95,7 +95,13 @@ export const ChangePassword = ({
         return <LoadingSpinnerFullPage />
       case Page.FAILURE:
       case Page.SUCCESS:
-        return <Button text='Done' onClick={onComplete} />
+        return (
+          <Button
+            text='Done'
+            onClick={onComplete}
+            type={ButtonType.PRIMARY_ALT}
+          />
+        )
       default:
         return (
           <ConfirmCredentials
--- a/src/containers/change-password/ConfirmCredentials.module.css
+++ b/src/containers/change-password/ConfirmCredentials.module.css
@@ -6,6 +6,7 @@
 .continueButton {
   display: block;
   margin: var(--spacing) auto 0;
+  font-size: var(--font-l);
 }
 .continueButton.isMobile {
   width: 100%;
--- a/src/containers/change-password/ConfirmCredentials.tsx
+++ b/src/containers/change-password/ConfirmCredentials.tsx
@@ -129,7 +129,7 @@ export const ConfirmCredentials = (props: ConfirmCredentialsProps) => {
           )
         }
         isDisabled={isSubmitDisabled}
-        type={isSubmitDisabled ? ButtonType.DISABLED : ButtonType.PRIMARY}
+        type={isSubmitDisabled ? ButtonType.DISABLED : ButtonType.PRIMARY_ALT}
         onClick={onSubmit}
       />
     </form>


src/containers/change-password/ChangePasswordModal.tsx
```diff
@@ -0,0 +1,48 @@
...
+      dispatch(changePage(Page.CONFIRM_CREDENTIALS))
+    }
+  }, [dispatch, showModal])
+
+  return (
+    <Modal
+      title={messages.title}
+      showTitleHeader
+      showDismissButton={allowClose}
+      dismissOnClickOutside={allowClose}
```
sliptype, line 41: I was still able to hit Escape to close the modal on the NEW_PASSWORD page

src/containers/change-password/ChangePassword.module.css
```diff
@@ -0,0 +1,38 @@
+.content:not(.isMobile) {
+  padding: 24px;
+  text-align: center;
+  --spacing: 24px;
```
raymondjacobson, line 4: niceee

src/containers/change-password/ChangePassword.tsx
```diff
@@ -0,0 +1,127 @@
...
+import { getChangePasswordStatus, getCurrentPage } from './store/selectors'
+import { changePage, changePassword, Page } from './store/slice'
+
+const messages = {
+  helpTexts: [
+    'Please enter your email and current password.',
+    'Create A New Password That Is\n Secure And Easy To Remember!',
+    'Changing Password, Please wait',
+    'Your Password Has Been Changed',
+    <span key={Page.FAILURE} className={styles.error}>
```
raymondjacobson, line 22: this is an interesting strategy. not entirely sure how much I like it -- maybe better to pull these out into strings and leave the layout for the component. the high level reason we pull messages out into a const is because someday we will probably want to i18n them

src/containers/change-password/ChangePasswordModal.tsx
raymondjacobson: yes lol -- unrelated bug we def need to fix

src/containers/change-password/store/slice.ts
```diff
@@ -0,0 +1,114 @@
...
+const slice = createSlice({
+  name: 'application/ui/changePassword',
+  initialState,
+  reducers: {
+    confirmCredentials: (
+      state,
+      _action: PayloadAction<{ email: string; password: string }>
+    ) => {
+      return {
+        ...state,
```
raymondjacobson, line 42: same comments here for this PR around whether we should be doing the immutablejs style of state transform that we're allowed to. maybe we should take this to product eng sync

src/containers/settings-page/components/desktop/SettingsCard.module.css
```diff
@@ -1,5 +1,8 @@
 .settingsCard {
-  height: 112px;
+  flex: 1 470px; /* 246px (description) + 48px (padding) + 176px (button) */
```
raymondjacobson, line 2: this layout is so much better lol. + the wrapping is excellent. good spot

src/containers/settings-page/components/desktop/SettingsPage.module.css
```diff
@@ -60,11 +56,11 @@
 .version span {
   line-height: 20px;
 }
-
+/*
```
raymondjacobson, line 59: remove?

raymondjacobson (approved):
Looks fantastic, and super functional! I like how well this flow came together.

Two minor design issues i noticed
1. this button should be PRIMARY_ALT. we basically don't use PRIMARY anywhere (maybe just 1 or two spots, and i think we should be eventually replacing those)
Screen Shot 2021-10-22 at 5 16 47 PM


2. There's a brief flash of the pink button when advancing from the first page of the change PW flow on web + mobile. Like the spinner spins for a bit as gray and then as state is reset, flashes pink before the page transitions

not blockers on initial merge

src/containers/change-password/store/slice.ts
rickyrombo: Oh yeah, old habits die hard whoops

src/containers/change-password/ChangePassword.tsx
rickyrombo: Yeah I don't love it either. I'll revisit

src/containers/change-password/ChangePasswordModal.tsx
rickyrombo: I ask because I wonder if the other bug leads to this one - for me the first escape brings up the other modal, and the second escape closes both of them

raymondjacobson: same

PR #620, opened 2021-10-21
[AUD-1000] Fix bad follower count bug

raymondjacobson:
### Description

Corresp. libs changes: https://github.com/AudiusProject/audius-protocol/pull/1988

This was a somewhat gnarly one. See inline comments for explanations of the changes.

### Dragons

### How Has This Been Tested?

- Made several against locally vs. stage w/ varying amounts of followed users
- Unit tests still run (and test the adjustUserField change)

### DIFF
--- a/src/common/store/cache/users/sagas.js
+++ b/src/common/store/cache/users/sagas.js
@@ -221,19 +221,14 @@ function* watchSyncLocalStorageUser() {
 // The cache respects the delta and merges the objects adding the field values
 export function* adjustUserField({ user, fieldName, delta }) {
   yield put(
-    cacheActions.update(
-      Kind.USERS,
-      [
-        {
-          id: user.user_id,
-          metadata: {
-            [fieldName]: delta
-          }
+    cacheActions.increment(Kind.USERS, [
+      {
+        id: user.user_id,
+        metadata: {
+          [fieldName]: delta
         }
-      ],
-      [],
-      /* isDelta */ true
-    )
+      }
+    ])
   )
 }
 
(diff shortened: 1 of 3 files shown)


src/common/store/cache/users/sagas.js
```diff
@@ -217,16 +217,23 @@ function* watchSyncLocalStorageUser() {
   yield takeEvery(cacheActions.UPDATE, syncLocalStorageUser)
 }
 
+// Adjusts a user's field in the cache by specifying an update as a delta.
+// The cache respects the delta and merges the objects adding the field values
 export function* adjustUserField({ user, fieldName, delta }) {
```
raymondjacobson, line 222: First important change. When following users, we call "adjustUserField" on our account to increment our "followee_count" by 1 for each user we have followed.

This saga is buggy. It effectively tries to be redux (pull state, update value w/ delta, write state). The problem is that between pulling state and updating state, because this is a saga, other stuff could have happened and that initial pulled state is not correct. This needs to be done synchronously, so I've added that to the reducer.

src/containers/sign-on/store/sagas.js
```diff
@@ -468,7 +467,11 @@ function* followArtists() {
     // The update user location depends on the user being discoverable in discprov
     // So we wait until both the user is indexed and the follow user actions are finished
     yield call(AudiusBackend.updateUserLocationTimezone)
-    yield call(fetchAccountAsync, { fromSource: true })
+
```
raymondjacobson, line 470: Second important change. We don't want to call `fetchAccountAsync` again here after following users because the aggregate_user view in discovery node will probably not have updated (yet).

Instead, we want to make sure that after we've followed everyone, our local storage persisted user is refreshed with the changes from `adjustUserField` that implicitly got called by all these follows

audius-infra:
Preview this change https://frontend-demo.audius.co/rj-aud-1000

rickyrombo (approved):

src/containers/sign-on/store/sagas.js
raymondjacobson: yeah. before the aggregate views (/ or got slow), this was not an issue because refetching would have the counts updated

PR #636, opened 2021-10-26
Consume latest libs with Solana Relay changes

piazzatron:
### Description

Consumes latest libs (currently unpublished, will update this PR with package-lock when it is)

### Dragons

None

### How Has This Been Tested?

Locally, a lot


audius-infra:
Preview this change https://frontend-demo.audius.co/piazz-rewards-libs-integration

PR #640, opened 2021-11-01
Add GUI to facilitate selection of different discovery node

sddioulde:
### Dragons

### How Has This Been Tested?

Screen Shot 2021-11-01 at 5 09 40 PM
Screen Shot 2021-11-01 at 5 10 08 PM

fixed typo in image :)


audius-infra:
Preview this change https://frontend-demo.audius.co/dn-selection-preview

src/containers/discovery-node-selection-preview/DiscoveryNodeSelectionPreview.tsx
```diff
@@ -0,0 +1,84 @@
+import React, { useState } from 'react'
+
+import { Button, ButtonType } from '@audius/stems'
+import { ChangeEvent } from 'react-slick/node_modules/@types/react'
+
+import { usePreviewHotkey } from 'hooks/useHotkey'
+
+import styles from './DiscoveryNodeSelectionPreview.module.css'
+
+const DISCOVERY_NODE_SELECTION_ENABLE_KEY =
```
raymondjacobson, line 10: can we just have 1 local storage var called like "enable-dev-mode" or something like that

src/containers/discovery-node-selection-preview/DiscoveryNodeSelectionPreview.tsx
```diff
@@ -0,0 +1,84 @@
...
+    68 /* d */,
+    DISCOVERY_NODE_SELECTION_ENABLE_KEY
+  )
+  const [endpoint, setEndpoint] = useState('')
+  const [success, setSuccess] = useState(false)
+  const [error, setError] = useState(false)
+
+  const handleDiscoveryNodeSelection = async () => {
+    const healthyEndpoints = (
+      await (await fetch('https://api.audius.co')).json()
```
raymondjacobson, line 24: i dont want to vendor api.audius.co like this into the client. libs has the capacity to go off of the chain to determine this -- but i actually think this tool would be most useful if we *dont* cross reference it for health. it would allow you to specific any node (even during dev)

src/hooks/useHotkey.ts
```diff
@@ -30,4 +30,23 @@ const useHotkeys = (mapping: Mapping) => {
   }, [mapping])
 }
 
+const ENABLE_DEV_MODE_KEY = 'enable-dev-mode'
+export const usePreviewHotkey = (keyCode: number) => {
```
raymondjacobson, line 34: nit useDevModeHotkey maybe

raymondjacobson (approved):
This is awesome! Follow up change would be to have this also set some artifact that would cause libs to not reselect elsewhere if the entered node is behind by blocks, but this already is super useful

raymondjacobson:
also, maybe just call this "Discovery Node Selection" not preview

audius-infra:
Preview this change https://frontend-demo.audius.co/dn-selection-preview

audius-infra:
Preview this change https://frontend-demo.audius.co/dn-selection-preview

PR #653, opened 2021-11-10
[AUD-1069] Fix missing repost on initial load

raymondjacobson:
### Description

Fix bug where a track does not appear to be reposted immediately after reposting and reloading. The source of this issue (and the reason probers have been flaky most of the time) is because the current user id (if applicable) is not passed to the handle/slug endpoint, so `has_current_user_reposted` does not return as true.

My suspicion for why this has gone unnoticed for so long is that likely the track is being fetched by another endpoint the majority of the time on production and that is overwriting the `has_current_user_x` field

### Dragons

### How Has This Been Tested?

* Locally vs. staging as a logged in account
* Locally vs. staging logged out


rickyrombo (approved):

audius-infra:
Preview this change https://frontend-demo.audius.co/rj-aud-1069

PR #656, opened 2021-11-11
Fix nav header margin when not logged in

sddioulde:
### Description

Fix nav header margin when not logged in. Also tweak fade in.

### Dragons

### How Has This Been Tested?

Locally without and with logged in user.


audius-infra:
Preview this change https://frontend-demo.audius.co/sd-fix-margin

raymondjacobson (approved):

PR #657, opened 2021-11-12
Deactivate Account

rickyrombo:
### Description

- Adds `/deactivate` page as a way to deactivate user accounts
- Adds client-side filtering of deactivated account content, such as filtering tracks from feeds, explore playlists, and trending, and showing tombstones for tracks in existing playlists and favorites, and showing tombstones of the profile in place of the user profile. A comprehensive list is in the "How Has This Been Tested?" section.

Recommend looking at individual commits to see the goal of each change.

### Dragons

- Deactivated users should never show up in UI anywhere, and same for their playlists and tracks.
- ❗ When deactivated user names do appear (eg when tracks are in existing playlists of other users), we need to make sure that there aren't any artist popovers (there shouldn't be)

### How Has This Been Tested?

Created a bunch of users on my local and had the deactivated one be very active. Manually tested the following, left unchecked for retesting (and you should test these too - ping me for the creds to the users I used on stage):
- [ ] Deactivated account reposts/saves don't show up in status texts on track tiles (desktop/mobile)
- [ ] Deactivated account remix parents don't show up on track pages (desktop/mobile)
- [ ] ~~Deactivated account activity doesn't show in feeds (with https://github.com/AudiusProject/audius-protocol/pull/2062)~~ 
    - @raymondjacobson and I decided that this requirement should be cut, as feed calculations are slow and this will make them much slower and items will fall out of the feed over time anyway.
- [ ] Deactivated account playlists and tracks don't show in feeds (desktop/mobile)
- [ ] Deactivated account tracks show as deleted in collections (desktop/mobile)
- [ ] Deactivated account collections are hidden from the nav (desktop)
- [ ] Deactivated account tracks, profiles, collections are hidden from search bar and search page (desktop/mobile)
- [ ] Deactivated account profiles don't show in follower/following modals (desktop) or lists (mobile)
- [ ] Deactivated account tracks redirect to user profiles (desktop/mobile)
- [ ] Deactivated account collections redirect to user profiles (desktop/mobile)
- [ ] Deactivated account tracks in playlists don't have artist name popover (desktop)
- [ ] Deactivated account profile page does not show bio, creation date, all social handles, all nav tabs, name, handle, verification status, badges, cover photo, profile photo, tracks (desktop/mobile)
- [ ] Deactivated account tracks don't show up in trending and the ranks don't jump
- [ ] Deactivated account tracks don't show up in underground trending and the ranks don't jump
- [ ] Deactivated account playlists don't show up in trending playlists and the ranks don't jump
- [ ] Deactivated account tracks don't show up in smart collections ("Heavy Rotation", "Let them DJ", "Best New Releases", "Under the Radar", "Remixables", "Most Loved", "Feeling Lucky?")
- [ ] Deactivated account playlists don't show up in top albums
- [ ]  Deactivated accounts don't show up in artist recommendations upon following another user
    - [ ]  For users with ≥ 200 followers
    - [ ]  For users with < 200 followers
- [ ]  Deactivated accounts don't show up in artist suggestions when the feed is empty (eg. user doesn't follow any artists)
- [ ]  Deactivated account albums don't show up in user's album favorites
- [ ]  Deactivated account tracks don't have Embed or View Artist Page options in track menus for track tables (eg Favorites, History)

TODO:
- [x] Test deactivated account tracks/playlists don't appear in trending page and/or explore pages
- [x] Create tombstone for deactivate account profile pages
- [x] ~~Create page for deactivate account~~
  - Cutting and saving for second PR, this one is large enough
- [x] ~~Handle deactivated user sign in~~
  - Cutting and saving for second PR

KNOWN ISSUES:
- [x] ~~Deactivated user profile pictures show up in repost/favorite previews~~
- [x] Explore remixables will show deactivated tracks
- [x] Deactivated account playlists flash in the side navbar before being removed

Adds the following metrics to amplitude:
- 'Deactivate Account: Page View' for when the `/deactivate` page is viewed
- 'Deactivate Account: Request' for when the user starts deactivating their account
- 'Deactivate Account: Success' for when the user successfully deactivated their account
- 'Deactivate Account: Failure' for when the user failed to deactivate their account


audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-tombstones

src/components/track/desktop/stats/Stats.tsx
```diff
@@ -80,9 +80,9 @@ const Stats = memo(
                 contentTitle={contentTitle}
                 // Map out all of the users so that the selector is cheaper inside the
                 // rendered component
-                userId1={slice[0] ? slice[0].user_id : undefined}
-                userId2={slice[1] ? slice[1].user_id : undefined}
-                userId3={slice[3] ? slice[2].user_id : undefined}
```
sddioulde, line 85: hmm i know this is removed code but this line is weird `userId3={slice[3] ? slice[2].user_id : undefined}`. we're checking slice[3] and grabbing slice[2]'s user id. calling out to make sure that you're not missing anything in your refactor.

src/containers/user-list/UserList.tsx
```diff
@@ -92,7 +92,10 @@ function mapStateToProps(state: AppState, ownProps: ConnectedUserListOwnProps) {
   const { hasMore, loading, userIds } = ownProps.stateSelector(state)
   const userId = getUserId(state)
   const usersMap: { [id: number]: User } = getUsers(state, { ids: userIds })
```
sddioulde, line 94: do you think it's worth adding a getNonDeactivatedUsers or getActiveUsers selector?

sddioulde: The other thing is: when do we care about deactivated users? If this is almost never then the above comment holds more weight and we just use that selector that returns only active users in most parts of the app. Curious what others think.

I see we've update the makeGetSearchPlaylists (and other similar ones) selector to include the deactivated filter.

src/components/track/desktop/ConnectedTrackTile.tsx
```diff
@@ -166,7 +166,7 @@ const ConnectedTrackTile = memo(
         includeShareToTikTok: !is_unlisted,
         includeTrackPage: true,
         isArtistPick: isArtistPick,
-        isDeleted: is_delete,
+        isDeleted: is_delete || !!user?.is_deactivated,
```
Kyle-Shanks, line 169: is the `!!` needed? `is_deactivated` should be a boolean right?

src/components/tracks-table/TracksTable.js
```diff
@@ -120,7 +120,7 @@ const repostButtonCell = (val, record, props) => {
 }
 
 const optionsButtonCell = (val, record, index, props) => {
-  const { is_delete: deleted } = record
+  const deleted = record.is_delete || !!record.user?.is_deactivated
```
sddioulde, line 123: any reason why some lines have the double bang operator while others don't? e.g. this one v. repostButtonCell

Kyle-Shanks:
Looks good to me so far. Only comment is wondering if there was a reason for sometimes using `!!`

Kyle-Shanks (commented):
Looks good to me so far. Only comment is wondering if there was a reason for sometimes using `!!`

src/containers/track-page/TrackPageProvider.tsx
```diff
@@ -409,10 +412,12 @@ class TrackPageProvider extends Component<
     })
     const canonicalUrl = user && track ? fullTrackPage(track.permalink) : ''
 
-    // If the track has a remix parent and it's not deleted.
+    // If the track has a remix parent and it's not deleted and the original's owner is not deactivated.
     const hasValidRemixParent =
       !!getRemixParentTrackId(track) &&
-      (!remixParentTrack || remixParentTrack.is_delete === false)
+      !!remixParentTrack &&
```
sddioulde, line 418: Did the logic change a bit here (outside of deactivated)?. In the previous code if there is !remixParentTrack is falsy it's still possible for hasValidRemixParent to be true. In your change that's no longer the case

src/containers/track-page/store/selectors.ts
```diff
@@ -22,10 +22,17 @@ export const getTrack = (state: AppState) => {
...
-  const id = getTrackId(state)
-  const cachedTrack = getCachedTrack(state, { id })
+  const cachedTrack = getTrack(state)
   const parentTrackId = cachedTrack?.remix_of?.tracks?.[0].parent_track_id
-  return parentTrackId ? getCachedTrack(state, { id: parentTrackId }) : null
+  if (parentTrackId) {
+    const parentTrack = getCachedTrack(state, { id: parentTrackId })
+    // Get user for deactivated status
+    const parentTrackUser = getCachedUser(state, { id: parentTrack?.owner_id })
+    if (parentTrack && parentTrackUser) {
```
sddioulde, line 31: Do we also want to check that parentTrackUser.is_deactivated is false or is that somehow taken care of?

sddioulde (commented):
Looks neat. Let me know if you have questions about my comments

src/components/track/desktop/stats/Stats.tsx
rickyrombo: Good callout, was wondering about this

src/containers/track-page/store/selectors.ts
rickyrombo: It's taken care of in TrackPageProvider

src/components/track/desktop/ConnectedTrackTile.tsx
rickyrombo: Yeah that's correct, I should be able to remove these now. Was thinking it could be null/undefined

src/containers/user-list/UserList.tsx
rickyrombo: I like this idea. I might have to look into it and double check this would have the same behavior as checking `is_deactivated` explicitly everywhere

src/components/tracks-table/TracksTable.js
rickyrombo: Some places (like here) pass the result value into the `isDeleted` prop of a component, which expects a boolean (and since I user `user?` it could be undefined). I'm curious though whether the user prop is ever actually undefined/null

src/components/track/desktop/ConnectedTrackTile.tsx
rickyrombo: Right, the `user` might be null/undefined (or is it possible?), so that's why I double banged

src/containers/track-page/TrackPageProvider.tsx
rickyrombo: Yeah the logic changed here, good catch, but I think it's more correct per the comment above? Not super familiar with this line.

src/containers/user-list/UserList.tsx
rickyrombo: Hmm although there are places in the PR where I getUsers for tracks that I then want to check for `is_deactivated` that instead I'd be checking for existence, and that seems to obscure what's happening, so maybe the two selector route?

src/components/track/desktop/stats/Stats.tsx
rickyrombo: CC: @raymondjacobson was this intentional?

src/components/track/desktop/stats/StatsText.tsx
```diff
@@ -118,9 +116,7 @@ const StatsText = memo(
 
 function mapStateToProps(state: AppState, ownProps: OwnProps) {
   return {
-    user1: getUser(state, { id: ownProps.userId1 }),
-    user2: getUser(state, { id: ownProps.userId2 }),
-    user3: getUser(state, { id: ownProps.userId3 })
+    users: Object.values(getUsers(state, { ids: ownProps.userIds }))
```
raymondjacobson, line 119: this was actually on purpose. we should play with the performance a bit, but the reason was that

`getUser` will return the exact cached entry (https://github.com/AudiusProject/audius-client/blob/f36aaf2dd555a6269edc7cca81b87935407e75ce/src/common/store/cache/users/selectors.ts#L15) that is the same object ref (https://github.com/AudiusProject/audius-client/blob/f36aaf2dd555a6269edc7cca81b87935407e75ce/src/common/store/cache/selectors.ts#L60) each time. Only when that individual user in the cache gets replaced, we would rerender here

versus `getUsers` which will return a *new* object each time: https://github.com/AudiusProject/audius-client/blob/f36aaf2dd555a6269edc7cca81b87935407e75ce/src/common/store/cache/users/selectors.ts#L43, which will cause a rerender every time a user is added to the cache (unfortunate).

Ideally we would connect this component all the way down to each user level to prevent this.

src/containers/search-bar/store/sagas.js
```diff
@@ -22,10 +22,18 @@ export function* getSearchResults(searchText) {
   })
 
   const { tracks, albums, playlists, users } = results
+  const checkedUsers = users.filter(t => !t.is_deactivated)
   const checkedTracks = (yield call(setTracksIsBlocked, tracks)).filter(
-    t => !t.is_delete && !t._blocked
+    t => !t.is_delete && !t._blocked && !t.user.is_deactivated
   )
-  return { users, tracks: checkedTracks, albums, playlists }
+  const checkedPlaylists = playlists.filter(t => !t.user?.is_deactivated)
```
raymondjacobson, line 29: since there's necessity to support this pattern for tracks, i kind of find this change to be super reasonable -- but maybe we should just have a generic helper that lives somewhere near-ish the cache that's like

checkUsers = (users: User[]) => users.filter(u => !u.is_deactivated)

and we just call that across the board (?)

src/containers/user-list/UserList.tsx
raymondjacobson: it almost feels like maybe we should have the User type be
`type User = ActiveUser | DeactivatedUser` and force us to handle both in future UI we build

just a thot

src/components/track/desktop/ConnectedTrackTile.tsx
raymondjacobson: in practice we should always have the user here. reason we have this is once upon a time we rendered the track tile w/o having the full user (yet). we should clean up these types, but this change seems ~reasonable for now.

FWIW, we have this weird fallback code above https://github.com/AudiusProject/audius-client/pull/657/commits/89ef4e1029082d7675a46f28f6be4efcbc08985c#diff-05e159c9c0c0c18c159b331ace712a645a34ba2a3b9f41357a25e267c82f44daR114

src/components/track/desktop/stats/StatsText.tsx
rickyrombo: Oof good callout - could we avoid rerenders by calling `ownProps.userIds.map(id => getUser(state, { id }))`

raymondjacobson: wouldn't the `.map` produce a new list each time as well?

raymondjacobson: i think you'd need to use reselect / similar to control the comparison

rickyrombo: yeah that's what I was guessing

audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-tombstones

src/components/track/desktop/stats/StatsText.tsx
rickyrombo: https://github.com/AudiusProject/audius-client/pull/657/commits/f8c55e44dbadb19b4b6650d4373a24363df5517b Should that do it?

rickyrombo: My guess is no, because it still depends on all of `getUsers` but maybe I'm wrong?

audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-tombstones

rickyrombo:
Added a few fixes (see commits) including proper tombstones and deployed to marcus.audius.co

audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-tombstones

audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-tombstones

audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-tombstones

src/components/track/desktop/ConnectedPlaylistTile.tsx
```diff
@@ -461,6 +466,10 @@ const ConnectedPlaylistTile = memo(
           : 'PLAYLIST'
         : undefined
 
+    // Failsafe check - should never get this far, lineups should filter deactivated playlists
+    if (isOwnerDeactivated) {
```
raymondjacobson, line 470: note to make this & ConnectedTrackTile the same

audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-tombstones

audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-tombstones

audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-tombstones

src/components/track/desktop/stats/Stats.tsx
```diff
@@ -47,7 +60,17 @@ const Stats = memo(
       [styles.large]: size === 'large'
     })
 
-    const slice = followeeActions.slice(0, MAX_REPOST_IMAGES)
+    const getFolloweeActionsUsers = useMemo(makeFolloweeActionsUsers, [])
```
sliptype, line 63: Why does this have to be a memo? Can you just define `getFolloweeActionUsers` outside the component?

src/containers/profile-page/components/DeactivatedProfileTombstone.tsx
```diff
@@ -0,0 +1,35 @@
...
+        type={ButtonType.PRIMARY_ALT}
+        size={ButtonSize.MEDIUM}
+        text={messages.buttonText}
+        rightIcon={<IconArrowWhite />}
+      />
+    </div>
+  )
+}
+
+DeactivatedProfileTombstone.defaults = { isMobile: false }
```
sliptype, line 35: No need to change but IMO we should prefer doing default arguments instead of the `defaults` property. Especially for functional components:

```typescript
export const DeactivatedProfileTombstone = ({
  goToRoute,
  isMobile = false
}: {
  goToRoute: (route: string) => void
  isMobile?: boolean
}) => {
...
}
```

src/store/lineup/sagas.js
```diff
@@ -31,11 +32,22 @@ const flatten = list =>
...
-
+  const users = yield select(getUsers)
   return tracksMetadata
     .map(metadata => {
       // If we said to remove deleted tracks and it is deleted, remove it
       if (removeDeleted && metadata.is_delete) return null
+      // If we said to remove deleted and the track/playlist owner is deactivated, remove it
+      else if (
+        removeDeleted &&
+        users[metadata.owner_id]?.is_deactivated === true
```
sliptype, line 43: Is the `=== true` necessary? What other truthy value could `is_deactivated` take?

sliptype (approved):
Nice work!

src/components/track/desktop/stats/Stats.tsx
rickyrombo: I think this is necessary as there will be several `Stats` components grabbing different users, and reselect would cause the selectors to all share the same cache (size of 1) if I didn't make a new one per instance.

> However, when the selector is used in multiple component instances and depends on the component's props, you need to ensure that each component instance gets its own selector instance (see here for a more thorough explanation of why this is necessary):

https://react-redux.js.org/api/hooks#using-memoizing-selectors

src/components/track/desktop/stats/StatsText.tsx
rickyrombo: Updated to use createShallowSelector and verified it reduced rerenders

raymondjacobson (approved):

audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-tombstones

audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-tombstones

audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-tombstones

src/containers/user-list/UserList.tsx
rickyrombo: That makes a lot of sense if we end up stripping properties from resources that are deactivated and deleted. Might follow up with that when we start doing that.

audius-infra:
Preview this change https://frontend-demo.audius.co/mjp-tombstones

PR #669, opened 2021-11-20
[AUD-1071] Improve Avenir font load to avoid FOUT

raymondjacobson:
### Description

On prod nowadays b/c our main css bundle is pretty heavy, we have a decent amount of FOUT (flash of unstyled text).
I think there are two camps around the base64 encoding of fonts like we do (I generally like it, primarily b/c it's self-contained), but regardless, we should be pulling the font in far earlier than we do in the main css bundle.

Some best practices stolen from 
https://css-tricks.com/how-to-load-fonts-in-a-way-that-fights-fout-and-makes-lighthouse-happy/

### Dragons

### How Has This Been Tested?

ray.audius.co

### DIFF
--- a/public/fonts.css
+++ b/public/fonts.css
@@ -5,7 +5,6 @@
 
 /**
  * Avenir font
- * The main DApp font, prefetched in css bundles before rendering
  */
 @font-face {
   font-family: 'Avenir Next LT Pro';
--- a/public/index.html
+++ b/public/index.html
@@ -3,6 +3,14 @@
   <head>
     <meta charset="utf-8">
 
+    <!-- Avenir Font -->
+    <link rel="preload" as="style" href="/fonts.css" />
+    <link rel="stylesheet" href="/fonts.css" media="print" onload="this.media='all'" />
+    <noscript>
+      <link rel="stylesheet"
+            href="/fonts.css" />
+    </noscript>
+
     <link rel="apple-touch-icon" sizes="180x180" href="%PUBLIC_URL%/favicons/apple-touch-icon.png">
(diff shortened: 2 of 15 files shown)


raymondjacobson:
@sliptype  let me know if you have any strong feelings here about alternative patterns

audius-infra:
Preview this change https://frontend-demo.audius.co/rj-aud-1071

jowlee (commented):
I thought we used to do this, but switched for some reason?

sliptype:
I'm a little confused on this, how are we getting FOUT if the fonts are included in the main bundle? Are we getting FOUC too? 

Taking the fonts out of the critical path is probably the right move, but doesn't it increase the potential for FOUC (especially with the lower priority via `media="print"`. I guess in practice requesting earlier probably means the fonts will be there before the react bundle is

raymondjacobson:
> I'm a little confused on this, how are we getting FOUT if the fonts are included in the main bundle? Are we getting FOUC too?
> 
> Taking the fonts out of the critical path is probably the right move, but doesn't it increase the potential for FOUC (especially with the lower priority via `media="print"`? I guess in practice requesting earlier probably means the fonts will be there before the react bundle is

I think it's b/c the fonts are currently included in the main css bundle, which is pretty hefty on initial load, sometimes the browser may not have finished fetching that before some JS comes in. I couldn't repro FOUT w/ this change, but down to sync a bit more on this strategy

sliptype (approved):
I can't repro FOUT w/ this either, so I think it's probably an improvement to make that main bundle a lil lighter

PR #676, opened 2021-11-24
Fix remixes with non-ascii characters

piazzatron:
### Description

- Setting remix parents would fail for tracks with non-ascii characters, because the slug was double encoded
- We would also show the encoded version of the title in the remix box, which is annoying

### Dragons

None

### How Has This Been Tested?

Locally against stage

n/a


audius-infra:
Preview this change https://frontend-demo.audius.co/aud-1113

rickyrombo (approved):

audius-infra:
Preview this change https://frontend-demo.audius.co/aud-1113

PR #679, opened 2021-11-29
[AUD-1068] Migrate history call to use new discovery API endpoint

raymondjacobson:
### Description

Proper history endpoint support was added in 
https://github.com/AudiusProject/audius-protocol/pull/2045

Migrate AudiusBackend history API to APIClient, add support for history & update the lineup

Note: the new endpoint supports up to limit =500, but this PR keeps parity at limit =100 for the history page. In the future, we will make the table paginate, but will probably not store indefinite history

### Dragons

### How Has This Been Tested?

Client locally vs. prod & verified history results are consistent.

### DIFF
--- a/src/containers/history-page/store/lineups/tracks/sagas.js
+++ b/src/containers/history-page/store/lineups/tracks/sagas.js
@@ -34,7 +34,7 @@ function* getHistoryTracks() {
       if (trackMetadata) {
         lineupTracks.push({
           ...trackMetadata,
-          dateListened: activity.track.listenDate
+          dateListened: activity.timestamp
         })
       }
     })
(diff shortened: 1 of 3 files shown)


audius-infra:
Preview this change https://frontend-demo.audius.co/rj-aud-1068

audius-infra:
Preview this change https://frontend-demo.audius.co/rj-aud-1068

jowlee (approved):
LGTM, worked for my staging acct. too
The only thing that seems different is that it previously de-duped the tracks in history, but no longer does that. 
Also, because history is sourced from plays which is indexed from chain, it takes a bit of time for a track to appear in history after you play the track ~ 20 seconds when I tried it - not a big deal though.

milindNshah (approved):
lgtm!

PR #665, opened 2021-11-18
remove segment

jowlee:
### Description
Replaces segment with ampitude
Adds adroll as well for page tracking

### How Has This Been Tested?
Ran against staging and saw the logs in amplitude

### How will this change be monitored?
amplitude logs

PR #682, opened 2021-11-30
Fix init track event

jowlee:
### Description
Bug in the amplitude tracking
The init event was not called correctly - bug from #665 
NOTE: we did not turn on the amplitude feature flag yet, so this code path is not executed

### Dragons
none

### How Has This Been Tested?
ran locally against prod and manually forced amplitude instead of using the feature flag

By looking for the amplitude events for init


sliptype (approved):
Good catch

PR #686, opened 2021-12-04
[ASI-604] higherPatch feature flags

SidSethi:
Feature flags for this libs PR - https://github.com/AudiusProject/audius-protocol/pull/2120


audius-infra:
Preview this change https://frontend-demo.audius.co/ss-libs-preferHigherPatch

audius-infra:
Preview this change https://demo.audius.co/ss-libs-preferHigherPatch

raymondjacobson (approved):
Can't think of any reason why getting different values for this flag between sessions would be an issue but figure i'd flag. probably session based is the right move like you did here.

PR #689, opened 2021-12-08
Your code sucks + I added Discord RPC to Electron Builds

MaleVTuber:
Literally the worst source code I've ever worked with in my entire life. Adding literally just Discord RPC to your app took me 2 weeks and a migrane. The source is fully functional and has been tested by at least 5 members on the Audius server, I can guarantee it works because I'm using it right now. Only things that could/should need to be changed is only starting the RPC Express server in `electron.js` if the user is currently on a Desktop build, and only sending the webhooks in the audio transport file on Desktop builds as well. Also, I think I removed the heavy load page, so you should fix that. It was really annoying ¯\_(ツ)_/¯

Also, I'm going to fix your dumb emoji issue. On behalf of myself and 99% of your users, you should probably put more time into improving your Desktop app instead of whatever stupid NFT project you were working on, because I can guarantee most of the people who use your service aren't here for an NFT Uploader :(

Additionally, the documentation for this repository is horrendous and I'm impressed literally _anyone_ has been able to make any changes at all to it without blowing up their computers.


MaleVTuber:
Also please ignore the python installer stuff, that also isn't important to this at all, and is specifically only for my build because I disabled autoupdater (also should've noted this lmao)

MaleVTuber:
eh I'll just keep my own repo tbh.

PR #690, opened 2021-12-08
[AUD-1115] Refactor remote config for use in audius-mobile-client

sliptype:
### Description

This PR:
* Moves remote config to `/common`
* Adds `remoteConfig` function that accepts environment-specific arguments and returns an instance. This is to allow our remote config logic to be used on other platforms (react native)
* Creates a web specific remoteConfig instance

Lots of file changes but most of the changes in `components` and `containers` are just updating imports to rely on the new remoteConfigInstance

### Dragons

Remote config initialization logic had to be updated to not depend on events. It's now fully promise and callback based. 

### How Has This Been Tested?

Tested that waiting for the optimizely data object works via the hooks and via the `waitForRemoteConfig` method.
Tested that remote config variables and feature flags are working properly


audius-infra:
Preview this change https://demo.audius.co/sk-aud-1115-remote-config

jowlee (approved):

PR #691, opened 2021-12-10
Ms download tracks mobile

milindNshah:
### Description
Trigger a mobile download onto audius-mobile-client when clicking the "download" button on the track page. 
- Create new type of message: DownloadTrackMessage
- When download button is clicked -> send that message to audius-mobile-client which will handle the rest. 

Designs
https://www.figma.com/file/ii6uNAdGGoQ7X7p7cENGhf/Mobile-Download-Buttons?node-id=101%3A141

### Dragons
- Easier to review by commit
- Drawer functionality was fully moved to the audius-mobile-client

### How Has This Been Tested?
Running through the download track flow from both web and mobile. See [MobilePR](https://github.com/AudiusProject/audius-mobile-client/pull/180) for more testing.  

Making sure the UI for the buttons looks good. 
| | Web | Mobile |
|-- | -- | -- |
| One button |Screen Shot 2021-12-04 at 3 46 38 PM|img1
| Two buttons |Img3|img4| 

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -137,6 +137,14 @@
           "version": "4.17.15",
           "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.15.tgz",
           "integrity": "sha512-8xOcRHvCjnocdS5cpwXQXVzmmh5e5+saE2QGoeQmbKmRS6J3VQppPOIt0MnmE+4xlZoumy0GPG0D0MVIQbNA1A=="
+        },
+        "proper-url-join": {
+          "version": "1.2.0",
+          "resolved": "https://registry.npmjs.org/proper-url-join/-/proper-url-join-1.2.0.tgz",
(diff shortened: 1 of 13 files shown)


sliptype (approved):
Nice work! Left a few comments. Also what is the reason for there being two download buttons?

milindNshah:
> Nice work! Left a few comments. Also what is the reason for there being two download buttons?

Oh yeah so I'm testing out the UI with two buttons because you can have multiple download buttons for different stems of the track (this file doesn't though). I just wanted to make sure the UI would look good when there is > 1 download button, there is no reason to have two of the same download button haha

Addressed the rest of the comments!

audius-infra:
Preview this change https://demo.audius.co/ms-download-tracks-mobile

raymondjacobson (approved):

PR #699, opened 2021-12-16
[AUD-1004] RN Trending Rewards Explainer

sliptype:
### Description

This PR:
* Disables Trending Rewards Explainer on native mobile (react native component will be shown instead)
* Updates remote config state to live in AppState rather than CommonState. `common` now exposes the slice, selectors, and sagas that are then used in the either client. 
 This is so both clients have separate pieces of state representing their own remote config instances
* Refactors mobile-overflow-modal state to use redux toolkit. This is to avoid a dependency on `makeReducer` which caused `common` do have a dependency on wormhole and broke the build when importing `common/store` into the mobile client

### Dragons

### How Has This Been Tested?

Tested that web TrendingRewardsExplainer is not shown on native mobile. Tested that remote config still works properly. Tested that overflow menus still work


Kyle-Shanks (approved):
LGTM!

milindNshah (approved):

PR #700, opened 2021-12-17
Stem files don't have a track.title, use filename instead

milindNshah:
### Description
Previously: No filename would show up when downloading stem version of a file on mobile
After this change: Filename shows up and track is saved to the user-specified directory

This was happening because stem files don't have a track name associated with them, they inherit the track name from the parent track file. This has already been incorporated into the "filename" within src/store/social/tracks/sagas.ts 

### How Has This Been Tested?
Tested downloading instrumental version & normal version on mobile. Filename now shows up.

### DIFF
--- a/src/services/audius-backend/TrackDownload.js
+++ b/src/services/audius-backend/TrackDownload.js
@@ -16,13 +16,13 @@ class TrackDownload {
     )
   }
 
-  static async downloadTrackMobile(cid, creatorNodeGateways, filename, title) {
+  static async downloadTrackMobile(cid, creatorNodeGateways, filename) {
     const urls = creatorNodeGateways.map(
       gateway => new URL(`${gateway}${cid}?filename=${filename}`)
     )
 
     const message = new DownloadTrackMessage({
-      title: title,
+      title: filename.split('.').slice(0, -1).join(''),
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://demo.audius.co/ms-download-track-fix-filename

src/services/audius-backend/TrackDownload.js
```diff
@@ -16,13 +16,13 @@ class TrackDownload {
...
 
-  static async downloadTrackMobile(cid, creatorNodeGateways, filename, title) {
+  static async downloadTrackMobile(cid, creatorNodeGateways, filename) {
     const urls = creatorNodeGateways.map(
       gateway => new URL(`${gateway}${cid}?filename=${filename}`)
     )
 
     const message = new DownloadTrackMessage({
-      title: title,
+      title: filename.split('.').slice(0, -1).join(''),
```
raymondjacobson, line 25: lgtm. i do agree that just passing filename all the way through would be better -- and requires the same amount of release. ok w/ merging this first -- this won't live forever as eventually the UI for downloaded stems should trigger via the react native track page.

raymondjacobson (approved):

PR #708, opened 2021-12-21
[AUD-1035] Hidden tracks

milindNshah:
### Description
Show hidden tracks in the artist page, only if you are the artist. 
1. Get hidden tracks with new headers request
2. Pipe track information down to track tile components
3. Display UI for hidden tracks (mobile & desktop)

Some additional UI info
- plays & share button will show up if it is enabled by user 
- all other stats, reposts, favorites will not show up because it is a hidden track

Linear Issue: https://linear.app/audius/issue/AUD-1035/show-hidden-tracks-on-profile-page-ui
Designs: https://www.figma.com/file/nD6xFuTNS4ICTv21j7bpRJ/Hidden-Tracks-On-Profile-Page?node-id=601%3A1372

### Dragons
- Easier to review by commit 

### How Has This Been Tested?
Desktop
Screen Shot 2021-12-21 at 12 02 07 PM

Mobile
Screen Shot 2021-12-21 at 12 02 28 PM


src/containers/profile-page/components/desktop/ProfilePage.tsx
```diff
@@ -399,7 +399,6 @@ const ProfilePage = ({
           ) : (
             <Lineup
               {...getLineupProps(artistTracks)}
-              count={profile.track_count}
```
milindNshah, line 402: The count will be determined by the TrackProvider based on the number of lineup tracks (this is already done for mobile)

src/components/track/TrackBannerIcon.tsx
```diff
@@ -0,0 +1,51 @@
...
+  return (
+    <div
+      className={cn(styles.artistPick, {
+        [styles.isMobile]: isMobile,
+        [styles.matrix]: isMatrixMode
+      })}
+    >
+      <div
+        className={cn(styles.container, {
+          [styles.star]: type === TrackBannerIconType.STAR,
```
milindNshah, line 42: Not sure this is the best way to go about conditional styling here, open to any suggestions!

audius-infra:
Preview this change https://demo.audius.co/ms-hidden-tracks

src/containers/lineup/LineupProvider.tsx
```diff
@@ -562,7 +562,7 @@ class LineupProvider extends PureComponent<CombinedProps, LineupProviderState> {
     if (
       isMetadataLoading &&
       lineup.hasMore &&
-      tiles.length < (count !== undefined ? count : MAX_TILES_COUNT) &&
+      tiles.length < (count !== undefined ? count : lineupCount) &&
```
milindNshah, line 565: This seems like the right change but I'm hesitant since this file is really confusing and I'm not sure what side-effects I might be causing cause of this. This change was put in to prevent infinite loading of tracks since count = profile.track_count doesn't account for hidden tracks so it keeps trying to get more even though no more tracks are there.

audius-infra:
Preview this change https://demo.audius.co/ms-hidden-tracks

audius-infra:
Preview this change https://demo.audius.co/ms-hidden-tracks

src/components/track/desktop/TrackTile.tsx
```diff
@@ -93,6 +96,42 @@ const TrackTile = memo(
...
       []
     )
+    const hideShare: boolean = fieldVisibility
+      ? fieldVisibility.share === false
+      : false
+    const hidePlays = fieldVisibility
+      ? fieldVisibility.play_count === false
+      : false
+
+    const shareButtonContainer = (
```
raymondjacobson, line 106: let's pull this into a function like so 
https://github.dev/AudiusProject/audius-client/blob/637c9b4f337c117ee54ecead959693e77990e825/src/containers/download-buttons/DownloadButtons.tsx#L89

src/components/track/mobile/BottomButtons.tsx
```diff
@@ -29,37 +31,61 @@ const messages = {
 const BottomButtons = (props: BottomButtonsProps) => {
   const { toast } = useContext(ToastContext)
 
-  return (
+  const repostButton = (
```
raymondjacobson, line 34: render fns for these too https://github.dev/AudiusProject/audius-client/blob/637c9b4f337c117ee54ecead959693e77990e825/src/containers/download-buttons/DownloadButtons.tsx#L89

src/containers/lineup/LineupProvider.tsx
raymondjacobson: We *really* need to rewrite lineup. @jowlee can you look into how safe this change may be?

I'm ok with this, and we shouldn't stop you from merging, but we should test pretty hard-core on staging.audius.co

src/containers/profile-page/store/lineups/tracks/sagas.js
```diff
@@ -43,14 +43,17 @@ function* getTracks({ offset, limit, payload }) {
   const sort = payload.sort === TracksSortMode.POPULAR ? 'plays' : 'date'
 
   if (user._artist_pick) {
+    const getUnlisted = true
```
raymondjacobson, line 46: some users may not have an artist pick, should we make the change on line 93 below?
```
const currentUser = yield select(getAccountUser)
if (currentUser.handle.toLowerCase() === handle.toLowercase()) {
  getUnlisted = true
}
```

src/services/audius-api-client/AudiusAPIClient.ts
```diff
@@ -886,9 +888,21 @@ class AudiusAPIClient {
       offset
     }
 
+    let headers = {}
+    if (encodedCurrentUserId && getUnlisted === true) {
```
raymondjacobson, line 892: you could just do (encodedCurrentUserId && getUnlisted)

src/services/AudiusBackend.js
```diff
@@ -2086,6 +2086,7 @@ class AudiusBackend {
   }
 
   static async signData() {
+    await waitForLibsInit()
```
raymondjacobson, line 2089: We should update the `Click to sign to authenticate ... ` message for this case because we're requesting discovery node.

I would either add an arg here for a custom message, or I would just make a separate version of this fn

static async signDiscoveryNodeRequest

raymondjacobson:
Mostly small stuff!

src/components/track/desktop/TrackTile.tsx
```diff
@@ -164,18 +219,28 @@ const TrackTile = memo(
             </div>
             <div className={styles.topRight}>
               {isArtistPick && (
-                <div className={styles.artistPickLabel}>
-                  <IconStar className={styles.iconStar} />
+                <div className={styles.topRightIconLabel}>
```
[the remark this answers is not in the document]
milindNshah: Yeah I think this whole render function needs a refactor haha

src/components/track/desktop/ConnectedTrackTile.tsx
```diff
@@ -105,6 +105,7 @@ const ConnectedTrackTile = memo(
       permalink,
       repost_count,
       save_count,
+      field_visibility,
```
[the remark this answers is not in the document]
milindNshah: I don't personally mind but doing that breaks away from the existing pattern. Happy to change them all though

src/containers/profile-page/store/lineups/tracks/sagas.js
milindNshah: Yes that was an oversight on my part, good catch!

src/components/track/mobile/BottomButtons.tsx
```diff
@@ -54,12 +68,32 @@ const BottomButtons = (props: BottomButtonsProps) => {
...
+    return (
       <MoreButton
         onClick={props.onClickOverflow}
         isDarkMode={props.isDarkMode}
         isMatrixMode={props.isMatrixMode}
       />
+    )
+  }
+
+  return props.isUnlisted ? (
```
raymondjacobson, line 86: tyty!

src/containers/profile-page/store/lineups/tracks/retrieveUserTracks.ts
```diff
@@ -9,21 +9,24 @@ type RetrieveUserTracksArgs = {
   sort?: 'date' | 'plays'
   offset?: number
   limit?: number
+  getUnlisted?: boolean
```
raymondjacobson, line 12: can we add a comment on `getUnlisted` that informs the caller that this will sign data & won't let you get anyone's unlisted tracks

raymondjacobson (approved):

PR #709, opened 2021-12-21
Retroactive Referrals

rickyrombo:
### Description

- Enables "Open in App" popover on sign-in page
- Copies current URL when "Open in App" is clicked (so that the mobile app can track referrals)
- Applies referral if user is already signed in and visits a referral link (and hasn't previously been referred)
- Applies referral if user signs in after visiting referral link (and hasn't previously been referred)

### Dragons

- ~~We always copy the full address from the open in app popover. Should that be specific to only the referral links?~~ Made this specific to sign up

### How Has This Been Tested?

- Manually, E2E on web, mobile web, Android mobile web => mobile app

**N/A**


src/containers/App.js
```diff
@@ -338,8 +338,12 @@ class App extends Component {
     }
 
     if (
-      prevProps.accountStatus === Status.LOADING &&
-      this.props.accountStatus !== Status.LOADING
+      (prevProps.accountStatus === Status.LOADING &&
```
raymondjacobson, line 341: since this was a pain to figure out, worth a comment or self-explaining variables

const accountStatusLoaded = prevProps.accountStatus === Status.LOADING && ...
const onSignUpPage = matchPath

if (accountStatusLoaded || onSignUpPage)

src/containers/app-redirect-popover/components/AppRedirectPopover.tsx
```diff
@@ -121,6 +121,7 @@ const AppRedirectPopover = ({
     onBeforeClickApp()
     const pathname = getPathname()
     const newHref = `https://redirect.audius.co${APP_REDIRECT}${pathname}`
+    navigator.clipboard.writeText(window.location.href)
```
raymondjacobson, line 124: comment on why/how this works (or link out to the doc)

src/containers/App.js
```diff
@@ -338,8 +338,12 @@ class App extends Component {
...
+        this.props.accountStatus !== Status.LOADING) ||
+      matchPath(getPathname(this.props.location), {
```
sliptype, line 343: I don't quiet understand why this is needed

sliptype (approved):
Nice! I just had one question:

raymondjacobson (approved):
odd that the package lock changed. want to undo that change from this PR?
`git fetch && git checkout origin/master package-lock.json`

Probably we should refresh it, but we should do it outside of this PR

src/containers/App.js
rickyrombo: I should add a comment. Essentially, the `isReady` flag gets sent into `AppRedirectPopover`'s `isEnabled` flag. Presumably this was used to wait until the rest of the UI was ready before loading the redirect popover. However, the `isReady` never gets set to true if there's no user to fetch, so the popover never shows on pages where there is no user (like sign-in). In theory, this could be a check for simply "is this route an auth route" but kept it narrow for the sign up referrer case.

audius-infra:
Preview this change https://demo.audius.co/mjp-retro-referrer

audius-infra:
Preview this change https://demo.audius.co/mjp-retro-referrer

audius-infra:
Preview this change https://demo.audius.co/mjp-retro-referrer

audius-infra:
Preview this change https://demo.audius.co/mjp-retro-referrer

PR #713, opened 2021-12-23
[AUD-1037] Send filename instead of title

milindNshah:
### Description
QOL fix, send filename instead of title to the mobile-client so the mobile-client doesn't have to generate it.

### How Has This Been Tested?
Ran on iOS simulator and iOS device to ensure download tracks still works

### DIFF
--- a/src/services/audius-backend/TrackDownload.js
+++ b/src/services/audius-backend/TrackDownload.js
@@ -22,8 +22,8 @@ class TrackDownload {
     )
 
     const message = new DownloadTrackMessage({
-      title: filename.split('.').slice(0, -1).join(''),
-      urls: urls
+      filename,
+      urls
     })
     message.send()
   }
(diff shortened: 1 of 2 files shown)


src/services/audius-backend/TrackDownload.js
```diff
@@ -22,7 +22,7 @@ class TrackDownload {
     )
 
     const message = new DownloadTrackMessage({
-      title: filename.split('.').slice(0, -1).join(''),
+      filename: filename,
```
raymondjacobson, line 25: nit, can just be
new DownloadTrackMessage({
  filename,
  urls
{

raymondjacobson (approved):

sliptype (approved):

audius-infra:
Preview this change https://demo.audius.co/ms-download-tracks-name-fix

PR #716, opened 2021-12-27
Updates SPL Audio UI 

jowlee:
### Description
Updates to the `/audio` page of the dapp to support changes for SPL wrapped audio
* Only show the solana address in the recieve modal
  * On first load, show the warning and store in localstorage to not show again
* On the send modal, only allow sending to a valid sol address
* Adds 2 remote config variables
  * Min send audio amount
  * Should use spl audio - which the UI toggles on
*  When transferring eth audio, wait for the transfer and show the this will take a minute text

**Note: These changes relay on libs changes as well 

Closes AUD-1134

### Dragons
This is touching audio send and should be looked at closely

### How Has This Been Tested?
Ran locally against prod

The relay is hooked up to a solana slack notifications.


src/containers/audio-rewards-page/components/SendInputBody.tsx
```diff
@@ -45,6 +51,7 @@ const messages = {
   addressMalformed: 'Please enter a valid address',
   addressRequired: 'Address is required',
   addressIsSelf: 'You cannot send $AUDIO to your own wallet!',
+  validSPLAddress: 'Please enter a valid Solana (SPL) wallet address',
```
raymondjacobson, line 54: do you think this terminology is correct? isn't an SPL wallet the derived token wallet? if you put that in versus the solana account wallet, would that work?

src/containers/audio-rewards-page/components/SendInputBody.tsx
```diff
@@ -54,49 +61,92 @@ type BalanceError =
...
-const isValidDestination = (wallet: WalletAddress) => {
+const isValidEthDestination = (wallet: WalletAddress) => {
   const libs = window.audiusLibs
   return libs.web3Manager.web3.utils.isAddress(wallet)
 }
 
+const isValidSolDestination = (wallet: WalletAddress) => {
+  const solanaweb3 = window.audiusLibs.solanaWeb3Manager.solanaWeb3
+  try {
+    const _ = new solanaweb3.PublicKey(wallet)
```
raymondjacobson, line 109: why the var?

src/services/LocalStorage.ts
```diff
@@ -14,7 +14,7 @@ const getValue = (key: string) => {
   return null
 }
 
-const getJSONValue = (key: string) => {
+export const getJSONValue = (key: string) => {
```
raymondjacobson, line 17: can we instead export methods to set what we care about in local storage directly? this makes native mobile harder

src/services/wallet-client/WalletClient.ts
```diff
@@ -26,6 +26,21 @@ class WalletClient {
     }
   }
 
+  async getCurrentWAudioBalance(): Promise<BNWei> {
+    try {
+      const balance = await AudiusBackend.getWAudioBalance()
+      return balance as BNWei
+    } catch (err) {
+      console.log(err)
```
raymondjacobson, line 34: console log right?

src/services/AudiusBackend.js
```diff
@@ -662,9 +662,15 @@ class AudiusBackend {
         account._artist_pick = body.pinnedTrackId || null
         account.twitterVerified = body.twitterVerified || false
         account.instagramVerified = body.instagramVerified || false
+      } catch (e) {
+        console.error(e)
+      }
+      try {
+        const userBank = await audiusLibs.solanaWeb3Manager.getUserBank()
```
raymondjacobson, line 669: is it necessary to wait for this on account load? this makes solana RPC part of the every page load deal. at the very lease, maybe we could do some smart promise.all?

src/services/LocalStorage.ts
raymondjacobson: https://github.com/AudiusProject/audius-mobile-client/pull/198/files#diff-fe4e9763ba51e5b01b39d69396a930770d4f43999f4be50458532a34f016808eR41
cc @sliptype

src/containers/audio-rewards-page/components/SendInputBody.tsx
```diff
@@ -54,49 +61,92 @@ type BalanceError =
...
+  solWallet: WalletAddress
 }
 
-const isValidDestination = (wallet: WalletAddress) => {
+const isValidEthDestination = (wallet: WalletAddress) => {
   const libs = window.audiusLibs
   return libs.web3Manager.web3.utils.isAddress(wallet)
 }
 
+const isValidSolDestination = (wallet: WalletAddress) => {
```
piazzatron, line 106: this is an interesting one - we should think through whether this is sufficient, or if we should actually check for the existence of a Token Account at this address first. Just sending to a regular Solana address will fail since it's not a token account created for our mint. would be good to see how Sol wallets (phantom) handle this

src/services/AudiusBackend.js
piazzatron: sanity check me on this one @jowlee but I think this is actually fine - getUsersBank() is only async bc the call to publicKey.createWithSeed is async due to the bump seed approach, and the Solana client shouldn't be doing any async work on init. Could use a comment

src/services/wallet-client/WalletClient.ts
```diff
@@ -26,6 +26,21 @@ class WalletClient {
     }
   }
 
+  async getCurrentWAudioBalance(): Promise<BNWei> {
+    try {
+      const balance = await AudiusBackend.getWAudioBalance()
+      return balance as BNWei
```
piazzatron, line 32: BNWei?? 👀

src/services/AudiusBackend.js
```diff
@@ -2522,6 +2528,25 @@ class AudiusBackend {
     }
   }
 
+  /**
+   * Make a request to fetch the sol wrapped audio balance of the the user
+   * @returns {Promise<BN>} balance
```
piazzatron, line 2533: `BN | null`

src/containers/audio-rewards-page/WalletModal.module.css
```diff
@@ -33,7 +37,7 @@
   font-size: 16px;
   line-height: 24px;
   text-align: center;
-  color: #9147CC;
+  color: #9147cc;
```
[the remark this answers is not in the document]
jowlee: It's dark --secondary - is there a way to force that? I can add a comment tho

src/containers/audio-rewards-page/WalletModal.tsx
```diff
@@ -42,10 +46,12 @@ import SendInputSuccess from './components/SendInputSuccess'
 import SendingModalBody from './components/SendingModalBody'
 import ModalDrawer from './components/modals/ModalDrawer'
 
+const { getRemoteVar } = remoteConfigInstance
 const DISCORD_URL = ' https://discord.gg/audius'
 
 const messages = {
   receive: 'Receive $AUDIO',
+  receiveSPL: 'Receive SPL $AUDIO',
```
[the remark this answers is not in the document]
jowlee: Screen Shot 2022-01-03 at 9 15 43 AM

It's described in the text

src/services/wallet-client/WalletClient.ts
jowlee: Yeah, it's `export type BNWei = Brand<BN, 'BNWei'>`
I think we did this so we know how many decimals we should be calculating with it. 
From libs, everything should be exposed in audio units with 18 decimals if with wei

src/services/wallet-client/WalletClient.ts
jowlee: Yeah, can change to error - will also change surrounding functions function.

src/containers/audio-rewards-page/WalletModal.tsx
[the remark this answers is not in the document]
raymondjacobson: Do you think this is the correct nomenclature though? isn't an SPL wallet *not* what you would paste here? b/c the SPL wallet would be the derived token account, not the root account

src/containers/audio-rewards-page/WalletModal.module.css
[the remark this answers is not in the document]
raymondjacobson: yeah let's just add a comment. we could add a whole subset of --static-dark-X colors though

src/containers/audio-rewards-page/WalletModal.tsx
```diff
@@ -186,6 +201,8 @@ const ModalContent = ({
   // TODO: user models need to have wallets
   const wallet = account.wallet as WalletAddress
 
+  const solWallet: string = (account as any).userBank?.toString()
```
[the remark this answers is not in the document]
jowlee: userbank is of type PublicKey which can be imported from solana web3, but for simplicity and to reduce deps because it is of common type user which is exported to mobile, will just store as string.

src/containers/audio-rewards-page/components/SendInputBody.tsx
jowlee: I figured it would be fine to check for valid solana address and just error on sending. 
This is a good point, can check what the behavior of phantom is.

audius-infra:
Preview this change https://demo.audius.co/jowlee-spl-audio

src/containers/audio-rewards-page/components/SendInputBody.tsx
jowlee: This will just error if it's not a valid public key

src/services/AudiusBackend.js
jowlee: Yes, you're right @piazzatron, it shouldn't make any rpc calls. 
To get the user bank, it takes in the eth address, and a couple solana constants and encodes it into a solana public key

src/containers/audio-rewards-page/components/SendInputBody.tsx
piazzatron: caught up on slack - system sol account will work since we'll derive the token account

src/common/services/remote-config/types.ts
```diff
@@ -104,7 +109,12 @@ export enum BooleanKeys {
   /**
    * Boolean to use amplitude as the metrics tracking.
    */
-  USE_AMPLITUDE = 'USE_AMPLITUDE'
+  USE_AMPLITUDE = 'USE_AMPLITUDE',
+
+  /**
+   * Boolean to use solana wrapped audio.
```
piazzatron, line 115: Nit: use in what sense? Would be nice to explain in comments that this is for which token we prefer in transfer etc

src/common/store/pages/token-dashboard/slice.ts
```diff
@@ -54,30 +54,52 @@ const slice = createSlice({
...
+        state.modalState?.stage !== 'SEND' ||
+        state.modalState.flowState.stage !== 'SENDING'
+      )
+        return
+
+      state.modalState.flowState = {
+        stage: 'AWAITING_CONVERTING_ETH_AUDIO_TO_SOL',
+        recipientWallet: state.modalState.flowState.recipientWallet,
+        amount: state.modalState.flowState.amount,
+        chain: state.modalState.flowState.chain
```
piazzatron, line 86: since this is using immer, do we need to set all these manually? I think we could just do `state.modalState.flowState.stage = ...` and move on?

src/common/store/pages/token-dashboard/slice.ts
```diff
@@ -54,30 +54,52 @@ const slice = createSlice({
...
+            'AWAITING_CONVERTING_ETH_AUDIO_TO_SOL')
       )
         return
 
       state.modalState.flowState = {
         stage: 'SENDING',
         recipientWallet: state.modalState.flowState.recipientWallet,
-        amount: state.modalState.flowState.amount
+        amount: state.modalState.flowState.amount,
+        chain: state.modalState.flowState.chain
```
piazzatron, line 102: same here

src/containers/audio-rewards-page/WalletModal.tsx
```diff
@@ -85,56 +92,67 @@ const AddWalletTitle = () => {
...
+    KEY_DISPLAY: () => {
+      const useSolSPLAudio = getRemoteVar(BooleanKeys.USE_SPL_AUDIO) as boolean
+      return (
+        <TitleWrapper
+          label={useSolSPLAudio ? messages.receiveSPL : messages.receive}
+        >
+          <IconReceive className={styles.receiveWrapper} />
+        </TitleWrapper>
+      )
+    }
```
piazzatron, line 109: nit: would this work as an IIFE, and we could avoid having to wrap everything in a function?

src/containers/audio-rewards-page/WalletModal.tsx
```diff
@@ -186,6 +204,8 @@ const ModalContent = ({
   // TODO: user models need to have wallets
   const wallet = account.wallet as WalletAddress
 
+  const solWallet = account.userBank!
```
piazzatron, line 207: let's avoid using forced unwrapping, it defeats the purpose of TS's nice optional handling. Lets either check this and return null here, or more ideally fix the account model if we know that the userBank will never be null

src/containers/audio-rewards-page/WalletModal.tsx
```diff
@@ -279,6 +303,10 @@ const shouldAllowDismiss = (modalState: Nullable<ModalState>) => {
     !(
       modalState.stage === 'CONNECT_WALLETS' &&
       modalState.flowState.stage === 'REMOVE_WALLET'
+    ) &&
+    !(
+      modalState.stage === 'SEND' &&
+      modalState.flowState.stage === 'AWAITING_CONVERTING_ETH_AUDIO_TO_SOL'
```
piazzatron, line 309: Comment a few lines above `// Allow dismiss on every stage except claiming` has been lying now for a while

src/containers/audio-rewards-page/components/ReceiveBody.tsx
```diff
@@ -1,30 +1,111 @@
...
 }
 
-const ReceiveBody = ({ wallet }: ReceiveBodyProps) => {
-  return (
-    <ModalBodyWrapper className={styles.container}>
-      <div className={styles.warning}>{messages.warning}</div>
-      <div className={styles.description}>
-        <div>{messages.warning2}</div>
-        <div>{messages.warning3}</div>
+const useLocalStorageClickedReceiveUnderstand = (): [boolean, () => void] => {
```
piazzatron, line 31: this guy is niceee

src/containers/audio-rewards-page/components/ReceiveBody.tsx
```diff
@@ -1,30 +1,111 @@
...
+import { getJSONValue, setJSONValue } from 'services/LocalStorage'
+import { remoteConfigInstance } from 'services/remote-config/remote-config-instance'
 
 import { ModalBodyWrapper } from '../WalletModal'
 
 import ClickableAddress from './ClickableAddress'
 import styles from './ReceiveBody.module.css'
 
-type ReceiveBodyProps = { wallet: WalletAddress }
+type ReceiveBodyProps = { wallet: WalletAddress; solWallet: WalletAddress }
```
piazzatron, line 16: Can we introduce a new type `SolanaWalletAddress`?

src/containers/audio-rewards-page/components/SendInputBody.tsx
piazzatron: I agree with Ray, I don't think this is right bc a valid Solana system account will work. Will post on slack

src/containers/audio-rewards-page/components/SendInputBody.tsx
piazzatron, line 106: Let's intro new Solana wallet type here

src/containers/audio-rewards-page/components/SendInputBody.tsx
```diff
@@ -54,49 +61,92 @@ type BalanceError =
...
+    console.log(err)
+    return false
+  }
+}
+
 const validateWallet = (
   wallet: Nullable<WalletAddress>,
-  ownWallet: WalletAddress
+  ownEthWallet: WalletAddress,
+  ownSolWallet: WalletAddress,
```
piazzatron, line 120: Here too

src/containers/audio-rewards-page/components/SendInputBody.tsx
```diff
@@ -164,25 +215,51 @@ const SendInputBody = ({
     [addressError, setAddressError, setDestinationAddress]
   )
 
+  const useSolSPLAudio = getRemoteVar(BooleanKeys.USE_SPL_AUDIO) as boolean
```
piazzatron, line 218: note to self, bummer we have to do these casts - is this to avoid the null case? should fix this at some later time

src/services/wallet-client/WalletClient.ts
piazzatron: yeah I know why we did it, I was surprised that we were coercing Solana Waudio balances to Wei. But I guess it makes sense

src/store/wallet/sagas.ts
```diff
@@ -29,22 +34,47 @@ const errors = {
 }
 
 function* sendAsync({
```
piazzatron, line 36: Let's add a comment somewhere capturing the new behavior of this function

src/store/wallet/sagas.ts
```diff
@@ -29,22 +34,47 @@ const errors = {
...
+      yield put(transferingEthAudioToSolWAudio())
+      yield call(walletClient.transferTokensFromEthToSol)
+    }
+
+    if (chain === Chain.Eth) {
+      yield call(() => walletClient.sendTokens(recipientWallet, weiBNAmount))
+    } else {
+      yield call(() =>
+        walletClient.sendWAudioTokens(recipientWallet, weiBNAmount)
+      )
```
piazzatron, line 76: The branching logic in this function is quite confusing imo. Could we refactor this to have some top level branch on chain type, and then the various amount conditions within that?

src/containers/audio-rewards-page/WalletModal.tsx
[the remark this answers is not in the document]
jowlee: resolved offline

src/containers/audio-rewards-page/components/SendInputBody.tsx
jowlee: resolved offline

src/common/store/pages/token-dashboard/slice.ts
jowlee: There's the typescript error because it doesn't know the rest of the shape of flowState here. 
`Type '"SENDING"' is not assignable to type '"AWAITING_CONFIRMATION" | "AWAITING_CONVERTING_ETH_AUDIO_TO_SOL"'.ts(2322)`

src/containers/audio-rewards-page/WalletModal.tsx
jowlee: the getRemoteVar would be invoked immediately and will not have the right value - which was my first attempt. 
I agree though that it's ugly.

audius-infra:
Preview this change https://demo.audius.co/jowlee-spl-audio

src/containers/audio-rewards-page/WalletModal.tsx
jowlee: I can write the null case, but this is the common pattern when we know it's the "current" account - it should be present but will always be optional - https://github.com/AudiusProject/audius-client/blob/c6224146c9d52796f1c96d2c0e631610148ee6cd/src/common/models/User.ts#L39-L52

src/containers/audio-rewards-page/WalletModal.tsx
raymondjacobson: we have a wait for remote config helper https://github.dev/AudiusProject/audius-client/blob/637c9b4f337c117ee54ecead959693e77990e825/src/services/remote-config/Provider.ts#L173

useful?

src/containers/audio-rewards-page/WalletModal.tsx
raymondjacobson: could we just return null like above if we don't have it?

audius-infra:
Preview this change https://demo.audius.co/jowlee-spl-audio

src/containers/audio-rewards-page/WalletModal.tsx
jowlee: it's async, so don't think there's a clean to use that here

jowlee: it's async, so don't think there's a clean to use that here

audius-infra:
Preview this change https://demo.audius.co/jowlee-spl-audio

PR #720, opened 2022-01-04
[AUD-1088] Fix CardLineup layout when only 1-2 cards present

dylanjeffers:
### Description

When there are two or fewer cards in a CardLineup, or an odd number of cards, the cards render incorrectly due to flex not positioning flex-items in a 2xn grid (the last card tries to take up the whole lower space, centering it). This fix addresses the case for two or fewer cards.

### How Has This Been Tested?

The issue can be reproduced on [staging](https://staging.audius.co/joey), and tested locally using the same user.

### DIFF
--- a/src/containers/lineup/CardLineup.module.css
+++ b/src/containers/lineup/CardLineup.module.css
@@ -18,6 +18,8 @@
   margin-bottom: 5px;
 }
 
-.emptyMobileContainer {
-  width: 174px;
+.emptyMobileCard {
+  width: 100%;
+  min-width: 174px;
+  max-width: 190px;
 }
--- a/src/containers/lineup/CardLineup.tsx
+++ b/src/containers/lineup/CardLineup.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { ReactElement } from 'react'
 
 import cn from 'classnames'
 import { connect } from 'react-redux'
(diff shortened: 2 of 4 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-aud-1088-fix-card-lineup-layout

raymondjacobson (approved):

src/containers/lineup/CardLineup.module.css
```diff
@@ -18,6 +18,8 @@
   margin-bottom: 5px;
 }
 
-.emptyMobileContainer {
-  width: 174px;
+.emptyMobileCard {
+  width: 100%;
+  min-width: 174px;
+  max-width: 190px;
```
raymondjacobson, line 24: how did you get the 190 number?

dylanjeffers: the above three come from the styles of the card: https://github.com/AudiusProject/audius-client/blob/46948cea5a468d2bfdb58d189139c1b6a672f29c/src/components/card/mobile/Card.module.css#L18 is there a good way for us to reference these styles? maybe make an empty card component and put it in the Card.tsx file?

src/containers/lineup/CardLineup.tsx
```diff
@@ -51,23 +51,35 @@ const DesktopCardContainer = ({
...
+        <EmptyMobileCard />
+        <EmptyMobileCard />
+      </>
+    )
+  } else if (cards.length === 2) {
+    emptyCards = <EmptyMobileCard />
+  } else if (cards.length % 2 === 1) {
+    emptyCards = <EmptyMobileCard />
+  }
+
```
sliptype, line 74: You could split this into a function to separate the logic and avoid mutating `emptyCards`:

```suggestion
const getEmptyCards = (cardLength: number) => {
  if (cardsLength === 1) {
    return (
      <>
        <EmptyMobileCard />
        <EmptyMobileCard />
      </>
    )
  } else if (cardsLength === 2) {
    return <EmptyMobileCard />
  } else if (cards.length % 2 === 1) {
    return <EmptyMobileCard />
  }
}
```

src/containers/lineup/CardLineup.module.css
raymondjacobson: Ah yeah -- might be nice to put the empty component there then and then also share via a css variable like https://github.dev/AudiusProject/audius-client/blob/637c9b4f337c117ee54ecead959693e77990e825/src/containers/App.module.css#L19

sliptype (approved):
Looks awesome! I have one stylistic preference but no need to change

src/containers/lineup/CardLineup.tsx
dylanjeffers: totally can do, was alternating between the two, thanks!

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1088-fix-card-lineup-layout

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1088-fix-card-lineup-layout

PR #721, opened 2022-01-04
[AUD-1148] Get visualizer working with 3 album art colors

milindNshah:
### Description
- Within UI state for tracks add fields for averageColor and dominantColors
- Make changes to dominantColorWorker to get 3 dominant colors from album art
- Pass dominant colors to visualizer through the visualizer provider

### Dragons
UI changes to design coming in different PR

### How Has This Been Tested?
Ran visualizer locally with these changes, works as expected! 

https://user-images.githubusercontent.com/12490332/148117476-218c7ad3-c2d0-4abd-99c2-10e400c2e876.mov

### DIFF
--- a/src/common/store/cache/tracks/sagas.js
+++ b/src/common/store/cache/tracks/sagas.js
@@ -408,12 +408,12 @@ function* watchFetchCoverArt() {
         cacheActions.update(Kind.TRACKS, [{ id: trackId, metadata: track }])
       )
 
-      const rgb = yield call(dominantColor, url)
-      console.log('got dominate', rgb)
+      const colors = yield call(dominantColor, url)
+      console.log({ colors })
       yield put(
         setColor({
           multihash,
-          color: rgb
+          color: colors
         })
       )
     } catch (e) {
--- a/src/common/store/cache/tracks/sagas.js
+++ b/src/common/store/cache/tracks/sagas.js
@@ -34,7 +34,7 @@ import { waitForBackendSetup } from 'store/backend/sagas'
 import * as confirmerActions from 'store/confirmer/actions'
 import { confirmTransaction } from 'store/confirmer/sagas'
 import { getCreatorNodeIPFSGateways } from 'utils/gatewayUtil'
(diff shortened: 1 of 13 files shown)


src/utils/visualizer/visualizer-1.js
```diff
@@ -173,19 +174,51 @@ let Visualizer1 = (function () {
...
 
+  function setDominantColors (colors) {
+    if (!colors || !colors[0]) return
+    
+    // Pull out 3 colors
+    const color1 = colors[0]
+    const color2 = colors[1] || color1
+    const color3 = colors[2] || color2
+
+    shader.uniforms.r1 = calcuateColorPixel(color1.r)
```
raymondjacobson, line 194: amaazing. i didn't know this would work

src/common/store/cache/tracks/sagas.js
```diff
@@ -410,11 +413,19 @@ function* watchFetchCoverArt() {
...
         })
       )
+
+      const dominantColors = yield call(dominantColor, url)
+      yield put(
+        setDominantColors({
+          multihash,
+          colors: dominantColors
+        })
+      )
```
raymondjacobson, line 428: This saga runs on every track fetch which is kinda unfortunate since most of the time you won't be playing the track.

Do you think we could fetch these dominant colors and call `setDominantColors` somewhere from the visualizer itself? Or perhaps when the track is played?

Alternatively, let's just use dominant color for the CurrentlyPlaying, NowPlaying, etc. which I think is far better than average color anyway and not call `yield call(averageRgb, url)` here at all.

wdyt?

audius-infra:
Preview this change https://demo.audius.co/ms-visualizer

raymondjacobson (approved):

dylanjeffers:
this is so cool!

raymondjacobson:
I'm going to rebase & merge this PR!

audius-infra:
Preview this change https://demo.audius.co/ms-visualizer

PR #722, opened 2022-01-04
[AUD-1034] Fix default CoverPhoto tiling

dylanjeffers:
### Description

Fixes default cover photo tiling, by overriding the default DynamicImage `background-size` from `cover` to `auto`. 
Also converts CoverPhoto to ts.

### How Has This Been Tested?

Go to any track with a default cover photo, modify screen size, and confirm it tiles and the resolution stays true.

### DIFF
--- a/src/components/general/CoverPhoto.tsx
+++ b/src/components/general/CoverPhoto.tsx
@@ -1,12 +1,12 @@
 import React, { memo, useState } from 'react'
 
 import cn from 'classnames'
-import PropTypes from 'prop-types'
+import { FileWithPreview } from 'react-dropzone'
 import Lottie from 'react-lottie'
 
 import loadingSpinner from 'assets/animations/loadingSpinner.json'
 import imageCoverPhotoBlank from 'assets/img/imageCoverPhotoBlank.jpg'
-import { WidthSizes } from 'common/models/ImageSizes'
+import { CoverPhotoSizes, WidthSizes } from 'common/models/ImageSizes'
 import DynamicImage from 'components/dynamic-image/DynamicImage'
 import ImageSelectionButton from 'components/image-selection/ImageSelectionButton'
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-aud-1034-fix-default-cover-photo-tiling

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1034-fix-default-cover-photo-tiling

src/components/general/CoverPhoto.tsx
```diff
@@ -17,29 +17,49 @@ const messages = {
   imageName: 'Cover Photo'
 }
 
-const CoverPhoto = props => {
+type CoverPhotoProps = {
+  userId: number | null
+  coverPhotoSizes: CoverPhotoSizes | null
```
piazzatron, line 22: lil nit: we have a nullable utility type we've been using for this stuff: `Nullable<number>`

piazzatron (approved):
LGTM 🚀

raymondjacobson:
I think this PR fixes the immediate issue of not tiling, but I think the resolution is wrong. This picture is probably @2x, so should probably be rendered at 300x300 instead of auto, but that causes some issues with the gradient overlay when it is tiled. Might need revisitation. This PR is a huge improvement & ty for typing in :)

dylanjeffers:
Ahh great callouts, yeah down to revisit sometime, can add a bug in linear for tracking too

raymondjacobson:
Screen Shot 2022-01-04 at 10 59 56 PM
Screen Shot 2022-01-04 at 10 59 41 PM

this is with setting background-size to 300x300 and removing the gradient (and i think w/ the overlay is what we'd want!) but tiling w/ the gradient overlay will be a bit tricky, probably need a separate layer & that will play interestingly with actual images too & the "dynamic image" fading from one to another

PR #723, opened 2022-01-04
Force libs make request on 404 of resource with optimistic api client

jowlee:
### Description
Relies on changes from - https://github.com/AudiusProject/audius-protocol/pull/2219
if the optimistic api client sees a 404, have it wait for the libs init and use the libs discovery make request method - which retries on another node

### How Has This Been Tested?
Ran manually

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -79,9 +79,9 @@
       }
     },
     "@audius/libs": {
-      "version": "1.2.42",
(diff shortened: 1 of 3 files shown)


audius-infra:
Preview this change https://demo.audius.co/jowlee-404-secondary

raymondjacobson (approved):

PR #727, opened 2022-01-05
Refactor IconButton to use button element

dylanjeffers:
### Description

Small change to ensure icon buttons use the button html attribute. This helps with a11y, and also enables a small refactor to add the `disabled` prop directly to the button element, which is a bit cleaner.

### Dragons

I wanted to separate this pr, since it affects a lot of elements on the page.

### DIFF
--- a/src/components/general/IconButton.module.css
+++ b/src/components/general/IconButton.module.css
@@ -1,4 +1,5 @@
 .container {
+  all: unset;
   cursor: pointer;
   display: flex;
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-aud-1187-icon-button-refactor

sliptype (approved):
Nice!!

raymondjacobson (approved):

PR #731, opened 2022-01-05
[AUD-1196] Add icons/styling options to ActionDrawer

dylanjeffers:
### Description

In order to implement the [ShareModal](https://www.figma.com/file/E2CeYYlz4vc8csivJyJlle/Share-Menu?node-id=501%3A97), the ActionDrawer needs custom styling and icon options.
- Adds custom styling for each action and actions in general
- Reimplements actions to use list and button elements for improved a11y
- Adds optional `id` prop which enables proper list/list-title relationship, improving a11y.

### Dragons

### How Has This Been Tested?

Here is an example leveraging the features in this pr: 
image

### DIFF
--- a/src/components/action-drawer/ActionDrawer.module.css
+++ b/src/components/action-drawer/ActionDrawer.module.css
@@ -41,6 +41,11 @@
   background-color: var(--neutral-light-9);
 }
 
+.actionIcon {
+  min-width: 35px;
+  display: inline-flex;
+}
+
 .destructiveAction {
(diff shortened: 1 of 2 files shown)


raymondjacobson (approved):
All looks good to me! Let's discuss some of these patterns at our product eng sync. We've iterated on the 'classes' thing a few times, but haven't fully reached consensus. Im a fan of the pattern. 

Also worth chatting about the id pattern too! easy adopt across the board 👍

PR #735, opened 2022-01-06
[AUD-1198] Add and connect ShareModal

dylanjeffers:
### Description

Adds a basic ShareModal and connects it to our store following the established modal patterns.

### DIFF
--- a/src/common/store/ui/share-modal/types.ts
+++ b/src/common/store/ui/share-modal/types.ts
@@ -4,7 +4,6 @@ import { ID } from 'common/models/Identifiers'
 import { Track } from 'common/models/Track'
 
 export type ShareModalState = {
-  isOpen: boolean
   track?: Track
 }
 
--- a/src/common/store/ui/share-modal/types.ts
+++ b/src/common/store/ui/share-modal/types.ts
@@ -1,0 +1,17 @@
+import { PayloadAction } from '@reduxjs/toolkit'
+
+import { ID } from 'common/models/Identifiers'
+import { Track } from 'common/models/Track'
+
+export type ShareModalState = {
+  isOpen: boolean
+  track?: Track
+}
+
+export type RequestOpenAction = PayloadAction<{
+  trackId: ID
+}>
+
+export type OpenAction = PayloadAction<{
+  track: Track
+}>
--- a/src/containers/share-modal/ShareModal.tsx
+++ b/src/containers/share-modal/ShareModal.tsx
@@ -1,0 +1,7 @@
+import React from 'react'
+
+import { ShareDrawer } from './components/ShareDrawer'
+
+export const ShareModal = () => {
+  return <ShareDrawer />
+}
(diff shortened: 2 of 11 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-aud-1198-add-connect-share-modal

src/common/store/ui/share-modal/types.ts
```diff
@@ -0,0 +1,17 @@
+import { PayloadAction } from '@reduxjs/toolkit'
+
+import { ID } from 'common/models/Identifiers'
+import { Track } from 'common/models/Track'
+
+export type ShareModalState = {
+  isOpen: boolean
```
sliptype, line 7: I don't think you need `isOpen` in this slice since that is handle by the modal slice

sliptype (approved):
Looks great!

src/containers/share-modal/ShareModal.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
+
+import { ShareDrawer } from './components/ShareDrawer'
+
+export const ShareModal = () => {
+  return <ShareDrawer />
```
raymondjacobson, line 6: is this going to do the switching logic for modal vs. drawer?

raymondjacobson (approved):

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1198-add-connect-share-modal

src/containers/share-modal/ShareModal.tsx
dylanjeffers: yeah exactly, just wanted to establish that here! does this seem like the right place for that?

raymondjacobson: A few ways we could do this I guess,
1. Like you've done here (I think fairly reasonable) with the useIsMobile() hook. Maybe worth calling this outer component something less UI-bound than Modal and just like src/containers/Share (ShareView?) - no strong opinion though
2. The provider pattern we've done for page/container level things like this: https://github.com/AudiusProject/audius-client/blob/master/src/containers/track-page/TrackPage.tsx#L19. I think this might be a bit verbose for what you need to do here
3. Some other new HOC component called DrawerOrModal (with a better name maybe) that handles this somewhat common usecase of modal in web, drawer in mobile

dylanjeffers: All great thoughts, I personally like the idea of naming it a more general UI name, cause one is a modal, and one is a drawer (though technically a modal?). I was hoping to try and keep the actual ui components closely tied to the redux state, following https://github.com/AudiusProject/audius-client/blob/master/src/containers/share-sound-to-tiktok-modal/ShareSoundToTikTokModal.tsx, how does that sound?

raymondjacobson: I think that's a good call. No blocker from me here!

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1198-add-connect-share-modal

PR #741, opened 2022-01-07
[AUD-1227] Create userbanks if needed

piazzatron:
### Description

Recover if no userbank exists

### Dragons

### How Has This Been Tested?

Tested with a broken account on staging

### DIFF
--- a/src/store/account/sagas.js
+++ b/src/store/account/sagas.js
@@ -38,6 +38,7 @@ import {
   clearAudiusAccount,
   clearAudiusAccountUser
 } from 'services/LocalStorage'
+import { createUserBankIfNeeded } from 'services/audius-backend/waudio'
 import { SignedIn } from 'services/native-mobile-interface/lifecycle'
 import { remoteConfigInstance } from 'services/remote-config/remote-config-instance'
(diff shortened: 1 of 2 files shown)


src/store/account/sagas.js
```diff
@@ -93,6 +94,7 @@ function* onFetchAccount(account) {
   yield fork(addPlaylistsNotInLibrary)
 
   yield fork(initAudioChecks)
+  yield call(createUserBankIfNeeded)
```
jowlee, line 97: Maybe we can put it here instead, because we try to fetch the user bank and attach to the current user obj in the store - so would be nice if it also created if didn't exist here

jowlee (approved):

audius-infra:
Preview this change https://demo.audius.co/aud-1227

src/store/account/sagas.js
jowlee: https://github.com/AudiusProject/audius-client/blob/3b414bba7a2c3f6fcfce9af3739a3ddb3e3e52b1/src/services/AudiusBackend.js#L666-L667

piazzatron: chatted on slack

PR #742, opened 2022-01-07
[AUD-1221]: Add share-to-titktok action to ShareDrawer

dylanjeffers:
### Description
Add share-to-titktok action to ShareDrawer for tracks owned by current user

### Dragons

Curious what reviewers think of the code choices in the `getActions` helper

### DIFF
--- a/src/containers/share-modal/components/ShareDrawer.tsx
+++ b/src/containers/share-modal/components/ShareDrawer.tsx
@@ -1,4 +1,4 @@
-import React, { useContext } from 'react'
+import React, { useCallback, useContext } from 'react'
 
 import {
   IconLink,
@@ -38,7 +38,7 @@ export const ShareDrawer = () => {
   const source = useSelector(getSource)
   const record = useRecord()
 
-  const getActions = () => {
+  const getActions = useCallback(() => {
     const isOwner = account && artist && account.user_id === artist.user_id
 
     const shareToTwitterAction = {
@@ -60,7 +60,7 @@ export const ShareDrawer = () => {
           )
         } else {
           console.error(
-            `Tried to share a track to twitter, but track and/or artist was missing`
+            'Tried to share a track to twitter, but track and/or artist was missing'
           )
         }
       }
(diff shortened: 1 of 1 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-aud-1221-share-to-tiktok-action

src/containers/share-modal/components/ShareDrawer.tsx
```diff
@@ -17,24 +29,64 @@ export const ShareDrawer = () => {
   const { toast } = useContext(ToastContext)
   const dispatch = useDispatch()
   const track = useSelector(getTrack)
+  const artist = useSelector((state: CommonState) =>
+    getUser(state, { id: track?.owner_id })
+  )
+  const account = useSelector(getAccountUser)
   const source = useSelector(getSource)
 
+  const getActions = () => {
```
rickyrombo, line 38: Does this need to be wrapped in a `useCallback` to not cause rerenders? Otherwise, love this

dylanjeffers: yeah it definitely could, the dep list is gonna be pretty huge which made me think how much will useCallback really optimize. Was talking with Sebastian a bit about this too, like the most optimized option would probably be doing more something like
```
<ActionDrawer>
  <ActionItem .... onClick={...} />
  {isOwner ? <ActionItem>Share To Tiktok</ActionItem> : null}
  ...
</ActionDrawer>
```
that way we are not having to recompute the array of actions a ton of times, and take advantage of general react pref stuff cause the props are closer to each subcomp.

Just thought i'd mention as an idea for the future! in meantime will do the useCallback

rickyrombo: That's a really good point re: deps. Love the idea of `<ActionDrawerItem>` - definitely a fan of child components over configs

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1221-share-to-tiktok-action

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1221-share-to-tiktok-action

rickyrombo (approved):

PR #744, opened 2022-01-10
[AUD-1224] Add DesktopShareModal

dylanjeffers:
### Description

Adds share modal for desktop

### Dragons

I wasn't able to compose stem button styles without using `!important`, not sure if im doing something wrong, or potentially a bug?

### DIFF
--- a/src/containers/share-modal/components/DesktopShareModal.module.css
+++ b/src/containers/share-modal/components/DesktopShareModal.module.css
@@ -1,0 +1,83 @@
+.headerContainer {
+  border-bottom-color: var(--neutral-light-8);
+}
+
+.modalBody {
+  display: flex;
+  overflow: hidden;
+  max-width: 528px;
+  width: 100%;
+  flex-direction: column;
+}
+
+.titleContainer {
+  font-size: var(--font-xl);
+  display: flex;
+  align-items: center;
+}
+
+.titleContainer svg {
+  height: var(--unit-6);
+  width: var(--unit-6);
+  margin-right: var(--unit-2);
+}
+
+.modalContent {
+  display: flex;
+  padding: 0px var(--unit-8) var(--unit-6);
+  flex-direction: column;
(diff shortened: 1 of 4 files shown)


src/containers/share-modal/components/DesktopShareModal.tsx
```diff
@@ -0,0 +1,94 @@
...
+      title={
+        <h2 className={styles.titleContainer}>
+          <IconShare />
+          Share Track
+        </h2>
+      }
+    >
+      <div className={styles.modalContent}>
+        <p className={styles.description}>
+          Spread the word! Share with your friends and fans!
```
sliptype, line 65: Let's put all copy into a `messages` object. The hope is that it will help us with i18n in the future

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1224-add-desktop-share-modal

src/containers/share-modal/components/DesktopShareModal.module.css
```diff
@@ -0,0 +1,83 @@
...
+  width: 100%;
+}
+
+.actionItemText {
+  font-size: var(--font-l) !important;
+  font-weight: var(--font-heavy) !important;
+}
+
+.twitterIcon path {
+  fill: var(--static-twitter-blue) !important;
```
sliptype, line 63: I think the important is required because of the way we are targeting the text in Stems:

https://github.com/AudiusProject/stems/blob/master/src/components/Button/Button.module.css#L84

Should be straightforward to reduce the specificity there. But we can leave the `!important`s for now

sliptype (approved):
Looks good!

src/containers/share-modal/components/DesktopShareModal.module.css
dylanjeffers: ahh yeah good call, will look into

PR #749, opened 2022-01-11
Fix ActionButton margins

dylanjeffers:
### Description

Fixes margins for action buttons. This regression occurred because a css selector expected the action buttons to be div elements, which changed due to IconButton now rendering a button element.

### How Has This Been Tested?

Before: 
image

After:
image


raymondjacobson (approved):
good catch

audius-infra:
Preview this change https://demo.audius.co/dj-fix-action-button-margin

PR #761, opened 2022-01-12
[AUD-1258] Share collections with ShareMenu

dylanjeffers:
### Description

The final piece for the share menu!
- Adds ability to share playlists and albums
- Refactors twitter shareText generation
- Fixes tiktok icon styles and removes !important by updating css specificity to override stems styles
### Dragons

The tiktok icon fixes are definitely a bit of a hack and prone to breakage if we update the icon

### How Has This Been Tested?
https://dylan.audius.co
Checking that each share button opens the share modal, it correctly shares the right type ie share track, share album..., and the generated links and twitter share text looks correct


audius-infra:
Preview this change https://demo.audius.co/dj-aud-1250-share-album-playlist

src/common/store/ui/share-modal/types.ts
```diff
@@ -1,39 +1,61 @@
...
 
 import { ShareSource } from 'common/models/Analytics'
+import { Collection } from 'common/models/Collection'
 import { ID } from 'common/models/Identifiers'
 import { Track } from 'common/models/Track'
 import { User } from 'common/models/User'
 import { Nullable } from 'common/utils/typeUtils'
 
-export type ShareType = 'track' | 'profile'
+export type ShareType = 'track' | 'profile' | 'album' | 'playlist'
```
milindNshah, line 10: Can you convert this to an enum instead and use ShareType.TRACK or ShareType.PROFILE, etc instead of the raw string values? :)

src/common/store/ui/share-modal/sagas.ts
```diff
@@ -1,33 +1,60 @@
...
-    )
-    yield put(open({ type, track, source, artist }))
-  } else {
-    const { profileId, source, type } = action.payload
-    const profile: User = yield select((state: AppState) =>
-      getUser(state, { id: profileId })
-    )
-    yield put(open({ type, profile, source }))
+  switch (action.payload.type) {
+    case 'track': {
```
milindNshah, line 24: Prefer usage of enums compared to raw strings (See comment below)

src/components/track/desktop/ConnectedPlaylistTile.tsx
```diff
@@ -535,7 +535,13 @@ function mapDispatchToProps(dispatch: Dispatch) {
     goToRoute: (route: string) => dispatch(pushRoute(route)),
     record: (event: TrackEvent) => dispatch(event),
     shareCollection: (id: ID) =>
-      dispatch(shareCollection(id, ShareSource.TILE)),
+      dispatch(
+        requestOpenShareModal({
+          type: 'collection',
```
milindNshah, line 540: Also replacing these calls with the enum value

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1250-share-album-playlist

src/containers/share-modal/ShareModal.tsx
```diff
@@ -31,56 +31,28 @@ export const ShareModal = () => {
...
-  )
   const account = useSelector(getAccountUser)
 
   const { isEnabled: isShareSoundToTikTokEnabled } = useFlag(
     FeatureFlags.SHARE_SOUND_TO_TIKTOK
   )
 
   const isOwner = Boolean(
-    account && trackArtist && account.user_id === trackArtist.user_id
+    content?.type === 'track' &&
```
milindNshah, line 42: Update to use enum

src/containers/share-modal/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+import { messages } from './messages'
+
+type ShareToTwitterEvent = Omit<ShareToTwitter, 'eventName' | 'source'>
+
+export const getTwitterShareText = (content: ShareModalContent) => {
+  let twitterText = ''
+  let link = ''
+  let analyticsEvent: ShareToTwitterEvent
+  switch (content.type) {
+    case 'track': {
```
milindNshah, line 19: Replace with enums here too

src/containers/share-modal/components/ShareDialog.module.css
```diff
@@ -57,34 +57,57 @@
...
+.actionButton .tikTokIcon path:nth-of-type(4) {
+  fill: #69c9d0;
+}
+
+.actionButton .tikTokIcon path:nth-of-type(5) {
+  fill: #69c9d0;
+}
+
+.actionButton .tikTokIconDark path {
+  fill: #fff;
```
milindNshah, line 96: fill: var(--static-white)

src/containers/share-modal/components/ShareDialog.tsx
```diff
@@ -72,23 +76,28 @@ export const ShareDialog = ({
...
               leftIcon={<IconTikTok {...iconProps} />}
               text={messages.tikTok}
-              textClassName={cn(styles.tikTokActionItemText, {
-                [styles.tikTokActionItemTextDark]: isDarkMode()
-              })}
+              iconClassName={
+                isDarkMode() ? styles.tikTokIconDark : styles.tikTokIcon
+              }
+              textClassName={
+                isDarkMode()
```
milindNshah, line 89: Nit: I think you can add this as a variable in colors.css, set values for both regular and dark mode, and then just use a single class with color: --var(tik-tok-icon-color) and not manually check for dark mode.

src/containers/share-modal/messages.ts
```diff
@@ -2,12 +2,23 @@ import { ShareType } from 'common/store/ui/share-modal/types'
...
 }
 
 export const messages = {
   modalTitle: (asset: ShareType) => `Share ${shareTypeMap[asset]}`,
   twitter: 'Share to Twitter',
   tikTok: 'Share Sound to TikTok',
-  copyLink: (asset: ShareType) => `Copy Link to ${shareTypeMap[asset]}`
+  copyLink: (asset: ShareType) => `Copy Link to ${shareTypeMap[asset]}`,
+  toast: (asset: ShareType) => `Copied Link to ${shareTypeMap[asset]}`,
+  trackShareText: (title: string, handle: string) =>
```
dylanjeffers, line 16: @milindNshah moved the twitterShareText here btw, this what you had in mind?

src/containers/share-modal/utils.ts
dylanjeffers, line 19: @milindNshah regarding using an enum here, it seems like we do both union types and enums, and i personally prefer union types on most cases cause its more lightweight. how about we have this as an item for a dev huddle, would love to have a discussion with everyone to figure out some guidelines?

milindNshah (approved):
Looks great! Left a few minor comments below - mainly to get rid of random strings in the code.

src/containers/share-modal/utils.ts
milindNshah: Sure sounds good, didn't see this before I left my comments but feel free to ignore for now

src/containers/share-modal/components/ShareDialog.tsx
dylanjeffers: ahh that makes a lot of sense. for context for this icon specifically i had to change the colors manually since the stems button is overriding the pre-defined fills in the svgs, in general, we dont need to reference the tikTok colors in css

PR #770, opened 2022-01-13
[AUD-1279] Add no-unused-vars to eslint and fix

dylanjeffers:
### Description

Adds no-unused-vars https://eslint.org/docs/rules/no-unused-vars#options, and fixes issues. This addition still ignores unused-args, which we can approach in another pr.


rickyrombo (approved):

.eslintrc.js
```diff
@@ -72,7 +72,15 @@ module.exports = {
     '@typescript-eslint/no-use-before-define': 'off',
     '@typescript-eslint/no-empty-function': 'off',
     '@typescript-eslint/no-var-requires': 'off',
-    '@typescript-eslint/no-unused-vars': 'off', // We should turn this one on soon
+    '@typescript-eslint/no-unused-vars': [
+      'error',
+      {
+        vars: 'all',
+        varsIgnorePattern: '[iI]gnored',
```
sliptype, line 79: What if there are multiple vars you want to ignore? I guess we could just ignore the rule on that line right?

I feel like it might be cleaner to not have any ignore pattern and just explicitly disable the rule on those lines, thoughts?

sliptype (approved):
Thank you for doing this!!

.eslintrc.js
dylanjeffers: so i think you can do
```
const varAIgnored = ....
const varBIgnored = ....
```
also cool with doing an explicit ignore too.
fwiw the reason i added this was for https://github.com/AudiusProject/audius-client/blob/b5f248fb295d4cff61a308931588191f5b88e84b/src/containers/audio-rewards-page/components/SendInputBody.tsx#L111, which is needed for a side-effect, but `new blah()` gives an error for not using `new` as a side-effect

Kyle-Shanks (approved):

.eslintrc.js
sliptype: Ah ok, makes sense since it's not an exact regex. I think that's fine then!

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1279-eslint-unused-vars

PR #772, opened 2022-01-13
Add profile completion optimistic update and refresh user challenges periodically

sddioulde:
### Description

Add hook that will return optimistic (client state) version of challenge completions. For now it only includes the profile-completion challenge as others are fine as-is.

Also fetch user challenges periodically to get latest DN challenge state for the user.

### Dragons

This means that if the profile completion is complete from the client perspective but incomplete from DN sides, the reward claim will fail. As-is, we just let it fail and the user can try again.

### How Has This Been Tested?

Manually, running client against stage.

### DIFF
--- a/src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
+++ b/src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
@@ -72,7 +72,7 @@ const RewardPanel = ({
     ? currentStepCountOverride!
     : challenge?.current_step_count || 0
   const isComplete = shouldOverrideCurrentStepCount
-    ? currentStepCountOverride === stepCount
+    ? currentStepCountOverride >= stepCount
     : !!challenge?.is_complete
 
   return (
--- a/src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
+++ b/src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
@@ -68,7 +68,9 @@ const RewardPanel = ({
 
   const challenge = userChallenges[id]
   const currentStepCount =
-    currentStepCountOverride || challenge?.current_step_count || 0
+    currentStepCountOverride === undefined
+      ? challenge?.current_step_count || 0
+      : currentStepCountOverride
   const isComplete = !!challenge?.is_complete
 
   return (
--- a/src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
+++ b/src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode, useEffect } from 'react'
+import React, { ReactNode, useEffect, useState } from 'react'
 
 import { ProgressBar } from '@audius/stems'
 import cn from 'classnames'
@@ -26,6 +26,7 @@ import styles from './RewardsTile.module.css'
 import ButtonWithArrow from './components/ButtonWithArrow'
 import { Tile } from './components/ExplainerTile'
 import { challengeRewardsConfig } from './config'
+import { useOptimisticChallengeCompletionStepCounts } from './hooks'
 
 const messages = {
   title: '$AUDIO REWARDS',
@@ -45,6 +46,7 @@ type RewardPanelProps = {
   stepCount: number
   openModal: (modalType: ChallengeRewardsModalType) => void
   id: ChallengeRewardID
+  currentStepCountOverride?: number
 }
 
 const RewardPanel = ({
@@ -55,7 +57,8 @@ const RewardPanel = ({
   openModal,
   progressLabel,
   icon,
-  stepCount
+  stepCount,
+  currentStepCountOverride
 }: RewardPanelProps) => {
   const wm = useWithMobileStyle(styles.mobile)
   const userChallenges = useSelector(getUserChallenges)
@@ -63,7 +66,19 @@ const RewardPanel = ({
   const openRewardModal = () => openModal(id)
 
   const challenge = userChallenges[id]
-  const currentStepCount = challenge?.current_step_count || 0
+  const currentStepCount =
+    currentStepCountOverride || challenge?.current_step_count || 0
+  if (
+    ['profile-completion', 'track-upload'].includes(
+      challenge?.challenge_id ?? ''
+    )
+  ) {
+    console.log({
+      challenge,
+      currentStepCountOverride,
+      stepCount: challenge?.current_step_count
+    })
+  }
   const isComplete = !!challenge?.is_complete
 
   return (
@@ -133,15 +148,35 @@ const useRewardIds = () => {
   return filteredRewards
 }
 
+const CHALLENGE_REFRESH_INTERVAL_MS = 5000
+
 const RewardsTile = ({ className }: RewardsTileProps) => {
   const setVisibility = useSetVisibility()
   const dispatch = useDispatch()
   const rewardIds = useRewardIds()
   const userChallengesLoading = useSelector(getUserChallengesLoading)
+  const currentStepCountOverrides = useOptimisticChallengeCompletionStepCounts()
+
+  const [haveChallengesLoaded, setHaveChallengesLoaded] = useState(false)
 
   useEffect(() => {
-    dispatch(fetchUserChallenges())
-  }, [dispatch])
+    if (!userChallengesLoading && !haveChallengesLoaded) {
+      setHaveChallengesLoaded(true)
+    }
+  }, [userChallengesLoading, haveChallengesLoaded])
+
+  useEffect(() => {
+    if (!haveChallengesLoaded) {
+      dispatch(fetchUserChallenges())
+    }
+    const interval = setInterval(() => {
+      dispatch(fetchUserChallenges())
+    }, CHALLENGE_REFRESH_INTERVAL_MS)
+
+    return () => {
+      clearInterval(interval)
+    }
+  }, [dispatch, haveChallengesLoaded])
 
   const openModal = (modalType: ChallengeRewardsModalType) => {
     dispatch(setChallengeRewardsModalType({ modalType }))
@@ -151,7 +186,12 @@ const RewardsTile = ({ className }: RewardsTileProps) => {
   const rewardsTiles = rewardIds
     .map(id => challengeRewardsConfig[id])
     .map(props => (
-      <RewardPanel {...props} openModal={openModal} key={props.id} />
+      <RewardPanel
+        {...props}
+        currentStepCountOverride={currentStepCountOverrides[props.id]}
+        openModal={openModal}
+        key={props.id}
+      />
     ))
 
   const wm = useWithMobileStyle(styles.mobile)
@@ -164,7 +204,11 @@ const RewardsTile = ({ className }: RewardsTileProps) => {
         <span>{messages.description2}</span>
       </div>
       <div className={styles.rewardsContainer}>
-        {userChallengesLoading ? <LoadingSpinner /> : rewardsTiles}
+        {userChallengesLoading && !haveChallengesLoaded ? (
+          <LoadingSpinner />
+        ) : (
+          rewardsTiles
+        )}
       </div>
     </Tile>
   )
--- a/src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
+++ b/src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
@@ -6,7 +6,7 @@ import { useDispatch, useSelector } from 'react-redux'
 
 import { useSetVisibility } from 'common/hooks/useModalState'
 import { ChallengeRewardID } from 'common/models/AudioRewards'
-import { IntKeys, StringKeys } from 'common/services/remote-config'
+import { StringKeys } from 'common/services/remote-config'
 import {
   getUserChallenges,
   getUserChallengesLoading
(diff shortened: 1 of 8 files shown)


src/containers/audio-rewards-page/ChallengeRewardsTile.tsx
```diff
@@ -133,15 +148,35 @@ const useRewardIds = () => {
   return filteredRewards
 }
 
+const CHALLENGE_REFRESH_INTERVAL_MS = 5000
```
sddioulde, line 151: what interval makes sense here? chose 5s because DN indexing challenges at a same rate

src/containers/audio-rewards-page/ChallengeRewardsTile.tsx
```diff
@@ -55,15 +57,28 @@ const RewardPanel = ({
...
   const challenge = userChallenges[id]
-  const currentStepCount = challenge?.current_step_count || 0
+  const currentStepCount =
+    currentStepCountOverride || challenge?.current_step_count || 0
+  if (
+    ['profile-completion', 'track-upload'].includes(
+      challenge?.challenge_id ?? ''
+    )
+  ) {
+    console.log({
```
sddioulde, line 76: will remove log

audius-infra:
Preview this change https://demo.audius.co/sd-optimistic-challenge-state

src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
```diff
@@ -55,15 +57,28 @@ const RewardPanel = ({
...
   const challenge = userChallenges[id]
-  const currentStepCount = challenge?.current_step_count || 0
+  const currentStepCount =
+    currentStepCountOverride || challenge?.current_step_count || 0
+  if (
+    ['profile-completion', 'track-upload'].includes(
+      challenge?.challenge_id ?? ''
+    )
+  ) {
+    console.log({
```
sddioulde, line 76: will remove log

src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
```diff
@@ -133,15 +148,35 @@ const useRewardIds = () => {
   return filteredRewards
 }
 
+const CHALLENGE_REFRESH_INTERVAL_MS = 5000
```
sddioulde, line 151: matches the DN challenge task interval, but do we want it this often?

audius-infra:
Preview this change https://demo.audius.co/sd-optimistic-challenge-state

src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
```diff
@@ -55,15 +57,28 @@ const RewardPanel = ({
...
 }: RewardPanelProps) => {
   const wm = useWithMobileStyle(styles.mobile)
   const userChallenges = useSelector(getUserChallenges)
 
   const openRewardModal = () => openModal(id)
 
   const challenge = userChallenges[id]
-  const currentStepCount = challenge?.current_step_count || 0
+  const currentStepCount =
+    currentStepCountOverride || challenge?.current_step_count || 0
```
piazzatron, line 70: I think there's a very subtle bug where if we decrement our step count to zero on the client but discovery is still showing > 0, we prefer discovery still. Realistically probably no impact but would be best to check explicitly against undefined bc of this

src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
piazzatron: seems like a good candidate for remote config. seems like a sensible starting value though

src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
```diff
@@ -133,15 +148,35 @@ const useRewardIds = () => {
...
+      dispatch(fetchUserChallenges())
+    }
+    const interval = setInterval(() => {
+      dispatch(fetchUserChallenges())
+    }, CHALLENGE_REFRESH_INTERVAL_MS)
+
+    return () => {
+      clearInterval(interval)
+    }
+  }, [dispatch, haveChallengesLoaded])
```
piazzatron, line 179: Might be nice to extract this functionality into a standalone hook with a comment explaining the refresh functionality here

src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
```diff
@@ -164,7 +204,11 @@ const RewardsTile = ({ className }: RewardsTileProps) => {
         <span>{messages.description2}</span>
       </div>
       <div className={styles.rewardsContainer}>
-        {userChallengesLoading ? <LoadingSpinner /> : rewardsTiles}
+        {userChallengesLoading && !haveChallengesLoaded ? (
```
piazzatron, line 207: why do we need this extra check?

src/pages/audio-rewards-page/components/modals/ChallengeRewards.tsx
```diff
@@ -161,7 +162,10 @@ const ChallengeRewardsBody = ({ dismissModal }: BodyProps) => {
     modalButtonInfo
   } = challengeRewardsConfig[modalType]
 
-  const currentStepCount = challenge?.current_step_count || 0
+  const currentStepCountOverrides = useOptimisticChallengeCompletionStepCounts()
+  const currentStepCount =
+    currentStepCountOverrides[modalType] || challenge?.current_step_count || 0
```
piazzatron, line 167: same comment here re: 0

piazzatron (approved):
Great stuff! The only thing I saw in my testing is that we should be refreshing the balance in addition to the challenge state on the rewards page - it looks like we already have code to do this (`pollForBalance`), so maybe the refresh rate is just too low? Let's just look into it before we merge this PR

src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
sddioulde: good catch

sddioulde:
interesting point about refreshing user balances. the pollForBalance is only called on the artist dashboard page but adding this logic for the audio rewards page too

audius-infra:
Preview this change https://demo.audius.co/sd-optimistic-challenge-state

audius-infra:
Preview this change https://demo.audius.co/sd-optimistic-challenge-state

audius-infra:
Preview this change https://demo.audius.co/sd-optimistic-challenge-state

audius-infra:
Preview this change https://demo.audius.co/sd-optimistic-challenge-state

audius-infra:
Preview this change https://demo.audius.co/sd-optimistic-challenge-state

audius-infra:
Preview this change https://demo.audius.co/sd-optimistic-challenge-state

src/pages/audio-rewards-page/components/modals/ChallengeRewards.tsx
```diff
@@ -163,10 +164,18 @@ const ChallengeRewardsBody = ({ dismissModal }: BodyProps) => {
...
+  const currentStepCountOverride = currentStepCountOverrides[modalType]
+  const shouldOverrideCurrentStepCount = currentStepCountOverride !== undefined
+  const currentStepCount = shouldOverrideCurrentStepCount
+    ? currentStepCountOverride!
+    : challenge?.current_step_count || 0
+
   const isIncomplete = currentStepCount === 0
-  const isComplete = challenge?.is_complete
-  const isInProgress = currentStepCount > 0 && !isComplete
+  const isInProgress = currentStepCount > 0 && currentStepCount !== stepCount
```
rickyrombo, line 175: This should be `>= stepCount`, I think?

src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
```diff
@@ -138,9 +147,22 @@ const RewardsTile = ({ className }: RewardsTileProps) => {
...
   useEffect(() => {
-    dispatch(fetchUserChallenges())
+    if (!userChallengesLoading && !haveChallengesLoaded) {
+      setHaveChallengesLoaded(true)
+    }
+  }, [userChallengesLoading, haveChallengesLoaded])
+
+  // poll for user challenges and user balance to refresh
+  useEffect(() => {
+    dispatch(refreshUserChallenges())
```
rickyrombo, line 161: nit: What's the benefit of a separate saga here vs something like `setInterval` and dispatch `getUserChallenges`?

rickyrombo: Or rather, `useInterval`: https://usehooks-ts.com/react-hook/use-interval

audius-infra:
Preview this change https://demo.audius.co/sd-optimistic-challenge-state

src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
```diff
@@ -55,16 +58,22 @@ const RewardPanel = ({
...
 
   const challenge = userChallenges[id]
-  const currentStepCount = challenge?.current_step_count || 0
-  const isComplete = !!challenge?.is_complete
+  const shouldOverrideCurrentStepCount = currentStepCountOverride !== undefined
+  const currentStepCount = shouldOverrideCurrentStepCount
+    ? currentStepCountOverride!
+    : challenge?.current_step_count || 0
+  const isComplete = shouldOverrideCurrentStepCount
+    ? currentStepCountOverride === stepCount
```
rickyrombo, line 75: Should this be `>=`?

src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
```diff
@@ -138,9 +147,22 @@ const RewardsTile = ({ className }: RewardsTileProps) => {
   const dispatch = useDispatch()
   const rewardIds = useRewardIds()
   const userChallengesLoading = useSelector(getUserChallengesLoading)
+  const currentStepCountOverrides = useOptimisticChallengeCompletionStepCounts()
+  const [haveChallengesLoaded, setHaveChallengesLoaded] = useState(false)
```
rickyrombo, line 151: Can `haveChallengesLoaded` be inferred (and maybe even `userChallengesLoading`)  by checking if `getUserChallenges` is truthy?

src/pages/audio-rewards-page/components/modals/ChallengeRewards.tsx
```diff
@@ -163,10 +164,18 @@ const ChallengeRewardsBody = ({ dismissModal }: BodyProps) => {
...
+  const isComplete = shouldOverrideCurrentStepCount
+    ? currentStepCountOverride === stepCount
```
rickyrombo, line 177: `>= stepCount`?

src/pages/audio-rewards-page/store/sagas.ts
```diff
@@ -69,6 +81,18 @@ function* claimChallengeRewardAsync(
 ) {
   const { claim, retryOnFailure } = action.payload
   const { specifier, challengeId, amount } = claim
+
+  // Do not proceed to claim if challenge is not complete from a DN perspective.
+  // This is possible because the client may optimistically set a challenge as complete
+  // even though the DN has not yet indexed the change that would mark the challenge as complete.
+  // In this case, we wait until the challenge is complete in the DN before claiming
+  yield call(
```
rickyrombo, line 89: 👍

src/pages/audio-rewards-page/hooks.ts
```diff
@@ -0,0 +1,21 @@
+import { useSelector } from 'react-redux'
+
+import { ChallengeRewardID } from 'common/models/AudioRewards'
+import { getCompletionStages } from 'components/profile-progress/store/selectors'
+
+type OptimisticChallengeCompletionResponse = Partial<
+  Record<ChallengeRewardID, number>
+>
+
+export const useOptimisticChallengeCompletionStepCounts = () => {
```
rickyrombo, line 10: Can this take a challenge object and return a challenge object with the step count updated as appropriate? Then we don't have to have the repeated override logic.

We could also shorten the name to `useOptimisticChallenge` then.

Something like:


```ts
export const useOptimisticChallenge(challenge: UserChallenge) => {
  const optimisticSteps = challengeStepCalculators[challenge.id]() // internal map of challenge step calculators
  if (optimisticSteps) {
      return { ...challenge, step_count: optimisticSteps }
  }
  return challenge
}
```

rickyrombo (changes requested):
Did you check to make sure this works if eg the user has more than 10 referrals or more than 3 songs uploaded? Worried about `=== step_count` vs `>= step_count`

Everything else looks good! Left some minor nits

src/pages/audio-rewards-page/components/modals/ChallengeRewards.tsx
sddioulde: nice yeah

src/pages/audio-rewards-page/components/modals/ChallengeRewards.tsx
sddioulde: < you mean. thanks

rickyrombo (approved):

src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
sddioulde: the empty object {} for user challenges could be tricky cuz edge case of no challenges active (all disabled) or sthg... there's probably other things we can do or additional checks if we wanted to move that direction

src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
piazzatron: makes sense

PR #774, opened 2022-01-13
Fix rewards progress exceeds total steps

rickyrombo:
### Description

- Prior to this change, if you referred more than the total number of referrals or uploaded more tracks than required for disbursement, the UI shows eg. 5/3 and as though it is in progress. 
- Also brings the text for the error to the `messages` object

Uses the `challenge.is_complete` member instead to determine challenge completion. Steps should never be complete without the challenge being complete, as they're pulled from the same response object.

### Dragons

N/A

### How Has This Been Tested?

Manually against stage

N/A


audius-infra:
Preview this change https://demo.audius.co/mjp-challenge-rewards-ui-fixes

PR #773, opened 2022-01-13
Add sagas for claiming challenge rewards

rickyrombo:
### Description

Adds sagas to handle claiming challenge rewards, as a precursor for native mobile "Claim Your Reward" button, and consolidation/refactoring with the current `ClaimRewardButton`

Depends on #771 as it needs to pop the Cognito Modal if that's the failure reason (and retries were allowed).

### Dragons

N/A

### How Has This Been Tested?

Tested manually by dispatching with Redux Dev Tools and ensuring the proper modals are triggered and actions are sent in response.

### How will this change be monitored?

Console errors will trigger sentry

PR #775, opened 2022-01-13
Decouple HCaptcha modal from Rewards

rickyrombo:
### Description

- Decouples the HCaptcha modal from rewards (note: depends on #773 going in in order to keep current behavior)
- fixes minor nit: Use `useCallback` instead of recreating a callback function every render

### Dragons

N/A

### How Has This Been Tested?

Tested manually against stage on top of #773 - HCaptcha modal only reopens if that was the source

N/A

### DIFF
--- a/src/pages/audio-rewards-page/components/modals/HCaptchaModal.tsx
+++ b/src/pages/audio-rewards-page/components/modals/HCaptchaModal.tsx
@@ -23,15 +23,17 @@ export const HCaptchaModal = () => {
   const [isOpen, setOpen] = useModalState('HCaptcha')
   const dispatch = useDispatch()
 
-  // here, we can also instead use a prop e.g. onClose to handle re-opening the reward modal if we want to decouple rewards and hcaptcha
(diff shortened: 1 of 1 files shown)


audius-infra:
Preview this change https://demo.audius.co/mjp-hcaptcha-modal

src/pages/audio-rewards-page/components/modals/HCaptchaModal.tsx
```diff
@@ -23,15 +23,17 @@ export const HCaptchaModal = () => {
...
-      setOpen(false)
-    }
-  }, [setRewardModalOpen, isOpen, setOpen])
+  const handleClose = useCallback(
+    (userInitiated = true) => {
+      if (isOpen) {
+        if (userInitiated) {
+          dispatch(setHCaptchaStatus({ status: HCaptchaStatus.USER_CLOSED }))
+        }
+        setOpen(false)
```
dylanjeffers, line 32: super nit, but handleClose can only be called when isOpen is true right? so i think we can remove the isOpen branch

sddioulde (approved):
lg

PR #776, opened 2022-01-13
Add basic helper for typing saga call return types

rickyrombo:
### Description

Adds a helper to make typing sagas easier.

Related: [AUD-1215](https://linear.app/audius/issue/AUD-1215/explore-typed-redux-saga)

Taken from:
https://spin.atomicobject.com/2020/08/06/redux-saga-call-effect/

Example:
```ts
const foo: CallReturnType<typeof func> = yield call(func, ...)
```

### Dragons

N/A

### How Has This Been Tested?

N/A

N/A

### DIFF
--- a/src/common/utils/sagaTypeUtils.ts
+++ b/src/common/utils/sagaTypeUtils.ts
@@ -1,0 +1,30 @@
+/**
+ * @file Utility methods for typing Sagas
+ * @author Drew Colthorp (@dcolthorp)
+ * Taken/Adapted from:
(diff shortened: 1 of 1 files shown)


audius-infra:
Preview this change https://demo.audius.co/mjp-type-saga-call

dylanjeffers (approved):
tried this out on a call, and it works great

rickyrombo:
Closing in favor of #784

PR #777, opened 2022-01-14
Use typed getCognitoSignature

rickyrombo:
### Description

Moves away from `AudiusBackend.getCognitoSignature()` to `getCognitoSignature` from `services/audius-backend/Cognito.ts`

### Dragons

Typed version throws exceptions instead of swallowing errors. Make sure all callers catch that.

### How Has This Been Tested?

Will test after other changes go in.

N/A


audius-infra:
Preview this change https://demo.audius.co/mjp-use-getCognitoSignature

src/pages/audio-rewards-page/components/ClaimRewardButton.tsx
```diff
@@ -168,67 +169,73 @@ const ClaimRewardButton = ({
...
-    })
+      flow.on('error', (event: FlowErrorEvent) => {
+        console.error(`Error claiming reward: Flow error! ${event.message}`)
+        dispatch(setClaimStatus({ status: ClaimStatus.ERROR }))
+      })
 
-    flow.open()
+      flow.open()
+    } catch (e) {
+      console.error('COGNITO: Failed to get signature.', e)
```
dylanjeffers, line 237: just wondering if there is a way to minimize the try/catch block here since its really just for the getCognitoSignature. potentially that function just throws the updated error with a console.log?

src/services/AudiusBackend.js
```diff
@@ -2064,26 +2064,6 @@ class AudiusBackend {
...
-      const response = await fetch(`${IDENTITY_SERVICE}/cognito_signature`, {
-        headers: {
-          'Content-Type': 'application/json',
-          [AuthHeaders.Message]: data,
-          [AuthHeaders.Signature]: signature
-        }
-      }).then(res => res.json())
-      return response
-    } catch (e) {
-      console.error(e)
```
dylanjeffers, line 2082: ah so we are catching here actially, and not rethrowing

sddioulde (approved):

audius-infra:
Preview this change https://demo.audius.co/mjp-use-getCognitoSignature

audius-infra:
Preview this change https://demo.audius.co/mjp-use-getCognitoSignature

PR #778, opened 2022-01-14
[AUD-1280] Move models from src to common

dylanjeffers:
### Description

Simple refactor to consolidate our models in one place. This also makes it easier to share models in audius-modile-client


audius-infra:
Preview this change https://demo.audius.co/dj-aud-1280-move-models-to-common

rickyrombo (approved):
Looks good to me - should check with @sliptype too in case these were excluded for a reason in the initial push

sliptype:
> Looks good to me - should check with @sliptype too in case these were excluded for a reason in the initial push

I've just been moving things incrementally as they are needed for features. But this sets us up to get rid of all the duplicate models in audius-mobile-client!

sliptype (approved):

PR #779, opened 2022-01-14
Replace year trending with all time

isaacsolo:
### Description

Use allTime instead of year for trending on mobile and desktop. Previously, mobile was said "this year" and fetched year data and desktop said "all time" but fetched year data as well.

Depends on trending change: https://github.com/AudiusProject/audius-protocol/pull/2267/files.

### Dragons

### How Has This Been Tested?

Tested against prod snapshot with change from https://github.com/AudiusProject/audius-protocol/pull/2267/files.

Screen Shot 2022-01-17 at 2 35 32 PM
Screen Shot 2022-01-17 at 2 28 06 PM

Test trending tabs on staging and prod.


audius-infra:
Preview this change https://demo.audius.co/is-trending-all-time

jowlee (changes requested):
In some places you're replacing `year` with `allTime` and in other places you're adding `allTime` in addition to the `year` as an option. 
ie. in the render for mobile & desktop - we never render year, only allTime so it looks like the client will ever use it? 
If we're fully replacing year with allTime, I think we should just remove year everywhere and map year to allTime when we get the trending if it's ever returned? Unless we wanted to be able to switch between the two later on?

isaacsolo:
> In some places you're replacing `year` with `allTime` and in other places you're adding `allTime` in addition to the `year` as an option. ie. in the render for mobile & desktop - we never render year, only allTime so it looks like the client will ever use it? If we're fully replacing year with allTime, I think we should just remove year everywhere and map year to allTime when we get the trending if it's ever returned? Unless we wanted to be able to switch between the two later on?

I left allTime in some places because we might use year sometime in the future, but I could just remove now and we could add it back in when the time comes. That would probably be less confusing.

audius-infra:
Preview this change https://demo.audius.co/is-trending-all-time

isaacsolo:
> LGTM - assuming the discovery changes for `/trending` endpoint with the allTime query param hasn't been deployed yet because I'm seeing week results for the all time tab

yeah changes are not deployed yet and we still have year results instead of all time for now.

audius-infra:
Preview this change https://demo.audius.co/is-trending-all-time

audius-infra:
Preview this change https://demo.audius.co/is-trending-all-time

PR #780, opened 2022-01-14
[AUD-1182 Visualizer performance

milindNshah:
### Description
- Some small UI changes
- Modified code for how dominant colors are picked
- Fixed visualizer lagging (reverted large parts of visualizer-1.js)
- Updated the VisualizerProvider to use hooks instead of classes
- Fixed bug where visualizer sometimes displayed default Dominant colors while switching between tracks

### How Has This Been Tested?
Ran on staging and personal domian.
- Basic tests to make sure UI is working as expected
- Stress tests with opening and closing visualizer a lot
- Stress tests with playing/pausing/next/prev with visualizer open

| Link | Threshold | Max Euclidean |
|--|--|--|
|https://milind.audius.co/isaacsolo/on-me-with-theo|T1|ME1||
|https://milind.audius.co/fuxwithitblog/lyny-x-chromonicci-want|T2|ME2||
|https://milind.audius.co/pacificpatterns/im-back|T3|ME3||
|https://milind.audius.co/Lifted/kryptt-elder-god|T4|ME4||
|https://milind.audius.co/destinuna/simpl-x-bonzaiii-seek|T5|ME5||
|https://milind.audius.co/alvinpresley/dil-ke-dairon-pe|T6|ME6|
|https://milind.audius.co/mysoundsucks_/panama|T7|ME7|
|https://milind.audius.co/bostn64/isoxo-click-bostn-flip|T8|ME8|
|https://milind.audius.co/heyjubie/angel-hair|T9|ME9|


audius-infra:
Preview this change https://demo.audius.co/ms-visualizer-performance

src/pages/visualizer/VisualizerProvider.tsx
```diff
@@ -119,67 +79,75 @@ class Visualizer extends Component<VisualizerProps, VisualizerState> {
...
-  }
+  // Update Colors
+  useEffect(() => {
+    if(dominantColors !== null) {
+      Visualizer1?.setDominantColors(dominantColors)
+    }
+  }, [isVisible, currentQueueItem])
+
+  // Rebind audio
+  useEffect(() => {
```
dylanjeffers, line 98: one thought is that we could create custom hooks with the name like `useRebindAudio`

dylanjeffers (commented):
These changes look great

src/utils/visualizer/visualizer-1.js
```diff
@@ -17,134 +17,133 @@ const setIdentity = require('gl-mat4/identity')
...
-    function createSegment () {
-      return newArray(steps).map((i, _, list) => {
-        const x = lerp(-1, 1, i / (list.length - 1))
-        return [ x, 0, 0 ]
-      })
-    }
+  let analyser = null
+  setDominantColors()
 
+  function show (darkMode) {
```
raymondjacobson, line 146: I know not your change, but generally, we prefer anonymous functions

const show = (darkMode) => { ... }

to the function notation because you don't have to worry about binding `this` ,etc. no need to change here, but wanted to call out :)

src/pages/visualizer/VisualizerProvider.tsx
```diff
@@ -119,67 +79,75 @@ class Visualizer extends Component<VisualizerProps, VisualizerState> {
...
+      setTimeout(() => {
+        setShowVisualizer(false)
+        Visualizer1?.hide()
+        recordClose()
+      }, 300)
+    }
+  }, [fadeVisualizer])
 
+  
+  const goToTrackPage = () => {
```
raymondjacobson, line 134: these probably could benefit from being useCallback-ified

```
const goToTrackPage = useCallback(() => {
...
}, [currentQueueItem, goToRoute])
```

raymondjacobson (approved):
ty SO much for refactors. didn't leave any comments on the distance algo itself b/c I imagine you'll update that. Looks great 🔥

audius-infra:
Preview this change https://demo.audius.co/ms-visualizer-performance

src/pages/visualizer/VisualizerProvider.tsx
milindNshah: Yeah I think that'd be a good idea if this logic was re-used but this is the only place this particular effect exists.

audius-infra:
Preview this change https://demo.audius.co/ms-visualizer-performance

src/pages/visualizer/VisualizerProvider.tsx
raymondjacobson: I think I generally agree w/ not breaking out a hook into its own useX unless it's 
1) super complicated
2) shared w/ other components

I do think comments for the useEffects are useful though

src/workers/dominantColor.worker.js
```diff
@@ -83,6 +86,48 @@ export default () => {
...
+        domColors,
+        selectFrom
+      )
+      domColors.push(selectFrom[indexOfNextDomColor])
+      selectFrom.splice(indexOfNextDomColor, 1)
+    }
+    return domColors
+  }
+
+  const findIndexOfMaxEuclideanDistance = (existing, selectFrom) => {
```
raymondjacobson, line 103: the leetcode goes to show. this is snazzy.
I don't think there's a huge perf gain for it, but you *could* probably memoize some of the euclidean distances along the way to save future computation if you see the same pair again

biggest advantage here is that this runs in a web worker so it has an isolated js context and wont block rendering on the main thread

raymondjacobson (approved):

src/pages/visualizer/VisualizerProvider.tsx
dylanjeffers: That all makes sense. Comments are great, agreed

PR #784, opened 2022-01-14
[AUD-1215] Add typed-redux-sagas

dylanjeffers:
### Description

Uses `typed-redux-sagas` package to improve our saga yield typings. Also uses `macro` extension to transpile away the typed-redux-saga wrapping, to avoid extra function calls. Implements this on the share-modal saga for an example

### Dragons

Now that things are typed correctly, is there a clean way to handle a selector returning null, or are the `if (!track) return` cases sufficient?


package.json
```diff
@@ -89,6 +89,7 @@
     "semver": "6.3.0",
     "simplebar-react": "2.1.0",
     "simplebar-react-legacy": "npm:simplebar-react@1.2.3",
+    "typed-redux-saga": "^1.3.1",
```
sliptype, line 92: Let's pin these new deps

sliptype (approved):
Niiice

Does this make https://github.com/AudiusProject/audius-client/pull/776 unnecessary? Or do we still need it for `call`?

Re the nullchecks, it's probably good to be explicit about them like this

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1215-typed-redux-sagas

dylanjeffers:
> Niiice
> 
> 
> 
> Does this make https://github.com/AudiusProject/audius-client/pull/776 unnecessary? Or do we still need it for `call`?
> 
> 
> 
> Re the nullchecks, it's probably good to be explicit about them like this

Was talking with marcus about it, and we believe it should, but haven't tested yet

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1215-typed-redux-sagas

rickyrombo (approved):

PR #791, opened 2022-01-18
Only override current step count if challenge is not complete

sddioulde:
### Description

Do not optimistically override the challenge current step count if the challenge is already/has already been complete

### Dragons

### How Has This Been Tested?

locally against staging


audius-infra:
Preview this change https://demo.audius.co/sd-complete-challenge

rickyrombo (approved):

PR #794, opened 2022-01-19
Cancel reward claim on modal close

rickyrombo:
### Description

Fixes a bug introduced by the move to using sagas for the claim rewards button. Since the button was being unmounted whenever the modal was closed, it didn't have this problem. But now, the claim is processed in the background and needs to be explicitly stopped on unmount.

Prior to this change, if the claim took sufficiently long, its resolution could come while the user has a different modal open. This could cause the user to see an error on the wrong modal, for instance.

Also removes unnecessary state for `displayError` and `setDisplayError`. As we no longer need to reset the state post-claim, we can simply check the claim state instead.

### Dragons

~The HCaptcha modal sets visibility to false for the challenge reward modal, but should not cancel the claim.~ ✅ 

### How Has This Been Tested?

Tested manually on web against staging

N/A

### DIFF
--- a/src/common/store/pages/audio-rewards/slice.ts
+++ b/src/common/store/pages/audio-rewards/slice.ts
@@ -107,13 +107,6 @@ const slice = createSlice({
       const { modalType } = action.payload
       state.challengeRewardsModalType = modalType
     },
-    setClaimStatus: (state, action: PayloadAction<{ status: ClaimStatus }>) => {
-      const { status } = action.payload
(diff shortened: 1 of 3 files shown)


audius-infra:
Preview this change https://demo.audius.co/mjp-cancel-claim-on-close

sddioulde (approved):
looks good

Kyle-Shanks (approved):
Nice 👌

PR #795, opened 2022-01-19
Consume UserChallenge amount instead of using config

rickyrombo:
### Description

Make the discovery nodes the source of truth for reward amounts.

### Dragons

N/A

### How Has This Been Tested?

Tested manually against stage

N/A


src/common/models/AudioRewards.ts
```diff
@@ -79,3 +80,15 @@ export type FlowSessionEvent =
   | FlowSessionResumeEvent
   | FlowSessionPassEvent
   | FlowSessionFailEvent
+
+/**
+ * @deprecated amounts should be pulled in directly from user challenges instead
```
rickyrombo, line 85: Still need config for the notifications.

Moved this here so at least notifications can share one source of truth between web/mobile

audius-infra:
Preview this change https://demo.audius.co/mjp-consume-amount

Kyle-Shanks (approved):

audius-infra:
Preview this change https://demo.audius.co/mjp-consume-amount

src/common/models/AudioRewards.ts
piazzatron: nice - might be nice to include this context in the comment so we remember why we're keeping this around

piazzatron (approved):

PR #796, opened 2022-01-19
Ms visualizer bug fixes

milindNshah:
### Description
Fixed some bugs
- [x]  Visualizer click goes to track page
- [x]  Firefox: visualizer icon not centered
- [x]  Visualizer album art is lower quality, upgrade to higher quality
- [x]  Use larger album art when selecting the colors

### How Has This Been Tested?
- Tested visualizer icon on firefox, now looks centered
- Tested image sizes for dominant colors being the same using this track from two different locations
    - https://milind.audius.co/trending?timeRange=month
    - https://milind.audius.co/dontlisten/heliosky-hold
- Tested out changes to prevent going to track page on visualizer click

### DIFF
--- a/src/common/store/cache/tracks/sagas.js
+++ b/src/common/store/cache/tracks/sagas.js
@@ -9,7 +9,7 @@ import {
 } from 'redux-saga/effects'
 
 import { Name } from 'common/models/Analytics'
-import { DefaultSizes } from 'common/models/ImageSizes'
+import { DefaultSizes, SquareSizes } from 'common/models/ImageSizes'
 import Kind from 'common/models/Kind'
 import Status from 'common/models/Status'
 import {
@@ -408,7 +408,13 @@ function* watchFetchCoverArt() {
         cacheActions.update(Kind.TRACKS, [{ id: trackId, metadata: track }])
       )
 
-      const dominantColors = yield call(dominantColor, url)
+      const largeImageUrl = yield call(
+        AudiusBackend.getImageUrl,
+        multihash,
+        SquareSizes.SIZE_1000_BY_1000,
+        gateways
+      )
+      const dominantColors = yield call(dominantColor, largeImageUrl)
       yield put(
         setDominantColors({
           multihash,
--- a/src/common/store/cache/tracks/sagas.js
+++ b/src/common/store/cache/tracks/sagas.js
@@ -408,13 +408,17 @@ function* watchFetchCoverArt() {
         cacheActions.update(Kind.TRACKS, [{ id: trackId, metadata: track }])
       )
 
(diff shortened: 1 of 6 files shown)


audius-infra:
Preview this change https://demo.audius.co/ms-visualizer-bug-fixes

audius-infra:
Preview this change https://demo.audius.co/ms-visualizer-bug-fixes

src/common/store/cache/tracks/sagas.js
```diff
@@ -408,7 +408,17 @@ function* watchFetchCoverArt() {
         cacheActions.update(Kind.TRACKS, [{ id: trackId, metadata: track }])
       )
 
-      const dominantColors = yield call(dominantColor, url)
+      let smallImageUrl = url
+      if (coverArtSize !== SquareSizes.SIZE_150_BY_150) {
```
raymondjacobson, line 412: I also don't think this is a super viable long-term solution -- we'll end up fetching two images a decent amt of the time (direct link to track page, etc.) -- it kind of undoes the work around having multiple image sizes.

Do you think we can we do this (or better the 1000x1000 request on opening of the visualizer?)

Or maybe we could check that the size is 480x480 here and only then fetch the dominant colors? 480x480 being the size of the now playing artwork.

I'm ok with merging this change for now since this action is handled async and won't block page loads or anything, but we should try to avoid doing this if we can.

src/components/nav/desktop/NavColumn.js
```diff
@@ -182,11 +182,15 @@ const NavColumn = ({
     if (route) goToRoute(route)
   }, [goToRoute, getTrackPageLink])
 
-  const onShowVisualizer = useCallback(() => {
-    if (NO_VISUALIZER_ROUTES.has(pathname)) return
+  const onShowVisualizer = useCallback(
+    e => {
+      e.stopPropagation()
```
raymondjacobson, line 187: I'd call stopPropagation after showVisualizer just to make sure that happens as uninterrupted as possible

raymondjacobson (approved):
I'm still noticing a little bit of weirdness around the fade animation sometimes not firing properly (pressing `v` or clicking the button). Maybe `display: none` gets set too early? Or perhaps we need to clear the setTimeout to set the display prop if the user interrupts the animation? The repro I'm seeing is if you press v and close the visualizer too early, the next time you open it, it snaps in rather than fading

src/common/store/cache/tracks/sagas.js
milindNshah: I'll try out the fetching on visualizer open in a separate PR

raymondjacobson: agree!

PR #797, opened 2022-01-19
[AUD-1281] Add config for referred challenge for notifications

rickyrombo:
### Description

Follows #795 

AUD-1281

Adds config for "referred" challenge despite it not having a modal, as it is needed for the title and amount of the corresponding notification on web.

### Dragons

N/A

### How Has This Been Tested?

Manually against staging

N/A

### DIFF
--- a/src/pages/audio-rewards-page/config.tsx
+++ b/src/pages/audio-rewards-page/config.tsx
@@ -102,7 +102,7 @@ export const challengeRewardsConfig: Record<
       `Invite your Friends! You’ll earn ${amount} $AUDIO for each friend who joins with your link (and they’ll get an $AUDIO too)`,
     progressLabel: '%0/%1 Invites',
     amount: amounts.referrals,
-    stepCount: 5,
+    stepCount: 1,
     panelButtonText: 'Invite your Friends',
     modalButtonInfo: {
       incomplete: null,
--- a/src/pages/audio-rewards-page/config.tsx
+++ b/src/pages/audio-rewards-page/config.tsx
@@ -92,6 +92,24 @@ export const challengeRewardsConfig: Record<
       complete: null
     }
   },
+  // This is used just for the notifications
+  referred: {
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://demo.audius.co/mjp-referred-notifs

Kyle-Shanks (commented):
IIRC Piazza had removed this from the config recently. Should probably check with him before merging this in case there's a reason it isn't needed

src/pages/audio-rewards-page/config.tsx
```diff
@@ -92,6 +92,24 @@ export const challengeRewardsConfig: Record<
...
+  referred: {
+    id: 'referrals' as ChallengeRewardID,
+    title: 'Invite your Friends',
+    icon: <i className='emoji large incoming-envelope' />,
+    description: amount => `Earn ${amount} $AUDIO, for you and your friend`,
+    fullDescription: amount =>
+      `Invite your Friends! You’ll earn ${amount} $AUDIO for each friend who joins with your link (and they’ll get an $AUDIO too)`,
+    progressLabel: '%0/%1 Invites',
+    amount: amounts.referrals,
+    stepCount: 5,
```
piazzatron, line 105: Referred should be a one-and-done reward, but I'm guessing no impact to the notif of having stepCount = 5?

piazzatron (approved):

piazzatron:
> IIRC Piazza had removed this from the config recently. Should probably check with him before merging this in case there's a reason it isn't needed



> IIRC Piazza had removed this from the config recently. Should probably check with him before merging this in case there's a reason it isn't needed

Moved it out since we never show a panel for this challenge, but turns out that broke notifications 😿

src/pages/audio-rewards-page/config.tsx
rickyrombo: yeah good point. would have nulled all the fields but they're required - i'll fix this anyway to prevent future bugs

PR #805, opened 2022-01-20
Add tests for claim rewards saga

rickyrombo:
### Description

Adds a bunch of little regression tests for the claimRewards sagas. (Mostly was an experiment to learning more about testing sagas)

### Dragons

### How Has This Been Tested?

These are tests

N/A


raymondjacobson:
😍😍😍

src/pages/audio-rewards-page/store/sagas.test.ts
```diff
@@ -0,0 +1,331 @@
...
+      args &&
+      args.length > 1 &&
+      args[1].toString().indexOf('User is blocked from claiming') < 0
+    ) {
+      oldConsoleError(args)
+    }
+  })
+})
+
+describe('Claim Rewards Async saga', () => {
```
raymondjacobson, line 106: Previously our convention had been to name this file

store.test.js because it is testing a bit more than just the saga itself.
https://github.com/AudiusProject/audius-client/blob/master/src/store/lineup/store.test.js

thoughts? i think consistency is more important than semantics

src/pages/audio-rewards-page/store/sagas.test.ts
```diff
@@ -0,0 +1,331 @@
...
+              args: [{ challengeId: testUserChallenge.challenge_id }]
+            }),
+            { is_completed: false }
+          ]
+        ])
+        // Assertions
+        .not.call(AudiusBackend.submitAndEvaluateAttestations)
+        .not.put(setUserChallengeDisbursed({ challengeId: 'connect-verified' }))
+        .not.put(claimChallengeRewardSucceeded())
+        .silentRun(5)
```
raymondjacobson, line 245: how did you pick these timeouts for silentRun?

raymondjacobson (approved):
awesome! I learned some things here

src/pages/audio-rewards-page/store/sagas.test.ts
rickyrombo: By reducing the timeout as much as I could without the tests failing 😅 probably a bit aggressive

src/pages/audio-rewards-page/store/sagas.test.ts
rickyrombo: Sure, I have no preference here really.

src/pages/audio-rewards-page/store/sagas.test.ts
raymondjacobson: be careful that it's not faster on your machine and this works the same on CI. maybe worth having some default sensible longer timeouts? i doubt we will fail because of timeouts for any good reason so we should just let the test run

rickyrombo: We can't run without timeouts because take every and takeLatest etc never terminate. The test plan defaults to having some longish timeouts, though, so we could stick to those.

Alternatively, I could export the individual sagas to work around takeLatest etc but then I'm modifying the source for the test

raymondjacobson: what if the default was just like 500 or something and we use that as a consistent config between tests?

rickyrombo: I think that is the default. I'll unset these manually configured timeouts and use that. We can tweak the default timeout separately (Imho that's wayy too long since it becomes the minimum time fo these tests)

raymondjacobson: i think having different timing for each test where someone will have to know how to tweak them / stand the chance of arbitrarily failing will cost us more time than a 500ms/test run, but maybe 100 is reasonable. feel free to just merge as is. good discussion

rickyrombo: One last note: AFAIK timeouts don't cause test failures. It outputs a warning and some of your asserts might not get attempted. (This is bad)

raymondjacobson: that seems really bad... sounds like picking a higher number is better even if the test is a little bit slower to avoid false negatives

rickyrombo: yeah. I should also not use silent so we can get those warnings. looks like default timeout is 250ms btw which is better: https://github.com/jfairbank/redux-saga-test-plan/blob/master/docs/integration-testing/timeout.md

audius-infra:
Preview this change https://demo.audius.co/mjp-test-rewards-saga

src/pages/audio-rewards-page/store/sagas.ts
```diff
@@ -306,9 +307,10 @@ function* userChallengePollingDaemon() {
     IntKeys.CHALLENGE_REFRESH_INTERVAL_AUDIO_PAGE_MS
   )!
   yield fork(function* () {
-    yield* visibilityPollingDaemon(fetchUserChallenges())
+    yield call(visibilityPollingDaemon, fetchUserChallenges())
```
rickyrombo, line 310: Changing these to `call`s let me mock them

src/pages/audio-rewards-page/store/sagas.ts
```diff
@@ -244,7 +244,8 @@ export function* watchFetchUserChallenges() {
     const currentUserId: number = yield select(getUserId)
 
     try {
-      const userChallenges: UserChallenge[] = yield apiClient.getUserChallenges(
+      const userChallenges: UserChallenge[] = yield call(
```
rickyrombo, line 247: Changing this to a `call` makes it mockable

src/pages/audio-rewards-page/store/sagas.ts
```diff
@@ -186,6 +185,7 @@ function* claimChallengeRewardAsync(
           default:
             throw new Error(`Unknown Error: ${response.error}`)
         }
+        yield put(claimChallengeRewardWaitForRetry(claim))
```
rickyrombo, line 188: Moved this to the bottom of the switch to prevent a "bug" where unknown errors would still put a retry in the store

src/services/audius-api-client/AudiusAPIClient.ts
```diff
@@ -1189,7 +1189,7 @@ class AudiusAPIClient {
     return encodedUserId ? decodeHashId(encodedUserId.toString()) : null
   }
 
-  async getUserChallenges({ userID }: GetUserChallengesArgs) {
+  getUserChallenges = async ({ userID }: GetUserChallengesArgs) => {
```
rickyrombo, line 1192: Since we use `this` in this function, I need to use arrow functions so that other callers don't override `this`. See: https://www.typescriptlang.org/play#example/this

I could also use `bind()`at the place where I use call but that doesn't let me mock it properly since it creates an entirely new function (and generally looks pretty ugly).

Could maybe sneak around this using `spy` but really curious what people think about using arrow functions here instead.

src/pages/audio-rewards-page/store/sagas.test.ts
rickyrombo: Well, actually those warnings aren't helpful, because I expect every test to timeout. I think maybe we lean towards exporting each individual generator we want to test, and that way we can avoid `takeEvery`/`takeLatest` and only test the generators they call underneath? Would help make the tests run faster, too.

audius-infra:
Preview this change https://demo.audius.co/mjp-test-rewards-saga

src/services/audius-api-client/AudiusAPIClient.ts
raymondjacobson: always use arrow imo. it's more normal function semantics and i like consistency

src/pages/audio-rewards-page/store/sagas.ts
dylanjeffers, line 247: can we try using yield* and typed-redux-saga for this file? does that mess up testing?

rickyrombo: Yeah we could - this branch predated that change but this file is probably a great candidate to start converting! I'll try it out and see how it affects testing

dylanjeffers (approved):
amazing job!!! this all looks great from what i can tell. i typically focus on just testing the ui, and let redux do the things to change it, but esp in our case that we are sharing redux logic across multiple apps, this type of testing makes even more sense

src/pages/audio-rewards-page/store/sagas.test.ts
dylanjeffers: yeah i like this idea

src/pages/audio-rewards-page/store/sagas.test.ts
raymondjacobson: i'm ok with that. sometimes the nuance of takeLatest/takeEvery is important to capture, but generally I agree that testing the called saga is better. let's come up w/ a consistent naming pattern

```
doThing* () {
}

watchDoThing* () {
  yield takeEvery(actions.thing, doThing)
}
```

raymondjacobson (approved):

rickyrombo:
Going to merge as is and follow up with some edits

PR #811, opened 2022-01-21
Upgrade redux-saga from v0.16.0 to v1.1.3

dylanjeffers:
### Description

Upgrades redux-saga from v0.16.0 to v1.x.x. This helps keep us up to date (we were on 2 year old version), and paves the way for typed-redux-saga to correctly type things! Note that a few of these changes are temporary type fixes which will be removed once typed-redux-saga is enabled.

### Dragons

There are many breaking changes introduced with v1 redux-saga. Below is a checklist derived from https://github.com/redux-saga/redux-saga/blob/master/CHANGELOG.md to ensure we cover all cases:

- [x]   errors thrown during put execution are no longer caught and swallowed, you need to catch them manually (based on our discussions this seems highly unlikely)
- [x]  errors thrown during cancellation process are no longer swallowed, you need to keep finally block fail-safe
- [x]   removed some deprecated APIs - takeEvery, takeLatest, throttle from the redux-saga entry point (they are and were importable from redux-saga/effects). put.sync and takem were removed.
- [x]   removing execution of an array of effects yield [...]. use all effect instead.
- [x] delay became an effect, old delay function (not effect!) can be imported from @redux-saga/delay-p
- [x]  put.resolve was changed to putResolve
- [x] take.maybe was changed to takeMaybe
- [x] changed API of runSaga - it no longer accepts subscribe option, you should create a channel (preferably stdChannel), pass it as channel argument to the runSaga API and communicate with through it with take and put methods
- [x] task.done getter was changed to be task.toPromise method
- [x] onError doesn't extend error with additional field sagaStack, but pass it as a property of second argument. before: onError: (e: Error), after: onError(e: Error, { sagaStack })
- [x] Effect shape, yielded to redux-saga middleware, is stabilized and declared now as a plain JavaScript object
- [x] channels private getters (takers and closed) got removed
- [x] {effects, utils} aren't imported from 'redux-saga' anymore. imports them from redux-saga/effects, redux-saga/utils
- [x] is helper should be imported from @redux-saga/is.
- [x] createMockTask, cloneableGenerator should be imported from @redux-saga/testing-utils
- [x] now race should be finished if any of effects resolved with END (by analogy with all)
- [x] cancel and join cannot receive variadic arguments anymore. so you have to rewrite cancel(...[tasks]) and join(...[tasks]) to cancel([tasks]) and join([tasks]) respectively. also calling cancel(...) returns a cancel-effect (before it may return an all effect), and calling join(...) returns a join-effect.
- [x] refactor effect structure from {[IO]: true, [type]: payload } to { [IO]: true, type, payload } to get rid of dynamic type property. Could affect you if implement custom monitor for saga effects.
- [x] channel and actionChannel have default buffer of buffers.expanding()
- [x] eventChannel does no longer accept matcher argument.
- [x] exported util of arrayOfDeffered got renamed to the correct arrayOfDeferred


src/store/lineup/sagas.js
```diff
@@ -1,4 +1,4 @@
-import { delay } from 'redux-saga'
+import delay from '@redux-saga/delay-p'
```
raymondjacobson, line 1: i think this should be now from @redux-saga/effects

https://redux-saga.js.org/docs/recipes/#debouncing

src/store/queue/sagas.ts
```diff
@@ -167,21 +174,22 @@ export function* watchPlay() {
...
 
-      const user: User = playActionTrack
+      const user: User | null = playActionTrack
         ? yield select(getUser, { id: playActionTrack.owner_id })
         : null
 
       // Skip deleted tracks
       if (
         (playActionTrack && playActionTrack.is_delete) ||
+        // @ts-ignore this is technically impossible with typing, but keeping for now to avoid potential runtime issues
```
raymondjacobson, line 184: better to do a ! operator w/ a comment?

src/store/sagas.js
```diff
@@ -151,5 +151,5 @@ export default function* rootSaga() {
   if (NATIVE_MOBILE) {
     sagas.push(initInterface)
   }
-  yield sagas.map(fork)
+  yield all(sagas.map(fork))
```
raymondjacobson, line 154: wonder if we actually need the fork

src/store/queue/sagas.ts
rickyrombo: Maybe I'm missing it but why is it impossible? Can't `user` be null?

rickyrombo (approved):
Looks good! Let's double check the `delay` import though and make sure we're importing from the right place though too.

sliptype (approved):
Looks good so far! A couple tests seem to have broken though: https://app.circleci.com/pipelines/github/AudiusProject/audius-client/3298/workflows/b0ff47b2-1a1b-4733-8403-1f1258a5ae72/jobs/19262

src/store/queue/sagas.ts
dylanjeffers: its a similar case to the ts demo today: user is type `never` here since its set to null if playActionTrack is false, which if looking at the if statement, will always be null since (playActionTrack && playActionTrack.is_delete) would only pass execution via || if it was false, therefore saying `user` would be type null, so checking `user.is_deactivated` is not a possible case.

ultimately, we can probably jut remove this || check, but wanted to keep the code just to make sure i dont break runtime execution, since the types are in a funky enough state that they may not be reliable till we get the `typed-redux-saga` stuff going. Thoughts?

rickyrombo: Can't the `select()` be undefined/null though if hte user isn't in the store?

dylanjeffers: so it boils down to `playActionTrack`, when that i falsy, user is null, so `(playActionTrack || user.is_deactivated)`, -> `(false || null.is_deactivated)` so ts types become `(boolean || never)`

rickyrombo: I don't think I follow. `playActionTrack` can be true, `getUser` could return undefined (so `user` is undefined) and `playActionTrack.is_delete` can be `false` all at the same time, no?

`if ((true && false) || undefined)`

dylanjeffers: oh youre so right! will update the comment for now. looks like user type is just "null", which is happening for unknown reasons atm, but can address when we convert to typed-redux-saga

src/store/lineup/sagas.js
dylanjeffers: nice catch, this was based off this change: delay became an effect, old delay function (not effect!) can be imported from @redux-saga/delay-p
so wasnt sure if switching from the function version to the effect version would break things?

audius-infra:
Preview this change https://demo.audius.co/dj-upgrade-redux-saga

src/store/sagas.js
dylanjeffers: just checked, looks like `all` doesn't fork automatically :(

src/store/lineup/sagas.js
raymondjacobson: i think the interface is the same so probably alright?

audius-infra:
Preview this change https://demo.audius.co/dj-upgrade-redux-saga

PR #813, opened 2022-01-24
[AUD-1353] Improve a11y to simplify probers e2e flows

dylanjeffers:
### Description

- Adds a11y fixes for core audius flows. This makes writing and maintaining our new cypress e2e tests much easier, since we can leverage https://testing-library.com/docs/cypress-testing-library
- Also includes performance fix for @audius-stems, which reduces its bundle size by over 70%. There are a11y fixes in that change too that make cypress tests easier to write.


src/pages/App.module.css
```diff
@@ -49,9 +49,6 @@ body {
   margin-bottom: 88px;
   width: 100%;
   text-align: center;
-  /* Show scrollbar always so the page doesn't jump around. */
-  overflow-y: scroll;
-  overflow-x: hidden;
```
sliptype, line 54: Not sure why this was needed but it seems kinda dangerous to remove

audius-infra:
Preview this change https://demo.audius.co/dj-probers-a11y

sliptype (approved):
Awesome! This is a great step in the right direction

src/pages/App.module.css
dylanjeffers: yes this was the one i should have called out in the dragons section. so for some reason cypress is complaining that some buttons get hidden by a parent specifying overflow: hidden. here is the error writedown that mentions it: https://docs.cypress.io/guides/references/error-messages#cy-failed-because-the-element-cannot-be-interacted-with
we can keep the style if we think this is a dangerous change, and i can just do a `force-click` in cypress for now. what do you think?

sliptype: I wish we had the history of this file bc I don't have enough context. But I think using the force click might be better than potentially breaking something here

src/components/play-bar/PlayButton.js
```diff
@@ -137,17 +142,28 @@ class PlayButton extends Component {
     const animation = (
       <div className={styles.animation}>
         <Lottie
+          ariaRole={null}
```
rickyrombo, line 145: Just double checking - should this be null?

audius-infra:
Preview this change https://demo.audius.co/dj-probers-a11y

rickyrombo (approved):
Love that we're getting some a11y support going!

Have you verified that these all look good with the new tags? I forget whether we have the right reset CSS to keep things consistent (don't want to have `h2`s accidentally making things bold or bigger, for example).

audius-infra:
Preview this change https://demo.audius.co/dj-probers-a11y

dylanjeffers:
> Love that we're getting some a11y support going!
> 
> 
> 
> Have you verified that these all look good with the new tags? I forget whether we have the right reset CSS to keep things consistent (don't want to have `h2`s accidentally making things bold or bigger, for example).

Great calls re h2, I was checking and it appeared to be the same, but I'll do a more rigorous check

src/components/play-bar/PlayButton.js
dylanjeffers: yes that is intentional! lottie incorrectly sets roles and labels, so this is a way to prevent that in the interim. long term, we should start using the more up-to-date (forked) version of lottie which fixes this

milindNshah (approved):
Looks great!

PR #815, opened 2022-01-24
[AUD-1357] Fix hash routing config

dylanjeffers:
### Description

Fixes issue where react-scripts doesn't interpolate missing config values, resulting in `const useHashRouting = '%REACT_APP_USE_HASH_ROUTING%'`, making it truthy instead of falsy, which causes strange hash-routing issues during account-recovery flow


raymondjacobson (approved):
Fantastic find

PR #816, opened 2022-01-25
Handle referrals-verified challenge

piazzatron:
### Description
- Support DN's new challenge type, referrals-verified, which is the same as a regular referral challenge but with a higher max referral count.
- DN now returns different lists of challenges per user (verified/not-verified), incorporate that into the display logic

### Dragons

None

### How Has This Been Tested?

Tested with verified + unverified accounts against staging

n/a


audius-infra:
Preview this change https://demo.audius.co/aud-1349

rickyrombo (approved):

PR #817, opened 2022-01-25
Add feature flag enable_spl_audio

jowlee:
### Description
Adds feature flag for enable spl audio and replaces the remote-config enable spl audio variable
If enabled
* adds the user_bank_spl audio to the user's balance
* Update the /audio page to display the solana user bank address instead of the user's eth wallet 
* Update the /audio page to send using sol audio instead of eth audio

Closes AUD-1335

### Dragons
none

### How Has This Been Tested?
Ran against stage


audius-infra:
Preview this change https://demo.audius.co/jowlee-audio-balance

Kyle-Shanks (approved):

PR #820, opened 2022-01-25
[AUD-1363] Add a11y fixes for e2e flows

dylanjeffers:
### Description
Adds various fixes needed for e2e selectors to correctly query and interact with audius-client.

### DIFF
--- a/src/components/data-entry/DropdownInput.js
+++ b/src/components/data-entry/DropdownInput.js
@@ -41,7 +41,9 @@ class DropdownInput extends Component {
       disabled,
       isRequired,
       error,
-      focused = this.state.focused
+      id,
+      focused = this.state.focused,
+      ...other
     } = this.props
     let { placeholder } = this.props
 
@@ -110,11 +112,14 @@ class DropdownInput extends Component {
     return (
       <div className={cn(styles.wrapper, style)}>
         {label ? (
-          <div className={cn(styles.label, labelStyle)}>{label}</div>
+          <label htmlFor={id} className={cn(styles.label, labelStyle)}>
+            {label}
+          </label>
         ) : null}
         <div className={styles.dropdownInput}>
           <Select
             {...defaultValueProp}
+            id={id}
             dropdownClassName={cn(styles.select, style)}
             showSearch
             disabled={disabled}
@@ -131,6 +136,7 @@ class DropdownInput extends Component {
             notFoundContent={''}
             getPopupContainer={popupContainer}
             onDropdownVisibleChange={this.onVisibleChange}
+            {...other}
           >
             {options}
           </Select>
--- a/src/components/data-entry/DropdownInput.js
+++ b/src/components/data-entry/DropdownInput.js
@@ -30,6 +30,7 @@ class DropdownInput extends Component {
 
   render() {
     const {
+      'aria-label': ariaLabel,
       menu,
       defaultValue,
       label,
@@ -42,8 +43,7 @@ class DropdownInput extends Component {
       isRequired,
       error,
       id,
-      focused = this.state.focused,
-      ...other
+      focused = this.state.focused
     } = this.props
     let { placeholder } = this.props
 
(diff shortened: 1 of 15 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-aud-1363-a11y-fixes

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1363-a11y-fixes

src/components/data-entry/FormTile.js
```diff
@@ -110,6 +111,7 @@ const BasicForm = props => {
           </div>
           <div className={styles.categorization}>
             <DropdownInput
+              aria-label='Pick a Genre'
```
milindNshah, line 114: Let's put these into a messages object!

src/components/track/GiantTrackTile.js
```diff
@@ -168,7 +168,7 @@ class GiantTrackTile extends PureComponent {
                     : ButtonType.COMMON
                 }
                 widthToHideText={BUTTON_COLLAPSE_WIDTHS.second}
-                text={isReposted ? 'REPOSTED' : 'REPOST'}
+                text={isReposted ? 'reposted' : 'repost'}
```
milindNshah, line 171: Messages object here too

src/components/track/desktop/TrackTile.tsx
```diff
@@ -121,6 +121,8 @@ const TrackTile = memo(
       )
     }
 
+    const repostLabel = isReposted ? 'Unrepost' : 'Repost'
```
milindNshah, line 124: Same here

src/components/data-entry/DropdownInput.js
```diff
@@ -41,7 +41,9 @@ class DropdownInput extends Component {
       disabled,
       isRequired,
       error,
-      focused = this.state.focused
+      id,
+      focused = this.state.focused,
+      ...other
```
milindNshah, line 46: Should we have this here? Should't we just identify all the other possible values instead including aria-label and not pass unknown props down?

milindNshah (approved):
Just wanted to confirm that the repost button and the FinishPage components that were converted from divs to buttons didn't mess up the look of the UI. Mind posting screenshots here? 

Left a few minor comments but otherwise LGTM!

src/components/data-entry/FormTile.js
dylanjeffers: oh yeah! nice

src/components/track/GiantTrackTile.js
dylanjeffers: sure thing

src/components/data-entry/DropdownInput.js
dylanjeffers: yeah we def could do that, i mostly draw inspiration from https://github.com/mui-org/material-ui/blob/5daaf9be5115b1cb777fb4cf3d9125baddfe5720/packages/mui-material/src/ButtonBase/ButtonBase.js#L361 that just makes it easier to compose elements, so DropdownInput can act like the antd select. but in this case since the intention is just aria-label, can just be more explicit about it!

src/pages/upload-page/components/FinishPage.module.css
```diff
@@ -10,14 +10,15 @@ sbuttons {
 }
 
 .continueButton {
+  all: unset;
```
dylanjeffers, line 13: @milindNshah so from what i can tell, this resets a button to look like a div. will get some screenshots to confirm as well

dylanjeffers:
> Just wanted to confirm that the repost button and the FinishPage components that were converted from divs to buttons didn't mess up the look of the UI. Mind posting screenshots here?
> 
> Left a few minor comments but otherwise LGTM!

Nice call on the button changes, here are some screenshots of before and after:

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1363-a11y-fixes

PR #822, opened 2022-01-26
Add amplitude user bank tracking

jowlee:
### Description
Adds amplitude tracks for user bank creation.
Since signup also creates a user bank account and is nested in libs, the track method is passed to the signup method which tracks it. Handled in https://github.com/AudiusProject/audius-protocol/pull/2387

Fixes AUD-1324

### Dragons
none

### How Has This Been Tested?
ran locally


audius-infra:
Preview this change https://demo.audius.co/jowlee-amplitude-user-bank

piazzatron (approved):

audius-infra:
Preview this change https://demo.audius.co/jowlee-amplitude-user-bank

PR #827, opened 2022-01-27
[AUD-1346] Add already claimed check and client side retries

raymondjacobson:
### Description

Add retry to UNKNOWN_ERROR in claiming (found this to be the most straightforward way to add to this flow).

Also add a state for the "ALREADY CLAIMED" case. We definitely don't want to retry here, and we should probably show a message. This would be redundant with the toast that says our bots did the work, but that toast may show up in a different context and the user may still try to claim. Want to confirm this behavior a bit with Forrest & Julian.

### Dragons

### How Has This Been Tested?

Locally against staging


audius-infra:
Preview this change https://demo.audius.co/rj-aud-1346

audius-infra:
Preview this change https://demo.audius.co/rj-aud-1346

src/pages/audio-rewards-page/store/sagas.ts
```diff
@@ -199,8 +204,13 @@ function* claimChallengeRewardAsync(
       yield put(claimChallengeRewardSucceeded())
     }
   } catch (e) {
-    console.error('Error claiming rewards:', e)
-    yield put(claimChallengeRewardFailed())
+    // Retry once in the case of generic failure, otherwise log error and abort
+    if (retryOnFailure) {
+      yield put(claimChallengeReward({ claim, retryOnFailure: false }))
```
rickyrombo, line 209: This will retry if a user is blocked.

rickyrombo: Maybe move this logic to default case/UNKNOWN_ERROR case instead?

src/pages/audio-rewards-page/store/sagas.ts
```diff
@@ -179,6 +180,10 @@ function* claimChallengeRewardAsync(
               setVisibility({ modal: COGNITO_MODAL_NAME, visible: true })
             )
             break
+          case FailureReason.ALREADY_DISBURSED:
+            yield put(claimChallengeRewardAlreadyClaimed())
+            // Return out and do not retry. We've already earned this reward.
+            return
```
rickyrombo, line 186: Instead of returning here, maybe move `yield put(claimChallengeRewardWaitForRetry(claim))` into the cognito/hcaptcha cases?

rickyrombo (changes requested):

src/pages/audio-rewards-page/components/modals/ChallengeRewards.tsx
```diff
@@ -293,6 +294,9 @@ const ChallengeRewardsBody = ({ dismissModal }: BodyProps) => {
     if (claimStatus === ClaimStatus.SUCCESS) {
       toast(messages.rewardClaimed, CLAIM_REWARD_TOAST_TIMEOUT_MILLIS)
     }
+    if (claimStatus === ClaimStatus.ALREADY_CLAIMED) {
+      toast(messages.rewardAlreadyClaimed, CLAIM_REWARD_TOAST_TIMEOUT_MILLIS)
+    }
```
piazzatron, line 299: Thanks for handling this one - was tracking it on linear as part of another issue but it slipped through the cracks. The way Saliou and I were discussing handling this was to just treat this the same as success case, but since we're polling for balance updates now, it would be odd if your balance had already increased and doesn't increase with the claim, so I think this solution makes sense.

src/pages/audio-rewards-page/store/sagas.ts
```diff
@@ -179,6 +180,10 @@ function* claimChallengeRewardAsync(
               setVisibility({ modal: COGNITO_MODAL_NAME, visible: true })
             )
             break
+          case FailureReason.ALREADY_DISBURSED:
```
piazzatron, line 183: We should also handle the `ALREADY_SENT` error - this one comes from the reward manager, and you'll see if it you submit after the challenge has completed but before DN has indexed it.

piazzatron (approved):

src/pages/audio-rewards-page/store/sagas.ts
raymondjacobson: sgtm!

src/pages/audio-rewards-page/store/sagas.ts
raymondjacobson: Sounds good, I also removed the default in the switch statement b/c we handle all the cases

audius-infra:
Preview this change https://demo.audius.co/rj-aud-1346

src/pages/audio-rewards-page/store/sagas.ts
```diff
@@ -176,19 +178,34 @@ function* claimChallengeRewardAsync(
...
+            yield put(claimChallengeRewardWaitForRetry(claim))
+            break
+          case FailureReason.ALREADY_DISBURSED:
+            yield put(claimChallengeRewardAlreadyClaimed())
+            break
+          case FailureReason.ALREADY_SENT:
+            if (retryOnFailure) {
+              // In this case, we should retry, but give a bit of time.
+              yield delay(2000)
+              yield put(claimChallengeReward({ claim, retryOnFailure: false }))
```
piazzatron, line 196: Don't think we should retry here - this is error means rewards manager has already created the account representing the transfer, so it's never going to succeed

rickyrombo (approved):

piazzatron (approved):
LGTM - I think for the sake of choppy Solana network conditions we'll eventually want some sort of retry loop in the saga with backoff, but can do think about this more later

audius-infra:
Preview this change https://demo.audius.co/rj-aud-1346

audius-infra:
Preview this change https://demo.audius.co/rj-aud-1346

PR #828, opened 2022-01-27
Add 'Claim Rewards' button in nav

rickyrombo:
### Description

Adds a "Claim Rewards" button to the nav after waiting for disbursement longer than the configured delay.

New `MANUAL_CLAIM_PROMPT_DELAY_MS` configures how long to delay before prompting the user to claim.

Also adds fade effects for the buttons and hover/active effects to make them button-like.

image

### Dragons

- Needs to handle the case where both the balance is at 0 and the user has unclaimed challenges
- Rewrites a lot of the existing logic for `NavAudio`

### How Has This Been Tested?

- Adds two additional test cases for testing that the `fetchUserChallenges` saga triggers the appropriate actions.
- Tested manually against stage by using the Redux Dev Tools to dispatch the events and set wallet balance.

N/A

### DIFF
--- a/src/components/nav/desktop/NavAudio.module.css
+++ b/src/components/nav/desktop/NavAudio.module.css
@@ -63,7 +63,18 @@
   letter-spacing: 0.1em;
   transition: all 0.07s ease-in-out;
 }
+
+.actionBubble:hover {
+  cursor: pointer;
+  transform: scale3d(1.04, 1.04, 1.04);
+}
+
+.actionBubble:active {
+  transform: scale3d(0.99, 0.99, 0.99);
+}
+
 .actionBubble.claimRewards {
+  transition: opacity 0.2s ease-in-out;
   background: var(--static-accent-green);
 }
 
--- a/src/components/nav/desktop/NavAudio.module.css
+++ b/src/components/nav/desktop/NavAudio.module.css
@@ -24,14 +24,6 @@
   opacity: 1;
 }
 
-.audio.show .earnAudio:hover {
-  transform: scale(1.03);
-}
-
-.audio.show .earnAudio:active {
-  transform: scale(0.97);
-}
-
 .audio.hasBalance {
   cursor: pointer;
 }
@@ -74,15 +66,14 @@
 
 .actionBubble:hover {
   cursor: pointer;
-  transform: scale3d(1.04, 1.04, 1.04);
+  transform: scale3d(1.03, 1.03, 1.03);
 }
 
 .actionBubble:active {
-  transform: scale3d(0.99, 0.99, 0.99);
+  transform: scale3d(0.97, 0.97, 0.97);
 }
 
 .actionBubble.claimRewards {
-  transition: opacity 0.2s ease-in-out;
   background: var(--static-accent-green);
 }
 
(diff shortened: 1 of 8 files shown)


audius-infra:
Preview this change https://demo.audius.co/mjp-nav-audio-claim

src/components/nav/desktop/NavAudio.module.css
```diff
@@ -49,26 +41,48 @@
...
   font-size: var(--font-2xs);
   font-weight: var(--font-bold);
   letter-spacing: 0.1em;
   transition: all 0.07s ease-in-out;
+  position: absolute;
+}
+
+.actionBubble:hover {
+  cursor: pointer;
+  transform: scale3d(1.03, 1.03, 1.03);
```
raymondjacobson, line 69: ty!

src/components/nav/desktop/NavAudio.tsx
```diff
@@ -37,17 +43,56 @@ const NavAudio = () => {
...
+          time => time && new Date().getTime() - time > claimPromptDelayMs
+        )
+      ) {
+        setBubbleType('claim')
+      } else if (nonNullTotalBalance && !positiveTotalBalance) {
+        setBubbleType('earn')
+      } else {
+        setBubbleType('empty')
+      }
+    }, 100)
```
raymondjacobson, line 67: 100 seems aggressive here, since this is running all the time. what if we did 500?

Also I would think that instead of the interval, you could just do this with a useEffect that listens for changes to `pendingAutoClaims` since every time that is updated it would be a new ref

raymondjacobson (approved):

piazzatron:
Per our discussion this morning, since this button should always appear I think DELAY_MS should be 0 in optimizely and in our feature flag defaults?

rickyrombo:
> Per our discussion this morning, since this button should always appear I think DELAY_MS should be 0 in optimizely and in our feature flag defaults?

Per our convo this morning can most of this PR go away and we just use the `is_complete && !is_disbursed` to show it? Or do we want to keep it with hopes of having auto disbursement improve and become the norm?

audius-infra:
Preview this change https://demo.audius.co/mjp-nav-audio-claim

piazzatron:
> > Per our discussion this morning, since this button should always appear I think DELAY_MS should be 0 in optimizely and in our feature flag defaults?
> 
> Per our convo this morning can most of this PR go away and we just use the `is_complete && !is_disbursed` to show it? Or do we want to keep it with hopes of having auto disbursement improve and become the norm?

I support most of this going away - if auto disbursement becomes good enough and if we opt to go w that, we can always reference this PR to bring it back!

rickyrombo:
> > > Per our discussion this morning, since this button should always appear I think DELAY_MS should be 0 in optimizely and in our feature flag defaults?
> > 
> > 
> > Per our convo this morning can most of this PR go away and we just use the `is_complete && !is_disbursed` to show it? Or do we want to keep it with hopes of having auto disbursement improve and become the norm?
> 
> I support most of this going away - if auto disbursement becomes good enough and if we opt to go w that, we can always reference this PR to bring it back!

I think what I'll do is merge this as-is, then follow up with another PR which we can revert later (or reference this) should we want to go back? That way this stays as a good snapshot vs becoming the new PR

src/components/nav/desktop/NavAudio.tsx
rickyrombo: see if you like my update that uses `setTimeout`?

PR #829, opened 2022-01-27
Check for social links being a string rather than truthy

rickyrombo:
### Description

Changes the check in AudiusBackend to check for strings rather than truthy values to avoid not updating to empty string. Also adds tikTokHandle to update user

### Dragons

N/A

### How Has This Been Tested?

Tested against stage by setting twitter handle to empty string

N/A

### DIFF
--- a/src/services/AudiusBackend.js
+++ b/src/services/AudiusBackend.js
@@ -1146,11 +1146,11 @@ class AudiusBackend {
       }
 
       if (
-        newMetadata.twitter_handle ||
-        newMetadata.instagram_handle ||
-        newMetadata.tiktok_handle ||
-        newMetadata.website ||
-        newMetadata.donation
+        typeof newMetadata.twitter_handle === 'string' ||
+        typeof newMetadata.instagram_handle === 'string' ||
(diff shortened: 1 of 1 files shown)


src/services/AudiusBackend.js
```diff
@@ -1146,11 +1146,11 @@ class AudiusBackend {
       }
 
       if (
-        newMetadata.twitter_handle ||
-        newMetadata.instagram_handle ||
-        newMetadata.tiktok_handle ||
-        newMetadata.website ||
-        newMetadata.donation
+        typeof newMetadata.twitter_handle === 'string' ||
```
raymondjacobson, line 1149: this seems good to me. I wonder if we should instead do a manual check on empty string and coerce that to null (not sure how the BE would handle, but maybe more semantically correct)

raymondjacobson (approved):

src/services/AudiusBackend.js
rickyrombo: I think the BE checks for nulls though, doesn't it? I think empty string is safer?

PR #830, opened 2022-01-27
Optimistically update ui for listen streak challenge

jowlee:
### Description
Optimistically update the listen streak challenges in the client
Closes AUD-1379

### How Has This Been Tested?
Locally against stage

### DIFF
--- a/src/common/store/pages/audio-rewards/slice.ts
+++ b/src/common/store/pages/audio-rewards/slice.ts
@@ -100,6 +100,19 @@ const slice = createSlice({
         is_disbursed: true
       }
     },
+    setUserChallengeCurrentStepCount: (
+      state,
+      action: PayloadAction<{
+        challengeId: ChallengeRewardID
+        stepCount: number
+      }>
+    ) => {
+      const { challengeId, stepCount } = action.payload
+      state.userChallengesOverrides[challengeId] = {
+        ...state.userChallengesOverrides[challengeId],
+        current_step_count: stepCount
+      }
+    },
     setTrendingRewardsModalType: (
       state,
       action: PayloadAction<{ modalType: TrendingRewardsModalType }>
@@ -200,7 +213,8 @@ export const {
   fetchCognitoFlowUrlFailed,
   fetchCognitoFlowUrlSucceeded,
   showRewardClaimedToast,
-  resetRewardClaimedToast
+  resetRewardClaimedToast,
+  setUserChallengeCurrentStepCount
 } = slice.actions
 
 export default slice
--- a/src/common/store/pages/audio-rewards/slice.ts
+++ b/src/common/store/pages/audio-rewards/slice.ts
@@ -126,11 +126,7 @@ const slice = createSlice({
         userChallengeOverride &&
         'current_step_count' in userChallengeOverride
       ) {
-        if (Object.keys(userChallengeOverride).length > 1) {
-          userChallengeOverride.current_step_count = undefined
-        } else {
-          state.userChallengesOverrides[challengeId] = undefined
-        }
+        delete userChallengeOverride.current_step_count
       }
     },
     setTrendingRewardsModalType: (
--- a/src/common/store/pages/audio-rewards/slice.ts
+++ b/src/common/store/pages/audio-rewards/slice.ts
@@ -121,17 +121,15 @@ const slice = createSlice({
       }>
     ) => {
       const { challengeId } = action.payload
-      const userChallengeOverride =
-        state.userChallengesOverrides[challengeId] || {}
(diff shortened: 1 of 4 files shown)


src/pages/audio-rewards-page/store/sagas.ts
```diff
@@ -284,6 +285,23 @@ export function* watchFetchUserChallenges() {
...
 }
 
+export function* updateOptimisticListenStreak() {
+  const listenStreakChallenge: ReturnType<typeof getUserChallenge> = yield select(
+    getUserChallenge,
+    {
+      challengeId: 'listen-streak'
+    }
+  )
+  if (listenStreakChallenge?.current_step_count === 0) {
```
rickyrombo, line 295: Do we also need a case to remove the override if the challenge reports a step count > 0? Like if a user never closes a session, this will be stuck at 1?

audius-infra:
Preview this change https://demo.audius.co/jowlee-listen-streak-optimistic-update

src/common/store/pages/audio-rewards/slice.ts
```diff
@@ -112,6 +113,27 @@ const slice = createSlice({
...
+      action: PayloadAction<{
+        challengeId: ChallengeRewardID
+      }>
+    ) => {
+      const { challengeId } = action.payload
+      const userChallengeOverride =
+        state.userChallengesOverrides[challengeId] || {}
+      if ('current_step_count' in userChallengeOverride) {
+        if (Object.keys(userChallengeOverride).length > 1) {
+          // eslint-disable-next-line
```
rickyrombo, line 127: Why is this disabled?

src/common/store/pages/audio-rewards/slice.ts
```diff
@@ -112,6 +113,27 @@ const slice = createSlice({
...
+          const { current_step_count, ...restProps } = userChallengeOverride
+          state.userChallengesOverrides[challengeId] = restProps
```
rickyrombo, line 129: This is neat. Would `current_step_count = undefined` also work though?

src/pages/audio-rewards-page/store/sagas.ts
```diff
@@ -271,6 +273,16 @@ export function* watchFetchUserChallenges() {
         ) {
           newDisbursement = true
         }
+        if (
+          (challengeOverrides?.current_step_count ?? 0) > 0 &&
+          challenge.current_step_count !== 0
```
rickyrombo, line 278: Might make more sense to `put(updateOptimisticListenStreak)` on success and handle the logic there. I don't _think_ this should be general for every challenge type?

src/common/store/pages/audio-rewards/slice.ts
jowlee: it's because the variable `current_step_count` is not used causing a lint error

rickyrombo: ahh gotcha, can we disable just that rule?

src/pages/audio-rewards-page/store/sagas.ts
```diff
@@ -284,6 +297,25 @@ export function* watchFetchUserChallenges() {
   })
 }
 
+function* watchUpdateOptimisticListenStreak() {
+  yield takeEvery(updateOptimisticListenStreak.type, function* () {
```
piazzatron, line 301: why not just have this watch `socialActions.RECORD_LISTEN`?

jowlee: We had a talk about re-using actions vs making new specified actions with new intentions a while ago and decided to make new actions - I'd actually prefer to use record_listen, but was trying to go along the make a new action model

src/common/store/pages/audio-rewards/slice.ts
jowlee: just set to undefined

audius-infra:
Preview this change https://demo.audius.co/jowlee-listen-streak-optimistic-update

audius-infra:
Preview this change https://demo.audius.co/jowlee-listen-streak-optimistic-update

src/pages/audio-rewards-page/store/sagas.ts
rickyrombo: I think if we're going full separate actions, we'd need to fire `socialActions.RECORD_LISTEN` and `updateOptimisticListenStreak` from the same place, rather than firing one from the other?

rickyrombo (approved):
I think it looks good. Separately, we/I should probably revisit having these all live in the `rewards-page` sagas and maybe make a root saga instead for some of the handling around fetching challenges, especially since the polling is app-wide

PR #833, opened 2022-01-28
Add type defs for App

dylanjeffers:
### Description

Adds type definition file for App.js to ensure consumers pass it the correct props


audius-infra:
Preview this change https://demo.audius.co/dj-fix-app-types

rickyrombo (approved):

PR #834, opened 2022-01-28
[AUD-985] Consolidate error flow

dylanjeffers:
### Description

Consolidates existing error flow and improves it by
 - Ensuring errors thrown in redux, and errors caught by React `ErrorBoundarys` are all ultimately handled by the `handleError` saga.
 - Properly implementing the "/error" route with react-router, instead of manual pathname checks that were happening in the App. Additionally, the error route behavior was divergent between saga errors and `ErrorBoundary` errors. For saga errors, we would push to "/error" route, and in app render, check pathname and conditionally render error page, whereas for ErrorBoundary errors, we would check if an error was logged to the state, and conditionally render error page (this would not change the pathname...).
 - Adds better ErrorBoundary component support (using `react-error-boundary` library), which can catch errors that occur in callbacks and other event handlers not inside redux. Before we essentially couldn't detect these errors!
 - Adds `react-router-last-location` to simplify getting our last location.

### Dragons

Things have worked as expected in my testing, but obviously with error handling we might get into weird edge-case states. I tried to keep functionality nearly the same for this pr. a follow up pr can make the "goBack" functionality use react-router instead of doing a full page refresh. This leads to much better performance as the page doesn't do a full refresh, but im worried it's too big of a change to also include here!

### Testing
On https://dylan.audius.co/ for testing. To trigger errors:
1. Go to /trending
2. Open console
3. Click the first track (should see "error from saga") this is a case where the "play action" puts a "handleError" action. This goes to the `handleError` saga
4. Click the second track (should see "error from ErrorBoundary"), this gets caught in the error boundary, which dispatches "handleError" action, which goes to the `handleError` saga

Confirm that in both cases, the pathname changes to /error, and we see the error page.


src/pages/App.js
```diff
@@ -430,24 +423,6 @@ class App extends Component {
...
 
-  static getDerivedStateFromError(_) {
-    return { didError: true }
-  }
-
-  componentDidCatch(error, errorInfo) {
-    try {
-      Sentry.withScope(scope => {
-        scope.setExtras(errorInfo)
-        Sentry.captureException(error)
```
rickyrombo, line 441: Worried that getting rid of this might break alerts, as the saga's `Sentry.captureException(new Error(action.message))` seems different than this? Maybe we do this code in the boundary to keep it consistent rather than dispatching the action for the saga.

src/pages/App.js
```diff
@@ -430,24 +423,6 @@ class App extends Component {
...
-      })
-    } catch (error) {
-      console.error(error.message)
-    }
-    // Dump the persistent cache on error, just in case that is causing some
-    // un-recoverable state.
-    clearAll()
```
rickyrombo, line 448: We also aren't clearing state anymore, is that important?

dylanjeffers: Both of these are great callouts. What I'm hoping to do is just be consistent with both flows, so do we think it makes sense to clear cache on a root saga error as well kinda thing?

src/pages/App.js
dylanjeffers: I see, is there a way to update the saga part to be consistent, or do you think that these call for different sentry implementations?

package.json
```diff
@@ -68,12 +68,14 @@
     "react-dates": "18.5.0",
     "react-dom": "17.0.2",
     "react-dropzone": "6.2.4",
+    "react-error-boundary": "^3.1.4",
```
sliptype, line 71: 📌

src/pages/App.js
sliptype: I think so, if something is erroring in a saga it could be cache related

sliptype: Wait, is the persistent cache even being used? @piazzatron

sliptype (approved):
Great find with `react-error-boundary`, we should use this in the mobile client too. Although it still looks like you have to manually handle errors in async functions and event handlers, but at least they can trigger the error boundary

raymondjacobson:
Mind spinning this up on a subdomain or creating some good space to test it/play with it?

package.json
dylanjeffers: totally, its better but not ideal, sounds like a plan

src/pages/App.js
rickyrombo: If they're different now, I think making it consistent will break the tracking/bucketing we have in sentry, but I'm not sure. CC: @raymondjacobson ? It might mean that besides making it consistent we'll also need to check and update sentry things?

dylanjeffers:
> Mind spinning this up on a subdomain or creating some good space to test it/play with it?

@raymondjacobson published to https://dylan.audius.co/

src/pages/App.js
dylanjeffers: @rickyrombo i think this might be the same actually. so the setExtras comes from errorInfo, which gets passed to `handleError` with `additionalInfo: errorInfo`, and the `captureException` gets passed correctly with `error.message`. am i missing something?

rickyrombo: They could very well be effectively the same, not super familiar with the Sentry API. I do see that in the sagas we call:
```ts
        Sentry.captureException(new Error(action.message))
```
which seems different than capturing the error object?

rickyrombo: Like with your change, instead of `Sentry.captureException(error)` it's essentially `Sentry.captureException(new Error(error.message))` which may or may not be different?

dylanjeffers: right, and from what i can tell that is the same, as long as the error class is the same, which it is

src/pages/App.js
piazzatron: persistent cache is not being used

PR #835, opened 2022-01-28
[AUD-985] Prevent full refresh on error

dylanjeffers:
### Description

Follow up pr that prevents a full page refresh when hitting the error page. The benefit is a faster TTI, but potentially we are in a faulty state? I don't think I know enough to determine!


raymondjacobson (approved):
Want sebastian's take definitely as well. My vote is we try this out & just soak on staging a bit extra

sliptype (approved):
I agree, there aren't too many errors that leave things in a faulty state. And in the worst case scenario most users will probably try a refresh anyway

PR #836, opened 2022-01-28
[AUD-1402] Move toast logic to redux

dylanjeffers:
### Description
- Makes it possible for redux actions to add toasts, which helps with providing relevant error messages
- Adjusts the handleError action to accept `shouldToast` option to show a toast on an error

### DIFF
--- a/src/common/store/errors/actions.ts
+++ b/src/common/store/errors/actions.ts
@@ -15,6 +15,7 @@ export type HandleErrorAction = {
   message: string
   shouldRedirect: boolean
   shouldReport: boolean
+  shouldToast?: boolean
   additionalInfo?: object
   level?: Level
 }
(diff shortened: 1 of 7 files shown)


sliptype (approved):
🥂 🍞

I guess ToastContext doesn't really need to be a context provider anymore, we could just dispatch toast actions where needed. But this way we don't have to update all the instances

dylanjeffers:
> 🥂 🍞
> 
> I guess ToastContext doesn't really need to be a context provider anymore, we could just dispatch toast actions where needed. But this way we don't have to update all the instances

Agreed! was considering doing that at some other point, but if its easy enough ill include it here

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1402-migrate-toasts-to-redux

Kyle-Shanks (approved):

PR #838, opened 2022-01-28
Fetch balance refresh on reward notification

jowlee:
### Description
When new notification are set, check for ChallengeReward type and trigger a balance refresh if found

Fixes AUD-1393

### How Has This Been Tested?
Ran locally against stage

### DIFF
--- a/src/components/notification/store/sagas.ts
+++ b/src/components/notification/store/sagas.ts
@@ -294,7 +294,7 @@ export function* parseAndProcessNotifications(
 /**
  * Run side effects for new notifications
  */
-export function* processNewNotifications(
+export function* handleNewNotifications(
   notifications: Notification[]
 ): Generator<any, void, any> {
   const hasRewardsNotification = notifications.some(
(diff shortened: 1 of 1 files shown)


src/components/notification/store/sagas.ts
```diff
@@ -515,6 +530,7 @@ export function* getNotifications(isFirstFetch: boolean) {
               hasMore
             )
           )
+          yield processNewNotifications(notificationItems)
```
sliptype, line 533: My only thing is maybe rename this to `handleNewNotifications` bc `process` makes it seem like we are manipulating the notif data in some way

sliptype (approved):
LGTM!

audius-infra:
Preview this change https://demo.audius.co/jowlee-fetch-balance-on-reward-notif

audius-infra:
Preview this change https://demo.audius.co/jowlee-fetch-balance-on-reward-notif

PR #839, opened 2022-01-28
Update useImageSize hook to optionally accept dispatch

Kyle-Shanks:
### Description
Update useImageSize hook to accept dispatch prop
Create new hooks for the web specific image hooks
Update all references to the web specific image hooks

### How Has This Been Tested?

Loaded up the client and nothing broke so it seems fine 🤷


audius-infra:
Preview this change https://demo.audius.co/kj-Update-useImageSize-hook-to-accept-dispatch

src/common/hooks/useImageSize.ts
```diff
@@ -145,7 +137,7 @@ export const useImageSize = <
     if (load && getPreviousId() !== id) {
       setPreviousId(id)
       // Request the desired size
-      dispatch(action(id, size))
+      dispatchProp ? dispatchProp(action(id, size)) : dispatch(action(id, size))
```
sliptype, line 140: Did this fix it on mobile? I feel like it might still break because we still have `useDispatch`

milindNshah (approved):
Looks good!

audius-infra:
Preview this change https://demo.audius.co/kj-Update-useImageSize-hook-to-accept-dispatch

src/common/hooks/useImageSize.ts
Kyle-Shanks: Did some testing, the error still appears to be popping up. Will do some more testing tomorrow, but this may not have worked 😞

sliptype: Removing the `useDispatch` will hopefully fix it, we will need to explicitly pass in dispatch on web

PR #840, opened 2022-01-28
Fix app test

dylanjeffers:
### Description

Not exactly sure why this was passing before, but it makes sense that it should have been breaking since we weren't passing in the required props for `App`

### DIFF
--- a/src/pages/App.test.tsx
+++ b/src/pages/App.test.tsx
@@ -1,9 +1,11 @@
-import React from 'react'
+import React, { createRef, MutableRefObject } from 'react'
 
 import { ConnectedRouter } from 'connected-react-router'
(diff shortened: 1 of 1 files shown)


raymondjacobson (approved):
much thank you. does this need to be .tsx vs. .ts since it just calls ReactDOM.render?

dylanjeffers:
> much thank you. does this need to be .tsx vs. .ts since it just calls ReactDOM.render?

I thought any time < /> syntax is used, .tsx is needed to properly transpile it?

rickyrombo (approved):

audius-infra:
Preview this change https://demo.audius.co/dj-fix-app-test

PR #853, opened 2022-02-02
Fix Heavy Rotation

rickyrombo:
### Description

`Explore.getTopUserListens` does _not_ return an array of `UserTrack`s. It hits identity and does not have user objects, but `userIds`. My guess is this failed to break on staging because the users would exist in the cache whenever I would visit the Heavy Rotation page?

This fix adds a fetch to get the users to check for deactivated accounts.

### Dragons

### How Has This Been Tested?

Tested against prod. Bug no longer repros for me.

### DIFF
--- a/src/pages/smart-collection/store/sagas.ts
+++ b/src/pages/smart-collection/store/sagas.ts
@@ -6,6 +6,7 @@ import Status from 'common/models/Status'
 import { Track, UserTrack, UserTrackMetadata } from 'common/models/Track'
 import { getAccountStatus, getUserId } from 'common/store/account/selectors'
 import { processAndCacheTracks } from 'common/store/cache/tracks/utils'
+import { fetchUsers as retrieveUsers } from 'common/store/cache/users/sagas'
 import { setSmartCollection } from 'pages/collection-page/store/actions'
 import Explore from 'services/audius-backend/Explore'
 import { waitForBackendSetup } from 'store/backend/sagas'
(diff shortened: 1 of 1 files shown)


src/pages/smart-collection/store/sagas.ts
```diff
@@ -29,8 +30,12 @@ const COLLECTIONS_LIMIT = 25
 function* fetchHeavyRotation() {
   const topListens = yield call(Explore.getTopUserListens)
 
+  const users = yield call(
+    retrieveUsers,
+    topListens.map((t: any) => t.userId)
+  )
   const trackIds = topListens
-    .filter((track: UserTrack) => !track.user.is_deactivated)
+    .filter((track: any) => !users[track.userId].is_deactivated)
```
jowlee, line 38: I think type should be `TopUserListen` from `/services/audius-backend/Explore.ts`

jowlee (approved):

audius-infra:
Preview this change https://demo.audius.co/mjp-retrieve-users

src/pages/smart-collection/store/sagas.ts
jowlee: Hmmm, still seeing the error, maybe check if track user id is in users maps
ie. `.filter((track: any) => !(users[track.userId]?.is_deactivated ?? true))`

audius-infra:
Preview this change https://demo.audius.co/mjp-retrieve-users

PR #854, opened 2022-02-02
Swap out old favorite/repost buttons in play bar

nicoback2:
### Description
Ticket: https://linear.app/audius/issue/AUD-1400/update-repostfavorite-buttons-on-play-bar

-Delete old favorite/repost buttons and replace existing references (on play bar and tracks table) w/ new favorite/repost buttons ('alt-buttons')
-Add "grow" hover state of those buttons to the components themselves 
buttons
buttons2

TODOs:
-Move new favorite/repost (+ other 'alt-buttons'?) buttons to Stems and clean up

### Dragons

### How Has This Been Tested?

Tested in browser


src/components/alt-button/FavoriteButton.module.css
```diff
@@ -1,46 +1,71 @@
 @keyframes scaleX {
-    0% {
-        transform: scale(1, 1);
```
nicoback2, line 3: Does this mean something is wrong with my Prettier/Eslint setup?

src/components/alt-button/RepostButton.module.css
```diff
@@ -32,6 +32,20 @@
 
 .button {
   all: unset;
+  transform: scale3d(1, 1, 1);
+  transition: all 0.12s ease-in-out;
+}
+
+.button:hover {
```
nicoback2, line 39: These hover states were being added by the parent before. Moved them here because I assumed we'd want the button to always/by default grow on hover. (same for favorite button)

src/components/alt-button/FavoriteButton.module.css
rickyrombo: I think you might need this in your VS Code settings?
```json
{
  "editor.tabSize": 2
}
```

rickyrombo: er, actually read the diff backwards. I know @dylanjeffers just added stylelint - what's the rule here with indents?

src/components/alt-button/FavoriteButton.tsx
```diff
@@ -1,3 +1,5 @@
+// TODO(nkang) - turn into button component like RepostButton, add accessibility
```
rickyrombo, line 1: TODOs tend to get lost. Should we make a linear issue?

src/components/alt-button/FavoriteButton.module.css
dylanjeffers: hey hey! so i have this in my .vscode/settings.json for the project:

rickyrombo (approved):
Looks good! Once you fix the conflicts the CI should push a demo up that I can play with, but from the images/video you posted it looks great!

src/components/alt-button/FavoriteButton.module.css
dylanjeffers: but actually i think this means everything is working for you, im surprised that file wasnt caught by stylelint already though...

dylanjeffers: oh actually... looks like its using tabs not spaces for stylelint, which i think we don't want to do... ill make a quick change!

rickyrombo: Also I think it is behind master by 8 commits, which means it's missing your stylelint change. :)

dylanjeffers: yeah maybe we rebase with master first and see where that takes us?

vicky-g:
omg gifs

src/components/play-bar/desktop/PlayBar.js
```diff
@@ -321,6 +320,9 @@ class PlayBar extends Component {
       playButtonStatus = 'play'
     }
 
+    const favoriteAndRepostIsDisabled = !uid || isOwner
```
dylanjeffers, line 323: i dont think we have a true standard as a team yet, but imo starting booleans with `is` makes it easier to identify, so `isFavoriteAndRepostDisabled`. what do you think?

src/components/alt-button/FavoriteButton.module.css
nicoback2: ohh gotcha thanks ill try that and see what happens

src/components/alt-button/FavoriteButton.tsx
nicoback2: Yeah I'll do that!

src/components/play-bar/desktop/PlayBar.js
rickyrombo: good catch, this is a standard! https://www.notion.so/audiusproject/Client-Javascript-Styleguide-4f34640139fc4a809c3765131bc597ff#d99d63f262f24990a83665bedd3b8610

raymondjacobson: 100% we should standardize around the verbs 

isReposted
hasUserFavorited
wenMoon

src/components/play-bar/desktop/PlayBar.js
```diff
@@ -321,6 +320,9 @@ class PlayBar extends Component {
...
+    const favoriteText = favorited ? 'Unfavorite' : 'Favorite'
```
dylanjeffers, line 324: In preparation for i18n, we are trying to put all user-facing text in a messages object, typically defined above the component, so something like
```typescript
const messages = {
  favorite: 'favorite',
  unfavorite: 'unfavorite'
}
```
and then used in the component like: 
```
const favoriteText = favorited ? messages.unfavorite : messages.favorite
```

src/components/play-bar/desktop/PlayBar.js
```diff
@@ -396,38 +398,49 @@ class PlayBar extends Component {
...
 
           <div className={styles.optionsRight}>
+            <VolumeBar
+              defaultValue={100}
+              granularity={VOLUME_GRANULARITY}
+              onChange={this.updateVolume}
+            />
+            <div className={styles.toggleRepostContainer}>
+              <Tooltip
+                text={repostText}
```
dylanjeffers, line 408: this might be a bit out of scope, but imo it is most web accessible to have text in lower case, and use `textTransform` css prop to do things like uppercase or capitalize. this might require a bit of an update to the Tooltip component, maybe have it by default have a `capitalize` text-transform, or allow a `className` prop to configure that. curious your thoughts!

src/components/play-bar/desktop/PlayBar.js
```diff
@@ -396,38 +398,49 @@ class PlayBar extends Component {
...
-                    favoritedClassName={cn(styles.favorited, {
-                      [styles.favoritedMatrix]: isMatrix
-                    })}
-                    disabled={!uid || isOwner}
-                    unfavoritedClassName={cn(styles.unfavorited, {
-                      [styles.unfavoritedMatrix]: isMatrix
-                    })}
-                    favorited={favorited}
+                    isDisabled={favoriteAndRepostIsDisabled}
+                    isMatrixMode={matrix}
```
dylanjeffers, line 436: so we are not consistent with this, but i also think it is best to derive as many props as possible in the child component, rather than have them passed down. in this case isMatrixMode, and isDarkMode is something that can grab itself with something like
```
import { isMatrix } from 'utils/theme/theme'
```

src/components/alt-button/FavoriteButton.tsx
```diff
@@ -84,7 +86,12 @@ const FavoriteButton = ({
   return (
     <div
       className={cn(
```
dylanjeffers, line 88: okay, also something that we should consider doing at some point, but just food for thought in case you were interested in tackling this, is that this should actually render a <button /> element. in case you weren't aware, rendering a div here is not the best cause a screen-reader or web-driver cant interact with divs that dont have a "tabIndex" (buttons have tabIndex by default). Also it makes the `disabled` prop much cleaner cause buttons have a `disabled` prop, that will prevent `onClick` from triggering.

src/components/play-bar/desktop/PlayBar.js
```diff
@@ -396,38 +398,49 @@ class PlayBar extends Component {
...
+              onChange={this.updateVolume}
+            />
+            <div className={styles.toggleRepostContainer}>
+              <Tooltip
+                text={repostText}
+                disabled={favoriteAndRepostIsDisabled}
+                mount='parent'
+                placement='top'
+              >
+                <span>
```
dylanjeffers, line 413: im guessing this <span> element is here because the tooltip needs something that can accept a ref to know how to anchor. if youre interested, we can update the repost/favorite buttons to accept a ref with `React.forwardRef`

src/components/play-bar/desktop/PlayBar.js
```diff
@@ -396,38 +398,49 @@ class PlayBar extends Component {
...
+                  <RepostButton
+                    aria-label={repostText}
```
dylanjeffers, line 415: love the aria-label!

src/components/alt-button/FavoriteButton.tsx
dylanjeffers: oh amazing, you're tracking this! ill remove my previous comment

dylanjeffers (approved):
looking great!

src/components/alt-button/FavoriteButton.module.css
nicoback2: looks like that fixed it ☺️

src/components/alt-button/FavoriteButton.tsx
nicoback2: yess sounds good. definitely want to make sure everything that goes into stems is properly accessible

nicoback2: https://linear.app/audius/issue/AUD-1432/turn-favoritebutton-into-actual-button-element-like-repostbutton-add

audius-infra:
Preview this change https://demo.audius.co/nkang--play-bar-btns

src/components/play-bar/desktop/PlayBar.js
nicoback2: gotcha, thanks! i think this pattern is fine until we figure out what we're going to use. sadly i do not have any suggestions there (the last i18n experience i had was at Airbnb where they had built an intense in house solution)

src/components/play-bar/desktop/PlayBar.js
nicoback2: Oh interesting, could you explain a bit more why it's more accessible to do it like that? (Hadn't heard of this before)

src/components/play-bar/desktop/PlayBar.js
nicoback2: Yeah this makes sense - I think this change makes sense to throw in the next PR (moving those buttons to Stems and cleaning them up) so I'll do that

nicoback2: https://linear.app/audius/issue/AUD-1434/move-alt-buttons-to-stems

src/components/play-bar/desktop/PlayBar.js
dylanjeffers: Yeah here is a stack overflow that explains the problem that can happen: https://stackoverflow.com/questions/53780653/voiceover-capitalized-small-words-are-read-as-abbreviations

Was looking at a few more sources, and sounds like modern screen readers can handle all caps okay now, so maybe it's not as important. Imo it still seems nice to have lowercase consistency and use transforms for stylistic choices like capitalize or uppercase, but what do you think?

audius-infra:
Preview this change https://demo.audius.co/nkang--play-bar-btns

src/components/play-bar/desktop/PlayBar.js
nicoback2: Ohh gotcha! 
My concern with what you're describing w/ regards to capitalization is that it could cause extra overhead / dev time without clear benefits, and that it might impede i18n efforts down the line since text-transform is applied universally [no matter the language or context](https://stackoverflow.com/questions/21050401/css-text-transform-capitalize-and-italian-spanish-portuguese-french-etc). To illustrate, if we were submitting this "Repost" text for translation, we'd probably provide context that the word is to be used on an action button to repost something - and the result might be capitalized or not depending on the language.

With regards to uppercase though I agree that applying `textTransform: uppercase` is the way to go there since it is purely stylistic in almost all circumstances + due to the screen reader issue. Should I add that to the client styleguide?

audius-infra:
Preview this change https://demo.audius.co/nkang--play-bar-btns

src/components/play-bar/desktop/PlayBar.js
dylanjeffers: yeah totally agree with your thoughts regarding capitalize vs uppercase! i think that would be a nice addition, maybe we can bring it up at the next engineering sync?

src/components/play-bar/desktop/PlayBar.js
dylanjeffers: ahh i see! so it's injecting props, nice find. so yeah keeping span is prob fine, but this probably also shows that we should be passing all "button" props through too, so maybe an action item in general in addition to forwardRef is allowing all button props to go through to button element. def a bit overkill in terms of eng effort though

src/components/play-bar/desktop/PlayBar.js
dylanjeffers: sounds like a plan!

PR #856, opened 2022-02-03
[AUD-1428] Upgrade antd

dylanjeffers:
### Description

Upgrades antd dependency to next major version, which includes a massive bundle-size improvement (they now properly tree-shake their icons)

Before:
@ant-design
+ 
(concaten

After: 
antd

### Dragons

There were slight core styling changes, so we should do a smoke test the components that use ant

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,39 +30,40 @@
       }
     },
     "@ant-design/colors": {
-      "version": "3.2.2",
-      "resolved": "https://registry.npmjs.org/@ant-design/colors/-/colors-3.2.2.tgz",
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-upgrade-ant

raymondjacobson:
We should not merge this PR. we've tried before and it broke a TON of UI. We should be removing usage of it (already on the roadmap).

dylanjeffers:
> We should not merge this PR. we've tried before and it broke a TON of UI. We should be removing usage of it (already on the roadmap).

Sounds good to me! fwiw from the little smoke testing i was doing things seemed more-or-less okay!

sliptype:
One of the main things we use antd for is tables. This branch is crashing for me on any page with a table like Favorites, Artist Dashboard, playlists

dylanjeffers:
Ahh I see, will give it a quick once over to see if it's an easy fix

PR #857, opened 2022-02-03
[AUD-1253] Move artist dashboard query to use discovery node

milindNshah:
### Description
Update artist dashboard to use just the retrieveUserTracks query onto discovery node. Then filter into listed and unlisted tracks. 

### How Has This Been Tested?
Opened /dashboard to confirm both types of tracks are still showing up (listed & unlisted)

### DIFF
--- a/src/pages/artist-dashboard-page/store/sagas.js
+++ b/src/pages/artist-dashboard-page/store/sagas.js
@@ -25,7 +25,6 @@ function* fetchDashboardAsync(action) {
       currentUserId: account.user_id,
       // TODO: This only supports up to 500, we need to redesign / paginate
       // the dashboard
-      // limit: account.track_count,
       getUnlisted: true
     }),
     call(AudiusBackend.getPlaylists, account.user_id, [])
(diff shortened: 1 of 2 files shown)


src/pages/artist-dashboard-page/store/sagas.js
```diff
@@ -27,12 +25,15 @@ function* fetchDashboardAsync(action) {
       currentUserId: account.user_id,
       // TODO: This only supports up to 500, we need to redesign / paginate
       // the dashboard
-      limit: account.track_count
+      // limit: account.track_count,
```
milindNshah, line 28: Had to remove limit because we don't know how many unlisted tracks a user can have. I can update it to be 500 if that makes sense too.

audius-infra:
Preview this change https://demo.audius.co/ms-artist-dashboard-query

src/pages/artist-dashboard-page/store/sagas.js
raymondjacobson: Default server side is 500? i think we can just leave no arg here then! 👍

raymondjacobson (approved):

raymondjacobson:
Tested on the demo site, looks fab!

PR #858, opened 2022-02-03
[AUD-1429] Clicking "Visit Track Page" from artist dashboard not redirecting bug fix

milindNshah:
### Description
Previously on the dashboard page, when looking at a specific track and clicking ... -> Visit Track Page, it would not redirect to the track page. This PR fixes that and will redirect to the track page now. 

### How Has This Been Tested?
By clicking visit track page on both hidden and public tracks on the artist dashboard page.


audius-infra:
Preview this change https://demo.audius.co/ms-fix-visit-track-page-dashboard

raymondjacobson (approved):
@rickyrombo  for final ack since permalinks are his brainchild

rickyrombo:
Oof! Good find! Thanks for fixing this!

rickyrombo (approved):

PR #863, opened 2022-02-04
[AUD-1442] Handle errors at root saga

raymondjacobson:
### Description

I don't know if there's a better way than `pushRoute` and recording to segment / amplitude separately. Would be awesome to be able to use the generic saga handling logic, but obviously this error transcends sagas.

This PR can be reviewed in two commits
1st -- type the stuffs
2nd -- the actual change

### Dragons

I typed some things

### How Has This Been Tested?

Locally vs. stage.
Worth playing around with demo a decent bit

Amplitude & sentry `ERROR_PAGE` event


audius-infra:
Preview this change https://demo.audius.co/rj-aud-1442

src/store/configureStore.ts
```diff
@@ -1,30 +1,43 @@
...
-import { getIsDeployedOnHost } from 'utils/clientUtil'
 import history from 'utils/history'
+import logger from 'utils/logger'
 import { ERROR_PAGE } from 'utils/route'
 
+import { AppState } from './types'
+
+declare global {
+  interface Window {
+    store: any
```
raymondjacobson, line 23: this can be `Store<RootState>` i think

src/store/configureStore.ts
```diff
@@ -33,11 +46,18 @@ const onSagaError = (error, errorInfo) => {
...
+    amplitudeTrack(ERROR_PAGE, errorInfo)
+  } catch {
+    // no-op
+  }
 }
 
 // Can't send up the entire Redux state b/c it's too fat
 // for Sentry to handle, and there is sensitive data
-const statePruner = state => {
+const statePruner = (state: AppState) => {
```
raymondjacobson, line 60: this should be RootState i think

src/store/configureStore.ts
dylanjeffers: yeah just tested and that works

src/store/routing/sagas.ts
```diff
@@ -93,7 +93,10 @@ function* handleHardwareBack() {
 }
 
 const sagas = () => {
-  const sagas: Array<() => void> = [trackLocation, handleHardwareBack]
+  const sagas: Array<() => Generator<any, void, any>> = [
```
dylanjeffers, line 96: I think it works to not include this explicit type

src/store/configureStore.ts
raymondjacobson: this doesn't work b/c it introduces a circular dep, and redux says best practices is to use typeOf in declaring root state:
https://redux.js.org/usage/usage-with-typescript#define-root-state-and-dispatch-types

src/store/configureStore.ts
```diff
@@ -1,30 +1,43 @@
...
+  }
+) => {
   console.error(`Caught saga error: ${error} ${errorInfo}`)
-  store.dispatch(replaceRoute(ERROR_PAGE))
+  store.dispatch(pushRoute(ERROR_PAGE))
 
-  if (!getIsDeployedOnHost()) return
+  // Try recording to sentry
   try {
     Sentry.withScope(scope => {
```
dylanjeffers, line 42: maybe we can try to reuse these blocks with the other error saga so the impl is consistent?

raymondjacobson: good idea

audius-infra:
Preview this change https://demo.audius.co/rj-aud-1442

dylanjeffers (approved):
looks great, nice to have the store typed!

raymondjacobson:
@dylanjeffers  would you mind taking a quick pass at the latest commit?

diff ended up being a bit larger than I'd hoped, so just want to make sure you agree w/ the types!
https://github.com/AudiusProject/audius-client/pull/863/commits/6dea8048242186f01bd7f5fe9ce0d9c32df7a4bc

raymondjacobson:
> yeah this looks great! a bit weird the root sagaError is in configureStore, but nbd!

Yeah I thought about this -- probably should move some of that stuff to /analytics/providers or /services, but this diff is already preeeety big

audius-infra:
Preview this change https://demo.audius.co/rj-aud-1442

audius-infra:
Preview this change https://demo.audius.co/rj-aud-1442

PR #864, opened 2022-02-04
Validate sol token address

jowlee:
### Description
Validate the recipient address when sending SOL 

### How Has This Been Tested?
Ran locally against prod trying to send to a user bank account that was created and not created


audius-infra:
Preview this change https://demo.audius.co/jowlee-validate-send-sol-token-acct

piazzatron (approved):

PR #868, opened 2022-02-04
[AUD-1421] Track download buttons in react native

sliptype:
### Description

This PR:
  * Moves `DownloadButtons/hooks.ts` to `common/hooks/useDownloadTrackButtons.ts`
  * Updates the hook to accept `useSelector` to avoid redux context issue (we probably need to sync dependency version so we don't have to continually do this)
  * Shows a toast on disabled download buttons on mobile (instead of a tooltip)

image

### Dragons

Tested on mobile web and native mobile

### How Has This Been Tested?


src/common/hooks/useDownloadTrackButtons.ts
```diff
@@ -1,44 +1,80 @@
...
   track: Track
 ) => !isOwner && !following && track.download?.requires_follow
 
-const useCurrentStems = (trackId: ID) => {
+const useCurrentStems = ({
+  trackId,
+  useSelector
+}: {
+  trackId: ID
+  useSelector: typeof reduxUseSelector
```
dylanjeffers, line 69: not sure if this makes sense, but could we do a runtime/compile-time check for which env we are, and use `useSelector` or `useSelectorWeb`? would probably mean moving `useSelectorWeb` over from `audius-mobile-client` or at least creating a duplicate?

src/store/application/ui/stemsUpload/sagas.ts
```diff
@@ -10,7 +10,10 @@ import { handleUploads } from 'pages/upload-page/store/sagas'
 import { createStemMetadata } from 'pages/upload-page/store/utils/stems'
 import { make } from 'store/analytics/actions'
 
-import { startStemUploads, stemUploadsSucceeded } from './slice'
+import {
+  startStemUploads,
+  stemUploadsSucceeded
+} from '../../../../common/store/stems-upload/slice'
```
dylanjeffers, line 16: absolute import

audius-infra:
Preview this change https://demo.audius.co/sk-aud-142-download-buttons

src/common/hooks/useDownloadTrackButtons.ts
sliptype: Passing `useSelector` in here isn't really for the purpose of using useSelectorWeb in mobile. I had to do this because of the weird issue where calling an external hook that calls a redux hook causes a `"could not find react-redux context value"` error. Passing in `useSelector` and/or `dispatch` alleviates this. I think the better solution would be to get all of our dependencies in sync like outlined here: https://github.com/reduxjs/react-redux/issues/1489


Apart from that, your question is still valid though. I think there could be cases where we want to use the same hook logic from mobile but sometimes with useSelector and sometimes with useSelector. So probably best not to bake that logic into the hook itself. Also that logic would go away when we finally sever the dependency on the web client

src/store/application/ui/stemsUpload/sagas.ts
sliptype: Thanks vscode.

Satya at it again

src/common/hooks/useDownloadTrackButtons.ts
dylanjeffers: ohhh i see i see, yeah would be down to look into the version issues at some point, just lmk. and yeah prob just hold off then until we remove dependency on web client

src/components/download-buttons/DownloadButtons.tsx
```diff
@@ -86,16 +88,26 @@ const DownloadButton = ({
     )
   }
 
-  const renderButton = () => {
-    const isDisabled =
-      state === ButtonState.PROCESSING || state === ButtonState.REQUIRES_FOLLOW
+  const handleOnClick = useCallback(() => {
+    if (isMobile && isDisabled) {
```
dylanjeffers, line 92: prob should do this after the isDisabled check?

src/components/download-buttons/DownloadButtons.tsx
```diff
@@ -146,14 +176,8 @@ const DownloadButtons = ({
...
-      {buttons.map(({ label, state, type, onClick }) => (
-        <DownloadButton
-          label={label}
-          state={state}
-          type={type}
-          key={label}
-          onClick={onClick}
-        />
+      {buttons.map(props => (
+        <DownloadButton {...props} key={props.label} />
```
dylanjeffers, line 180: nice

dylanjeffers (approved):
lgtm

audius-infra:
Preview this change https://demo.audius.co/sk-aud-142-download-buttons

audius-infra:
Preview this change https://demo.audius.co/sk-aud-142-download-buttons

milindNshah (commented):
looks good!

PR #871, opened 2022-02-05
Move profile progress selectors to common

rickyrombo:
### Description

- Move profile progress selectors to common
- Fix broken profile meter, since the state for profile moved into pages
- Convert the selectors to TypeScript

### Dragons

- When converting to TypeScript, opted to change to using `_cover_photo_sizes` and `_profile_picture_sizes` instead of `updatedCoverPhoto` and `updatedProfilePicture` as these types don't exist on `User` and they are functionally equivalent.

### How Has This Been Tested?


dylanjeffers (approved):
Good good

PR #872, opened 2022-02-05
Convert optimistic challenges hooks into selector

rickyrombo:
### Description

Converts the audio rewards page hooks into a selector. Avoids problems with importing them on mobile, and also seems to conceptually make sense.

Also moves the types to the proper models file.

### Dragons

- Removes the `useOptimisticChallenge` hook, as it seems it wasn't being used.

### How Has This Been Tested?


audius-infra:
Preview this change https://demo.audius.co/mjp-hooks-to-selectors

src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
```diff
@@ -59,11 +62,11 @@ const RewardPanel = ({
   stepCount
 }: RewardPanelProps) => {
   const wm = useWithMobileStyle(styles.mobile)
-  const userChallenges = useSelector(getUserChallenges)
+  const userChallenges = useSelector(getOptimisticUserChallenges)
```
piazzatron, line 65: Could also have a second version of this selector that takes in an id, and calls this one and indexes into it

piazzatron (approved):
Looks gr8

src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
rickyrombo: yeah, we had something similar before in the hooks, but it wasn't getting used so I figured i'd not bring it over

PR #873, opened 2022-02-07
[AUD-1433] Allow prod and staging electron to be run simultaneously

raymondjacobson:
### Description

Allow prod and staging desktop apps to be run at the same time. Two changes changes allow this:
1. Unique "scheme" per app (audius:// and audius-staging://). The same scheme causes the apps to share the same local storage.
2. Unique "productName" per app (overridden in package.json). This allows both apps to be open simultaneously without blocking each other from trying to open the other.

### Dragons

First major structural electron change in a while

### How Has This Been Tested?

Deploying staging app at a version ahead, installing both, etc.

Screen Shot 2022-02-05 at 12 25 07 PM


src/electron.js
```diff
@@ -29,30 +25,53 @@ const Environment = Object.freeze({
...
+  const configFile = path.resolve(app.getAppPath(), 'electronConfig.json')
+  try {
+    // Needs to be readFileSync because async js here will take
+    // too long, leading the scheme to be decided after the app has
+    // registered the scheme.
+    const data = fs.readFileSync(configFile)
+    const config = JSON.parse(data)
+    env = config.env
+  } catch (e) {
+    console.error('No config file found', e)
```
nicoback2, line 46: Nit: Maybe specify that the error could have occurred because something was wrong with the config file content

nicoback2 (approved):
LGTM!

src/electron.js
raymondjacobson: 👌 yes!

sliptype (approved):
🍻

dylanjeffers (approved):
nice change, going to use this a lot, thanks

audius-infra:
Preview this change https://demo.audius.co/rj-aud-1433

PR #877, opened 2022-02-08
Adds retry claim with exponential backoff

jowlee:
### Description
Adds max retry count to claiming logic (instead of a single retry) with exponential backoff

### Dragons
check that it can't infinite loop or break the claim flow

### How Has This Been Tested?
Ran locally against stage and manually forced `AudiusBackend.submitAndEvaluateAttestations` to return unknown error
Also wrote a couple tests

### DIFF
--- a/src/pages/audio-rewards-page/store/store.test.ts
+++ b/src/pages/audio-rewards-page/store/store.test.ts
@@ -92,6 +92,8 @@ const retryClaimProvisions: StaticProvider[] = [
   [select(getClaimToRetry), testClaim]
 ]
 
+const MAX_CLAIM_RETRIES = 5
+
 const expectedRequestArgs = {
   ...testClaim,
   encodedUserId: undefined,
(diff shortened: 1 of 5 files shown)


audius-infra:
Preview this change https://demo.audius.co/jowlee-retry-exponential

src/pages/audio-rewards-page/store/store.test.ts
```diff
@@ -281,6 +284,68 @@ describe('Rewards Page Sagas', () => {
...
+        expectSaga(saga)
+          .dispatch(
+            claimChallengeReward({
+              claim: testClaim,
+              retryOnFailure: true
+            })
+          )
+          .provide([
+            ...claimAsyncProvisions,
+            { call: provideDelay },
```
rickyrombo, line 300: just fyi I _think_ you should be able to do a matcher here with `[call.fn(delayP), null]` if you want to avoid the extra helper method

rickyrombo (approved):

audius-infra:
Preview this change https://demo.audius.co/jowlee-retry-exponential

PR #879, opened 2022-02-09
Fix tracking create user bank

jowlee:
### Description
Track event in the wrong spot

### How Has This Been Tested?
ran against staging


audius-infra:
Preview this change https://demo.audius.co/jowlee-fix-tracking

piazzatron (approved):

PR #887, opened 2022-02-10
Fix sentry state pruner

jowlee:
### Description
* Fixes the type `AppState` to remove application > pages > profile bc it was moved to the `CommonState` pages > profile 
* Updates sentry to allow deeper nested objects as part of the additional data
* Updates the sentry statepruner to point to the updated profile page state

### How Has This Been Tested?
manually triggered error locally against stage and saw it in sentry

### DIFF
--- a/src/services/sentry.ts
+++ b/src/services/sentry.ts
@@ -29,6 +29,7 @@ export const initializeSentry = () => {
       })
     ],
 
+    normalizeDepth: 5,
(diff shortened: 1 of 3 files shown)


raymondjacobson (approved):

audius-infra:
Preview this change https://demo.audius.co/jowlee-sentry-err

PR #894, opened 2022-02-12
[AUD-1504] Fix optimistic disbursement of challenge rewards

rickyrombo:
### Description

Optimistic disbursement was broken causing the green pill to linger until the next poll for fetching user challenges.

This PR:
- Updates the default polling period (on claiming rewards) for `is_complete` to be much higher, and makes both the polling period and frequency configurable via optimizely
- Fixes the `claimableAmount` calculation to be 0 if a challenge is not complete nor an aggregate challenge, since we aren't adding the specifiers the `disbursed` list we can't rely on the `undisbursed` challenges list sans `disbursed` list to give us the proper claimable amount.
  - An alternative here would be to track the `disbursed` list instead, and make the `undisbursed` endpoint more entrenched in user code. I avoided this deliberately to make adjusting this code with a DN update to include nested specifiers for aggregate challenges on the userChallenge endpoint possible in the future with less headache but that might not be the correct call.
- Ensures that non-aggregate challenges always get optimistically set as disbursed when disbursed, even if they're not marked `is_complete` in the store (since "profile-completion" might be optimistically complete, and subsequently optimistically disbursed).

### Dragons

### How Has This Been Tested?

Locally against stage using a proxy/commenting out the actual attestations and observing the pill removal is immediate for the "profile-completion" challenge.

Added logging around the polling for `is_complete` so that if a failure occurs for attestation we hopefully will have that in the breadcrumbs.


audius-infra:
Preview this change https://demo.audius.co/mjp-wait-for-completion

piazzatron (approved):

PR #896, opened 2022-02-12
Add reward analytics

piazzatron:
### Description

**Needs latest libs**

- Adds sentry, amplitude, identity reporting

### Dragons

### How Has This Been Tested?

### DIFF
--- a/src/services/audius-backend/Rewards.ts
+++ b/src/services/audius-backend/Rewards.ts
@@ -1,0 +1,96 @@
+import { isMobile, isElectron } from 'utils/clientUtil'
+
+// TODO: make the sending attestation stuff fire-and-forget
+export class ClientRewardsReporter {
+  source: 'mobile' | 'electron' | 'web'
+  libs: any
+
+  constructor(libs: any) {
+    this.source = isMobile() ? 'mobile' : isElectron() ? 'electron' : 'web'
+    this.libs = libs
+  }
+
+  async reportSuccess({
+    userId,
+    challengeId,
+    amount,
+    specifier
+  }: {
+    userId: string
+    challengeId: string
+    amount: number
+    specifier: string
+  }) {
+    try {
+      await this.libs.Rewards.sendAttestationResult({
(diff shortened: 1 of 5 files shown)


audius-infra:
Preview this change https://demo.audius.co/piazz-working-on-analytics-wip

src/services/audius-backend/Rewards.ts
```diff
@@ -0,0 +1,149 @@
...
+    specifier
+  }: {
+    userId: string
+    challengeId: string
+    amount: number
+    specifier: string
+  }) {
+    ;(async () => {
+      try {
+        await track(Name.REWARDS_CLAIM_SUCCESS, {
```
raymondjacobson, line 34: not sure the benefit of awaiting here at all. since it's fire & forget, you probably don't want to fail to call `sendAttestationResult` if the `track` call itself fails.

since maybe you want the log, maybe this should be two async methods? no need to change, just musing

raymondjacobson (approved):

src/services/audius-backend/Rewards.ts
piazzatron: Yeah I think you're right - no need for async will remove

audius-infra:
Preview this change https://demo.audius.co/piazz-working-on-analytics-wip

src/services/audius-backend/Rewards.ts
piazzatron: JK we need it, otherwise the try catch doesn't catch the exceptions

raymondjacobson: My point here was more that
```
try {
  await track
  await send
}
```

will lead to `send` never getting called if `track` throws
so probably better to do

```
try {
  await track
} catch () {}
try {
  await send
} catch () {}

```

or wrapping them in each in the async call to not do serially

```
async () => {
  try {
    await track
  } catch () { }
}()

async () => {
  try {
     await send
  } catch () { }
}()
```

PR #900, opened 2022-02-14
[AUD-1517] Fix ProfileScreen scroll

dylanjeffers:
### Description

Fixes scroll issue with profile page (where inner flatlists dont scroll the profile header), using some interesting node_module modification trickery. 

Essentially, we need to nest a flatlist in a scrollview, so the entire page scrolls when the flatlist scrolls. Unfortunately that causes a warning that scrollviews should never wrap flatlists. The workaround is to nest a flatlist in a flatlist, which works great. 

Unfortunately though, `react-navigation's` `MaterialTopTabNavigator` fails to render it's content when rendered inside a `ScrollView` or `FlatList`. The underlying issue has to do with a `react-navigation` peerDep, `react-native-tab-view`. The proposed workaround, discussed [here](https://github.com/satya164/react-native-tab-view/pull/1198) and implemented [here](https://github.com/knockaway/react-native-tab-view/tree/forked-3.0.2) still haven't been merged, over six months since developed. 

While we wait, we will use the `patch-package` package to hotwire our version of `react-native-tab-view` with the fix

before:
image

after: 
image


audius-infra:
Preview this change https://demo.audius.co/dj-aud-1517-fix-profile-page-scroll

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1517-fix-profile-page-scroll

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1517-fix-profile-page-scroll

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1517-fix-profile-page-scroll

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1517-fix-profile-page-scroll

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1517-fix-profile-page-scroll

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1517-fix-profile-page-scroll

packages/mobile/patches/react-native-tab-view+3.1.1.patch
```diff
@@ -0,0 +1,45 @@
+diff --git a/node_modules/react-native-tab-view/src/SceneView.tsx b/node_modules/react-native-tab-view/src/SceneView.tsx
```
raymondjacobson, line 1: this is so powerful. wow. very cool.

Is it possible to somehow reference in the patch file the fix PR (or maybe we keep a patches/README.md). Just so in the future if that PR does get merged / released, we know which patch we can get rid of. wdyt?

raymondjacobson (approved):

raymondjacobson:
Not sure if we're tracking, but the font & spacing of the tabs looks off to me

dylanjeffers:
> Not sure if we're tracking, but the font & spacing of the tabs looks off to me

not tracking, but great call, will add it to a linear for overall pixel-perfect run-through

packages/mobile/patches/react-native-tab-view+3.1.1.patch
dylanjeffers: yeah this is a great call! ill see if its part of their options, otherwise will add a readme for it

PR #902, opened 2022-02-14
[AUD-1342] Fix whitespace clipping at bottom of CollectibleDetailsDrawer

dylanjeffers:
### Description

Fixes whitespace clipping at the bottom of the CollectibleDetailsDrawer

Before: 
image

After: 
OWNED +
Pasted Graphic 1


raymondjacobson (approved):
Maybe a separate issue, but i feel that even on these full-screen drawers, we don't want any "safe area" padding at the bottom (check out our other drawers, or even the drawers on google calendar). They usually just show text below the bottom-swipe bar on iOS

PR #904, opened 2022-02-15
Add RewardsBanner, `as` prop to Tile

dylanjeffers:
### Description

- Adds the RewardsBanner
- Adds `as` prop to Tile, to enable rendering a LinearGradient Tile. Uses some cool TS type-args to get tab-completable extra props to Tile based on the `as` prop component props.
image


sliptype (approved):
This is really cool! I have seen some component frameworks follow this same pattern. I can't really think of any other situations we would use `as` on `Tile` but maybe we will come across something

PR #907, opened 2022-02-15
[AUD-1527] Migrate trending store to common

dylanjeffers:
### Description

Moves trending store to common to use in mobile


audius-infra:
Preview this change https://demo.audius.co/dj-aud-1527-migrate-trending-to-common

sliptype (approved):

PR #918, opened 2022-02-16
Bump client version to 1.1

Kyle-Shanks:
### Description

Minor version bump

### DIFF
--- a/packages/web/package-lock.json
+++ b/packages/web/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "audius-client",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


sliptype:
We haven't really discussed this, but we should keep the package versions in sync using `npx lerna version patch`: https://github.com/lerna/lerna#fixedlocked-mode-default

audius-infra:
Preview this change https://demo.audius.co/kj-bump-minor-to-1.1

dylanjeffers:
just wondering, what would be the case of minor vs patch updates?

raymondjacobson:
@sliptype  is this PR good to merge as is or will lerna do something different?

sliptype:
@raymondjacobson Lerna would update the version in `lerna.json` and `mobile`, mainly important if we ever want to publish these packages. But it also makes sense to keep all the versions in sync if we want to name the release branch after that version IMO

PR #922, opened 2022-02-16
[AUD-1473] Add TrendingFilterDrawer

dylanjeffers:
### Description

- Adds TrendingFilterDrawer and hooks it up with the TrendingFilterButton
- Adds SearchInput component, which we can re-use for search page, genre filter, and other searches/filtering

https://user-images.githubusercontent.com/8230000/154361791-c758cfa4-aa19-4c06-8272-3d443616522f.mov


audius-infra:
Preview this change https://demo.audius.co/dj-aud-1473-trending-filter-drawer

packages/mobile/src/screens/trending-screen/TrendingFilterButton.tsx
```diff
@@ -14,14 +24,22 @@ const useStyles = makeStyles(({ spacing, palette }) => ({
...
   const styles = useStyles()
+  const dispatchWeb = useDispatchWeb()
+  const trendingGenre = useSelectorWeb(getTrendingGenre) ?? Genre.ALL
+
+  const handlePress = useCallback(() => {
+    dispatchWeb(setVisibility({ modal: MODAL_NAME, visible: true }))
+  }, [dispatchWeb])
+
   return (
     <Button
```
sliptype, line 38: Should we just use `HeaderButton` here?

packages/web/src/common/utils/genres.ts
```diff
@@ -83,6 +84,7 @@ export const getCanonicalName = (genre: Genre | any) => {
 }
 
 export const GENRES = [
+  Genre.ALL,
```
sliptype, line 87: Does this make sense here or should we extend `GENRES` specifically for the filter drawer

sliptype (approved):
Looks awesome!

packages/mobile/src/screens/trending-screen/TrendingFilterButton.tsx
dylanjeffers: yes will update

packages/web/src/common/utils/genres.ts
dylanjeffers: good question, for now ill just extend it in filter drawer so we dont get issues with web. i was thinking long term we add it and update all our filters, and instead of "null" meaning All, we make this the initialValue for the state too

packages/mobile/src/screens/trending-screen/TrendingFilterButton.tsx
dylanjeffers: btw including this in another pr that will do work merging screen-header, header, and header-button

PR #923, opened 2022-02-16
Consolidate audio-rewards

dylanjeffers:
### Description

- Moves common audio-rewards code into `components/audio-rewards`
- Simplifies usage of AudioBadges with IconAudioBadge
- Renames some helpers to be more clear
- Prep work for TierExplainerDrawer

### DIFF
--- a/packages/mobile/src/components/audio-rewards/IconAudioBadge.tsx
+++ b/packages/mobile/src/components/audio-rewards/IconAudioBadge.tsx
@@ -1,0 +1,29 @@
+import { ComponentType } from 'react'
+
+import { SvgProps } from 'react-native-svg'
+
+import IconBronzeBadge from 'app/assets/images/IconBronzeBadge.svg'
+import IconGoldBadge from 'app/assets/images/IconGoldBadge.svg'
+import IconPlatinumBadge from 'app/assets/images/IconPlatinumBadge.svg'
+import IconSilverBadge from 'app/assets/images/IconSilverBadge.svg'
+
+import { AudioTier } from './audioTier'
+
+const audioTierMap: Record<AudioTier, ComponentType<SvgProps> | null> = {
+  none: null,
+  bronze: IconBronzeBadge,
+  silver: IconSilverBadge,
+  gold: IconGoldBadge,
+  platinum: IconPlatinumBadge
+}
+
+type IconBadgeProps = SvgProps & {
+  tier: AudioTier
+}
+
+export const IconAudioBadge = (props: IconBadgeProps) => {
+  const { tier, ...other } = props
+  const AudioBadge = audioTierMap[tier]
+
+  return AudioBadge ? <AudioBadge {...other} /> : null
+}
--- a/packages/mobile/src/components/audio-rewards/audioTier.ts
+++ b/packages/mobile/src/components/audio-rewards/audioTier.ts
@@ -5,7 +5,7 @@ export const WEI = new BN('1000000000000000000')
 
 export type AudioTier = 'none' | 'bronze' | 'silver' | 'gold' | 'platinum'
 
-export const audioTierRequirements: { tier: AudioTier; minAudio: BN }[] = [
+const audioTierRequirements: { tier: AudioTier; minAudio: BN }[] = [
   {
     tier: 'platinum',
     minAudio: new BN('100000')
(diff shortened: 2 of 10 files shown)


packages/mobile/src/components/audio-rewards/IconAudioBadge.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { SvgProps } from 'react-native-svg'
+
+import IconBronzeBadge from 'app/assets/images/IconBronzeBadge.svg'
+import IconGoldBadge from 'app/assets/images/IconGoldBadge.svg'
+import IconPlatinumBadge from 'app/assets/images/IconPlatinumBadge.svg'
+import IconSilverBadge from 'app/assets/images/IconSilverBadge.svg'
+
+import { AudioTier } from './audioTier'
+
+const audioTierMap: Record<AudioTier, ComponentType<SvgProps> | null> = {
```
nicoback2, line 12: nit - should we UPPER_SNAKE_CASE constants like these? `AUDIO_TIER_MAP`

packages/mobile/src/components/audio-rewards/audioTier.ts
```diff
@@ -3,8 +3,9 @@ import BN from 'bn.js'
 
 export const WEI = new BN('1000000000000000000')
 
-export type BadgeTier = 'none' | 'bronze' | 'silver' | 'gold' | 'platinum'
-export const badgeTiers: { tier: BadgeTier; minAudio: BN }[] = [
+export type AudioTier = 'none' | 'bronze' | 'silver' | 'gold' | 'platinum'
+
+export const audioTierRequirements: { tier: AudioTier; minAudio: BN }[] = [
```
nicoback2, line 8: same comment here re: upper snake casing constants, not sure if that's a thing in this codebase though

packages/mobile/src/components/audio-rewards/audioTier.ts
```diff
@@ -27,31 +28,29 @@ export const badgeTiers: { tier: BadgeTier; minAudio: BN }[] = [
...
 ]
 
-export const getTierLevel = (tier: BadgeTier) => {
-  return tierLevels.indexOf(tier)
+export const getAudioTierRank = (tier: AudioTier) => {
+  return audioTierOrder.indexOf(tier)
 }
 
-export const getBadgeTier = (
+export const getUserAudioTier = (
```
nicoback2, line 43: might be worth adding a test for this util - up to you though depending on whether you think it's prone to regressions

nicoback2 (approved):
lgtm 🕺

packages/mobile/src/components/audio-rewards/IconAudioBadge.tsx
dylanjeffers: yeah def agree!

packages/mobile/src/components/audio-rewards/audioTier.ts
dylanjeffers: yeah ill have a look around as well

packages/mobile/src/components/audio-rewards/audioTier.ts
dylanjeffers: oh tests... would love to add them in mobile at some point, we don't have any im pretty sure 🗡️

packages/mobile/src/components/audio-rewards/audioTier.ts
nicoback2: I also like to use `as const` with "constant" objects and arrays like this and `audioTierMap` above to prevent them from being accidentally modified by consumers. Not sure if that will mess up your types here though

dylanjeffers: oh im looking at this: https://github.com/airbnb/javascript/#naming--uppercase so maybe this one we upper-case, but not the first one since it's not exported?

packages/mobile/src/components/audio-rewards/audioTier.ts
nicoback2: ohhh haha

packages/mobile/src/components/audio-rewards/audioTier.ts
nicoback2: ah! i like that, makes sense to me

nicoback2: shall we add to Audius style guide? (i still have yet to add the stuff we talked about in product sync rip)

dylanjeffers: yeah just added it!

PR #926, opened 2022-02-17
Combine Header and ScreenHeader, refactor HeaderButton

dylanjeffers:
### Description

- Combines Header and ScreenHeader, using ScreenHeader implementation
- Refactors HeaderButton, and uses it everywhere


sliptype (approved):
Looking clean 🔥

PR #927, opened 2022-02-17
Follows you indicator

stereosteve:
### Description

Follows You badge on the web profile + artist popover.

Status quo:

image

image


stereosteve:
API PR: https://github.com/AudiusProject/audius-protocol/pull/2514

packages/web/src/components/artist/ArtistCard.module.css
```diff
@@ -92,6 +92,7 @@
 }
 
 .artistHandle {
+  display: inline-block;
```
sliptype, line 95: What's the behavior when the handle is really long? Does the badge wrap to the next line?

packages/web/src/components/user-badges/FollowsYouBadge.tsx
```diff
@@ -0,0 +1,6 @@
+import React from 'react'
+import styles from './FollowsYouBadge.module.css'
+
+export default function FollowsYouBadge() {
+  return <div className={styles.badge}>FOLLOWS YOU</div>
+}
```
sliptype, line 6: Make sure you have Prettier installed, and if you're using vscode you can set format on save to true!

sliptype (approved):
Awesome work!

I was wondering if we use this same badge pattern anywhere else, and if it makes sense to make a reusable component. But I didn't see anything right away, so this is probably fine for now!

packages/web/src/components/user-badges/FollowsYouBadge.tsx
```diff
@@ -0,0 +1,6 @@
+import React from 'react'
+import styles from './FollowsYouBadge.module.css'
+
+export default function FollowsYouBadge() {
+  return <div className={styles.badge}>FOLLOWS YOU</div>
```
dylanjeffers, line 5: we are following a convention to put all text in a `messages` object at the top of each file so it'll be easier for i18n at some point. so something like `const messages = {followsYouText: 'follows you'}`

Also we are trying to avoid UPPERCASE text since some screenreaders read out each letter :D. The way we avoid this is adding `text-transform: uppercase;` to its styles.

And one last minor thing, does it make sense to have this as a span so that it can stay inline with the h2 text?

dylanjeffers (approved):
awesome job! one other callout, are we checking how it looks in matrix mode (that's a theme)?

packages/web/src/components/artist/ArtistCard.module.css
stereosteve: Went with option 1... margin-right on the handle so when it wraps it's all left aligned.

audius-infra:
Preview this change https://demo.audius.co/follows-you-indicator

packages/web/src/pages/profile-page/components/desktop/ProfileWrapping.js
```diff
@@ -349,7 +350,10 @@ const ProfileWrapping = props => {
                 onChange={props.onUpdateName}
                 userId={props.userId}
               />
-              <h2 className={styles.handle}>{props.handle}</h2>
+              <div className={styles.handleWrapper}>
+                <h2 className={styles.handle}>{props.handle}</h2>
+                {props.doesFollowCurrentUser && <FollowsYouBadge />}
```
dylanjeffers, line 355: also, we have a new preference for using terneries instead of && for conditional rendering: https://www.notion.so/audiusproject/Client-Javascript-Styleguide-4f34640139fc4a809c3765131bc597ff#f1c89a4e0843430bbf98f0ce53dbcf14

audius-infra:
Preview this change https://demo.audius.co/follows-you-indicator

PR #932, opened 2022-02-17
Check for empty endpoint config

rickyrombo:
### Description

Allows us to remove the config from optimizely by setting it to be empty and allowing libs to select for us.

### Dragons

### How Has This Been Tested?

Tested against stage with stage having empty endpoints config.

We have an amplitude dashboard [here](https://analytics.amplitude.com/audius/dashboard/0ebjc1d), and one of those charts shows [error counts by type](https://analytics.amplitude.com/audius/chart/76f22tw). If we see a spike of INSUFFICIENT_DISCOVERY_NODES when setting the config to an empty string, we know that the change is not working as intended and we need to instead remove the variable in optimizely entirely.

If we don't actively watch the chart, there's an alert set up to send a slack message (and an email to me) if the claim failure rate goes above 80% (we're working on fixing some outstanding unknown errors still and will tweak that threshold)

### DIFF
--- a/packages/web/src/pages/audio-rewards-page/store/sagas.ts
+++ b/packages/web/src/pages/audio-rewards-page/store/sagas.ts
@@ -202,7 +202,7 @@ function* claimChallengeRewardAsync(
   // When endpoints is unset, `submitAndEvaluateAttestations` picks for us
   const endpoints =
     rewardsAttestationEndpoints && rewardsAttestationEndpoints !== ''
-      ? rewardsAttestationEndpoints?.split(',')
+      ? rewardsAttestationEndpoints.split(',')
       : null
   const hasConfig =
     oracleEthAddress &&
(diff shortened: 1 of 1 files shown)


audius-infra:
Preview this change https://demo.audius.co/mjp-optimizely-dn-selection

raymondjacobson (approved):

raymondjacobson:
How does one monitor this change

rickyrombo:
> How does one monitor this change

I'm so glad you asked! We have an amplitude dashboard [here](https://analytics.amplitude.com/audius/dashboard/0ebjc1d), and one of those charts shows [error counts by type](https://analytics.amplitude.com/audius/chart/76f22tw). If we see a spike of `INSUFFICIENT_DISCOVERY_NODES` when setting the config to an empty string, we know that the change is not working as intended and we need to instead remove the variable in optimizely entirely.

If we don't actively watch the chart, there's an alert set up to send a slack message (and an email to me) if the claim failure rate goes above 80% (we're working on fixing some outstanding unknown errors still and will tweak that threshold)

PR #936, opened 2022-02-17
Add remote config for libs discovery block diff

jowlee:
### Description
Updates libs
Adds remote config for libs discovery block diff


audius-infra:
Preview this change https://demo.audius.co/jowlee-libs-discovery-block-diff

sddioulde (approved):

PR #941, opened 2022-02-18
Add size prop to ArtistCard Stats

stereosteve:
### Description

fixes look of stats on artist card


audius-infra:
Preview this change https://demo.audius.co/fix-artist-card-stats

sliptype (approved):
ezpz

PR #953, opened 2022-02-18
Migrate lineup action to ts, implement play/pause in lineup

dylanjeffers:
### Description

- Migrates lineup actions to ts to make it easier to import and use lineup actions
- Implements play/pause actions in lineup using new types
- Removes redundant play/pause implementations

### Dragons

We may want the ability to know when a track is played/paused from outside the lineup. Rather than keep the `playTrack` `pauseTrack` props, thinking we add`onPlay`, `onPause` callbacks at that point?

### DIFF
--- a/packages/mobile/src/components/lineup/Lineup.tsx
+++ b/packages/mobile/src/components/lineup/Lineup.tsx
@@ -99,8 +99,6 @@ export const Lineup = ({
   lineup,
   loadMore,
   header,
-  playTrack,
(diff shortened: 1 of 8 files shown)


sliptype (approved):
Love it! This is a relic of how the web lineup was implemented. Lots of unnecessary prop drilling and indirection

And yeah, I don't think there are any places we need those callbacks right now. But can add in the future

PR #956, opened 2022-02-19
Migrate trending-underground to common

dylanjeffers:
### Description

- Moves trending-underground to common
- Types lineup reducer
- Updates other reducer types


audius-infra:
Preview this change https://demo.audius.co/dj-migrate-underground-to-common

PR #964, opened 2022-02-20
Add feature flag to disable sign up confirmation

sliptype:
### Description

* Adds `DISABLE_SIGN_UP_CONFIRMATION` feature flag. Hoping this allows more users to successfully sign up while most discprovs are behind
* Adds `SIGN_IN_WITH_INCOMPLETE_ACCOUNT` event to track users who get into bad sign up state

### Dragons

This could potentially break many things. Since we aren't confirming, the user is potentially missing data until they get indexed. This especially causes problems when signing out immediately and trying to sign back in. We added analytics to determine if more users end up in this bad state

### How Has This Been Tested?

Tested social actions, creating playlists, etc while in an unindexed state and things seem to work pretty well

New analytics event. Monitoring sign up success rate

### DIFF
--- a/packages/web/src/common/models/Analytics.ts
+++ b/packages/web/src/common/models/Analytics.ts
@@ -35,6 +35,7 @@ export enum Name {
   // Sign in
   SIGN_IN_OPEN = 'Sign In: Open',
   SIGN_IN_FINISH = 'Sign In: Finish',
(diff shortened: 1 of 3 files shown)


audius-infra:
Preview this change https://demo.audius.co/sign-up-no-confirm

raymondjacobson (approved):

dylanjeffers (approved):

PR #965, opened 2022-02-20
Debug error from getLocation call

sddioulde:
### Description

To better see why we get null locations

### Dragons

### How Has This Been Tested?

logs and sentry

### DIFF
--- a/packages/web/src/services/Location.ts
+++ b/packages/web/src/services/Location.ts
@@ -1,8 +1,5 @@
 /* global fetch */
 
-import { Level } from 'common/store/errors/level'
-import { reportToSentry } from 'common/store/errors/reportToSentry'
-
 type Location = {
   asn: string
   city: string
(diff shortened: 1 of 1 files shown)


packages/web/src/services/Location.ts
```diff
@@ -26,6 +29,20 @@ export const getLocation = async (): Promise<Location | null> => {
...
+        (e as any).message
+      }`
+    )
+    await reportToSentry({
+      level: Level.Error,
+      error: new Error('IPAPIError'),
+      name: 'IPAPIError',
+      additionalInfo: {
+        error: e,
+        errorMessage: (e as any).message
```
raymondjacobson, line 43: (e as any).message could throw. better (e as any)?.message ?? null

packages/web/src/services/Location.ts
```diff
@@ -26,6 +29,20 @@ export const getLocation = async (): Promise<Location | null> => {
     const res = await fetch('https://ipapi.co/json/')
     return res.json()
   } catch (e) {
+    console.debug(
```
raymondjacobson, line 32: if you just do console.error it will go to sentry. will simplify this code a bit. we send console.error there

sddioulde: nice

audius-infra:
Preview this change https://demo.audius.co/sd-location-error

raymondjacobson (approved):

audius-infra:
Preview this change https://demo.audius.co/sd-location-error

PR #968, opened 2022-02-22
[AUD-1549] Add UndergroundTrendingScreen

dylanjeffers:
### Description

Adds trending underground screen and connects it to tile


packages/mobile/src/components/app-navigator/BottomTabNavigator.tsx
```diff
@@ -61,7 +62,13 @@ const TrendingStackScreen = createStackScreen<TrendingStackParamList>(Stack => (
 ))
 
 const ExploreStackScreen = createStackScreen<ExploreStackParamList>(Stack => (
-  <Stack.Screen name='explore-stack' component={ExploreScreen} />
+  <>
+    <Stack.Screen name='explore-stack' component={ExploreScreen} />
+    <Stack.Screen
+      name='TrendingUnderground'
```
Kyle-Shanks, line 68: should we keep kebab-case here?

Kyle-Shanks (approved):

packages/mobile/src/components/app-navigator/BottomTabNavigator.tsx
dylanjeffers: chatted with sebastian about it at some point, we are gonna move to PascalCase cause that is what react-native examples default to

Kyle-Shanks: sounds good 👍

PR #970, opened 2022-02-23
[AUD-1595] Refactor TabSlider to SegmentedControl

dylanjeffers:
### Description

- Renames TabSlider to SegmentedControl
- Refactors SegmentedControl to use StylesProp
- Fixes `fullWidth` mode


packages/mobile/src/components/core/SegmentedControl.tsx
```diff
@@ -0,0 +1,211 @@
...
+    backgroundColor: palette.neutralLight7,
+    flexDirection: 'row',
+    alignItems: 'center',
+    padding: OFFSET,
+    paddingRight: 0
+  },
+  tab: {
+    elevation: 3,
+    paddingTop: 10,
+    paddingHorizontal: spacing(4),
```
nicoback2, line 69: why use spacing here and not for `paddingTop`?

packages/mobile/src/components/core/SegmentedControl.tsx
```diff
@@ -0,0 +1,211 @@
...
+  TextStyle,
+  View,
+  ViewStyle,
+  Text
+} from 'react-native'
+
+import { makeStyles, StylesProps } from 'app/styles'
+
+// Note, offset is the inner padding of the container div
+const OFFSET = 3
```
nicoback2, line 16: Why is this in a constant but not other values?

nicoback2 (approved):

sliptype (approved):

packages/mobile/src/components/core/SegmentedControl.tsx
dylanjeffers: Missed it good catch

PR #977, opened 2022-02-23
[AUD-734] Add mobile Favorites Page

Kyle-Shanks:
### Description
* Move saved-page state to common
* Add TrackList component (basic version to unblock work on settings page, collections screen, etc.)
* Update FavoritesScreen tabs to display actual data

### Dragons

There are a lot of things used, but would love a look over to make sure that I didn't redo any work/components that were already done.

### How Has This Been Tested?

Lots of manual testing, need to test data on staging again, but it worked before


audius-infra:
Preview this change https://demo.audius.co/kj-Add-mobile-Favorites-page

.gitignore
```diff
@@ -1,9 +1,58 @@
...
+/coverage
 
-node_modules
+# build
+/build
+/build-development
+/build-demo
+/build-staging
+/build-mobile-staging
+/build-ci
```
dylanjeffers, line 22: so for this .gitignore, i thinkk most of this stuff got moved to the packages/web and packages/mobile sections. mind double checking on that?

.gitignore
```diff
@@ -1,9 +1,58 @@
...
+/build-production
```
dylanjeffers, line 23: so for this .gitignore, i thinkk most of this stuff got moved to the packages/web and packages/mobile sections. mind double checking on that?

packages/mobile/src/components/core/SearchInput.tsx
```diff
@@ -28,11 +29,13 @@ const useStyles = makeStyles(({ typography, palette, spacing }) => ({
   }
 }))
 
-type SearchInputProps = TextInputProps
+type SearchInputProps = TextInputProps & {
+  Icon?: React.FC<SvgProps>
```
dylanjeffers, line 33: ComponentType<SvgProps> I believe is the best here, since React.FC has implicit children prop

packages/mobile/src/components/track-list/TablePlayButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+  svg: {},
+  hideDefault: {}
+}))
+
+export const TablePlayButton = ({
+  playing = false,
+  paused,
+  hideDefault = true,
+  onPress = () => {},
+  styles: propStyles = {
```
dylanjeffers, line 29: i like the default arg here, what we have been doing in other componets is just `stylesProp?.root`, what do you think?

packages/mobile/src/components/track-list/TablePlayButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+    root: {},
+    svg: {}
+  }
+}: TablePlayButtonProps) => {
+  const styles = useStyles()
+
+  return (
+    <TouchableOpacity style={[styles.root, propStyles.root]} onPress={onPress}>
+      {playing && !paused ? (
+        <View>
```
dylanjeffers, line 39: Do we need the containing views?

packages/mobile/src/components/track-list/TrackList.tsx
```diff
@@ -0,0 +1,116 @@
...
+            styles.divider,
+            hideDivider(idx) ? styles.hideDivider : {},
+            noDividerMargin ? styles.noMarginDivider : {}
+          ]}
+        />
+      ) : null}
+      <TrackListItem
+        key={track.trackId}
+        index={idx}
+        trackId={track.trackId}
```
dylanjeffers, line 88: can we do {...track} to get most of the props through?

packages/mobile/src/components/track-list/TrackList.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const hideDivider = (idx: number) =>
+    activeIndex >= 0 && (activeIndex === idx || activeIndex === idx - 1)
+
+  const renderTrack = ({ item: track, index: idx }) => (
+    <View key={track.uid}>
+      {showDivider && (showTopDivider || idx > 0) ? (
+        <View
+          style={[
+            styles.divider,
+            hideDivider(idx) ? styles.hideDivider : {},
```
dylanjeffers, line 80: hideDivider(idx) && styles.hideDivider also works. if we want to do ternery like we do with rendering, i think can just return null for else case, instead of empty object

packages/mobile/src/components/track-list/TrackList.tsx
```diff
@@ -0,0 +1,116 @@
...
+  showTopDivider,
+  togglePlay,
+  trackItemAction,
+  tracks
+}: TrackListProps) => {
+  const styles = useStyles()
+
+  // The dividers above and belove the active track should be hidden
+  const activeIndex = tracks.findIndex(track => track.isActive)
+  const hideDivider = (idx: number) =>
```
dylanjeffers, line 71: this can prob be a computed val inside the renderTrack function?

packages/mobile/src/components/track-list/TrackList.tsx
```diff
@@ -0,0 +1,116 @@
...
+    activeIndex >= 0 && (activeIndex === idx || activeIndex === idx - 1)
+
+  const renderTrack = ({ item: track, index: idx }) => (
```
dylanjeffers, line 74: personally i like keeping index instead of idx, just for semantics, what do you think?

dylanjeffers, line 74: also careful of the implicit any for track and index here!

packages/mobile/src/components/track-list/TrackListItem.tsx
```diff
@@ -0,0 +1,313 @@
...
+  playButtonContainer: {
+    marginRight: spacing(4)
+  }
+}))
+
+const ArtworkIcon = ({ isLoading, isPlaying }: ArtworkIconProps) => {
+  const [artworkHeight, setArtworkHeight] = useState(0)
+  const [artworkWidth, setArtworkWidth] = useState(0)
+  const styles = useStyles()
+  let artworkIcon
```
dylanjeffers, line 114: nit but maybe we move the artworkIcon right before the if / else

packages/mobile/src/components/track-list/TrackListItem.tsx
```diff
@@ -0,0 +1,313 @@
...
+
+type ArtworkProps = {
+  trackId: ID
+  isLoading: boolean
+  isActive?: boolean
+  isPlaying: boolean
+  coverArtSizes: CoverArtSizes
+}
+
+const Artwork = ({
```
dylanjeffers, line 170: would love to know if we have a component similar to this that we can share, some kinda `TrackCoverArt` component?

packages/mobile/src/components/track-list/TrackListItem.tsx
```diff
@@ -0,0 +1,313 @@
...
+            {trackTitle}
+            {messages.deleted}
+          </Text>
+          <Text numberOfLines={1} style={styles.artistName}>
+            {artistName}
+            <UserBadges user={user} badgeSize={12} hideName />
+          </Text>
+        </View>
+        {onSaveTrack && trackItemAction === 'save' && (
+          <TouchableOpacity style={styles.iconContainer} onPress={onSaveTrack}>
```
dylanjeffers, line 302: maybe use IconButton here?

packages/mobile/src/screens/explore-screen/tabs/ArtistsTab.tsx
```diff
@@ -27,20 +27,10 @@ const useStyles = makeStyles(({ spacing }) => ({
...
-    const users = state.users.entries
-    return userIds.map(id => users[id].metadata).filter(Boolean)
-  }
-}
-
 export const ArtistsTab = ({ navigation }: Props) => {
   const styles = useStyles()
-  const userIds = useSelectorWeb(getExploreUsers)
-  const profiles = useSelectorWeb(makeGetUsers(userIds))
+  const getExplore = makeGetExplore()
```
dylanjeffers, line 32: lets move this outside the component so useSelectoWeb is always referencing the same selector function.

packages/mobile/src/screens/explore-screen/tabs/PlaylistsTab.tsx
```diff
@@ -26,29 +26,10 @@ const useStyles = makeStyles(({ spacing }) => ({
...
-        user: users[collection.playlist_owner_id]?.metadata ?? {}
-      }))
-  }
-}
-
 export const PlaylistsTab = ({ navigation }: Props) => {
   const styles = useStyles()
-  const playlistIds = useSelectorWeb(getExplorePlaylists)
-  const playlists = useSelectorWeb(makeGetFullPlaylists(playlistIds))
+  const getExplore = makeGetExplore()
```
dylanjeffers, line 31: same!

packages/mobile/src/screens/favorites-screen/AlbumsTab.tsx
```diff
@@ -0,0 +1,59 @@
...
+const messages = {
+  emptyTabText: "You haven't favorited any albums yet.",
+  inputPlaceholder: 'Filter Albums'
+}
+
+export const AlbumsTab = () => {
+  const [filterValue, setFilterValue] = useState('')
+  const user = useSelectorWeb(getAccountWithAlbums)
+
+  const matchesFilter = (playlist: any) => {
```
dylanjeffers, line 21: let's avoid any

packages/mobile/src/screens/favorites-screen/EmptyTab.tsx
```diff
@@ -0,0 +1,82 @@
...
+    fontSize: 16,
+    lineHeight: 26,
+    letterSpacing: 0.2,
+    marginBottom: spacing(4),
+    color: palette.neutral,
+    textAlign: 'center'
+  }
+}))
+
+export const EmptyTab = ({ message }: EmptyTabProps) => {
```
dylanjeffers, line 38: This is great! we have a basic impl of this in the profile page, so ill def use this one instead cause its more built out

packages/mobile/src/screens/favorites-screen/FilterInput.tsx
```diff
@@ -0,0 +1,40 @@
...
+    padding: spacing(2)
+  },
+  input: {
+    backgroundColor: palette.neutralLight10
+  }
+}))
+
+export const FilterInput = ({
+  value,
+  onChange = () => {},
```
dylanjeffers, line 25: i imagine onChange should be a required prop?

packages/mobile/src/screens/favorites-screen/FilterInput.tsx
```diff
@@ -0,0 +1,40 @@
...
+  placeholder = ''
```
dylanjeffers, line 26: and yeah similarly, i wouldnt worry about a deault value hre, just use searchinput defaults?

packages/mobile/src/screens/favorites-screen/FilterInput.tsx
```diff
@@ -0,0 +1,40 @@
+import IconFilter from 'app/assets/images/iconFilter.svg'
+import { SearchInput, Tile } from 'app/components/core'
+import { makeStyles } from 'app/styles'
+
+type FilterInputProps = {
```
dylanjeffers, line 5: id prob type this: FilterInputProps = SearchInputProps

packages/mobile/src/screens/favorites-screen/TracksTab.tsx
```diff
@@ -0,0 +1,197 @@
...
+    const matchValue = filterValue.toLowerCase()
+    return (
+      track.trackTitle.toLowerCase().indexOf(matchValue) > -1 ||
+      track.artistName.toLowerCase().indexOf(matchValue) > -1
+    )
+  }
+
+  const trackList = tracksWithUsers
+    .map(track => ({
+      isLoading: false,
```
dylanjeffers, line 107: ah i see, i think this is fine, but i wonder if it makes sense to not have to rename all the props and instead just pass down the track unmodified to those inner components? it could simplify their api too cause they won't have like 20 props, and instead just have `track`, and like `isLoading` or something?

we can def handle this later though

dylanjeffers (approved):
Wow what a huge lift, great job!!

.gitignore
Kyle-Shanks: Yea for sure, I had things like the builds and the public scripts top level showing up in my git diff so I just copied over another gitignore that we use. I can give it a once over and filter it down 👍

packages/mobile/src/components/track-list/TablePlayButton.tsx
Kyle-Shanks: Yea I should have checked, sounds good to me

.gitignore
sliptype: I think this happens when switching from `master` to `main`. Git doesn't delete those files for some reason even though they don't exist on `main`. The better solution is to manually delete those files (since we won't be using `master` anymore). That way we don't need to change the gitignore

Kyle-Shanks: good point, I'll just remove and revert the gitignore 👍

audius-infra:
Preview this change https://demo.audius.co/kj-Add-mobile-Favorites-page

sliptype (approved):
Big changes, looks great!

packages/mobile/src/components/track-list/TrackListItem.tsx
```diff
@@ -0,0 +1,306 @@
...
+        styles.trackContainer,
+        isActive ? styles.trackContainerActive : {}
+      ]}
+    >
+      <TouchableOpacity
+        style={styles.trackInnerContainer}
+        onPress={onClickTrack}
+      >
+        {coverArtSizes ? (
+          <Artwork
```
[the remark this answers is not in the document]
Kyle-Shanks: Listening History in Settings page per Ray 👍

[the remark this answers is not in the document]
sliptype: Ah dope, might be cool to include on other track lists in the future!

audius-infra:
Preview this change https://demo.audius.co/kj-Add-mobile-Favorites-page

PR #990, opened 2022-02-26
[AUD-1611] Fix airplay and cast for native

raymondjacobson:
### Description

This change fixes the cast integration with the common store and also addresses the local storage issue in common.
Curious for feedback on this approach, but I figured that abstracting the complexity around storage away from the sagas and reducers themselves is the way to go.

One annoying thing is of course that AsyncStorage is async and local storage interaction is not, so local storage really can't be used in a common store reducer. I'm guessing that there will be more platform-specific context to be passing in for system APIs in the future though.

### Dragons

Not actively used yet

### How Has This Been Tested?

Locally vs. staging -- opening / setting cast options.

Mobile staging app

### DIFF
--- a/packages/web/src/store/reducers.ts
+++ b/packages/web/src/store/reducers.ts
@@ -47,10 +47,14 @@ import player from 'store/player/slice'
 import playlistLibrary from 'store/playlist-library/slice'
 import reachability from 'store/reachability/reducer'
 
+import { webStoreContext } from './storeContext'
+
+export const commonStoreReducers = clientStoreReducers(webStoreContext)
+
 const createRootReducer = (routeHistory: History) =>
   combineReducers({
-    // Client store
-    ...clientStoreReducers,
+    // Common store
+    ...commonStoreReducers,
 
     // Router
     router: connectRouter(routeHistory),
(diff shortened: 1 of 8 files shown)


packages/web/src/store/reducers.ts
```diff
@@ -47,10 +47,14 @@ import player from 'store/player/slice'
 import playlistLibrary from 'store/playlist-library/slice'
 import reachability from 'store/reachability/reducer'
 
+import { webStoreContext } from './storeContext'
+
+export const commonStoreReducers = clientStoreReducers(webStoreContext)
```
raymondjacobson, line 52: The crux of this PR is that common store reducers (and sagas below) are initialized with a context that has local storage handlers (and could have other things).

audius-infra:
Preview this change https://demo.audius.co/rj-aud-1611

raymondjacobson:
> 

Very much stole it. Anticipating that other sagas will need access to some other "system" APIs in the future

dylanjeffers (approved):
Yeah this makes a lot of sense to me, and cool that it's following patterns similar to remote-config

PR #992, opened 2022-02-28
[AUD-1610] Fix ProfileReposts lineup

dylanjeffers:
### Description

Fixes issue where ProfileReposts lineup doesn't load

- implements `selfLoad` prop for lineups to trigger initial load
- converts profile store to typescript


audius-infra:
Preview this change https://demo.audius.co/dj-aud-1610-web-profile-nav

sliptype (approved):
Nice! I was meaning to do this but didn't think we would need it until the web dependency is removed

PR #996, opened 2022-02-28
[AUD-1629] Connect ProfileScreen

dylanjeffers:
### Description

Connects profile screen actions
- Subscribe/follow buttons
- Linkify all the links
- EditProfile
- Various UI fixes
- Upgrades react-native-image-picker (wasn't working properly on the older version)

### Dragons

There is a pretty big hack added in the BaseNavigator to not dispatch "onBack" for edit-profile screen since that isn't actually a route in mobile. We might want to think of a better long term solution


packages/mobile/src/screens/profile-screen/ExpandableBio.tsx
```diff
@@ -83,13 +84,15 @@ export const ExpandableBio = ({ profile }: ExpandableBioProps) => {
...
       <View>
         {bio ? (
-          <Text
-            numberOfLines={fullBioHeight && !isExpanded ? 2 : 0}
-            style={styles.bio}
-            onLayout={handleBioLayout}
-          >
-            {bio}
-          </Text>
+          <HyperLink>
```
sliptype, line 87: Does this not need a prop to handle the link press?

sliptype (approved):
Good stuff! Yeah we def need to go back and fix those types in `EditProfileScreen`

packages/mobile/src/screens/profile-screen/ExpandableBio.tsx
dylanjeffers: updated!

PR #1000, opened 2022-03-01
[AUD-972] Fill out explainer and tiers of audio page

raymondjacobson:
### Description

Add additional UI to /audio page. Does not connect the challenges yet.

### Dragons

### How Has This Been Tested?

https://user-images.githubusercontent.com/2731362/156215061-8fb5582a-171f-45b6-8c17-410cc968d1c2.mp4

Mobile native release


packages/mobile/src/components/mobile-upload-drawer/MobileUploadDrawer.tsx
```diff
@@ -32,7 +32,7 @@ const useStyles = makeStyles(({ typography, spacing }) => ({
     width: 80
   },
   title: {
-    fontSize: typography.fontSize.xxxl,
+    fontSize: typography.fontSize.xxxxl,
```
raymondjacobson, line 35: the web client does 4xl for this. I kinda like that more, but not sure if this was intentional

sliptype: I'm on board with `4xl`. Thoughts @dylanjeffers ?

packages/mobile/src/screens/audio-screen/Tier.tsx
```diff
@@ -0,0 +1,142 @@
...
+    width: '100%',
+    justifyContent: 'center',
+    alignItems: 'center',
+    paddingVertical: spacing(10),
+    paddingHorizontal: spacing(4),
+    borderColor: palette.neutralLight7,
+    borderWidth: 2,
+    borderRadius: spacing(4),
+    backgroundColor: palette.neutralLight10,
+    flexDirection: 'column'
```
sliptype, line 33: Column is the default flex direction in RN

packages/mobile/src/components/mobile-upload-drawer/MobileUploadDrawer.tsx
sliptype: I wonder if it makes sense to have a function like `spacing(2)` for font size too

sliptype (approved):
Looks great!!

packages/mobile/src/screens/audio-screen/Tier.tsx
raymondjacobson: ty

raymondjacobson: well we have the h1,h2,h3... styles, which over time, we should figure out just how to keep 1-1 w/ julian's design so we're never off cannon

packages/mobile/src/components/mobile-upload-drawer/MobileUploadDrawer.tsx
dylanjeffers: i would have loved that, however... numbers as the first value in js :/

packages/mobile/src/screens/audio-screen/Tier.tsx
dylanjeffers: yeah totally, and just a headsup im totally guessing on the existing values, so we should all def work to define them more rigorously

PR #1001, opened 2022-03-01
[AUD-1632] ProfileScreen performance

dylanjeffers:
### Description

Cuts down profile page rerenders from 60 (at the root) to 3 for the entire profile stack
- Uses hook selector hybrids to share equality functions, but open to defining whatever pattern we want.

### Dragons

There are assumptions made with the selectors that prevent potential re-renders that are necessary for some subcomponents to function (ie expecting nested profile data that gets added in a separate re-render)

### How Has This Been Tested?

Ran through various profiles to ensure things look the same

If there seems to be data missing on the profile page, we will look into existing selectors as one of the first points of triage

### DIFF
--- a/packages/mobile/src/hooks/selectors.ts
+++ b/packages/mobile/src/hooks/selectors.ts
@@ -1,0 +1,19 @@
+import { getAccountUser } from 'audius-client/src/common/store/account/selectors'
+import { getProfileUser } from 'audius-client/src/common/store/pages/profile/selectors'
+import { isEmpty } from 'lodash'
+
+import { useSelectorWeb } from './useSelectorWeb'
+
+export const useProfile = () => {
+  return useSelectorWeb(
+    getProfileUser,
+    (a, b) =>
+      a?.user_id === b?.user_id &&
+      isEmpty(a?._cover_photo_sizes) === isEmpty(b?._cover_photo_sizes) &&
+      isEmpty(a?._profile_picture_sizes) === isEmpty(b?._profile_picture_sizes)
+  )
+}
+
+export const useAccountUser = () => {
+  return useSelectorWeb(getAccountUser, (a, b) => a?.user_id === b?.user_id)
+}
--- a/packages/mobile/src/hooks/selectors.ts
+++ b/packages/mobile/src/hooks/selectors.ts
@@ -1,19 +1,13 @@
 import { getAccountUser } from 'audius-client/src/common/store/account/selectors'
 import { getProfileUser } from 'audius-client/src/common/store/pages/profile/selectors'
-import { isEmpty } from 'lodash'
+import { isEqual } from 'lodash'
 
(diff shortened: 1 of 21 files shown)


packages/mobile/src/hooks/selectors.ts
```diff
@@ -0,0 +1,19 @@
+import { getAccountUser } from 'audius-client/src/common/store/account/selectors'
+import { getProfileUser } from 'audius-client/src/common/store/pages/profile/selectors'
+import { isEmpty } from 'lodash'
+
+import { useSelectorWeb } from './useSelectorWeb'
+
+export const useProfile = () => {
+  return useSelectorWeb(
+    getProfileUser,
+    (a, b) =>
```
sliptype, line 10: My only concern here is that these equality functions may no longer be necessary when we remove the web dependency, because reference equality will work correctly then

sliptype (approved):
This should be fine, I do think using `_.isEqual` would be safer though. When I ran some benchmarks on it last time it was only taking ~3ms to compare big cache objects

packages/mobile/src/hooks/selectors.ts
dylanjeffers: ill def try the `isEqual` now and see how many re-renders that results in. i think these choices were less to do with selector performance, and more to do with number of re-renders, cause the root only needs user_id and a few other constants, and dont want it rerendering for things like track length increasing

raymondjacobson: At the very least, maybe we could make a linear to track this and write
`TODO(AUD-1234): This selector is not necessary when common store is vendored in native`

packages/mobile/src/screens/profile-screen/TracksTab.tsx
```diff
@@ -1,15 +1,29 @@
...
+import { useProfileTracksLineup } from './selectors'
 
 export const TracksTab = () => {
-  const { profile } = useSelectorWeb(getProfile)
-  const lineup = useSelectorWeb(getProfileTracksLineup)
+  const profile = useProfile()
+  const lineup = useProfileTracksLineup()
+  const dispatchWeb = useDispatchWeb()
+
+  const loadMore = useCallback(
```
raymondjacobson, line 17: Just curious, was this an oversight before to not call loadMore?

raymondjacobson (approved):

packages/mobile/src/screens/profile-screen/TracksTab.tsx
dylanjeffers: i added a default `loadMore` implementation in lineup that i thought covered everything, so i removed this after that, but turns out some `fetchLinups` need extra data

dylanjeffers:
as a headsup @sliptype, went with isEqual for profile page, and things seem to be working fine, a few extra re-renders, but prob fine overall

PR #1002, opened 2022-03-01
Fix profile track-tab loadMore

dylanjeffers:
### Description

Fixes issue where default loadMore wasn't including userId, leading to a load error for users with more than 12 tracks
- also patches type error

### DIFF
--- a/packages/mobile/src/screens/profile-screen/TracksTab.tsx
+++ b/packages/mobile/src/screens/profile-screen/TracksTab.tsx
@@ -17,9 +17,10 @@ export const TracksTab = () => {
 
   const loadMore = useCallback(
     (offset: number, limit: number) => {
+      if (!profile) return
       dispatchWeb(
         tracksActions.fetchLineupMetadatas(offset, limit, false, {
-          userId: profile?.user_id
+          userId: profile.user_id
         })
       )
     },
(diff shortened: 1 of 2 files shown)


packages/mobile/src/screens/profile-screen/TracksTab.tsx
```diff
@@ -10,6 +13,18 @@ import { getProfile } from './selectors'
 export const TracksTab = () => {
   const { profile } = useSelectorWeb(getProfile)
   const lineup = useSelectorWeb(getProfileTracksLineup)
+  const dispatchWeb = useDispatchWeb()
+
+  const loadMore = useCallback(
+    (offset: number, limit: number) => {
+      dispatchWeb(
+        tracksActions.fetchLineupMetadatas(offset, limit, false, {
```
raymondjacobson, line 21: what happens if this is null?

raymondjacobson (approved):

packages/mobile/src/screens/profile-screen/TracksTab.tsx
dylanjeffers: ah, we do get the existing issue, so will add a check here just in case :)

PR #1007, opened 2022-03-02
[AUD-1634] Conditionally render native drawers

dylanjeffers:
### Description

Improves native performance by conditionally rending drawers. This improves a few things:
  1. Initial render is 3x faster
  2. Average app rerenders are down 30% as drawers don't continuously rerender due to state changes
  3. Overall dom elements are down 50%. This should help with memory and general react performance.

As to the true extent of whether this helps perceivable perf, it's hard to know, but i think taken together this will prob help!

- Additionally changed the OverflowMenu open/close functionality to use the common modal slice.

### Dragons

- the modal slice `visible` boolean was changed to `boolean | 'closing'` which is pretty hacky. This is needed to wait for drawer close animation to finish before unmounting the drawer. I wanted to avoid modifying modal slice too much, since web still uses `boolean`, but if we think there is a better way to represent this (like changing visible from boolean to status variable), i can time-box to see how long it would take
- the Drawers file is a bit complicated now, and it's not super clear how it works. If needed, can move these drawers to another file and clarify their usage.

### How Has This Been Tested?

Most of the changes are on the native side. Drawers open and close like before!
Since this affects the common modal slice, I also tested the desktop/mobile menus, and they also work like before.


audius-infra:
Preview this change https://demo.audius.co/dj-aud-1634-drawer-perf

packages/mobile/src/components/drawer/NativeDrawer.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { useNativeDrawer } from 'app/hooks/useDrawer'
+import { Drawer as DrawerName } from 'app/store/drawers/slice'
+
+import Drawer, { DrawerProps } from './Drawer'
+
+type NativeDrawerProps = SetOptional<DrawerProps, 'isOpen' | 'onClose'> & {
+  drawerName: DrawerName
+}
+
+export const NativeDrawer = (props: NativeDrawerProps) => {
```
raymondjacobson, line 14: Would be helpful to have a component comment  that explains NativeDrawer vs. Drawer vs. AppDrawer
/**
 *
 */

packages/mobile/src/hooks/useDrawer.ts
```diff
@@ -14,11 +16,32 @@ import { Drawer, setVisibility } from 'app/store/drawers/slice'
...
-    dispatch(setVisibility({ drawer, visible }))
+  const setIsOpen = (visible: boolean, closed?: boolean) => {
+    let visibleStatus = visible ? true : ('closing' as const)
+    if (closed) {
+      visibleStatus = false
+    }
+    dispatch(setVisibility({ drawer, visible: visibleStatus }))
+  }
+
+  return [isOpen === true, setIsOpen, isOpen]
```
raymondjacobson, line 30: Can we rename `isOpen` to something like state/status/similar? this was a bit confusing to me at first

packages/mobile/src/Drawers.tsx
```diff
@@ -19,32 +23,91 @@ import { TransferAudioMobileDrawer } from 'app/components/transfer-audio-mobile-
...
+  CollectibleDetails: CollectibleDetailsDrawer,
+  DeactivateAccountConfirmation: DeactivateAccountConfirmationDrawer,
+  Cognito: CognitoDrawer,
+  FeedFilter: FeedFilterDrawer,
+  TrendingGenreSelection: TrendingFilterDrawer,
+  MobileUpload: MobileUploadDrawer,
+  Overflow: OverflowMenuDrawer
+  /* Disable the audio breakdown drawer until we get
+   * the feature flags to work for native mobile
+   * AudioBreakdown: AudioBreakdownDrawer
```
sliptype, line 75: Outside of the scope of this PR, but the `AudioBreakdownDrawer` can be enabled now

packages/mobile/src/hooks/useDrawer.ts
```diff
@@ -14,11 +16,32 @@ import { Drawer, setVisibility } from 'app/store/drawers/slice'
  */
 export const useDrawer = (
```
sliptype, line 17: So it looks like `useDrawer` is only used for NowPlaying right?

Can we maybe rename `useNativeDrawer` since that will be the default once we aren't dependent on web?

sliptype (approved):
This is awesome! I agree some more comments around the various Drawer components and hooks would be good. If I were looking at this for the first time trying to figure out how to implement/add a drawer I would be pretty confused lol

packages/mobile/src/Drawers.tsx
```diff
@@ -19,32 +23,91 @@ import { TransferAudioMobileDrawer } from 'app/components/transfer-audio-mobile-
...
+        )
+      })}
+      {nativeDrawers.map(([drawerName, Drawer]) => (
+        <NativeDrawer
+          key={drawerName}
+          drawerName={drawerName}
+          drawer={Drawer}
+        />
+      ))}
       <AddToPlaylistDrawer />
```
raymondjacobson, line 110: why is add to playlist separated?

raymondjacobson (approved):
I merged my drawer PR, but pls lmk if i can help w/ rebase

packages/mobile/src/hooks/useDrawer.ts
dylanjeffers: yeah down

packages/mobile/src/hooks/useDrawer.ts
dylanjeffers: yeah exactly, and sure can def update

packages/mobile/src/Drawers.tsx
dylanjeffers: so the drawers that use their own slice to manage open/closed cant use the `useDrawer` hooks etc. im down to try and migrate this one though

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1634-drawer-perf

PR #1009, opened 2022-03-02
[AUD-1639] conditionally render NotificationDrawer

dylanjeffers:
### Description

Conditionally renders the notifications drawer which should help with render performance. The numbers I am seeing:
- 50% fewer overall dom elements
- 20 or so fewer rerenders

 also refactored hoc to hooks


packages/mobile/src/components/notifications/Notifications.tsx
```diff
@@ -69,27 +67,44 @@ const styles = StyleSheet.create({
...
   const isStackOpen = useIsStackOpen()
+  const [drawerStatus, setDrawerStatus] = useState<DrawerStatus>('closed')
+
+  useEffect(() => {
+    setDrawerStatus(isOpen ? 'open' : 'closed')
+    if (isOpen) {
+      setDrawerStatus('open')
+    } else {
+      setDrawerStatus('closed')
+    }
```
sliptype: We probably don't need to do this twice right?

packages/mobile/src/components/notifications/Notifications.tsx
```diff
@@ -281,18 +297,20 @@ const Notifications = ({
...
       }
     }
-  }, [isOpen, slideIn, anchorRoute, pathname, setAnchorRoute])
+  }, [drawerStatus, slideIn, anchorRoute, pathname, setAnchorRoute])
 
   const containerStyle = useTheme(styles.container, {
     backgroundColor: 'background'
   })
 
+  console.log(drawerStatus)
```
sliptype, line 312: Log

sliptype (approved):
Dope!

PR #1010, opened 2022-03-03
[AUD-1616] Disallow identical track uploads in same session

piazzatron:
### Description

- Selecting identical tracks (either in the same drag-and-drop session, or already selected and in the state) will skip duplicates

### Dragons

None

### How Has This Been Tested?

Tested locally

Looking for error rate of MULTI_TRACK_UPLOAD decreasing


raymondjacobson (approved):
Nice nice nice.
Non-blocking on this PR, but wonder if a toast would be nice for UX. can be separate convo

nicoback2 (approved):

audius-infra:
Preview this change https://demo.audius.co/piazz-fix-identical-uploads

PR #1011, opened 2022-03-03
[AUD-1640] Add null renderer for native

dylanjeffers:
### Description

Uses a null-renderer for native web to improve performance by preventing the web app from writing to the dom.
- Replaces usage of useLocation with useHistory. For some reason useLocation wasn't updating.
- Adds stage/prod mobile env

### Dragons

There may still be issues with the null renderer that we haven't triggered yet
We had to downgrade to react 16x to be compatible with the null-renderer. Going to try again to see if 17x can work now that the overall implementation has been determined

### How Has This Been Tested?

Running the app in mobile stage mode, validating that the web app doesn't render anything if you go to it in browser, and validating the mobile app work as expected


packages/web/src/index.tsx
```diff
@@ -1,6 +1,8 @@
 import React from 'react'
 
 import ReactDOM from 'react-dom'
+// @ts-ignore
```
raymondjacobson, line 4: why the ignore? can we add a d.ts file?

raymondjacobson (approved):
So k00000L.

Would you mind also testing running the app with hash routing enabled and make sure page transitions work?

packages/web/src/index.tsx
dylanjeffers: yeah we def can, just found the declarations file :)

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1640-null-renderer

packages/web/package.json
```diff
@@ -62,17 +62,18 @@
     "promise.allsettled": "1.0.4",
     "prop-types": "15.7.2",
     "raw-loader": "0.5.1",
-    "react": "17.0.2",
+    "react": "^16.14.0",
```
sliptype, line 65: We can run into issues with react erroring because of multiple versions of it in the resolution tree. That happened with stems a lot before we synced the versions. If we can continue to use `17.0.2` that would be great!

sliptype (approved):
Big 🧠  move

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1640-null-renderer

packages/web/.env/.env.mobile.prod
```diff
@@ -0,0 +1,2 @@
+REACT_APP_NATIVE_MOBILE=true
+NATIVE_NAVIGATION_ENABLED=false
```
sliptype, line 2: `REACT_APP_NATIVE_NAVIGATION_ENABLED`

PR #1013, opened 2022-03-03
[AUD-1638] Partial web -> mobile state syncing

sliptype:
### Description

After getting some conflicting signals, I was finally able to confirm that the expensive part of the state syncing operation is sending large amounts of data over the message bridge. Here is an average case of the times in ms for each leg of the trip:

image

Sending only the changed top-level entries of the state over the bridge, the `Web -> Mobile` time was decreased by ~75%. This specific case was for an update in the `pages` slice

image

We could probably further improve this by flattening the redux state, but that might get into diminishing returns

Also includes:
  * Rename `clientStore` to `commonState` for clarity
  * Added some useCallbacks to prevent rerenders on FeedScreen

### Dragons

Using reference equality on the web layer to determine which slices have changed, so theoretically the mobile state should never get out of sync. But something to be aware of if we do run into something weird

### How Has This Been Tested?

Tested on ios simulator and device

Not deployed yet

### DIFF
--- a/packages/web/src/store/configureStore.ts
+++ b/packages/web/src/store/configureStore.ts
@@ -1,6 +1,6 @@
 import * as Sentry from '@sentry/browser'
 import { routerMiddleware, push as pushRoute } from 'connected-react-router'
-import { pick } from 'lodash'
+import { isEmpty, pick, pickBy } from 'lodash'
 import { createStore, applyMiddleware, Action, Store } from 'redux'
 import { composeWithDevTools } from 'redux-devtools-extension/logOnlyInProduction'
 import createSagaMiddleware from 'redux-saga'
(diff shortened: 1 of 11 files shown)


audius-infra:
Preview this change https://demo.audius.co/sk-aud-1638-sync-latency

packages/web/src/store/configureStore.ts
```diff
@@ -116,28 +116,41 @@ const middlewares = applyMiddleware(
...
+        previousState = state
+        return
+      }
+
+      // Subsequently only sync the changed slices
+      const stateToSync = pickBy(
+        state,
+        (value, key) =>
+          commonStateKeys.includes(key) &&
+          value !== previousState[key as keyof RootState]
```
dylanjeffers, line 145: so if im understanding correctly, only sync the objects at the top level that have changed?

dylanjeffers (approved):
amazing investigation and solution

packages/web/src/store/configureStore.ts
sliptype: Exactly! We could use flatten the state to get even more benefit here

PR #1014, opened 2022-03-03
Remove logic to individually enable RN screens

sliptype:
### Description

This PR removes extra logic at the AppNavigator level to allow enabling certain RN screens individually. Since we decided to roll all the screens out at once, this is a little more performant

### Dragons

Wraps the BottomTabNavigator in a check for if the navigation is enabled, this prevents RN screens from rendering when they are supposed to be disabled. Shouldn't cause issues though

### How Has This Been Tested?

Manually on ios simulator

Not deployed yet

### DIFF
--- a/packages/mobile/src/components/app-navigator/AppNavigator.tsx
+++ b/packages/mobile/src/components/app-navigator/AppNavigator.tsx
@@ -1,8 +1,7 @@
-import { useCallback, useMemo, useState } from 'react'
+import { useMemo } from 'react'
 
-import { useNavigationState } from '@react-navigation/native'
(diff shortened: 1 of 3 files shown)


dylanjeffers (approved):
Than you for handling this! I was trying to but got a bit worried that i might mess up the existing native app functionality. Assuming this still handles that?

sliptype:
> Than you for handling this! I was trying to but got a bit worried that i might mess up the existing native app functionality. Assuming this still handles that?

Yup! The only thing that we really have to account for is the signon flow which is handled with the `isAuthed` check

PR #1015, opened 2022-03-03
[AUD-1361] Improve bottom tab button performance

sliptype:
### Description

* Fixes bug where buttons get stuck in active state
* Fixes some rerenders triggered by updated anon function references
* Converts `common` slice to non-reduxjstoolkit to avoid `immer.produce` performance overhead

### Dragons

Potentially breaking changes to navigation

### How Has This Been Tested?

Test on simulator and ios device

N/A


audius-infra:
Preview this change https://demo.audius.co/sk-aud-1361-bottom-button-performance

packages/mobile/ios/Podfile.lock
```diff
@@ -324,7 +324,7 @@ PODS:
     - react-native-google-cast/RNGoogleCast
   - react-native-google-cast/RNGoogleCast (3.4.1):
     - React
-  - react-native-image-picker (4.7.3):
+  - react-native-image-picker (4.0.6):
```
dylanjeffers, line 327: run an `npm i` first, and then a pod install which should remove this file

packages/mobile/src/components/app-navigator/AppNavigator.tsx
```diff
@@ -72,9 +72,8 @@ const AppNavigator = () => {
             name='main'
             navigationKey='main'
             options={{ headerShown: false }}
-          >
-            {() => (IS_MAIN_NAVIGATION_ENABLED ? <BottomTabNavigator /> : null)}
-          </Stack.Screen>
+            component={IS_MAIN_NAVIGATION_ENABLED ? BottomTabNavigator : View}
```
dylanjeffers, line 75: nicee

packages/mobile/src/components/app-navigator/BottomTabNavigator.tsx
```diff
@@ -74,12 +74,9 @@ const ExploreStackScreen = createStackScreen<ExploreStackParamList>(Stack => (
...
       component={TrendingUndergroundScreen}
     />
-    <Stack.Screen
-      name='UnderTheRadar'
-      component={() => (
-        <SmartCollectionScreen smartCollection={UNDER_THE_RADAR} />
-      )}
-    />
+    <Stack.Screen name='UnderTheRadar'>
+      {() => <SmartCollectionScreen smartCollection={UNDER_THE_RADAR} />}
```
dylanjeffers, line 78: yeah good catch here, wondering if we just create an under-the-radar screen maybe?

packages/mobile/src/store/common/reducer.ts
```diff
@@ -0,0 +1,19 @@
+import { CommonState } from 'audius-client/src/common/store'
+
+import { CommonActions, RECEIVE } from './actions'
+
+// Not using reduxjs/toolkit here because immer performance was less than optimal
+// for this reducer that gets hit very often
+export const reducer = (
```
dylanjeffers, line 7: from what ive seen, its still standard to do default export for reducer, but def a nit

dylanjeffers (approved):
awesome job, its much cleaner now too

packages/mobile/src/components/app-navigator/BottomTabNavigator.tsx
sliptype: Probably a good idea, we will need to create them for each smart collection though

packages/mobile/src/store/common/reducer.ts
sliptype: I say we do named exports across the board

PR #1018, opened 2022-03-04
[AUD-1642]: Reduce web-app rerenders

dylanjeffers:
### Description

Reduces web-app rerenders to improve native app performance
- Prevents web modals from rendering, which reduces rerenders by 50%, and total dom elements by 60%
- Reduces profile-page specific rerenders by 50% using createDeepEqualSelector
- improves stems reducer
- memozies feed config which saves a render

### Dragons
I removed a console.warn which was always triggering since we wait for data to arrive. what do you think we should do @sliptype ?

### DIFF
--- a/packages/web/src/common/store/account/selectors.ts
+++ b/packages/web/src/common/store/account/selectors.ts
@@ -3,7 +3,6 @@ import { createSelector } from 'reselect'
 import { CommonState } from 'common/store'
 import { getCollections } from 'common/store/cache/collections/selectors'
 import { getUser, getUsers } from 'common/store/cache/users/selectors'
-import { createDeepEqualSelector } from 'common/utils/selectorHelpers'
 import { removeNullable } from 'common/utils/typeUtils'
 
 import { AccountCollection } from './reducer'
@@ -30,11 +29,9 @@ export const getAccountToCache = (state: CommonState) => ({
   hasFavoritedItem: state.account.hasFavoritedItem
 })
 
-export const getAccountUser = createDeepEqualSelector(
+export const getAccountUser = createSelector(
   [internalGetAccountUser],
-  user => {
-    return user
-  }
+  user => user
 )
 export const getUserHandle = createSelector([internalGetAccountUser], user =>
   user ? user.handle : null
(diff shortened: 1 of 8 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-default-equality-selector

audius-infra:
Preview this change https://demo.audius.co/dj-default-equality-selector

nicoback2 (approved):
🔥

packages/web/src/common/store/account/selectors.ts
```diff
@@ -29,9 +30,11 @@ export const getAccountToCache = (state: CommonState) => ({
   hasFavoritedItem: state.account.hasFavoritedItem
 })
 
-export const getAccountUser = createSelector(
+export const getAccountUser = createDeepEqualSelector(
```
sliptype, line 33: Is this needed because the user object reference changes?

packages/web/src/pages/Modals.tsx
```diff
@@ -25,11 +25,15 @@ import RewardsModals from 'pages/audio-rewards-page/components/modals/RewardsMod
...
 
 const NATIVE_MOBILE = process.env.REACT_APP_NATIVE_MOBILE
+const NATIVE_NAVIGATION_ENABLED =
+  process.env.REACT_APP_NATIVE_NAVIGATION_ENABLED === 'true'
 
 const Modals = () => {
   const client = getClient()
   const isMobileClient = client === Client.MOBILE
 
+  if (NATIVE_NAVIGATION_ENABLED) return null
```
sliptype, line 35: This could be risky right? If a modal has data fetching effects in it for example

sliptype (approved):
I'm a little worried about not rendering the modals but let me know your thoughts!

packages/web/src/pages/Modals.tsx
dylanjeffers: yeah thats what im wondering as well... cause none of the modals have an actual route associated so i thought maybe wed be fine. but yeah maybe we just see if suddenly we are missing important data on the native modals?

sliptype: Yeah let's keep an eye on it, should be caught in qa if there is an issue

packages/web/src/common/store/account/selectors.ts
dylanjeffers: i think youre right on this one, can double check

PR #1019, opened 2022-03-04
[AUD-972] Finish $AUDIO mobile page

raymondjacobson:
### Description

Finishes /audio page on mobile. This is unf a big PR because there was a lot to wire up.
Notably I did not also solve for the fact that we have two JSON objects (one in mobile and one in web that represent challenges. I felt that was asking for too much in this PR). Things are somewhat more consolidated here though, so we can continue to refactor.

### Dragons

It's all new basically, but touched a few web audio page things to refactor.

### How Has This Been Tested?

https://user-images.githubusercontent.com/2731362/156687250-cae5eeb2-2b01-4e91-9d60-2d832c0a2317.mp4

Mobile staging release

### DIFF
--- a/packages/mobile/src/components/challenge-rewards-drawer/ChallengeRewardsDrawerProvider.tsx
+++ b/packages/mobile/src/components/challenge-rewards-drawer/ChallengeRewardsDrawerProvider.tsx
@@ -10,35 +10,19 @@ import {
   getClaimStatus
 } from 'audius-client/src/common/store/pages/audio-rewards/selectors'
 import {
-  ChallengeRewardsModalType,
   claimChallengeReward,
   ClaimStatus,
   resetAndCancelClaimReward
 } from 'audius-client/src/common/store/pages/audio-rewards/slice'
 import { setVisibility } from 'audius-client/src/common/store/ui/modals/slice'
 import { Maybe } from 'audius-client/src/common/utils/typeUtils'
-import {
-  ACCOUNT_VERIFICATION_SETTINGS_PAGE,
-  AUDIO_PAGE,
-  TRENDING_PAGE,
-  UPLOAD_PAGE
-} from 'audius-client/src/utils/route'
-import { isEqual } from 'lodash'
-import { ImageSourcePropType } from 'react-native'
+import { AUDIO_PAGE } from 'audius-client/src/utils/route'
 
-import Headphone from 'app/assets/images/emojis/headphone.png'
-import IncomingEnvelope from 'app/assets/images/emojis/incoming-envelope.png'
-import LoveLetter from 'app/assets/images/emojis/love-letter.png'
-import MobilePhoneWithArrow from 'app/assets/images/emojis/mobile-phone-with-arrow.png'
-import MultipleMusicalNotes from 'app/assets/images/emojis/multiple-musical-notes.png'
-import WhiteHeavyCheckMark from 'app/assets/images/emojis/white-heavy-check-mark.png'
-import IconArrow from 'app/assets/images/iconArrow.svg'
-import IconCheck from 'app/assets/images/iconCheck.svg'
-import IconUpload from 'app/assets/images/iconUpload.svg'
 import { useDispatchWeb } from 'app/hooks/useDispatchWeb'
 import { usePushRouteWeb } from 'app/hooks/usePushRouteWeb'
 import { useRemoteVar } from 'app/hooks/useRemoteConfig'
 import { useSelectorWeb } from 'app/hooks/useSelectorWeb'
+import { challengesConfig } from 'app/utils/challenges'
 
 import Button, { ButtonType } from '../button'
 import { useDrawerState } from '../drawer/AppDrawer'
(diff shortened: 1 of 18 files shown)


raymondjacobson:
Noted that the SegmentedView on the trending drawers is whack, but those were already there

audius-infra:
Preview this change https://demo.audius.co/rj-aud-972-finish

packages/mobile/src/utils/challenges.tsx
```diff
@@ -0,0 +1,252 @@
...
+import LoveLetter from 'app/assets/images/emojis/love-letter.png'
+import MobilePhoneWithArrow from 'app/assets/images/emojis/mobile-phone-with-arrow.png'
+import MultipleMusicalNotes from 'app/assets/images/emojis/multiple-musical-notes.png'
+import NerdFace from 'app/assets/images/emojis/nerd-face.png'
+import WhiteHeavyCheckMark from 'app/assets/images/emojis/white-heavy-check-mark.png'
+import IconArrow from 'app/assets/images/iconArrow.svg'
+import IconCheck from 'app/assets/images/iconCheck.svg'
+import IconUpload from 'app/assets/images/iconUpload.svg'
+
+// TODO: These should be programmatic based on amount, but historically have not been
```
rickyrombo, line 24: The `description` (and `fullDescription`) is programmatic based on amount on web - we should pull that in here as well too. Mobile didn't have it prior for the drawers and I think we can probably leave it that way for now since it uses `fullDescription` which hardly references amount and the actual amount is used in the same modal, but since the audio page uses `description` and is the first thing seen I think it'd be good to make sure it's always accurate

packages/mobile/src/utils/challenges.tsx
```diff
@@ -0,0 +1,252 @@
+import {
```
rickyrombo, line 1: This really doesn't feel like a "util" but I saw your PR description - probably ok to leave for now and move to "common" somewhere later?

Semi relatedly, could also see grouping "challenges page" stuff (like the config here and the drawers too) in the audio-screen folder? It just seems too far away/spread out from where it's used/useful

packages/web/src/pages/audio-rewards-page/AudioRewardsPage.tsx
```diff
@@ -28,7 +28,7 @@ import ExplainerTile from './components/ExplainerTile'
 
 export const messages = {
   title: '$AUDIO & Rewards',
-  description: 'View important stats like plays, reposts, and more.'
+  description: 'Complete tasks to earn $AUDIO tokens!'
```
rickyrombo, line 31: 😨  nice catch

packages/mobile/src/components/challenge-rewards-drawer/ChallengeRewardsDrawerProvider.tsx
```diff
@@ -49,146 +33,18 @@ import { ProfileCompletionChecks } from './ProfileCompletionChecks'
...
-    }
-  }
-}
 const styles = {
   button: {
     width: '100%'
   }
 }
 
-const renderUploadIcon = color => <IconUpload fill={color} />
```
rickyrombo, line 190: oops, that was redundant - thanks!

rickyrombo (approved):

packages/mobile/src/components/challenge-rewards-drawer/ChallengeRewardsDrawerProvider.tsx
```diff
@@ -10,35 +10,19 @@ import {
...
-import IconArrow from 'app/assets/images/iconArrow.svg'
-import IconCheck from 'app/assets/images/iconCheck.svg'
-import IconUpload from 'app/assets/images/iconUpload.svg'
 import { useDispatchWeb } from 'app/hooks/useDispatchWeb'
 import { usePushRouteWeb } from 'app/hooks/usePushRouteWeb'
 import { useRemoteVar } from 'app/hooks/useRemoteConfig'
 import { useSelectorWeb } from 'app/hooks/useSelectorWeb'
+import { challengesConfig } from 'app/utils/challenges'
 
 import Button, { ButtonType } from '../button'
```
dylanjeffers, line 27: would it be too much of a lift to us the core button?

dylanjeffers (approved):
wow amazing job!

packages/mobile/src/components/challenge-rewards-drawer/ChallengeRewardsDrawerProvider.tsx
raymondjacobson: I can do this as a sep PR, this isn't new here!

packages/mobile/src/utils/challenges.tsx
raymondjacobson: I think this requires a bit heavier lift b/c we currently include those descriptions in the objects that contain the web icons. I want to make a separate change where we consolidate all of these into one shared object.

packages/mobile/src/utils/challenges.tsx
raymondjacobson: the drawer gets mounted at the root level though, so it feels kind of weird for it to depend on audio-screen for me. I think a shared web + mobile config needs to go in common/challenges/config.ts, potentially as a class where mobile + web can construct it with ReactNode's for the corresponding/correct images

audius-infra:
Preview this change https://demo.audius.co/rj-aud-972-finish

packages/mobile/src/utils/challenges.tsx
rickyrombo: Yeah def will take more work to consolidate - I was thinking literal copy/paste to use functions for description/fullDescription instead of importing for now until we break out into a consolidated approach?

raymondjacobson: Yeah that's fair. I can submit a subsequent PR. I just figured that because drawers already had hard-coded amts on mobile, this was a ~wash

rickyrombo: Yeah, true. I think they have the benefit of having the true amount in big purple letters right next to it, so even if the text gets out of date it's pretty obvious what it is. Also I think every challenge uses the amount in the normal description vs the full description only mentions it for a couple

PR #1022, opened 2022-03-04
Fix settings-page when null-rendering

dylanjeffers:
### Description

Fixes two issues when loading settings page using null-renderer:
 - React.lazy seems to not work
 - refs will sometimes not be attached and their values will be null.

### Dragons

We may have a future issue where we can't use react.lazy in places that we would like. The fix here will need to be build-time webpack code-mod to either use `React.lazy(....)` or `require(...).default` https://webpack.js.org/plugins/define-plugin/

### DIFF
--- a/packages/web/src/components/data-entry/TabSlider.js
+++ b/packages/web/src/components/data-entry/TabSlider.js
@@ -31,11 +31,12 @@ const TabSlider = props => {
       selectedRefIdx = 0
     }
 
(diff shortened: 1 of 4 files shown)


sliptype (approved):
Dope!

audius-infra:
Preview this change https://demo.audius.co/dj-null-renderer-fixes

PR #1024, opened 2022-03-05
[AUD-732] Finish react native scrubber

raymondjacobson:
### Description

Finishes connecting the scrubber.
There are probably a couple bugs remaining that have to do with timing/etc. Will keep an eye on them, but wanted to get this out first!

At a high level --
1. Move scrubber int components/scrubber
2. Break scrubber into `<Scrubber>` and `<Slider>`, the later of which does not contain the timestamp start & timestamp end bit because we need to memoize the component that does the gesture stuff.

### Dragons

### How Has This Been Tested?

https://user-images.githubusercontent.com/2731362/156902320-b9ea6c30-e3c7-4adc-a0b5-afb71da1a981.mp4


packages/mobile/src/components/now-playing-drawer/PlayBar.tsx
```diff
@@ -134,9 +134,9 @@ export const PlayBar = ({
...
   useEffect(() => {
     setInterval(() => {
-      const { currentTime, playableDuration } = global.progress
-      if (playableDuration !== undefined) {
-        setPercentComplete(currentTime / playableDuration)
+      const { currentTime, seekableDuration } = global.progress
+      if (seekableDuration !== undefined) {
+        setPercentComplete(currentTime / seekableDuration)
       } else {
         setPercentComplete(0)
```
dylanjeffers, line 141: should we have a cleanup effect?

packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
```diff
@@ -219,10 +236,16 @@ const NowPlayingDrawer = ({
               <TrackInfo track={track} user={user} />
             </View>
             <View style={styles.scrubberContainer}>
-              <Scrubber mediaKey='1' />
+              <Scrubber
+                mediaKey={`${mediaKey}`}
+                isPlaying={isPlaying}
+                onPressIn={() => setIsGestureEnabled(false)}
```
dylanjeffers, line 242: if we make these callbacks, i think scrubber then doesn't need to be memoized?

packages/mobile/src/components/scrubber/Scrubber.tsx
```diff
@@ -0,0 +1,169 @@
...
+      },
+      [onPressOut, setIsDragging, dispatch, duration]
+    )
+
+    // Register an interval to update the start / end timestamps (duration)
+    const seekInterval = useRef<NodeJS.Timeout | null>(null)
+
+    const onDrag = useCallback(
+      percentComplete => {
+        if (global.progress) {
```
dylanjeffers, line 119: if with an && maybe?

packages/mobile/src/components/scrubber/Scrubber.tsx
```diff
@@ -0,0 +1,169 @@
...
+        }
+      },
+      [duration, setDragSeconds]
+    )
+
+    useEffect(() => {
+      setDragSeconds(formatSeconds(0))
+      setTimeout(() => {
+        setDragSeconds(null)
+      }, SCRUB_RELEASE_TIMEOUT_MS)
```
dylanjeffers, line 132: cleanup timeout?

packages/mobile/src/components/scrubber/Scrubber.tsx
```diff
@@ -0,0 +1,169 @@
...
+    }, [mediaKey])
+
+    useEffect(() => {
+      if (seekInterval.current) clearInterval(seekInterval.current)
+      if (!isDragging) {
+        seekInterval.current = setInterval(() => {
```
dylanjeffers, line 138: when it comes to setting a ref with an interval, do we just need a cleanup effect at the end here that simply returns clearInterval(seekInterval.current)?

packages/mobile/src/components/scrubber/Slider.tsx
```diff
@@ -0,0 +1,364 @@
...
+        width: 0,
+        height: 1
+      },
+      shadowOpacity: 0.5,
+      shadowRadius: 2,
+      elevation: 3
+    }
+  })
+
+const attachToDx = (animation: Animated.Value, newValue: number) => (
```
dylanjeffers, line 69: may make sense to share this with drawer and other places that might use it?

packages/mobile/src/components/scrubber/Slider.tsx
```diff
@@ -0,0 +1,364 @@
...
+          toValue: railWidth,
+          duration: timeRemaining,
+          useNativeDriver: true
+        })
+        currentAnimation.current.start()
+      },
+      [translationAnim, railWidth]
+    )
+
+    const pause = useCallback(() => {
```
dylanjeffers, line 170: onPause and onPlay maybe?

dylanjeffers (approved):
really good job, i learned a lot reading this pr

packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
raymondjacobson: Yep you're right. Only was memoizing b/c the Slider used to be the same component. Ty

packages/mobile/src/components/scrubber/Slider.tsx
raymondjacobson: This isn't a responder though -- it's the actual verb. Calling this fn actually triggers the "playing" or "pausing" of the animation. Wdyt?

dylanjeffers: ah word yeah down with it!

packages/mobile/src/components/scrubber/Scrubber.tsx
raymondjacobson: I don't think this is strictly necessary b/c it's not an interval and it's called on an action versus passively in an effect. I think the better solution would be a throttle/debounce, but i don't think it's warranted right now

dylanjeffers: oh yeah timeout not interval, nice sgsg

PR #1026, opened 2022-03-07
[AUD-1646] Various native fixes

dylanjeffers:
### Description

Adds fixes for a few native issues in prep for beta-testing

- Fixes drawer onClose animation. Before drawers would prematurely unrender causing a sudden jump in content.
- Fixes trending lineup loading tiles, which were missing
- Fixes blank cover-photo for profiles
- Fixes button heights so they stay the same when user clicks. Before the content above and below them would shift when user clicks
- Fixes rewards header icon size to be closer to existing size
- Fixes upload-track button spacing
- Fixes card spacing and height

### DIFF
--- a/packages/mobile/src/components/core/Button.tsx
+++ b/packages/mobile/src/components/core/Button.tsx
@@ -245,9 +245,7 @@ export const Button = (props: ButtonProps) => {
     [onPressOut, handlePressOutScale, handlePressOutColor]
   )
 
-  /*
-   * Ensures button take up a static height even when scaling
-   */
+  // Ensures button takes up a static height even when scaling
   const handleRootLayout = useCallback(
     (event: LayoutChangeEvent) => {
       rootHeightRef.current = event.nativeEvent.layout.height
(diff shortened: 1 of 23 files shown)


packages/mobile/src/components/core/Button.tsx
```diff
@@ -243,6 +245,16 @@ export const Button = (props: ButtonProps) => {
     [onPressOut, handlePressOutScale, handlePressOutColor]
   )
 
+  /*
+   * Ensures button take up a static height even when scaling
+   */
+  const handleRootLayout = useCallback(
```
sliptype, line 251: Good catch, I wonder if this could/should be incorporated in the `usePressScaleAnimation` hook?

sliptype (approved):

packages/mobile/src/components/core/Button.tsx
dylanjeffers: yeah i wonder as well, how would that hook figure out the height of the root element i guess?

PR #1027, opened 2022-03-07
[AUD-1657] Improve native header perf

dylanjeffers:
### Description

Using a custom header has lead to performance issues when transitioning between screens. The reason for this has been tracked in react-navigation issues, however there are no workarounds from what ive seen. One solution is to use the react-navigation header, and simply customize it. This pr attempts to move all custom header components into the correct navigation `screenOptions` object. This has resulted in much better header perf!

Before:

https://user-images.githubusercontent.com/8230000/157105918-92cd2bdd-9e36-41e6-9972-eeb10b266833.mov

After: 

https://user-images.githubusercontent.com/8230000/157105934-b92ff15d-3c9c-4623-bec0-8a079e974c7a.mov


packages/mobile/src/components/app-navigator/BaseStackNavigator.tsx
```diff
@@ -41,21 +38,26 @@ export const BaseStackNavigator = ({
...
-        headerStyle: { height: 87 },
-        headerMode: 'float'
-      }}
-      screenListeners={({ navigation }) => ({
+      screenOptions={screenOptions}
+      screenListeners={() => ({
         beforeRemove: e => {
           // hack for now to prevent pop for some pages
-          if (!e.target?.includes('EditProfile')) {
+          console.log(
```
sliptype, line 48: Log

sliptype: (and one below)

packages/mobile/src/components/core/Screen.tsx
```diff
@@ -6,6 +6,13 @@ import { Animated, StyleProp, View, ViewStyle } from 'react-native'
 
 import { makeStyles } from 'app/styles'
 
+const removeFalsy = (obj: Record<string, any>) =>
```
sliptype, line 9: Lodash `pickBy` would be useful here. Also this is checking specifically for undefined, so the name could be more specific

sliptype (approved):
Very slick

packages/mobile/src/components/core/Screen.tsx
dylanjeffers: ah thank you! i was tryna figure out which lodash thing would work here. and totally agreed!

PR #1029, opened 2022-03-07
[AUD-1648] RN Scroll To Top

sliptype:
### Description

* Adds custom `useScrollToTop` hook that supports nested tab navigators (unlike the one provided by react-navigation)
* Triggers scroll to top on long tab press and focused tab press
* Does not scroll to top on profile tab page tab press

https://user-images.githubusercontent.com/19916043/157122124-7c97c617-bc55-478e-acab-e483c3d11f7e.mov

### Dragons

Attaching listeners to a few events, they get cleaned up but could have slight performance/memory impact

### How Has This Been Tested?

iOS simulator

Not released yet

### DIFF
--- a/packages/mobile/src/components/animated-button/AnimatedButtonProvider.tsx
+++ b/packages/mobile/src/components/animated-button/AnimatedButtonProvider.tsx
@@ -10,6 +10,7 @@ import { useColor } from 'app/utils/theme'
 export type BaseAnimatedButtonProps = {
   iconIndex?: number
   onPress?: GestureResponderHandler
+  onLongPress?: GestureResponderHandler
   isActive?: boolean
   isDisabled?: boolean
   showUnderlay?: boolean
@@ -27,6 +28,7 @@ const AnimatedButton = ({
   iconIndex: externalIconIndex,
   iconJSON,
   onPress,
+  onLongPress,
   isActive,
   isDisabled = false,
   showUnderlay = false,
@@ -81,25 +83,23 @@ const AnimatedButton = ({
     externalIconIndex
   ])
 
-  const handleClick = useCallback(() => {
-    if (isDisabled) {
-      return
-    }
-
+  const handlePress = useCallback(() => {
     if (hasMultipleStates || !isActive) {
       setIsPlaying(true)
       animationRef.current?.play()
     }
 
     onPress?.()
-  }, [
-    isDisabled,
-    onPress,
-    isActive,
-    setIsPlaying,
-    hasMultipleStates,
-    animationRef
-  ])
+  }, [onPress, isActive, setIsPlaying, hasMultipleStates, animationRef])
+
+  const handleLongPress = useCallback(() => {
+    if (onLongPress) {
+      onLongPress()
+      return
+    }
+
+    handlePress()
+  }, [onLongPress, handlePress])
 
   // For multi state buttons, when `isActive` flips, trigger
   // the animation to run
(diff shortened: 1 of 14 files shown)


packages/mobile/src/components/animated-button/AnimatedButtonProvider.tsx
```diff
@@ -81,25 +83,23 @@ const AnimatedButton = ({
...
-    setIsPlaying,
-    hasMultipleStates,
-    animationRef
-  ])
+  }, [onPress, isActive, setIsPlaying, hasMultipleStates, animationRef])
+
+  const handleLongPress = useCallback(() => {
+    if (onLongPress) {
+      onLongPress()
+      return
```
dylanjeffers, line 98: maybe if/else?

packages/mobile/src/hooks/useScrollToTop.ts
```diff
@@ -0,0 +1,62 @@
...
+ * see: https://github.com/react-navigation/react-navigation/issues/8586
+ */
+export const useScrollToTop = (
+  scrollToTop: () => void,
+  disableTopTabScroll = false
+) => {
+  const navigation = useNavigation()
+
+  useFocusEffect(
+    useCallback(() => {
```
dylanjeffers, line 23: can we nest hooks like that?

packages/mobile/src/screens/explore-screen/tabs/ForYouTab/ForYouTab.tsx
```diff
@@ -60,12 +63,20 @@ const tiles = [
...
+  useScrollToTop(() => {
+    ref.current?.scrollTo({
+      y: 0,
+      animated: true
+    })
+  })
+
   return (
-    <ScrollView style={styles.tabContainer}>
+    <ScrollView style={styles.tabContainer} ref={ref}>
```
dylanjeffers, line 75: would it make sense to have special scroll to top flatlist/scrollview components at some point that have useScrollToTop baked in?

packages/mobile/src/screens/explore-screen/tabs/ForYouTab/ForYouTab.tsx
```diff
@@ -60,12 +63,20 @@ const tiles = [
...
       <TabInfo header={messages.infoHeader} text={messages.infoText} />
       <View style={styles.contentContainer}>
         {tiles.map(tile => (
-          <ColorTile style={styles.tile} key={tile.variant} {...tile} />
+          <ColorTile style={styles.tile} key={tile.title} {...tile} />
```
dylanjeffers, line 79: nice find

dylanjeffers (approved):
really awesome work

packages/mobile/src/hooks/useScrollToTop.ts
sliptype: Yeah! The execution order isn't changing, this is essentially like assigning the result of `useCallback` to a a variable and passing that to `useFocusEffect` so it's ok

dylanjeffers: ah word, yeah that makes sense

packages/mobile/src/screens/explore-screen/tabs/ForYouTab/ForYouTab.tsx
sliptype: Yeah I was going back and forth on this. We would need one for flatlist, section list, and scroll view. And then we don't necessarily want every instance to support scroll to top. So i opted for just using the hook in relevant places. But I could be convinced to make custom components for sure

PR #1034, opened 2022-03-08
Implement expanding/collapsing folders

nicoback2:
### Description
-Show/hide playlist folder contents
expand

### Dragons
I refactored the playlist library component a bit to enable the library contents nesting - need to make sure this doesn't break anything. I tested things pretty thoroughly on staging (no playlists/folders, playlists with updates, playlists/folders with long names, etc) but wasn't quite sure how to test the `playlistsNotInLibrary` functionality (will comment in PR).

### How Has This Been Tested?
Tested in browser using fake data for folder contents

### DIFF
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -83,6 +83,7 @@
     "react-twitter-auth": "0.0.13",
     "react-twitter-embed": "3.0.3",
     "react-use-gesture": "7.0.0-beta.5",
+    "react-use-measure": "^2.1.1",
     "redux": "4.0.5",
     "redux-devtools-extension": "2.13.8",
     "redux-saga": "1.1.3",
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -83,7 +83,7 @@
     "react-twitter-auth": "0.0.13",
     "react-twitter-embed": "3.0.3",
     "react-use-gesture": "7.0.0-beta.5",
-    "react-use-measure": "^2.1.1",
+    "react-use-measure": "2.1.1",
     "redux": "4.0.5",
     "redux-devtools-extension": "2.13.8",
     "redux-saga": "1.1.3",
--- a/packages/web/src/components/nav/desktop/PlaylistFolderNavItem.tsx
+++ b/packages/web/src/components/nav/desktop/PlaylistFolderNavItem.tsx
@@ -7,7 +7,8 @@ import {
   IconKebabHorizontal
 } from '@audius/stems'
 import cn from 'classnames'
-import { isEmpty } from 'lodash'
+import { useSpring, animated } from 'react-spring'
+import useMeasure from 'react-use-measure'
 
 import { ID } from 'common/models/Identifiers'
 import { PlaylistLibraryFolder } from 'common/models/PlaylistLibrary'
@@ -68,25 +69,36 @@ const FolderNavLink = ({
   )
 }
 
-export const PlaylistFolderNavItem = ({
-  folder,
-  hasUpdate = false,
-  dragging,
-  draggingKind,
-  onClickEdit
-}: {
+type PlaylistFolderNavItemProps = {
   folder: PlaylistLibraryFolder
   hasUpdate: boolean
   dragging: boolean
   draggingKind: string
   onClickEdit: (folderId: string) => void
-}) => {
-  const { id, name, contents } = folder
+  children?: React.ReactNode
+}
+
+export const PlaylistFolderNavItem = ({
+  folder,
+  hasUpdate = false,
+  dragging,
+  draggingKind,
+  onClickEdit,
+  children
+}: PlaylistFolderNavItemProps) => {
+  const { id, name } = folder
   const isDroppableKind =
     draggingKind === 'track' ||
     draggingKind === 'playlist' ||
     draggingKind === 'playlist-folder'
   const [isHovering, setIsHovering] = useState(false)
+  const [isExpanded, setIsExpanded] = useState(false)
+  const [ref, bounds] = useMeasure()
+  const contentsStyle = useSpring({
+    height: isExpanded ? bounds.height : 0,
+    opacity: isExpanded ? 100 : 0,
+    overflow: 'hidden'
+  })
 
   return (
     <Droppable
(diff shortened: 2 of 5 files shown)


packages/web/src/components/nav/desktop/PlaylistLibrary.tsx
```diff
@@ -39,6 +43,54 @@ type PlaylistLibraryProps = {
...
+          case 'playlist': {
+            const playlist = playlists[content.playlist_id]
+            return renderPlaylist(playlist)
+          }
+          case 'temp_playlist': {
+            try {
+              const playlist = playlists[parseInt(content.playlist_id)]
+              return renderPlaylist(playlist)
+            } catch (e) {
+              console.debug(e)
```
nicoback2, line 81: kept this since it was here before but not sure why?

packages/web/src/components/nav/desktop/PlaylistLibrary.tsx
```diff
@@ -149,10 +188,67 @@ const PlaylistLibrary = ({
...
+      )
+    },
+    [record, onClickNavLinkWithAccount]
+  )
+
   // Iterate over playlist library and render out available explore/smart
   // playlists and ordered playlists. Remaining playlists that are unordered
   // are rendered aftewards by sort order.
-  const playlistsNotInLibrary = { ...playlists }
+  const playlistsNotInLibrary = useMemo(() => {
```
nicoback2, line 232: Would appreciate a closer look at this chunk in particular - I'm not sure what situation this was originally put here to address?

audius-infra:
Preview this change https://demo.audius.co/nkang--expand-folder

packages/web/src/components/nav/desktop/PlaylistFolderNavItem.tsx
```diff
@@ -68,25 +69,36 @@ const FolderNavLink = ({
...
     draggingKind === 'playlist' ||
     draggingKind === 'playlist-folder'
   const [isHovering, setIsHovering] = useState(false)
+  const [isExpanded, setIsExpanded] = useState(false)
+  const [ref, bounds] = useMeasure()
+  const contentsStyle = useSpring({
+    height: isExpanded ? bounds.height : 0,
+    opacity: isExpanded ? 100 : 0,
+    overflow: 'hidden'
+  })
```
nicoback2, line 101: Planning on making a `<Collapse>` utility component for Stems using this pattern and then will just use that component here https://linear.app/audius/issue/AUD-1664/add-collapse-utility-component

audius-infra:
Preview this change https://demo.audius.co/nkang--expand-folder

packages/web/package.json
```diff
@@ -83,6 +83,7 @@
     "react-twitter-auth": "0.0.13",
     "react-twitter-embed": "3.0.3",
     "react-use-gesture": "7.0.0-beta.5",
+    "react-use-measure": "^2.1.1",
```
raymondjacobson, line 86: pin dep pls!

packages/web/src/components/nav/desktop/PlaylistLibrary.tsx
```diff
@@ -42,6 +44,47 @@ type PlaylistLibraryProps = {
...
 }
 
+type LibraryContentsLevelProps = {
+  contents: PlaylistLibraryType['contents']
+  renderPlaylist: (playlistId: number) => void
+  renderExplorePlaylist: (playlistId: SmartCollectionVariant) => void
+  renderFolder: (folder: PlaylistLibraryFolder) => void
+}
+
+// Function component for rendering a single level of the playlist library. Playlist library consists of up to two content levels (root + inside a folder)
```
raymondjacobson, line 54: super ultra nit, but let's do component comments in format

```
/*
 *
 */
```

https://www.notion.so/audiusproject/Client-Javascript-Styleguide-4f34640139fc4a809c3765131bc597ff#5ec92a9e0f9741a8a61e5b71d528b74c

packages/web/src/components/nav/desktop/PlaylistLibrary.tsx
raymondjacobson: I don't think this actually could ever throw given how it's returned. parseInt just returns NaN and doesn't throw. I think save to remove.

packages/web/src/components/nav/desktop/PlaylistLibrary.tsx
```diff
@@ -128,8 +173,8 @@ const PlaylistLibrary = ({
     },
     [record, onClickNavLinkWithAccount]
   )
-
-  const renderPlaylist = (playlist: AccountCollection) => {
+  const renderPlaylist = (playlistId: number) => {
```
raymondjacobson, line 176: we have an `ID` type (alias) in common/models

packages/web/src/components/nav/desktop/PlaylistLibrary.tsx
```diff
@@ -157,10 +202,56 @@ const PlaylistLibrary = ({
...
+            />
+          </div>
+        )}
+      </PlaylistFolderNavItem>
+    )
+  }
+
+  const playlistsNotInLibrary = useMemo(() => {
+    const result = { ...playlists }
+    function helpComputePlaylistsNotInLibrary(
```
raymondjacobson, line 231: can we do anon function style? `const helpCompute... = () => {`

packages/web/src/components/nav/desktop/PlaylistLibrary.tsx
raymondjacobson: If for some reason you have a playlist that is not in your library, we do want to render it in the left nav, so the point of this was to make sure that it shows up (maybe a write failed, etc., we still want to be absolutely sure you see all your playlists). I think this recursion makes sense, just removing from the library as we go. should be very cheap for the normal user. Maybe a function comment explaining that intent would be good. (Or pulling this into a helper w/ the other ones and adding a test case)

raymondjacobson (approved):

audius-infra:
Preview this change https://demo.audius.co/nkang--expand-folder

PR #1037, opened 2022-03-08
[AUD-1667] Improve user-list performance

dylanjeffers:
### Description

- Improves user-list performance by reducing rerenders and caching user-lists between reloads
- Fixes follow-button follow source
- Removes unused components

Also addresses [AUD-1662]

Before:

https://user-images.githubusercontent.com/8230000/157311532-abf794b3-1cb7-4d47-9548-11f09958276c.mov

After: 

https://user-images.githubusercontent.com/8230000/157312213-3a98f960-b3ba-4f6e-a6b6-9fb694038f69.mov

### DIFF
--- a/packages/mobile/src/screens/user-list-screen/UserList.tsx
+++ b/packages/mobile/src/screens/user-list-screen/UserList.tsx
@@ -1,9 +1,13 @@
-import { useCallback, useEffect, useMemo, useRef } from 'react'
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 
+import { useFocusEffect, useIsFocused } from '@react-navigation/native'
 import { CommonState } from 'audius-client/src/common/store'
 import { getUserId } from 'audius-client/src/common/store/account/selectors'
 import { getUsers } from 'audius-client/src/common/store/cache/users/selectors'
-import { loadMore } from 'audius-client/src/common/store/user-list/actions'
+import {
+  loadMore,
+  setLoading
+} from 'audius-client/src/common/store/user-list/actions'
 import { UserListStoreState } from 'audius-client/src/common/store/user-list/types'
 import { isEqual } from 'lodash'
 import { FlatList } from 'react-native'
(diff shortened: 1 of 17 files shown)


packages/mobile/src/screens/user-list-screen/UserList.tsx
```diff
@@ -47,16 +59,26 @@ export const UserList = (props: UserListProps) => {
...
+  }, [users])
+
   const handleEndReached = useCallback(() => {
     if (hasMore) {
       dispatchWeb(loadMore(tag))
     }
   }, [hasMore, dispatchWeb, tag])
 
+  if (loading && cachedUsers.current.length === 0) {
+    return <LoadingSpinner style={styles.spinner} />
```
raymondjacobson, line 75: ty. does the loading spinner show at the bottom now too? i think that might also be missing

packages/mobile/src/screens/user-list-screen/UserList.tsx
```diff
@@ -47,16 +59,26 @@ export const UserList = (props: UserListProps) => {
     [usersMap, userIds]
   )
 
+  useEffect(() => {
+    if (users.length !== 0) {
+      cachedUsers.current = users
```
raymondjacobson, line 64: this cache doesn't persist b/w "types" of user list screens right? I think it might. If you go from track => reposters => track => favorites

sliptype (approved):
The After video didn't load but I already saw it in action :)

raymondjacobson (approved):

packages/mobile/src/screens/user-list-screen/UserList.tsx
dylanjeffers: i don't think so cause all each screen renders their own `UserList`!

raymondjacobson: what if you go to a different track's userlist modal for reposters?

packages/mobile/src/screens/user-list-screen/UserList.tsx
dylanjeffers: ahh like a loading spinner at the bottom when we have content, and are loading more? can def add

raymondjacobson: yep! we do that on mobile web currently (or at least we should), and definitely on desktop web

PR #1039, opened 2022-03-09
[AUD-1681] Add profile-screen pull-to-refresh

dylanjeffers:
### Description

Adds pull to refresh to profile screen

### Dragons

We are getting the same error that exists on current app, which is a pull-to-refresh results in empty profile picture and cover art:

https://user-images.githubusercontent.com/8230000/157353282-2c642ab6-81d0-4065-801e-74fc662aa9c4.mov

### DIFF
--- a/packages/mobile/src/components/core/VirtualizedScrollView.tsx
+++ b/packages/mobile/src/components/core/VirtualizedScrollView.tsx
@@ -1,22 +1,26 @@
 import { ReactElement, useRef } from 'react'
 
-import { StyleProp, ViewStyle, FlatList } from 'react-native'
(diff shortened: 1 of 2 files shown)


raymondjacobson (approved):
Nice! This is definitely a step towards it, but we do have a fairly complex / nice expanding coverphoto fade operation that we should try to nail with react native. Good to merge this for now though.

dylanjeffers:
> Nice! This is definitely a step towards it, but we do have a fairly complex / nice expanding coverphoto fade operation that we should try to nail with react native. Good to merge this for now though.

yeah def agreed! this was just a first pass, cause i know there is a spike to figure out custom pull-to-refresh with RN for feed as well

PR #1041, opened 2022-03-09
[AUD-1669] Add tile press animation

dylanjeffers:
### Description

Adds press animation when tile is pressable.

### Dragons

This affects all tiles that are pressable. It seems like a fair assumption that any tile with onPress should have an animation, though.

### DIFF
--- a/packages/mobile/src/components/core/Tile.tsx
+++ b/packages/mobile/src/components/core/Tile.tsx
@@ -39,6 +39,7 @@ type TileOwnProps<
   TileComponentType extends ComponentType = ComponentType
 > = TilePressableProps & {
   children: ReactNode
+  scaleTo?: number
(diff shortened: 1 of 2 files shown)


packages/mobile/src/components/core/Tile.tsx
```diff
@@ -71,24 +74,52 @@ export const Tile = <
     ...other
   } = props
 
+  const {
+    scale,
+    handlePressIn: handlePressInScale,
+    handlePressOut: handlePressOutScale
+  } = usePressScaleAnimation()
```
raymondjacobson, line 81: Probably good to pass the `scale` prop all the way through from a Tile prop so that callers can customize. I'm sure we'll want to tweak it based on the size of some objects at some point in time

raymondjacobson (approved):

PR #1043, opened 2022-03-09
[AUD-1552, AUD-1555, AUD-1556, AUD-1557] Add smart collection screens

Kyle-Shanks:
### Description

* Updated some types and did some plumbing for the mood screens, collections screens, and smart collection screens
* Added smart collection screen to the BottomTabNavigator

[AUD-1552]
[AUD-1555]
[AUD-1556]
[AUD-1557]

### Dragons

N/A

### How Has This Been Tested?

Manually tested, all of the smart collection screens work and render tracks properly

N/A

### DIFF
--- a/packages/mobile/src/components/app-navigator/BottomTabNavigator.tsx
+++ b/packages/mobile/src/components/app-navigator/BottomTabNavigator.tsx
@@ -10,7 +10,13 @@ import NowPlayingDrawer from 'app/components/now-playing-drawer/NowPlayingDrawer
 import { AudioScreen } from 'app/screens/audio-screen'
 import { EditProfileScreen } from 'app/screens/edit-profile-screen'
 import ExploreScreen from 'app/screens/explore-screen'
-import { UNDER_THE_RADAR } from 'app/screens/explore-screen/smartCollections'
+import {
+  UNDER_THE_RADAR,
+  BEST_NEW_RELEASES,
+  REMIXABLES,
+  FEELING_LUCKY,
+  MOST_LOVED
+} from 'app/screens/explore-screen/smartCollections'
 import { TrendingUndergroundScreen } from 'app/screens/explore-screen/tabs/ForYouTab'
 import FavoritesScreen from 'app/screens/favorites-screen'
 import { FeedScreen } from 'app/screens/feed-screen'
@@ -33,6 +39,14 @@ import {
   ProfileStackParamList,
   TrendingStackParamList
 } from './types'
+
+const smartCollections = [
+  UNDER_THE_RADAR,
+  BEST_NEW_RELEASES,
+  REMIXABLES,
+  MOST_LOVED,
+  FEELING_LUCKY
+]
 
 const styles = StyleSheet.create({
   tabNavigator: {
@@ -75,9 +89,11 @@ const ExploreStackScreen = createStackScreen<ExploreStackParamList>(Stack => (
       name='TrendingUnderground'
       component={TrendingUndergroundScreen}
     />
-    <Stack.Screen name='UnderTheRadar'>
-      {() => <SmartCollectionScreen smartCollection={UNDER_THE_RADAR} />}
-    </Stack.Screen>
+    {smartCollections.map(collection => (
+      <Stack.Screen name={collection.screen} key={collection.screen}>
+        {() => <SmartCollectionScreen smartCollection={collection} />}
+      </Stack.Screen>
+    ))}
   </>
 ))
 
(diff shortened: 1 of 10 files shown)


packages/mobile/src/components/details-tile/DetailsTile.tsx
```diff
@@ -347,6 +326,22 @@ export const DetailsTile = ({
         >
           {renderDetailLabels()}
         </View>
+        <View style={styles.descriptionContainer}>
```
sliptype, line 329: This is gonna change the layout of the Tracks/Playlist screens as well. I talked to Julian about this and he said he was ok with keeping the details labels consistent, so I think we should continue rendering them below the description

sliptype (approved):
Awesome work! We're getting so close to having all the UI done now 🎉

packages/mobile/src/components/details-tile/DetailsTile.tsx
```diff
@@ -292,7 +281,9 @@ export const DetailsTile = ({
         ) : null}
         <View style={styles.buttonSection}>
           <Button
-            styles={{ text: styles.playButtonText }}
+            styles={{
```
dylanjeffers, line 284: ultra nit, can keep on on line :D

packages/mobile/src/components/details-tile/DetailsTile.tsx
```diff
@@ -347,6 +326,22 @@ export const DetailsTile = ({
         >
           {renderDetailLabels()}
         </View>
+        <View style={styles.descriptionContainer}>
+          {description ? (
+            <HyperLink
```
dylanjeffers, line 331: we have a new core hyperlink that comes with the correct styling and clicking external link now!

packages/mobile/src/components/track-list/TrackListItem.tsx
```diff
@@ -197,7 +197,12 @@ export const TrackListItem = ({
           />
         ) : isActive && !isDeleted ? (
           <View style={styles.playButtonContainer}>
-            <TablePlayButton playing paused={!isPlaying} hideDefault={false} />
+            <TablePlayButton
+              playing
+              paused={!isPlaying}
+              hideDefault={false}
+              onPress={onClickTrack}
```
dylanjeffers, line 204: prob not have "click" in the function name, so maybe onPressTrack or handlePressTrack?

packages/mobile/src/components/details-tile/DetailsTile.tsx
```diff
@@ -37,124 +29,121 @@ const messages = {
...
+
+  infoValue: {
+    flexShrink: 1,
+    lineHeight: 14,
+    color: palette.neutral,
+    fontSize: 14
+  },
+
+  infoIcon: {
+    marginTop: -spacing(1)
```
dylanjeffers, line 140: oh thats cool

packages/mobile/src/components/details-tile/DetailsTile.tsx
Kyle-Shanks: Ah right, I was supposed to check in to see why this was backwards but we were in standup. Will update 👍

packages/mobile/src/components/app-navigator/BottomTabNavigator.tsx
```diff
@@ -75,9 +89,11 @@ const ExploreStackScreen = createStackScreen<ExploreStackParamList>(Stack => (
       name='TrendingUnderground'
       component={TrendingUndergroundScreen}
     />
-    <Stack.Screen name='UnderTheRadar'>
-      {() => <SmartCollectionScreen smartCollection={UNDER_THE_RADAR} />}
-    </Stack.Screen>
+    {smartCollections.map(collection => (
+      <Stack.Screen name={collection.screen} key={collection.screen}>
+        {() => <SmartCollectionScreen smartCollection={collection} />}
```
dylanjeffers, line 94: so one thing im thinking we can do, is add `collection` to the `initialParams` to the SmartCollectionScreen, to avoid having to do a child render, so something like

```
<Stack.Screen name={name} key={name} component={SmartCollectionScreen} initialParams={{collection}} />
```
then we can capture those in the smart collection screen like 
```
const {params} = useRoute()
```
i suppose one of the problems here is that we don't know which route (screen name) we are on which is needed as a type arg to useRoute... but food for thought

dylanjeffers (approved):
great work!

dylanjeffers:
oh and @Kyle-Shanks lets update the pr title. I guess we can't put all the tickets in there, but maybe one? up to you

packages/mobile/src/components/details-tile/DetailsTile.tsx
Kyle-Shanks: I don't have an thing that I'm currently looking at to test this, but will update when doing deprecated component pass

dylanjeffers: word, and also fwiw i think its okay to keep deprecated component like text around if you want, cause i don't thing the new one has all the capabilities we want, just a headsup if it becomes difficult

PR #1044, opened 2022-03-09
[AUD-1603] Fix overflow menu nav, close now-playing drawer on navigate

dylanjeffers:
### Description

Fixes two issues:
1. Overflow menu `View Track/Collection Page` and `View Artist Page` were not hooked up with native navigation
2. When clicking overflow menu items that trigger navigation, the NowPlayingDrawer would not close

### Dragons

It is def a bit interesting the way this pr implements this fix. Curious of folks thoughts

### DIFF
--- a/packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
+++ b/packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
@@ -13,6 +13,7 @@ import {
 } from 'react-native'
 import { useSafeAreaInsets } from 'react-native-safe-area-context'
 import { useSelector } from 'react-redux'
+import { usePrevious } from 'react-use'
 
 import Drawer from 'app/components/drawer'
 import { Scrubber } from 'app/components/scrubber'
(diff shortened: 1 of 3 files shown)


sliptype (approved):
Makes sense to me! The closing of a drawer always kind of has 2 branches so it's good to see that reflected in the code now

packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
```diff
@@ -92,18 +95,31 @@ const NowPlayingDrawer = ({
...
-  const onDrawerClose = useCallback(() => {
+  const handleDrawerClose = useCallback(() => {
     springToValue(playBarOpacityAnim, 1, DrawerAnimationStyle.SPRINGY)
-    onClose()
     setIsPlayBarShowing(true)
     onCloseProp()
-  }, [setIsPlayBarShowing, onClose, onCloseProp, playBarOpacityAnim])
+  }, [setIsPlayBarShowing, onCloseProp, playBarOpacityAnim])
+
+  const handleDrawerCloseFromSwipe = useCallback(() => {
```
raymondjacobson, line 111: Nice! We haven't implemented clicking on the track artwork/title/artist name yet, so we'll re-use this fn there too

raymondjacobson (approved):

packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
raymondjacobson: Or rather, the other close version

PR #1045, opened 2022-03-09
[AUD-1625] TrackScreen performance and DynamicImage fix

sliptype:
### Description

This PR:
  * Reduces rerenders related to loading the track screen
  * Fixes the scroll indicator on Lineup based pages
  * Fixes the flashing of images on Track/Collection screens

### Dragons

Added some pretty specific selector equality functions around track screen, if omitted data is needed in the future it will be important to update those

### How Has This Been Tested?
ios simulator

when released

### DIFF
--- a/packages/mobile/src/components/details-tile/DetailsTile.tsx
+++ b/packages/mobile/src/components/details-tile/DetailsTile.tsx
@@ -1,4 +1,4 @@
-import { useCallback } from 'react'
+import { useCallback, useState } from 'react'
 
 import { Name } from 'audius-client/src/common/models/Analytics'
 import { getUserId } from 'audius-client/src/common/store/account/selectors'
@@ -250,13 +250,13 @@ export const DetailsTile = ({
   const imageElement = coSign ? (
     <CoSign size={Size.LARGE}>
       <DynamicImage
-        source={{ uri: imageUrl }}
+        source={imageUrl ? { uri: imageUrl } : undefined}
         styles={{ image: styles.coverArt as ImageStyle }}
       />
     </CoSign>
   ) : (
     <DynamicImage
-      source={{ uri: imageUrl }}
+      source={imageUrl ? { uri: imageUrl } : undefined}
       styles={{ image: styles.coverArt as ImageStyle }}
     />
   )
--- a/packages/mobile/src/components/details-tile/DetailsTile.tsx
+++ b/packages/mobile/src/components/details-tile/DetailsTile.tsx
@@ -248,7 +248,7 @@ export const DetailsTile = ({
   }
 
   const imageElement = coSign ? (
-    <CoSign size={Size.LARGE} style={styles.coverArt}>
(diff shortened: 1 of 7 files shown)


packages/mobile/src/components/details-tile/DetailsTile.tsx
```diff
@@ -248,15 +248,15 @@ export const DetailsTile = ({
   }
 
   const imageElement = coSign ? (
-    <CoSign size={Size.LARGE} style={styles.coverArt}>
+    <CoSign size={Size.LARGE}>
       <DynamicImage
-        source={{ uri: imageUrl }}
+        source={imageUrl ? { uri: imageUrl } : undefined}
```
dylanjeffers, line 253: nice check here, so would we need to do this for all dynamic images that can take a uri that may be undefined?

packages/mobile/src/screens/feed-screen/FeedScreen.tsx
```diff
@@ -27,7 +27,10 @@ const messages = {
 
 export const FeedScreen = () => {
   const dispatchWeb = useDispatchWeb()
-  const feedLineup = useSelectorWeb(getFeedLineup, isEqual)
+  const feedLineup = useSelectorWeb(getFeedLineup, (a, b) => {
+    const omitUneeded = o => omit(o, ['inView'])
```
dylanjeffers, line 31: nice find here, i wonder if we want this in other places too, can def check profile

packages/mobile/src/screens/track-screen/TrackScreen.tsx
```diff
@@ -114,9 +115,27 @@ const TrackScreenMainContent = ({
...
   const { params } = useRoute<'Track'>()
-  const track = useSelectorWeb(state => getTrack(state, params))
-  const user = useSelectorWeb(state => getUser(state, { id: track?.owner_id }))
-  const lineup = useSelectorWeb(getMoreByArtistLineup, isEqual)
+  const track = useSelectorWeb(
+    state => getTrack(state, params),
+    // Omitting uneeded fields from the equality check because they are
+    // causing extra renders when added to the `track` object
+    (a, b) => {
+      const omitUneeded = o => omit(o, ['_stems', '_remix_parents'])
```
dylanjeffers, line 123: awesome

dylanjeffers (approved):
awesome finds, and i particularly like the comments for each equality check

packages/mobile/src/components/details-tile/DetailsTile.tsx
sliptype: Yeah, otherwise it doesn't load the fallback gradient. I thought about baking it into DynamicImage but ImageSourceType is fairly wide. It could still be done though

dylanjeffers: yeah totally agree its hard cause of the type ^^

PR #1046, opened 2022-03-09
[AUD-1623] Fix LineupTile layout shift

sliptype:
### Description

* Adjusts height of LineupTileStats to avoid height shift when replacing skeleton with LineupTile
* Apply width to container of LineupTileActionButtons to prevent the buttons from moving during render (The lottie animation buttons take a bit longer to load)

### Dragons

There will still be shift when loading Collections in a lineup, this is expected and current behavior

### How Has This Been Tested?

iOS simulator

n/a


dylanjeffers (approved):
sweet!

PR #1051, opened 2022-03-10
[AUD-1551, AUD-1554] Add LetThemDJ and TopAlbums screens for explore screen

Kyle-Shanks:
### Description

* Add LetThemDJScreen
* Add TopAlbumsScreen

[AUD-1551] [AUD-1554]

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A

### DIFF
--- a/packages/mobile/src/screens/explore-screen/tabs/ForYouTab/LetThemDJScreen.tsx
+++ b/packages/mobile/src/screens/explore-screen/tabs/ForYouTab/LetThemDJScreen.tsx
@@ -1,0 +1,35 @@
+import Status from 'audius-client/src/common/models/Status'
+import {
+  getCollections,
+  getStatus
+} from 'audius-client/src/common/store/pages/explore/exploreCollections/selectors'
+import { ExploreCollectionsVariant } from 'audius-client/src/common/store/pages/explore/types'
+import { isEqual } from 'lodash'
+
+import { CollectionList } from 'app/components/collection-list'
+import { Screen } from 'app/components/core'
+import { Header } from 'app/components/header'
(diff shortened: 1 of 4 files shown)


packages/mobile/src/screens/explore-screen/tabs/ForYouTab/LetThemDJScreen.tsx
```diff
@@ -0,0 +1,35 @@
...
+    isEqual
+  )
+  const exploreData = useSelectorWeb(state =>
+    getCollections(state, { variant: ExploreCollectionsVariant.LET_THEM_DJ })
+  )
+
+  return (
+    <Screen>
+      <Header text={LET_THEM_DJ.title} />
+      <WithLoader loading={status === Status.LOADING}>
```
dylanjeffers, line 30: this is cool! WithLoader implies this is an hoc, will think about a good name... cant really come up with anything atm! im also down to just have 
```
 loading ? <ScreenSpinner/> : <ScreenComponent />
```
just have the spinner with the special styles ya know? but i like this too, just thinking of all the options.

dylanjeffers (approved):
perfect

packages/mobile/src/screens/explore-screen/tabs/ForYouTab/LetThemDJScreen.tsx
Kyle-Shanks: Yea I realized that the `With__` name has meaning to it, but was the first thing that I thought of. I'll change it later if we do come up with a better name 👌

PR #1052, opened 2022-03-10
Add commas to numbers

rickyrombo:
### Description

Adds commas to amount/step numbers in rewards page/modal

Also fixes bug with the progress text on mobile

### Dragons

### How Has This Been Tested?

Tested manually on web desktop and iOS native

### DIFF
--- a/packages/mobile/src/components/challenge-rewards-drawer/ChallengeRewardsDrawer.tsx
+++ b/packages/mobile/src/components/challenge-rewards-drawer/ChallengeRewardsDrawer.tsx
@@ -2,6 +2,7 @@ import React from 'react'
 
 import { UserChallengeState } from 'audius-client/src/common/models/AudioRewards'
 import { ClaimStatus } from 'audius-client/src/common/store/pages/audio-rewards/slice'
+import { fillString } from 'audius-client/src/common/utils/fillString'
 import { formatNumberCommas } from 'audius-client/src/common/utils/formatUtil'
 import { StyleSheet, View, ImageSourcePropType } from 'react-native'
 
@@ -201,9 +202,11 @@ export const ChallengeRewardsDrawer = ({
   const statusText = hasCompleted
     ? messages.complete
     : isInProgress
-    ? `${formatNumberCommas(currentStep)}/${formatNumberCommas(
-        stepCount
-      )} ${progressLabel}`
+    ? fillString(
+        progressLabel,
+        formatNumberCommas(currentStep),
+        formatNumberCommas(stepCount)
+      )
     : messages.incomplete
 
   const claimedAmountText = `(${formatNumberCommas(claimedAmount)} ${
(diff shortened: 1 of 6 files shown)


audius-infra:
Preview this change https://demo.audius.co/mjp-commas

piazzatron (approved):

packages/web/src/pages/audio-rewards-page/components/modals/ChallengeRewards.tsx
```diff
@@ -245,7 +246,7 @@ const ChallengeRewardsBody = ({ dismissModal }: BodyProps) => {
   const progressReward = (
     <div className={wm(styles.progressReward)}>
       <h3>Reward</h3>
-      <h2>{challenge?.totalAmount}</h2>
+      <h2>{formatNumberCommas(challenge?.totalAmount || '')}</h2>
```
raymondjacobson, line 249: this fn isn't typed, but i think it takes a num instead of a string, so the empty string is maybe not the right undefined default (though it may work).

perhaps the fn itself could just take undefined and if undefined return ''

packages/mobile/src/components/challenge-rewards-drawer/ChallengeRewardsDrawer.tsx
```diff
@@ -200,11 +202,19 @@ export const ChallengeRewardsDrawer = ({
   const statusText = hasCompleted
     ? messages.complete
     : isInProgress
-    ? `${currentStep}/${stepCount} ${progressLabel}`
+    ? fillString(
```
rickyrombo, line 205: FYI @raymondjacobson I think this changed to use `fillString` instead of just being the label. (Originally it was just "Invites Accepted" before your change - now it's "%0/%1 Invites Accepted")

raymondjacobson: Oh shoot, thank you for finding this

raymondjacobson (approved):

audius-infra:
Preview this change https://demo.audius.co/mjp-commas

packages/web/src/pages/audio-rewards-page/components/modals/ChallengeRewards.tsx
rickyrombo: It takes either - if the arg is a number, it calls `toString()` first anyway.

I think it makes sense to have `Earn up to  $AUDIO` in the worst case rather than `Earn up to 0 $AUDIO`?

raymondjacobson: don't get me wrong, I don't think a default of 0 is correct, but since the arg is called num, I think this is confusing. We should probably just type the util haha.

audius-infra:
Preview this change https://demo.audius.co/mjp-commas

audius-infra:
Preview this change https://demo.audius.co/mjp-commas

packages/web/src/pages/audio-rewards-page/components/modals/ChallengeRewards.tsx
rickyrombo: It has JSDoc, which TypeScript used to warn me that the arg isn't allowed to be undefined. I didn't find it confusing fwiw. also just realizing this should be ?? '' not || '' probably

audius-infra:
Preview this change https://demo.audius.co/mjp-commas

PR #1054, opened 2022-03-11
[AUD-1684]: ProfileScreen performance

dylanjeffers:
### Description

Uses new selector equality and selectors to drastically minimize profile rerenders and push them as far down the render tree as possible.

Experimenting with `lazy: true` for profile tabs. This helps with navigation animation by quieting down other tabs.

A few improvements to Lineup
 - Show skeleton on initial mount, which improves perceieved perf as the lineups don't suddenly show skeletons
 - Check the `limit` prop before rendering skeletons or loading more. this allows us to immediately show `ListEmptyComponent`, which is useful for our EmptyProfileTiles

A bit of a hack adding `initalParams: {handle: "accountUser"}}` which is a way to immediately tell account profile tab to render with account user data

### DIFF
--- a/packages/mobile/src/components/lineup/Lineup.tsx
+++ b/packages/mobile/src/components/lineup/Lineup.tsx
@@ -122,7 +122,8 @@ export const Lineup = ({
   listKey,
   selfLoad,
   includeLineupStatus,
-  ...scrollViewProps
+  limit = Infinity,
+  ...sectionListProps
 }: LineupProps) => {
   const dispatchWeb = useDispatchWeb()
   const ref = useRef<SectionList>(null)
@@ -158,6 +159,7 @@ export const Lineup = ({
       lineupLength < countOrDefault &&
       // Page item count doesn't exceed current offset
       (page === 0 || pageItemCount <= offset) &&
+      entries.length < limit &&
       (includeLineupStatus ? status !== Status.LOADING : true)
 
     if (shouldLoadMore) {
@@ -183,7 +185,8 @@ export const Lineup = ({
     loadMore,
     dispatchWeb,
     pageItemCount,
-    includeLineupStatus
+    includeLineupStatus,
+    limit
   ])
 
   useEffectOnce(() => {
@@ -277,10 +280,11 @@ export const Lineup = ({
 
     const getSkeletonCount = () => {
       const shouldCalculateSkeletons =
+        items.length < limit &&
         // Lineup has more items to load
         hasMore &&
-        // Data is loading
-        isMetadataLoading &&
+        // Data is loading or about to start
+        (items.length === 0 || isMetadataLoading) &&
         // There are fewer items than the max count
         items.length < countOrDefault
 
@@ -321,10 +325,14 @@ export const Lineup = ({
       ]
     }
 
+    const data = [...items, ...skeletonItems]
+
+    if (data.length === 0) return []
+
     return [
       {
         delineate: false,
-        data: [...items, ...skeletonItems]
+        data: data
       }
     ]
   }, [
@@ -336,12 +344,13 @@ export const Lineup = ({
     pageItemCount,
     leadingElementId,
     showLeadingElementArtistPick,
-    start
+    start,
+    limit
   ])
 
   return (
     <SectionList
-      {...scrollViewProps}
+      {...sectionListProps}
       ref={ref}
       ListHeaderComponent={header}
       ListFooterComponent={<View style={{ height: 160 }} />}
--- a/packages/mobile/src/components/lineup/Lineup.tsx
+++ b/packages/mobile/src/components/lineup/Lineup.tsx
@@ -121,6 +121,7 @@ export const Lineup = ({
   variant = LineupVariant.MAIN,
   listKey,
   selfLoad,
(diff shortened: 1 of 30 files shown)


packages/mobile/src/components/lineup/Lineup.tsx
```diff
@@ -319,10 +325,14 @@ export const Lineup = ({
...
 
+    const data = [...items, ...skeletonItems]
+
+    if (data.length === 0) return []
+
     return [
       {
         delineate: false,
-        data: [...items, ...skeletonItems]
+        data: data
```
raymondjacobson, line 335: just data :)

packages/mobile/src/components/lineup/types.ts
```diff
@@ -67,7 +67,7 @@ export type LineupProps = {
    * A header to display at the top of the lineup,
    * will scroll with the rest of the content
    */
-  header?: SectionListProps<any>['ListHeaderComponent']
+  header?: SectionListProps<unknown>['ListHeaderComponent']
```
raymondjacobson, line 70: why does this have to be unknown? (certainly better than any, but just curious)

packages/mobile/src/components/lineup/types.ts
```diff
@@ -111,4 +111,13 @@ export type LineupProps = {
    * another VirtualizedList.
    */
   listKey?: string
-} & Pick<ScrollViewProps, 'showsVerticalScrollIndicator'>
+
+  /**
+   * When `true` don't load more while lineup status is `LOADING`.
```
raymondjacobson, line 116: do we ever want to have this false?

packages/mobile/src/screens/profile-screen/ProfileInfo.tsx
```diff
@@ -62,26 +62,38 @@ const messages = {
...
+  ])
+
+  const {
+    name,
+    handle,
+    does_current_user_follow,
+    does_follow_current_user
+  } = profile
+
+  const isOwner = useSelectorWeb(getIsOwner)
```
raymondjacobson, line 86: we could just match on handle equality here instead of a useSelectorWeb right? that would save 1 rerender

packages/mobile/src/screens/profile-screen/SubscribeButton.tsx
```diff
@@ -33,7 +33,7 @@ export const SubscribeButton = (props: SubscribeButtonProps) => {
   const styles = useStyles()
   const { profile } = props
   const { user_id } = profile
-  const { isSubscribed } = useSelectorWeb(getProfile)
+  const isSubscribed = useSelectorWeb(getIsSubscribed)
```
raymondjacobson, line 36: +1 specificity

packages/web/src/common/store/pages/profile/selectors.ts
```diff
@@ -39,6 +40,37 @@ export const getProfileFeedLineup = (state: CommonState) =>
 export const getProfileTracksLineup = (state: CommonState) =>
   state.pages.profile.tracks
 
+export const getProfileCollections = createDeepEqualSelector(
```
raymondjacobson, line 43: The selectors in this file scare me a bit. There must be a better way. This is certainly good though, and not a new pattern.

raymondjacobson (approved):

packages/mobile/src/components/lineup/types.ts
dylanjeffers: yeah just updated to unknown to be safer, but ultimately that type-arg should be the value of our `section` const, which we can def determine and add here if we feel the need

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1684-profile-perf-v2

packages/mobile/src/screens/profile-screen/ProfileInfo.tsx
dylanjeffers: so how do we get the owner's handle/id, just do `useSelectorWeb(getUserId)`?

packages/web/src/common/store/pages/profile/selectors.ts
dylanjeffers: yeah this file is wild! essentially I just stripped out the relevant parts for collections from `getProfileUser`

packages/mobile/src/screens/profile-screen/ProfileInfo.tsx
raymondjacobson: oh nvm ... i misread. you do need that somehow. it is something we don't need to constantly pull and maybe could be in RN state itself, but nvm!!!!!!

packages/mobile/src/components/lineup/types.ts
dylanjeffers: agreed, we added the status check thinking we'd always want to do that. It ended up breaking a few lineups though: trending month/all-time, and reposts. just to be safe added this config in to solve the specific case for trackscreen. we should prob get to the bottom of this though. happy to do that now if you think its worth it

raymondjacobson: gotcha gotcha. non-blocking for now! probably lower pri

PR #1055, opened 2022-03-11
Add fp

piazzatron:
### Dragons

None

### How Has This Been Tested?

- Tested on stage


packages/web/src/services/fingerprint/FingerprintClient.ts
```diff
@@ -0,0 +1,74 @@
...
+    console.log('Initializing Fingerprint client')
+    try {
+      const fp = await FingerprintJS.load({
+        apiKey: this.apiKey,
+        endpoint: this.endpoint
+      })
+      console.log(`Fingerprint loaded`)
+      this.fingerprint = fp
+    } catch (e) {
+      console.error(`Error initializaing fingerprint client: ${e}`)
```
raymondjacobson, line 28: initializing

packages/web/src/services/fingerprint/FingerprintClient.ts
```diff
@@ -0,0 +1,74 @@
...
+  async identify(userId: number) {
+    if (!this.fingerprint) {
+      console.warn('Fingerprint client not yet initted')
+      return
+    }
+    try {
+      const origin = isMobile() ? 'mobile' : isElectron() ? 'desktop' : 'web'
+
+      // First, see if we've fingerprinted this user before
+      const response = await fetch(
```
raymondjacobson, line 41: Non-blocking, but we could probably avoid an extra network req if we remember in local storage that we have done a fprint

raymondjacobson (approved):

audius-infra:
Preview this change https://demo.audius.co/piazz-fingerprint

sddioulde (approved):
lg

packages/web/src/services/fingerprint/FingerprintClient.ts
piazzatron: interesting idea - would be a bit problematic in the case that you switch browsers / use incognito, and we'd end up fingerprinting multiple times. going to leave as if for now

raymondjacobson: don't we want to fp again if you switch browsers?

audius-infra:
Preview this change https://demo.audius.co/piazz-fingerprint

packages/web/src/services/fingerprint/FingerprintClient.ts
piazzatron: discussed online – fp is stable between browsers

PR #1060, opened 2022-03-11
Record IP if not recent

sddioulde:
### Description
Record IP if not recent

### Dragons

None

### How Has This Been Tested?

Manually

Elsewhere on identity

### DIFF
--- a/packages/web/src/services/AudiusBackend.js
+++ b/packages/web/src/services/AudiusBackend.js
@@ -2452,6 +2452,25 @@ class AudiusBackend {
     }
   }
 
+  static async recordIP() {
+    await waitForLibsInit()
+    const account = audiusLibs.Account.getCurrentUser()
+    if (!account) return { error: true }
+
+    try {
+      const { data, signature } = await AudiusBackend.signData()
+      return fetch(`${IDENTITY_SERVICE}/record_ip`, {
+        headers: {
+          [AuthHeaders.Message]: data,
+          [AuthHeaders.Signature]: signature
+        }
+      }).then(res => res.json())
+    } catch (err) {
+      console.error(err.message)
+      return { error: true }
+    }
+  }
+
   static async getRandomFeePayer() {
     await waitForLibsInit()
     try {
--- a/packages/web/src/services/AudiusBackend.js
+++ b/packages/web/src/services/AudiusBackend.js
@@ -2452,25 +2452,6 @@ class AudiusBackend {
     }
   }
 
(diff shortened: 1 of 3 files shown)


audius-infra:
Preview this change https://demo.audius.co/sd-id

packages/web/src/services/AudiusBackend.js
```diff
@@ -2452,6 +2452,25 @@ class AudiusBackend {
     }
   }
 
+  static async recordIP() {
```
piazzatron, line 2455: let's make a new file `RecordIP.ts` in the `services/audius-backend` folder for this guy, so we're not adding more code to this poor untyped file

packages/web/src/store/account/sagas.js
```diff
@@ -69,6 +69,25 @@ import mobileSagas, { setHasSignedInOnMobile } from './mobileSagas'
 
 const NATIVE_MOBILE = process.env.REACT_APP_NATIVE_MOBILE
 
+const IP_STORAGE_KEY = 'user-ip-timestamp'
+
+function* recordIPIfNotRecent() {
+  const twentyFourHours = 24 * 60 * 60 * 1000
```
piazzatron, line 75: nit: probably shouldn't name this variable the duration of the interval itself, because that means we can't change the interval without changing the variable name

packages/web/src/store/account/sagas.js
```diff
@@ -69,6 +69,25 @@ import mobileSagas, { setHasSignedInOnMobile } from './mobileSagas'
...
+  const now = Date.now()
+  const twentyFourHoursAgo = now - twentyFourHours
+  const storedIPStr = window.localStorage.getItem(IP_STORAGE_KEY)
+  const storedIP = storedIPStr && JSON.parse(storedIPStr)
+  if (!storedIP || storedIP.timestamp < twentyFourHoursAgo) {
```
piazzatron, line 80: Should we include user_id / handle in this localstorage object, so that if a single user has multiple accounts on a machine we assign IPs for all of them?

piazzatron (approved):

packages/web/src/store/account/sagas.js
sddioulde: good idea

audius-infra:
Preview this change https://demo.audius.co/sd-id

audius-infra:
Preview this change https://demo.audius.co/sd-id

PR #1068, opened 2022-03-14
[AUD-1683] Improve app layout and navigation structure

dylanjeffers:
### Description

- Renames AppNavigator to RootScreen
- Renames BottomTabNavigator to AppScreen
- Moves SplashScreen to screens
- Modularizes AppScreenTabs and types
- Improves WebApp height and NativeApp Height
- Removes Screen padding fix as a result of fixing the webapp/nativeapp heights
- Removes various legacy/unused components

### Dragons

The naming conventions for some of the files im not too sure about, but overall i think the restructure does make more sense. curious on yall's thoughts.

### How Has This Been Tested?

Ran iPhone 13 and iPhone 8 simulators and checked each page


packages/mobile/src/App.tsx
```diff
@@ -76,12 +70,11 @@ const App = () => {
         <ToastContextProvider>
           <ErrorBoundary>
             <WebRefContextProvider>
-              <WebAppManager webApp={<WebApp webRef={webRef} />}>
+              <WebAppManager webRef={webRef}>
                 <ThemeProvider>
                   <NavigationContainer>
                     <GoogleCast webRef={webRef} />
-                    <AppNavigator />
-                    <SearchScreenLegacy />
```
sliptype, line 84: So this means the webview version of the app will no longer work right? I'm wondering if we should maybe merge this into a branch so we don't prevent ourselves from pushing bugfixes the the current app

We ran into that case last week with the fingerprint stuff needing to be released to prod

packages/mobile/src/screens/app-screen/AppTabScreen/AppTabScreen.tsx
```diff
@@ -15,35 +19,49 @@ import {
...
 /**
- * This is the base stack that includes common screens
+ * This is the base tab screen that includes common screens
  * like track and profile
  */
-export const BaseStackNavigator = ({
-  baseScreen,
-  Stack
-}: BaseNavigatorProps) => {
+export const AppTabScreen = ({ baseScreen, Stack }: AppTabScreenProps) => {
```
sliptype, line 58: I'm a little hesitant on this name because this stack navigator doesn't _necessarily_ need to be in a tab, it just so happens to be. And the root component is a Navigator, so I feel like naming this *Something*Navigator would make more sense?

packages/mobile/src/screens/app-screen/AppTabScreen/useTabScreenOptions.tsx
```diff
@@ -5,31 +5,25 @@ import {
...
-import { IconButton } from '../core'
+import { AppScreenParamList } from '../AppScreen'
 
-import { TopBarArrowBack } from './TopBarArrowBack'
-import { MainParamList, SearchParamList } from './types'
+import { AppTabScreenParamList } from './AppTabScreen'
 
 const useStyles = makeStyles(({ palette, spacing, typography }) => ({
-  root: {
-    height: Platform.OS === 'ios' ? 86 : 55,
```
sliptype, line 27: This wasn't needed anymore?

sliptype (commented):
Looks great! Love the structure. We should probably discuss if/when we want to merge this into main since it sets the native UI as default

packages/mobile/src/App.tsx
dylanjeffers: @sliptype so totally agreed, however i was talking with @raymondjacobson whether we should keep things backwards compatible, and he seemed to think no. like all the notification changes that were just merged are not backwards-compat with webview version i beleive

packages/mobile/src/screens/app-screen/AppTabScreen/AppTabScreen.tsx
dylanjeffers: yeah i def here the idea for navigator, but maybe same argument that the screen just so happens to be a navigator? i was also considering putting all the "tab" stuff in the actually FeedScreen, ExploreScreen screens, since it prob makes sense for those screens to compose all the parts of the explore/feed experience?

packages/mobile/src/screens/app-screen/AppTabScreen/useTabScreenOptions.tsx
dylanjeffers: with the safeview i think we are covered now. plus the 86 was quite tall on legacy iPhones

packages/mobile/src/App.tsx
sliptype: Ah ok fair enough, we'll do it live 😤

packages/mobile/src/screens/app-screen/AppTabScreen/AppTabScreen.tsx
sliptype: Yeah that makes sense too! I'm good with that

sliptype (approved):

packages/mobile/src/screens/app-screen/AppTabScreen/AppTabScreen.tsx
dylanjeffers: word, ill move those over then and we can see how that looks

audius-infra:
Preview this change https://demo.audius.co/dj-aud-1683-app-layout

PR #1073, opened 2022-03-15
[AUD-1724] Fix useImageHook default bug on native

sliptype:
### Description

* Fixes bug where `useImageSize` hook returns the default (statically imported) image which gets assigned as the `uri` string of a source prop
* Refactors native image hooks to reduce duplication
* Updates `DynamicImage` to accept a `uri` prop instead of `source`. This allows for easier equality and null checks. We are only using `DynamicImage` for remote images anyway, and with `Image.resolveAssetSource` we still have a way to use it with static images if we need to

### Dragons

Looks like updating cover and profile pictures from the profile screen is broken, seems to be broken on `main` too: https://linear.app/audius/issue/AUD-1731/cannot-update-coverprofile-picture-from-profile-screen

Images need another pass entirely on native. Because of the way our image fetch sagas work, each image is fetched twice. Also we are not making use of the caching abilities of `Image`: https://linear.app/audius/issue/AUD-1729/fix-duplicate-image-fetch

### How Has This Been Tested?

iOS simulator

Verify that images continue to load properly


packages/mobile/src/hooks/useImageSize.ts
```diff
@@ -0,0 +1,46 @@
...
+      'action' | 'dispatch' | 'defaultImage' | 'sizes' | 'size'
+    > & {
+      sizes: Nullable<ImageSizesObject<Sizes>>
+      size: Sizes
+    }
+  ) => {
+    const dispatch = useDispatchWeb() as typeof useDispatch
+
+    // This resolves a statically imported image into a uri
+    const defaultImage = Image.resolveAssetSource(defaultImageSource).uri
```
dylanjeffers, line 37: Nice

dylanjeffers (approved):
This looks awesome!! Also pull to refresh messes up the photos in profile screen, which I also see in web, so maybe there is another thing funky with the sagas

PR #1074, opened 2022-03-15
[AUD-1687] Reduce AnimatedButton rerenders and move to core

sliptype:
### Description
* Refactors `AnimatedButtonProvider` into `AnimatedButton` and moves to `core`
* Reduces rerenders by memoizing props in RepostButton and FavoriteButton

### Dragons

This removes all the logic in the provider that was setting refs of the iconJSONs, although I don't think that was doing anything anyway because the inner component wasn't memoized

### How Has This Been Tested?

iOS simulator, animated buttons work and there are fewer rerenders

Verify button animations continue to work


packages/mobile/src/components/favorite-button/FavoriteButton.tsx
```diff
@@ -16,29 +15,31 @@ const styles = StyleSheet.create({
...
-  AnimatedButtonProviderProps,
+  AnimatedButtonProps,
   'iconLightJSON' | 'iconDarkJSON' | 'isDarkMode'
 >
 
-const FavoriteButton = ({ isActive, ...props }: FavoriteButtonProps) => {
-  const themeVariant = useThemeVariant()
-  const isDarkMode = themeVariant === Theme.DARK
+export const FavoriteButton = ({ isActive, ...props }: FavoriteButtonProps) => {
+  const wrapperStyle = useMemo(
```
dylanjeffers, line 27: Interesting that this needed to be memorized, makes sense but annoying

dylanjeffers (approved):
So much better, crushing it

PR #1077, opened 2022-03-15
[AUD-1636] Add hidden track header on native

sliptype:
### Description

* Adds hidden track header

Screen Shot 2022-03-15 at 2 47 23 PM

### Dragons

N/A

### How Has This Been Tested?
ios sim

N/A

### DIFF
--- a/packages/mobile/src/screens/track-screen/TrackScreenDetailsTile.tsx
+++ b/packages/mobile/src/screens/track-screen/TrackScreenDetailsTile.tsx
@@ -41,6 +41,7 @@ import { open as openOverflowMenu } from 'common/store/ui/mobile-overflow-menu/s
 import { Image, Pressable, View } from 'react-native'
 import { useSelector } from 'react-redux'
 
+import IconHidden from 'app/assets/images/iconHidden.svg'
 import { Text } from 'app/components/core'
 import { DetailsTile } from 'app/components/details-tile'
 import { DetailsTileDetail } from 'app/components/details-tile/types'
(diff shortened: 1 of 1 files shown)


packages/mobile/src/screens/track-screen/TrackScreenDetailsTile.tsx
```diff
@@ -96,7 +99,18 @@ const useStyles = makeStyles(({ palette, spacing }) => ({
...
+  },
+
+  hiddenTrackLabel: {
+    marginTop: spacing(1),
+    marginLeft: spacing(2),
+    color: palette.accentOrange,
+    fontFamily: typography.fontByWeight.demiBold,
+    fontSize: 14,
+    letterSpacing: 2,
+    textTransform: 'uppercase'
```
Kyle-Shanks, line 113: It seems weird that we have variants for the Text component, but still need to add custom styles in a lot of places. Maybe we should revisit the props to make it more flexible?

Kyle-Shanks (approved):

packages/mobile/src/screens/track-screen/TrackScreenDetailsTile.tsx
sliptype: I agree, we actually deprecated our Text component for this reason. Decided to use styles exclusively bc it will be the most flexible, and is still convenient with the `typography` stuff Dylan added

So going forward let's just use `Text` from `react-native`

PR #1080, opened 2022-03-16
Fix album draggable bug

nicoback2:
### Description
Before:
Could drag album tile to playlist library, which would be a no op
Could not drag album tile to "Favorites" list

After:
Can only drag album tile to "Favorites" list

### Dragons

### How Has This Been Tested?
Tested in browser

### DIFF
--- a/packages/web/src/components/lineup/CardLineup.tsx
+++ b/packages/web/src/components/lineup/CardLineup.tsx
@@ -37,7 +37,7 @@ const DesktopCardContainer = ({
             <Draggable
               key={`draggable-${card.props.playlistId}`}
               text={card.props.primaryText}
-              kind={card.props.isAlbum ? 'album' : 'playlist'}
+              kind={card.props.isPlaylist ? 'playlist' : 'album'}
               id={card.props.playlistId}
               link={card.props.link}
             >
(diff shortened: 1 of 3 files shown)


packages/web/src/components/lineup/CardLineup.tsx
```diff
@@ -37,7 +37,7 @@ const DesktopCardContainer = ({
             <Draggable
               key={`draggable-${card.props.playlistId}`}
               text={card.props.primaryText}
-              kind={card.props.isAlbum ? 'album' : 'playlist'}
+              kind={card.props.isPlaylist ? 'playlist' : 'album'}
```
nicoback2, line 40: This is the real fix right here - the desktop CollectionsPage component was passing in the `isPlaylist` prop but this component was looking for an `isAlbum` prop https://github.com/AudiusProject/audius-client/blob/d8cb1f704d43647389ddb58664ea743b59e23bd9/packages/web/src/pages/explore-page/components/desktop/CollectionsPage.tsx#L88-L89

I updated the mobile `Card` and `CollectionsPage` components to have the `isPlaylist` prop as well so that they match the desktop ones

dylanjeffers (approved):
nice catch

PR #1082, opened 2022-03-16
Fix types for AudioBreakdownDrawer equality check

sliptype:
### Description

* Fixes type for AudioBreakdownDrawer equality check

### Dragons

Could change selector behavior

### How Has This Been Tested?

iOS sim

Ensure audio breakdown drawer continues to work


sliptype:
This was to fix lint on CI, but it isn't failing anymore so I'm closing this for now

PR #1085, opened 2022-03-17
Fix local storage temp playlist bug

nicoback2:
### Description
Before:
1. Create a playlist
2. Quickly reorder the playlist or add it to a folder. This triggers a cache + local storage update with the new library order (which includes the temp playlist).
3. Refresh
4. Result: Temp playlist gets stuck in local storage, which prevents any user playlist library updates from being persisted because we wait for all temp playlists to be resolved before updating the user metadata

After:
Temp playlists never get put in local storage which prevents the corrupted state described above. 

### Dragons

### How Has This Been Tested?
Tested in browser using repro steps above


raymondjacobson (approved):
tysm!!

PR #1092, opened 2022-03-21
[AUD-1727] Implement collapsible tab profile screen

raymondjacobson:
### Description

Implement  collapsible tab profile screen (Twitter-style).
This PR implements (with some necessary changes) V2 of this library: https://github.com/PedroBern/react-native-collapsible-tab-view/tree/v2.

Note v3 & v4 are super cool, but require re-animated V2 (which we don't have), and also do not play super nicely with react-navigation (yet!). They're working on those things though it still sounds like :)

One caveat:
- Touch targets on the "header" aren't scrollable. It sounds like Reddit behaves similarly, but Twitter is nicer in that you can still scroll on those touch targets.  The weapon here is unfortunately a bit blunt: https://github.com/PedroBern/react-native-collapsible-tab-view/tree/v2#scroll-on-header but there may be some more we can do.

### Dragons

Touches main rendering of profile screen

### How Has This Been Tested?

* iOS: Locally with sim and on-device to play with scrolling
* This has not been extensively tested on android


raymondjacobson:
> nice, all makes a bunch of sense, def a bit annoying that everything has to have a collapsible variant. i wonder if it's performant to use a collapsible context just to prevent prop drilling?

this was a great suggestion

PR #1098, opened 2022-03-22
[AUD-1738] Hide profile button until profile is available

sliptype:
### Description

Prevents a flash of outdated data on profile screens by delaying the render of the profile button and tracks lineup until the profile is loaded

### Dragons

The button pops in after the rest of the page, along with the user images etc. Could be kinda jarring

Had to add a prop to the `TracksTab` component which could impact performance

### How Has This Been Tested?

iOS sim

Testflight qa


dylanjeffers (approved):
interesting, so basically wait till the profile user comes back before rendering the button? does this change perf at all with rerendering due to `getProfileUser` ?

packages/mobile/src/screens/profile-screen/ProfileTabNavigator.tsx
```diff
@@ -48,7 +49,7 @@ export const ProfileTabNavigator = ({
   const trackScreen = collapsibleTabScreen({
     name: 'Tracks',
     Icon: IconNote,
-    component: TracksTab,
+    component: () => <TracksTab isProfileLoaded={isProfileLoaded} />,
```
dylanjeffers, line 52: I believe this leads to perf issues, since this component is not memoized, i think the solution is to provide children to the tab like we did with BottomTabNavigator back in the day. Thinking what we might be able to do is just add the `isProfileLoaded` selector into TracksTab?

dylanjeffers (approved):
looks good, let's figure out the component prop perf thing tho

packages/mobile/src/screens/profile-screen/ProfileTabNavigator.tsx
sliptype: I was able to set `handle` on initialParams and pull that out in `TracksTab` using `useRoute`. Let me know what you think!

dylanjeffers: yeah that's perf

packages/mobile/src/screens/profile-screen/selectors.ts
```diff
@@ -49,3 +50,14 @@ export const getIsOwner = createSelector(
   [getProfileUserId, getUserId],
   (profileUserId, accountUserId) => profileUserId === accountUserId
 )
+
+export const useIsProfileLoaded = () => {
+  const { params } = useRoute<'Profile'>()
```
dylanjeffers, line 55: yeah nice

PR #1102, opened 2022-03-22
[AUD-1740] Add now playing links

sliptype:
### Description

Makes the artwork, title, and artist name pressable on the now playing drawer
Also fixes an issue with the title overflowing

image

### Dragons

Makes the title a maximum of 2 lines to match the current app. This causes the layout to shift slightly but everything remains on screen and looks reasonable

### How Has This Been Tested?

iOS sim

Testflight QA


Kyle-Shanks (approved):
Looks good. Could make updates to use `useStyles` if we don't want to have to update this later

PR #1105, opened 2022-03-23
[AUD-1749] Update explore screen layout to match design

Kyle-Shanks:
### Description

[AUD-1749]

* Small update to fix the layout to match the original design

### Dragons

N/A

### How Has This Been Tested?

Manually tested, looks good

simulator_screenshot_D03B048A-50CB-4ADA-82C4-60669B6E9FF0

N/A

### DIFF
--- a/packages/mobile/src/screens/app-screen/ExploreTabScreen.tsx
+++ b/packages/mobile/src/screens/app-screen/ExploreTabScreen.tsx
@@ -17,6 +17,7 @@ import {
 import {
   LetThemDJScreen,
   TopAlbumsScreen,
+  TrendingPlaylistsScreen,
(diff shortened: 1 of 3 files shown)


sliptype (approved):
Dope! One other thing I noticed on the new explore screen is that the ColorTile press animation doesn't match what we had on the old on. The new one changes opacity while the old one animates the size, like a TrackTile

Kyle-Shanks:
> Dope! One other thing I noticed on the new explore screen is that the ColorTile press animation doesn't match what we had on the old on. The new one changes opacity while the old one animates the size, like a TrackTile

Yes, I missed that part the first time. Should be fixed now 👍

PR #1109, opened 2022-03-23
[AUD-1761, AUD-1670] Add shadow offset and multiple shadows

dylanjeffers:
### Description

- Adds shadow inset to Tile
- Adds layered shadows to simulate multiple box-shadows

Before:
image

First attempt:
image

Final:
image

### Dragons

- There may be a perf issue doing this, now that each tile renders 3x the shadows
- I followed our box-shadow specs for each shadow exactly, but it def looks funky, so we should modify things before merge

### DIFF
--- a/packages/mobile/src/components/core/Tile/Tile.tsx
+++ b/packages/mobile/src/components/core/Tile/Tile.tsx
@@ -127,10 +127,10 @@ export const Tile = <
       />
       <TileShadow
         offset={[0, 2]}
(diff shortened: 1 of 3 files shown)


Kyle-Shanks (approved):
noice

Kyle-Shanks:
Yea definitely seems darker than before and a bit thick compared to our shadows on web. code looks good to me though

PR #1112, opened 2022-03-24
[AUD-1637] Add horizontal padding to tabs

sliptype:
### Description

* Adds horizontal padding to tabs to match figma https://www.figma.com/file/qGwU0Nd3yrchciX8R3fjRW/Mobile-Tab-Bar?node-id=2%3A58
* Padding decreases as number of tabs increases. at 2 tabs it's 32px, and 5 tabs it's 8px
image

### Dragons

Touches the tab indicator width, need to keep and eye on the tabs to see if we catch anything funky

### How Has This Been Tested?

iOS sim

TestFlight QA

### DIFF
--- a/packages/mobile/src/components/top-tab-bar/TopTabBar.tsx
+++ b/packages/mobile/src/components/top-tab-bar/TopTabBar.tsx
@@ -14,6 +14,8 @@ const INDICATOR_STRETCH_FACTOR = 0.35
 // The higher the number, the more accurate the line stretch will follow the sin function
 const INDICATOR_ANIM_GRANULARITY = 10
 
+const HORIZONTAL_PADDING = 8
+
 const getSinAnimationRanges = (len: number) => {
   const inputRange: number[] = []
   const outputRange: number[] = []
@@ -34,7 +36,6 @@ const getSinAnimationRanges = (len: number) => {
 const useStyles = makeStyles(({ palette, typography, spacing }) => ({
   tabBarContainer: {
     elevation: 3,
-    flexDirection: 'row',
     marginBottom: -1 * spacing(1),
     paddingBottom: spacing(1),
     position: 'relative',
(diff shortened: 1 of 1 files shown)


packages/mobile/src/components/top-tab-bar/TopTabBar.tsx
```diff
@@ -115,43 +125,47 @@ export const TopTabBar = ({ state, descriptors, navigation, position }) => {
...
-            >
-              <Animated.View style={{ opacity }}>{icon}</Animated.View>
-              <Animated.Text style={[styles.tabText, { opacity }]}>
-                {label}
-              </Animated.Text>
-            </TouchableOpacity>
-          </View>
-        )
-      })}
+      <View
```
dylanjeffers, line 128: just wondering if we can combine the styles with tabBarContainer and tabsContainer and just have one View?

sliptype: No because the outer one contains the tab indicator and setting the background color of that would include the background of the indicator

dylanjeffers: ah word, thanks for clarifying

packages/mobile/src/components/top-tab-bar/TopTabBar.tsx
```diff
@@ -102,11 +114,9 @@ export const TopTabBar = ({ state, descriptors, navigation, position }) => {
     })
   }
 
-  const tabWidth = 100 / state.routes.length
-
   const left = position.interpolate({
     inputRange: [0, state.routes.length],
-    outputRange: [0, screenWidth]
+    outputRange: [horizontalPadding, screenWidth - horizontalPadding]
```
dylanjeffers, line 119: this is really cool!

dylanjeffers (approved):
really cool approach here, nice job. one idea that might be cool, is to pass the tab-count into the useStyles hook, and then we can do the computation in the style hook, though looks like maybe you need to use that value in other places?

packages/mobile/src/components/top-tab-bar/TopTabBar.tsx
sliptype: Yeah unfortunately we have to use that value for the animation values too. Definitely like the approach of passing things into useStyles though, will use in the future

PR #1117, opened 2022-03-24
[AUD-1783] Fix scrubber bug

sliptype:
### Description

* Fixes a bug where the timestamp changing width due to longer times would shift the scrubber and break dragging
Verified that scrubber & timestamp styling works with times up to 59:59:59

https://user-images.githubusercontent.com/19916043/159995555-14532b14-ce94-48a4-992d-1154c290d51b.mov

* Adds now playing play button scale animation AUD-1767

### Dragons

N/A

### How Has This Been Tested?

iOS sim

TestFlight QA

### DIFF
--- a/packages/mobile/src/components/core/AnimatedButton.tsx
+++ b/packages/mobile/src/components/core/AnimatedButton.tsx
@@ -21,8 +21,12 @@ import { usePrevious } from 'react-use'
 import { GestureResponderHandler } from 'app/types/gesture'
 import { Theme, useThemeVariant } from 'app/utils/theme'
 
-export type BaseAnimatedButtonProps = {
+type IconJSON = any
+
+export type AnimatedButtonProps = {
   iconIndex?: number
+  iconDarkJSON: IconJSON | IconJSON[]
+  iconLightJSON: IconJSON | IconJSON[]
   isActive?: boolean
   isDisabled?: boolean
   onLongPress?: GestureResponderHandler
@@ -30,14 +34,7 @@ export type BaseAnimatedButtonProps = {
   renderUnderlay?: (state: PressableStateCallbackType) => ReactNode
   style?: PressableProps['style']
   wrapperStyle?: StyleProp<ViewStyle>
-}
-
-type IconJSON = any
-
-export type AnimatedButtonProps = {
-  iconDarkJSON: IconJSON | IconJSON[]
-  iconLightJSON: IconJSON | IconJSON[]
-} & BaseAnimatedButtonProps
+} & PressableProps
 
 export const AnimatedButton = ({
   iconIndex: externalIconIndex,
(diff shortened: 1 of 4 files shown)


packages/mobile/src/components/scrubber/Scrubber.tsx
```diff
@@ -150,7 +151,11 @@ export const Scrubber = ({
 
   return (
     <View style={styles.root}>
-      <Text style={styles.timestamp} weight='regular' numberOfLines={1}>
+      <Text
+        style={[styles.timestamp, { textAlign: 'right' }]}
```
dylanjeffers, line 155: can we add textAlign to timestamp style?

packages/mobile/src/components/now-playing-drawer/AudioControls.tsx
```diff
@@ -72,6 +73,12 @@ export const AudioControls = ({ onNext, onPrevious }: AudioControlsProps) => {
   const shuffleEnabled = useSelectorWeb(getShuffle)
   const repeatMode = useSelectorWeb(getRepeat)
 
+  const {
```
dylanjeffers, line 76: is this for a different ticket?

dylanjeffers (approved):
just want to double check about the animated button changes

packages/mobile/src/components/now-playing-drawer/AudioControls.tsx
sliptype: Yeah AUD-1767, I put it in the description. Could have made a separate PR but thought it was quick enough to knock out while in the now playing drawer

dylanjeffers: ah i see! let me add a comment real quick then

packages/mobile/src/components/scrubber/Scrubber.tsx
sliptype: Yes but the right hand time stamp also uses the same style. So there would need to be a `timestampLeft`. I figured since it was just the alignment it would be ok to leave like this. What do you think?

packages/mobile/src/components/core/AnimatedButton.tsx
```diff
@@ -21,23 +21,20 @@ import { usePrevious } from 'react-use'
 import { GestureResponderHandler } from 'app/types/gesture'
 import { Theme, useThemeVariant } from 'app/utils/theme'
 
-export type BaseAnimatedButtonProps = {
+type IconJSON = any
```
dylanjeffers, line 24: oh i guess we already had `any` but does `JSON` type work?

sliptype: `JSON` doesn't work but I was able to pull the type off the Lottie props!

PR #1120, opened 2022-03-24
[AUD-1753] Add seek buttons for podcasts on mobile

Kyle-Shanks:
### Description
[AUD-1753]

Added the seek forward and seek back buttons for podcasts in the mobile now playing drawer

### Dragons

N/A

### How Has This Been Tested?

Manually tested, it works

Simulator Screen Shot - iPhone 13 - 2022-03-24 at 16 58 25

N/A


raymondjacobson (approved):
noice

packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
```diff
@@ -209,12 +216,30 @@ const NowPlayingDrawer = ({
   }, [trackId])
 
   const onNext = useCallback(() => {
-    setMediaKey(mediaKey => mediaKey + 1)
-  }, [setMediaKey])
+    if (track?.genre === Genre.PODCASTS) {
+      if (global.progress) {
+        const { currentTime } = global.progress
+        const newPosition = currentTime + SKIP_DURATION_SEC
+        dispatch(seek({ type: SEEK, seconds: Math.max(0, newPosition) }))
```
sliptype, line 223: What happens if the seconds exceed the length of the podcast?

sliptype (approved):
Looks great! One question about seeking forward

packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
Kyle-Shanks: ah good catch. yea should be min with the track duration

PR #1124, opened 2022-03-24
[AUD-1775] Improve search results loading state

dylanjeffers:
### Description

Fixes a few loading issue with search results:

- Fixes case where "no results" briefly shows up in between the loading component and the results list.
- Fixes case where popping back to search results shows old list, then loading component, then list again. For now we are following what prod mobile does, which is show a loading component on pop-back.
- Fixes case where search-results tracks have a secondary loading status (inside tracks.status) that is slightly different from searchStatus that was not being referenced, resulting in extra flash of "no-results" every time.


raymondjacobson (approved):
Learned something! cool! curious to see how it does in release

PR #1128, opened 2022-03-25
[AUD-1778] Improve now playing drawer performance

raymondjacobson:
### Description

Improves now playing drawer performance by
* Removing a lot of state updates
* Sharing translation anim between drawer & bottom bar
* Adding bottom "padding" elsewhere, which in its currently implementation is kinda jank

### Dragons

### How Has This Been Tested?

### DIFF
--- a/packages/mobile/src/components/drawer/Drawer.tsx
+++ b/packages/mobile/src/components/drawer/Drawer.tsx
@@ -2,7 +2,6 @@ import { ReactNode, useCallback, useEffect, useRef, useState } from 'react'
 
 import {
   Animated,
-  Dimensions,
   GestureResponderEvent,
   Image,
   ImageSourcePropType,
@@ -25,6 +24,8 @@ import { ThemeColors, useThemedStyles } from 'app/hooks/useThemedStyles'
 import { attachToDy } from 'app/utils/animation'
 import { useColor } from 'app/utils/theme'
 
+import { FULL_DRAWER_HEIGHT } from './constants'
+
 const MAX_SHADOW_OPACITY = 0.15
 const ON_MOVE_RESPONDER_DY = 10
 const MOVE_CUTOFF_CLOSE = 0.8
@@ -33,8 +34,6 @@ const BACKGROUND_OPACITY = 0.5
 
 // Controls the amount of friction in swiping when overflowing up or down
 const OVERFLOW_FRICTION = 4
-
-const { height } = Dimensions.get('window')
 
 const createStyles = (zIndex = 5, shouldAnimateShadow = true) => (
   themeColors: ThemeColors
@@ -184,9 +183,10 @@ export type DrawerProps = {
    */
   initialOffsetPosition?: number
   /**
-   * Whether or not the drawer is open partially (to the initial offset)
+   * Whether or not the drawer should close to the initial offset. i.e.
+   * has it been opened to the offset once?
    */
-  isOpenToInitialOffset?: boolean
+  shouldCloseToInitialOffset?: boolean
   /**
    * Whether or not we should be showing rounded or straight borders at
    * the initial offset if one is enabled.
@@ -311,7 +311,7 @@ export const Drawer: DrawerComponent = ({
   isGestureSupported = true,
   animationStyle = DrawerAnimationStyle.SPRINGY,
   initialOffsetPosition = 0,
-  isOpenToInitialOffset,
+  shouldCloseToInitialOffset,
   shouldHaveRoundedBordersAtInitialOffset = false,
   zIndex,
   drawerStyle,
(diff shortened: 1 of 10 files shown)


packages/mobile/src/screens/app-screen/AppTabBar.tsx
```diff
@@ -12,39 +12,12 @@ export const AppTabBar = (props: TabBarProps) => {
   // Set handlers for the NowPlayingDrawer and BottomTabBar
   // When the drawer is open, the bottom bar should hide (animated away).
   // When the drawer is closed, the bottom bar should reappear (animated in).
```
raymondjacobson, line 14: update comment

sliptype (approved):
This looks good. IMO it's not too bad to have the offset in the core components. But we could always create another wrapper component not in core that supplies the padding constant

raymondjacobson:
> This looks good. IMO it's not too bad to have the offset in the core components. But we could always create another wrapper component not in core that supplies the padding constant

Dramatically changed what we were doing here. But few things I also learned @dylanjeffers 
* We cannot animate the bottom "margin" as we were discussing. TranslateY actually has no impact on the layout of other elements in relation to the one we're animating, so `height: 20, translateY: -20` actually creates a ghost element at translateY: 0. My solution instead was to listen to the pan responder
* We were incurring a lot of drawer chunkiness by firing "onClose" at the same time as you panRelease. (Makes sense, we don't want to update the state of the drawer being open while you're releasing it). The solution is to add the onFinished callback and set open state after that. However b/c we use spring physics, it may take a while before the finish gets called since we're asymptotically bouncing toward steady state. Tweaking `restDisplacementThreshold` doesn't do as much as I hoped. So the better solution that's implemented here is to capture the "intent" of opening the drawer in a ref and then letting state get properly reflected when the animation settles. Or at least, I found this to be better. Code is still complex tho!
* Added the velocity to the panRelease scroll. I don't think it does a whole lot, but hard to tell w/o a full productionized build I think.

I really do think this PR is going to be a huge improvement!!

raymondjacobson:
Also.. another thing I learned here. We should really try to rely on 1 animation and interpolating that animation for anything we want to be remotely tied together. Really makes things simpler & removes the question of extra rerenders polluting things

packages/mobile/src/components/drawer/Drawer.tsx
```diff
@@ -299,42 +311,75 @@ export const Drawer: DrawerComponent = ({
...
   const borderRadiusAnim = useRef(new Animated.Value(BORDER_RADIUS)).current
   const backgroundOpacityAnim = useRef(new Animated.Value(0)).current
 
+  // Capture the intent of opening the drawer (for use in pan responder release).
+  // Generally when releasing a pan responder, we don't want to update state until
+  // the resulting animation has completed. The completion of that animation may be
+  // well after the user has started yet another pan gesture though, and we definitely
+  // want to capture the users previous intent so that our next pan gesture can be
+  // handled properly.
+  const isOpenIntent = useRef(isOpen)
```
dylanjeffers, line 349: this is cool and makes sense

packages/mobile/src/components/drawer/Drawer.tsx
```diff
@@ -489,11 +556,18 @@ export const Drawer: DrawerComponent = ({
...
               ? BORDER_RADIUS
               : BORDER_RADIUS * percentOpen * 5
-            const newBorderRadius = Math.min(
+            let newBorderRadius = Math.min(
               borderRadiusInitialOffset,
-              BORDER_RADIUS,
-              BORDER_RADIUS * 2 * (1 - percentOpen)
+              BORDER_RADIUS
             )
+            // On non-iOS platforms, bring the border radius back to 0 at the fully open state
```
dylanjeffers, line 563: cool addition!

packages/mobile/src/components/drawer/Drawer.tsx
```diff
@@ -506,30 +580,35 @@ export const Drawer: DrawerComponent = ({
...
         if (
           gestureState.vy > 0 &&
-          gestureState.moveY > height - MOVE_CUTOFF_CLOSE * drawerHeight
+          gestureState.moveY >
+            FULL_DRAWER_HEIGHT - MOVE_CUTOFF_CLOSE * drawerHeight
         ) {
-          if (isOpenToInitialOffset) {
-            slideOut(initialOffsetOpenPosition)
+          if (shouldCloseToInitialOffset) {
+            slideOut(initialOffsetOpenPosition, gestureState.vy, onClose)
```
dylanjeffers, line 595: ah so we put onClose here to avoid a state change till after the animation?

packages/mobile/src/components/drawer/Drawer.tsx
```diff
@@ -506,30 +580,35 @@ export const Drawer: DrawerComponent = ({
...
+            isOpenIntent.current = false
             borderRadiusAnim.setValue(0)
-            onClose()
           } else {
-            slideOut(initialPosition)
-            onClose()
+            slideOut(initialPosition, gestureState.vy, onClose)
```
dylanjeffers, line 599: so are there cases where a quick slideOut, then slideIn animation would want us to cancel the onClose call, or a guess that is handled with the "finished" boolean in the animation onFinished callback?

packages/mobile/src/components/drawer/Drawer.tsx
```diff
@@ -469,16 +533,19 @@ export const Drawer: DrawerComponent = ({
           if (onPercentOpen) onPercentOpen(percentOpen)
         } else if (gestureState.dy < 0) {
           // Dragging upwards
-          const percentOpen = (-1 * gestureState.dy) / drawerHeight
+          const percentOpen =
+            (-1 *
+              (isOpenIntent.current ? -1 * drawerHeight : gestureState.dy)) /
```
dylanjeffers, line 538: could i get a little bit of understanding what this ternery is doing?

dylanjeffers (approved):
soooo much better, great job! just a few comments for my own understanding, but code wise, looks great

packages/mobile/src/components/drawer/Drawer.tsx
raymondjacobson: yee

packages/mobile/src/components/drawer/Drawer.tsx
raymondjacobson: precisely! the finished thing is pretty cool

packages/mobile/src/components/drawer/Drawer.tsx
raymondjacobson: will add a comment. essentially this makes percent bounded to [0,1] where it wasn't before. kinda an unrelated bug that got worse by sharing more animation state. overscrolling being exposed means wed do stuff like change the opacity of the play bar to <0

PR #1129, opened 2022-03-25
Add drag auto scroll capability for custom scrollbar containers

nicoback2:
### Description
autoscroll

Sadly the new scrollbar component doesn't play well with native drag and drop scrolling - see https://github.com/mdbootstrap/perfect-scrollbar/issues/100 and https://stackoverflow.com/questions/55482184/auto-scroll-while-dragging-an-item-with-perfect-scrollbar-and-sortablejs.

This means that when you drag something to the top or bottom of the nav container, it doesn't scroll automatically like it used to.
This PR adds back that functionality manually (+ a shadow indicator when you're drag-scrolling up or down).

### Dragons

### How Has This Been Tested?
Tested in browser

### DIFF
--- a/packages/web/src/components/drag-autoscroller/DragAutoscroller.tsx
+++ b/packages/web/src/components/drag-autoscroller/DragAutoscroller.tsx
@@ -1,0 +1,96 @@
+import React, { useRef, useState } from 'react'
+
+import throttle from 'lodash/throttle'
+
+export type DragAutoscrollerProps = {
+  children: React.ReactNode
+  containerBoundaries: {
+    top: number
+    bottom: number
+    left: number
+    right: number
+  }
+  updateScrollTopPosition: (difference: number) => void
+  onChangeDragScrollingDirection: (
+    newDirection: 'up' | 'down' | undefined
+  ) => void
+}
+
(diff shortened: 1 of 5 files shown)


packages/web/src/components/drag-autoscroller/DragAutoscroller.tsx
```diff
@@ -0,0 +1,96 @@
+import React, { useRef, useState } from 'react'
```
nicoback2, line 1: Wonder if this should go in Stems instead? 🤔

audius-infra:
Preview this change https://demo.audius.co/nkang--autoscoll-dnd

packages/web/src/components/drag-autoscroller/DragAutoscroller.tsx
rickyrombo: I think that makes sense. We can start here though and graduate it if we see it needed elsewhere

rickyrombo (approved):
Some nits for the "magic numbers" but otherwise looks good!

raymondjacobson:
pls bug julian for the UI if you haven't already!

audius-infra:
Preview this change https://demo.audius.co/nkang--autoscoll-dnd

PR #1130, opened 2022-03-25
[AUD-1779] Improve track tile click animation fluidity

dylanjeffers:
### Description

Improves track tile click animation by delaying the cpu intensive tasks (queueing) until after the animation finishes

### Dragons

- Obviously this slightly slows down play/pause responsiveness. In testing it seems okay, but something to watch out for.
- I also considered adding the `setImmediate` to all tile `onPress` actions since that would ensure all animations are smooth. Curious what folks think about that idea and its tradeoffs


dylanjeffers:
> curious to play with on-device. i think the biggest thing would just be to see if once we are on react native reloaded, whether this is necessary stil

yeah good calls, added a comment that we should consider removing once reloaded is active

PR #1140, opened 2022-03-28
[AUD-1671] Prevent Drawer from popping to top of screen

sliptype:
### Description

BIG PR

* Fixes a bug where the drawer would pop to the top of the screen. This usually happened when lots of other processing was going on and the `onLayout` callback wasn't called immediately

### Dragons

Affects drawers, but the actual height should always get set and therefore this shouldn't be dangerous

### How Has This Been Tested?

iOS sim

TestFlight QA


dylanjeffers (approved):
interesting, nice

PR #1144, opened 2022-03-29
[AUD-1676, AUD-1677] Update dark mode colors 

Kyle-Shanks:
### Description
[AUD-1676] [AUD-1677]

* Update tile component shadows to pull from theme
* Update verified icon to have a white check mark
* Small fix for dividers in track list

### Dragons

N/A

### How Has This Been Tested?

Manually tested, things look good

simulator_screenshot_4A176436-A148-4506-BDEC-FB14805707AD

N/A


packages/mobile/src/components/core/Tile/Tile.tsx
```diff
@@ -114,21 +116,23 @@ export const Tile = <
       <TileShadow
         offset={[0, 0]}
         distance={1}
-        startColor='rgba(133, 129, 153, 0.1)'
+        // Approximately 10% opacity on neutral
+        startColor={themeColors.neutral + '1a'}
```
raymondjacobson, line 120: nice. we should chat with julian sometime about how to do shadows in an easier way for us

raymondjacobson (approved):

packages/mobile/src/components/core/Tile/Tile.tsx
Kyle-Shanks: yea i agree. should maybe have specific colors built into the themes for the shadows, or a helper function to extract the shadow values from the theme. something like that

PR #1145, opened 2022-03-29
[AUD-1739] Add missing haptics

dylanjeffers:
### Description

Adds missing haptics to favorite and reposts buttons, and segmented control


sliptype (approved):
Easy win! Nice

PR #1146, opened 2022-03-30
[AUD-1766] NowPlayingDrawer layout

sliptype:
### Description

* Makes the NowPlayingDrawer layout more dynamic to fix display on smaller screens
iPhone 13:
image

iPhone 8:
image
* Fixes full screen drawers
* Fixes issues with NowPlaying translation offsets and height
* Disables StatusBar hiding on screens with smaller top insets to avoid layout shifting during animation
* Swaps type parameters for `makeStyles`
* Fixes text color for RewardsBanner in dark mode

### Dragons

Flex box layout could have unexpected issues on different screen sizes

### How Has This Been Tested?

iOS sim and iOS device

TestFlight QA


packages/mobile/src/components/drawer/Drawer.tsx
```diff
@@ -651,6 +658,35 @@ export const Drawer: DrawerComponent = ({
     }
   }
 
+  const renderContent = () => {
+    const Component = disableSafeAreaView ? View : SafeAreaView
+
+    return (
+      <Component
```
dylanjeffers, line 665: maybe ViewComponent?

packages/mobile/src/components/drawer/Drawer.tsx
```diff
@@ -651,6 +658,35 @@ export const Drawer: DrawerComponent = ({
...
+        style={isFullscreen ? styles.fullScreenContent : styles.content}
+        onLayout={(event: LayoutChangeEvent) => {
+          if (!isFullscreen) {
+            const { height } = event.nativeEvent.layout
+            setDrawerHeight(height)
+          }
+        }}
+        {...(disableSafeAreaView
```
dylanjeffers, line 673: might be nice to have this as a const before the return just to clean up the jsx here? also not sure if you can do `const viewProps = disableSafeAreaView ? null : {edges: ....}`

packages/mobile/src/components/now-playing-drawer/Artwork.tsx
```diff
@@ -20,20 +20,21 @@ const createStyles = (themeColors: ThemeColors) =>
     root: {
       marginLeft: spacing,
       marginRight: spacing,
-      width: dimensions.width - spacing * 2,
-      height: dimensions.width - spacing * 2
+      maxHeight: dimensions.width - spacing * 2,
+      alignSelf: 'center'
```
dylanjeffers, line 24: ah that's cool, what dos this do ultimately?

packages/mobile/src/components/now-playing-drawer/Artwork.tsx
```diff
@@ -20,20 +20,21 @@ const createStyles = (themeColors: ThemeColors) =>
...
     image: {
+      alignSelf: 'center',
       borderRadius: 8,
       borderColor: themeColors.white,
       borderWidth: 2,
       overflow: 'hidden',
       height: '100%',
-      width: '100%'
+      width: '100%',
+      aspectRatio: 1
```
dylanjeffers, line 37: ah that's cool.

packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
```diff
@@ -41,28 +41,37 @@ import { TitleBar } from './TitleBar'
 import { TrackInfo } from './TrackInfo'
 import { PLAY_BAR_HEIGHT } from './constants'
 
-const combinedBottomAreaHeight = BOTTOM_BAR_HEIGHT + PLAY_BAR_HEIGHT
-
 const STATUS_BAR_FADE_CUTOFF = 0.6
 const SKIP_DURATION_SEC = 15
+// If the top screen inset is greater than this,
+// the status bar will be hidden when the drawer is open
+const INSET_STATUS_BAR_HIDE_THRESHOLD = 20
```
dylanjeffers, line 48: cool, so is this like "hide for phones without notches"?

packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
```diff
@@ -41,28 +41,37 @@ import { TitleBar } from './TitleBar'
...
 
 const useStyles = makeStyles(({ spacing }) => ({
   container: {
     paddingTop: 0,
-    height: FULL_DRAWER_HEIGHT - combinedBottomAreaHeight
+    height: FULL_DRAWER_HEIGHT,
+    justifyContent: 'space-evenly'
+  },
+  playBarContainer: {
```
dylanjeffers, line 56: ah, and is this so playbar is not calculated in overall drawer height?

dylanjeffers (approved):
wow really awesome work!

packages/mobile/src/components/now-playing-drawer/Artwork.tsx
sliptype: It allows the image to scale to its container but never bigger than the width of the screen

packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
sliptype: Yeah, on small screens (without a notch) the drag would get interrupted because the status bar added/removed extra space at the top

packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
sliptype: Pretty much, this makes it so there isn't a big blank space at the top of the drawer where the playbar would be. Important for devices without a notch

PR #1149, opened 2022-03-31
Update AnimatedButton and colorize icons passed to animated buttons

Kyle-Shanks:
### Description
* Colorize all of the icons
* Updated AnimatedButton to only accept one json for the icon since it is colorized

### Dragons

N/A

### How Has This Been Tested?

Manually Tested, colors are looking good 👍 
Simulator Screen Shot - iPhone 13 - 2022-03-31 at 14 45 46
simulator_screenshot_EACA8840-29FC-4961-B916-9EC54C10D26D

N/A


Kyle-Shanks:
> This is sick! I do think it's kinda a lot to have all that colorize config in the components themselves, might be cool to do a `colorizedIcons` folder containing individual colorized icons. But I'm fine with this for now!

Yea I agree, I'll move them now actually 👍

PR #1151, opened 2022-04-01
Consume libs 1.2.98

sddioulde:
### Description

For this change https://github.com/AudiusProject/audius-protocol/pull/2801

### Dragons

that rewards error is handled correctly

### How Has This Been Tested?

locally against staging

flow already has existing logs and analytics


audius-infra:
Preview this change https://demo.audius.co/sd-consume-libs

PR #1152, opened 2022-04-01
[AUD-1805] Remove lazy screen option for profile

dylanjeffers:
### Description

Removes lazy option which removes strange profile scroll bug

### DIFF
--- a/packages/mobile/src/components/top-tab-bar/CollapsibleTopTabNavigator.tsx
+++ b/packages/mobile/src/components/top-tab-bar/CollapsibleTopTabNavigator.tsx
@@ -69,9 +69,6 @@ type CollapsibleTabNavigatorProps = {
   initialScreenName?: string
   children: ReactNode
   screenOptions?: MaterialTopTabNavigationOptions
-  collapsibleOptions?: {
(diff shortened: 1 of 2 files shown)


packages/mobile/src/components/top-tab-bar/CollapsibleTopTabNavigator.tsx
```diff
@@ -88,7 +92,8 @@ export const CollapsibleTabNavigator = ({
         // Lazy empirically helps with reducing "content offset" jumping.
         // Potentially related issue (not from the v2 codebase though)
         // https://github.com/PedroBern/react-native-collapsible-tab-view/pull/120
-        lazy: true
+        lazy: true,
+        ...collapsibleOptions
```
raymondjacobson, line 96: id just default false

raymondjacobson (approved):

packages/mobile/src/components/top-tab-bar/CollapsibleTopTabNavigator.tsx
dylanjeffers: word!

PR #1154, opened 2022-04-01
[AUD-1781] Fix issue with feed not loading after sign in

sliptype:
### Description

* Fixes the issue with the feed not loading after sign in

This was due to the strange boolean logic we have to switch between `SignOnStack` and `MainStack`. `authed` is always true initially because the dapp has not yet loaded, which causes the feed screen to render while the splash screen is still visible. The feed lineup uses `selfLoad` which is fine when the account already exists in state, but when signing in the feed has already tried to load with no user and failed. Then no new fetch is performed on the lineup

We should probably improve the boolean logic there, could be as simple as checking `signedIn`. But trying that resulted in a flash of blank screen after the splash screen because the feed screen takes a second to render

The kinda hacky fix here is to set `selfLoad` based on `signedIn` so the feed renders initially but doesn't load until the user signs in

### Dragons

This should be pretty harmless because we require all users to sign in to see the feed

### How Has This Been Tested?

iOS sim. Ran through sign on, and sign up flows

TestFlight QA

### DIFF
--- a/packages/mobile/src/screens/feed-screen/FeedScreen.tsx
+++ b/packages/mobile/src/screens/feed-screen/FeedScreen.tsx
@@ -9,12 +9,14 @@ import {
 } from 'audius-client/src/common/store/pages/feed/selectors'
 import { setVisibility } from 'audius-client/src/common/store/ui/modals/slice'
 import { isEqual, omit } from 'lodash'
+import { useSelector } from 'react-redux'
 
 import { Screen } from 'app/components/core'
 import { Header } from 'app/components/header'
 import { Lineup } from 'app/components/lineup'
 import { useDispatchWeb } from 'app/hooks/useDispatchWeb'
 import { useSelectorWeb } from 'app/hooks/useSelectorWeb'
+import { getIsSignedIn } from 'app/store/lifecycle/selectors'
 import { make, track } from 'app/utils/analytics'
 
 import { FeedFilterButton } from './FeedFilterButton'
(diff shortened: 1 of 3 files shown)


packages/mobile/src/screens/feed-screen/FeedScreen.tsx
```diff
@@ -72,7 +75,7 @@ export const FeedScreen = () => {
         loadMore={loadMore}
         refresh={handleRefresh}
         refreshing={isRefreshing}
-        selfLoad
+        selfLoad={!!signedIn}
```
dylanjeffers, line 78: so for this basically it will self load all times except when initially signing in? or when it becomes true does that cause a selfLoad fetch only when we are actually signed in?

packages/mobile/src/screens/root-screen/RootScreen.tsx
```diff
@@ -78,6 +78,9 @@ export const RootScreen = () => {
   const onSignUp = useSelector(getOnSignUp)
   const isAccountAvailable = useSelector(getAccountAvailable)
 
+  // This check is overly complicated and should probably just check `signedIn`.
```
dylanjeffers, line 81: nice comment

dylanjeffers (approved):

packages/mobile/src/screens/feed-screen/FeedScreen.tsx
sliptype: Ok so I also ended up changing from `useEffectOnce` to `useEffect` in Lineup with a dependency on `selfLoad`. 

In the case where you are already signed in:
The lineup will render while the splash screen is showing. Then, once signedIn is true, it will fetch the data.

In the case where you aren't signed in yet:
The lineup will also render initially while the splash screen is showing, but is replaced by the sign on screen (this is due to the wonky logic I was referencing). Then, once signedIn is true, it will fetch the data.

PR #1155, opened 2022-04-01
[AUD-1810] Add docs to run local libs in audius-client

dylanjeffers:
### Description

Adds docs to run libs locally in audius-client, which has changed slightly due to libs now needing to be transpiled.


audius-infra:
Preview this change https://demo.audius.co/dj-aud-1810-libs-docs

sddioulde (approved):

PR #1159, opened 2022-04-04
[AUD-1784] Fix ReachabilityBar height

dylanjeffers:
### Description

Fixes case where small purple line of the reachability bar is showing on smaller devices

### DIFF
--- a/packages/mobile/src/components/reachability-bar/ReachabilityBar.tsx
+++ b/packages/mobile/src/components/reachability-bar/ReachabilityBar.tsx
@@ -17,6 +17,8 @@ const springConfig = {
   friction: 20
 }
 
+const barHeight = 40
+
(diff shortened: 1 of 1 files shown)


raymondjacobson (approved):
lgtm! im not sure how this was an off by one error tho?

dylanjeffers:
> lgtm! im not sure how this was an off by one error tho?

I think it had to do with padding(2.5) somehow being off by one? capping the height at 40 solves the problem though, and is prob more correct. what you think?

raymondjacobson:
> > lgtm! im not sure how this was an off by one error tho?
> 
> I think it had to do with padding(2.5) somehow being off by one? capping the height at 40 solves the problem though, and is prob more correct. what you think?

lgtm!

PR #1160, opened 2022-04-04
[AUD-1806] Update SegmentedControl component to set initial width and left position

Kyle-Shanks:
### Description
[AUD-1806]

* Updated the SegmentedControl component to set initial width and left position
* Changed optionsWidths to use the state hook so that the initial position would be correct

### Dragons

N/a

### How Has This Been Tested?

Manually tested

N/a


sliptype (approved):
Nice!!

PR #1162, opened 2022-04-05
Allow confirmer calls to be parallelized and squashed + allow consolidation of success callbacks

nicoback2:
### Description
Confirmer optimizations!

-Currently, confirmer calls are grouped by entity (`uid`). This PR allows confirmer calls to be further sub-grouped by operation (`operationId`). The requester can then specify that calls in the same operation sub-group may be parallelized or squashable (_squashable_ meaning that only the first and last call must be called, whatever's in the middle can be squashed). Additionally, the requester can specify that only the success callback last resolved confirm in an operation group be invoked.

The options detailed above are passed in through a new object options param to `requestConfirmation`, `ConfirmationOptions`. 

The four actions modified to use these new optimization options are: 
1. Add track to playlist (parallelize + use only last success callback)
2. Remove track from playlist (parallelize + use only last success callback)
3. Reorder track in a playlist (squash)
4. Reorder playlists in the library ('update profile') (squash)

Follow ups:
-Type the confirmer saga without the use of `any`s
-Use `ConfirmationOptions` for additional flows (e.g. squashing favorite/unfavorite calls?)

### Dragons

This is a pretty risky change since it has the potential to screw up core flows if there are bugs. If implemented correctly, there should be no change to existing confirmation flows besides the four that were outlined above.

### How Has This Been Tested?

Tested in browser, some unit tests
-Adding/removing/reordering tracks in playlists in any order
-Reordering playlists in left nav
-Performing other flows in product that trigger request confirmations

Monitoring Sentry after deployment

confirmergif

confirmergif2


packages/web/src/common/store/cache/collections/sagas.js
```diff
@@ -1,3 +1,4 @@
+import isEqual from 'lodash/isEqual'
```
raymondjacobson, line 1: we should settle on this convention vs. what we do elsewhere https://github.com/AudiusProject/audius-client/blob/6488d30511a49d0c379bfdc2651f09c6d775d189/packages/mobile/src/hooks/selectors.ts#L2

packages/web/src/common/store/cache/collections/sagas.js
```diff
@@ -448,20 +452,49 @@ function* confirmAddTrackToPlaylist(userId, playlistId, trackId, count) {
...
+        ))[0]
+
+        const playlist = yield select(getCollection, { id: playlistId })
+
+        /** Since "add track" calls are parallelized, tracks may be added
+         * out of order. Here we check if tracks made it in the intended order;
+         * if not, we reorder them into the correct order.
+         */
+        if (
+          confirmedPlaylist.playlist_contents.track_ids.length ===
```
raymondjacobson, line 470: might be nice & self documenting to pull these out into vars

```
const numberOfTracksMatch = confirmedPlaylist.playlist_contents.track_ids.length == playlist.playlist-contents.track_ids.length
const confirmedPlaylistHasTracks = ...


if (numberOfTracksMatch && confirmedPlaylistHasTracks) {...}

```

packages/web/src/common/store/cache/collections/utils/retrieveCollections.ts
```diff
@@ -86,7 +86,7 @@ export function* retrieveTracksForCollections(
  * Retrieves a single collection via API client
  * @param playlistId
  */
-function* retrieveCollection(playlistId: ID) {
+export function* retrieveCollection(playlistId: ID) {
```
raymondjacobson, line 89: wow surprised this wasn't exported before

packages/web/src/components/confirmer-preview/ConfirmerPreview.tsx
```diff
@@ -11,6 +11,7 @@ type ConfirmerPreviewProps = {} & ReturnType<typeof mapStateToProps>
 
 const ConfirmerPreview = ({ confirmer }: ConfirmerPreviewProps) => {
   const entities = Object.keys(confirmer.confirm)
+  console.log(confirmer.confirm, 'confirmer contents')
```
raymondjacobson, line 14: 🪵

packages/web/src/store/confirmer/actions.ts
```diff
@@ -0,0 +1,198 @@
...
+      successCall,
+      failCall,
+      previousResultSelector,
+      timeoutMillis,
+      confirmationOptions
+    }
+  }
+)
+
+/* Private */
```
raymondjacobson, line 99: wdyt of `_prefix` notation?

packages/web/src/store/confirmer/actions.ts
```diff
@@ -0,0 +1,198 @@
...
+
+/** Enqueues a confirmation call to make, e.g. "create playlist with a temp id, wait for playlist to appear in discprov." */
```
raymondjacobson, line 101: nit discovery node (just to keep on disambiguating down the line)

packages/web/src/common/store/cache/collections/sagas.js
```diff
@@ -473,7 +506,13 @@ function* confirmAddTrackToPlaylist(userId, playlistId, trackId, count) {
           )
         )
       },
-      result => (result.playlist_id ? result.playlist_id : playlistId)
+      result => (result.playlist_id ? result.playlist_id : playlistId),
+      undefined,
+      {
+        operationId: 'ADD_TRACK',
```
raymondjacobson, line 512: I wonder if we could infer ADD_TRACK from something? or at least pull it in from an action type or something to make it bound

packages/web/src/pages/profile-page/sagas.js
```diff
@@ -436,7 +436,10 @@ function* confirmUpdateProfile(userId, metadata) {
       },
       function* () {
         yield put(profileActions.updateProfileFailed())
-      }
+      },
+      undefined,
+      undefined,
+      { operationId: 'OVERWRITE', squashable: true }
```
raymondjacobson, line 442: eventually playlist & track updates should be able to be squashable too. will that be a problem if they are also of operationId `OVERWRITE`?

raymondjacobson (approved):
so great!!

packages/web/src/store/confirmer/actions.ts
```diff
@@ -0,0 +1,198 @@
...
+  useOnlyLastSuccessCall
+}: ConfirmationOptions) => {
+  if (
+    (squashable === true ||
+      parallelizable === true ||
+      useOnlyLastSuccessCall === true) &&
+    operationId == null
+  ) {
+    throw new Error(
+      'Programming error - must pass `operationId` in confirmation options if `squashable` or `parallelizable` is also passed.'
```
sddioulde, line 34: nit: ...`squashable` or `parallelizable` or `useOnlyLastSuccessCall`

nicoback2:
tysm for review @raymondjacobson and @sddioulde 🙏 🙏 will revise and merge today!

packages/web/src/pages/profile-page/sagas.js
nicoback2: nope will be fine since they will have different entity ids (i.e. be in different "confirm groups")! calls are grouped by `uid` first then `operationId`

packages/web/src/store/confirmer/actions.ts
nicoback2: i like it! right now you can't tell what's private vs not without having to scroll a bunch - will add in the prefix

packages/web/src/components/confirmer-preview/ConfirmerPreview.tsx
nicoback2: nooo

packages/web/src/common/store/cache/collections/utils/retrieveCollections.ts
nicoback2: yeah I think until now we've always use the `retrieveCollections` function (for which this one is a helper) below which tries to get the playlist from cache first/automatically caches the result after

packages/web/src/common/store/cache/collections/sagas.js
nicoback2: ok will do - i used the other way to avoid importing the full library (since idt we have tree shaking? cc @dylanjeffers) but since we use `import { x } from 'lodash'` in a bunch of places it wouldn't make a difference on bundle size anyway

packages/web/src/common/store/cache/collections/sagas.js
nicoback2: ohh yeah ill explore this a it. Although, we might want to keep this free so that multiple actions can use the same operation id (if whatever confirm calls they make can be grouped together)

packages/web/src/common/store/cache/collections/sagas.js
raymondjacobson: i think import { x } from lo dash does tree shake. but actually what i'm saying is i like yours better :)

packages/web/src/common/store/cache/collections/sagas.js
raymondjacobson: usually i'm no enum, but enum might be nice for op id

packages/web/src/common/store/cache/collections/sagas.js
dylanjeffers: Personally I like import {} since we have a babel plugin that modifies them into import blah from lodash/blah

raymondjacobson: sgtm!

audius-infra:
Preview this change https://demo.audius.co/nkang--confirmer

PR #1167, opened 2022-04-06
[AUD-1817] Fix PlayBar opacity

dylanjeffers:
### Description

Fixes issue where PlayBar has opacity < 1 at resting point since opacity interp is not taking into place it's height relative to the NowPlayingDrawer height.

### Dragons

Down to make `NOW_PLAYING_HEIGHT - PLAY_BAR_HEIGHT a constant, but not sure of the name


raymondjacobson (approved):
interp

PR #1170, opened 2022-04-06
Fix libs browser config

dylanjeffers:
### Description

Bumps libs to a version where libs browser config does not exclude a dependency that is required.


audius-infra:
Preview this change https://demo.audius.co/dj-fix-stage-libs

raymondjacobson (approved):

PR #1171, opened 2022-04-06
[AUD-1800] Remove pull-to-refresh on track screen

dylanjeffers:
### Description

Removes pull-to-refresh on track screen. Also goes the extra step and removes refresh indicator for any section-list that doesn't explicitly pass down an `onRefresh` function. Curious if the way it was implemented is best!

### DIFF
--- a/packages/mobile/src/components/core/SectionList.tsx
+++ b/packages/mobile/src/components/core/SectionList.tsx
@@ -40,10 +40,8 @@ const useCollapsibleSectionListScene = (sceneName: string) => {
   }
 }
 
-const CollapsibleSectionList = ({
-  sceneName,
-  ...other
-}: CollapsibleSectionListProps) => {
+const CollapsibleSectionList = (props: CollapsibleSectionListProps) => {
+  const { sceneName, ...other } = props
   const { refreshing, onRefresh, scrollY: collapsibleScrollAnim } = useContext(
     CollapsibleTabNavigatorContext
   )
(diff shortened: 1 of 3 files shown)


packages/mobile/src/components/core/SectionList.tsx
```diff
@@ -53,25 +51,28 @@ const CollapsibleSectionList = ({
...
   )
 }
 
 const AnimatedSectionList = forwardRef<RNSectionList, SectionListProps>(
   function AnimatedSectionList(
-    { refreshing, onRefresh, ...other },
+    props,
     ref: MutableRefObject<RNSectionList<any, DefaultSectionT> | null>
   ) {
+    const { refreshing, onRefresh, ...other } = props
```
sliptype, line 75: Why prefer destructuring in the function body?

sliptype (approved):
Looks good! Should we do the same for `FlatList`?

dylanjeffers:
> Looks good! Should we do the same for `FlatList`?

Definitely, good call

packages/mobile/src/components/core/SectionList.tsx
dylanjeffers: i like how it cleans up the `const Blah = () => ` part of things. Also sometimes I've had cases where we want the full `props` object and just need to reference a part of props in body, sometimes destructure some and create ...other, so this way it can be the same approach, if that makes sense? it also cleans up docs: 
image
image

sliptype: Yeah I definitely have run into the case of needing the whole props object, I think being consistent make sense

PR #1172, opened 2022-04-06
[AUD-1824] Improve track-tile skeleton transition

dylanjeffers:
### Description

Improves the transition handoff from track-tile skeleton to track-tile content by starting the content fade-in at 50%, which is a similar color to the skeleton gray color. Happy to continue adjusting if it seems like it could be better!

https://user-images.githubusercontent.com/8230000/162079739-bb37081a-f8ac-451f-906b-5f368b81c531.mov

With the dynamic image opacity change:

https://user-images.githubusercontent.com/8230000/162080902-7ceebc81-8406-4c2d-8121-2569d00d228c.mov

### DIFF
--- a/packages/mobile/src/components/core/DynamicImage.tsx
+++ b/packages/mobile/src/components/core/DynamicImage.tsx
@@ -104,8 +104,8 @@ export const DynamicImage = memo(function DynamicImage({
   const [firstImage, setFirstImage] = useState<string>()
   const [secondImage, setSecondImage] = useState<string>()
 
-  const firstOpacity = useRef(new Animated.Value(0)).current
-  const secondOpacity = useRef(new Animated.Value(0)).current
+  const firstOpacity = useRef(new Animated.Value(0.5)).current
+  const secondOpacity = useRef(new Animated.Value(0.5)).current
 
   const [isFirstImageActive, setIsFirstImageActive] = useState(true)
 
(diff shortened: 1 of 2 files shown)


sliptype (approved):
I like this! I think the most jarring part of the transition is the dynamic image going completely white for a frame

Kyle-Shanks (approved):
I agree with Sebastian, this looks better for sure though

dylanjeffers:
> I like this! I think the most jarring part of the transition is the dynamic image going completely white for a frame

ah yeah great call, can see what i can do with dynamic image!

sliptype:
The Dynamic Image looks good but this might be a breaking change when two images swap. We only swap when new image sizes get loaded, but you could set up a contrived test that swaps between two images to make sure that transition is still smooth

dylanjeffers:
> The Dynamic Image looks good but this might be a breaking change when two images swap. We only swap when new image sizes get loaded, but you could set up a contrived test that swaps between two images to make sure that transition is still smooth

So i thinkk this happens when we upload an image in edit profile, here's what it looked like, looks aight to me i think!

https://user-images.githubusercontent.com/8230000/162081964-a989c91a-4e1c-4c7b-9787-f6eb9188da77.mov

raymondjacobson:
This is fab

PR #1174, opened 2022-04-07
[AUD-1820] Prevent Tiles from triggering when scrolling

dylanjeffers:
### Description

When scrolling through a lineup, occasionally the user is redirected to a track or artist screen, or the app mistakenly plays a track. A fix for this is discussed here: https://github.com/callstack/react-native-pager-view/issues/424#issuecomment-1022399796 and implemented in this PR.

- Creates a base Pressable component that prevents `onPress` when it perceives a user scroll.
- Uses base Pressable in Tile, and TileMetadata. Thinking we can start with this, and then add it in more areas as needed!


packages/mobile/src/components/core/Pressable.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const handlePress = useCallback(
+    (event: GestureResponderEvent) => {
+      if (!activeTouchPositionRef.current) return
+
+      const { pageX, pageY } = event.nativeEvent
+
+      const absX = Math.abs(activeTouchPositionRef.current.pageX - pageX)
+      const absY = Math.abs(activeTouchPositionRef.current.pageY - pageY)
+
+      const dragged = absX > 2 || absY > 2
```
raymondjacobson, line 41: worth pulling 2 into a const so it's clear that it's the abs pos change for scroll?  separately kinda curious why not 1 px

raymondjacobson (approved):

packages/mobile/src/components/core/Pressable.tsx
dylanjeffers: yeah no idea on why 2, that's what they used for the example, will extract and clarify the values here too, prob gonna use dx and dy instead of absX absY

raymondjacobson (approved):

PR #1175, opened 2022-04-07
Remove react-native-shadow-2 from tiles to improve performance

sliptype:
### Description

* Removes `react-native-shadow-2` shadows from tiles to improve performance of lineups, especially in tabs

After (on left) and before (on right)
image

Still using `react-native-shadow-2` where we need colored and more custom shadows

### Dragons

N/A

### How Has This Been Tested?

iOS simulator

TestFlight QA


dylanjeffers (approved):
Nice, great find, interested to see how much smoother app is on device

PR #1180, opened 2022-04-07
[AUD-1807] Fix Chromecast permissions

sliptype:
### Description

* Uses the provided `CastButton` to trigger the cast dialog so that network permissions are properly obtained
image
image
image
image

### Dragons

There isn't a good way to render the provided `CastButton` in the background and trigger it. So we are kinda tied to the icon and animations provided out of the box. However, with styling the button looks pretty similar to the rest of the UI

### How Has This Been Tested?

iOS device in release config

TestFlight QA


raymondjacobson (approved):
lgtm! thx dawg

PR #1181, opened 2022-04-07
[AUD-1787] [WIP] Scrollable profile header

sliptype:
### Description

* Removes `react-native-collapsible-tabs` in favor of a top level flatlist implementation

Pros:
  * Can scroll the profile header, even on pressables!
  * Buttery smooth tab transitions

Cons:
  * Tab content lists do not have independent scroll (Can this be solved?)

### Dragons

Couple broken things:
  * long pressing the bottom tab or pressing the currently active tab is not scrolling to top (should be an easy fix)
  * In some cases the bottom of the lineup stops loading, this might have to do with the constants defined in `Lineup` and how they are interacting with the offset created by the tabs/header

### How Has This Been Tested?

iOS sim

Not necessarily being merged


packages/mobile/src/components/top-tab-bar/CollapsibleTopTabNavigator.tsx
```diff
@@ -69,34 +37,72 @@ type CollapsibleTabNavigatorProps = {
...
-        ...screenOptions
+    <FlatList
+      data={[0, 1, 2]}
+      scrollAnim={animatedValue}
+      renderItem={({ index }) => {
+        if (index === 0) {
+          return renderHeader() as any
+        }
+        if (index === 1) {
+          return <PortalHost name={tabBarPortalHostName} />
```
sliptype, line 81: This is the fun part

PR #1183, opened 2022-04-08
Upgrade libs to 1.2.105, adds constants patch

dylanjeffers:
### Description

Upgrades libs to latest, and includes fixes for constants that were imported directly from libs/src, which isn't supported anymore.


audius-infra:
Preview this change https://demo.audius.co/dj-upgrade-libs

raymondjacobson (approved):
I don't love, but the comment is good

PR #1186, opened 2022-04-08
Move apple site association to /.well-known

sliptype:
### Description

Currently the app banner isn't showing in production because apple is unable to fetch our site association file. Moving to `/.well-known` should fix this

image

image

### Dragons

I had to make a slight change to the way the build was copied when testing via `dot-audius-dot-co`. I believe no change is necessary for the prod workers site because the build is moved rather than copied

### How Has This Been Tested?

Locally, and on a workers site on `sebastian.audius.co`

Check that validator passes on staging & prod once deployed


audius-infra:
Preview this change https://demo.audius.co/sk-move-apple-site-association

PR #1188, opened 2022-04-08
[AUD-1858] Move isPlayingUid logic back into TrackTile and CollectionTile and fix Collectiontile play-pause logic

Kyle-Shanks:
### Description
[AUD-1858]

### Dragons

There may be unintended regression added to certain flows with the TrackTile. Please keep an eye out for that

### How Has This Been Tested?

Manually tested. Play-pause for collection tile works and the title switched to primary color

N/A

### DIFF
--- a/packages/mobile/src/components/lineup-tile/TrackTile.tsx
+++ b/packages/mobile/src/components/lineup-tile/TrackTile.tsx
@@ -27,12 +27,15 @@ import { requestOpen as requestOpenShareModal } from 'audius-client/src/common/s
 import { RepostType } from 'audius-client/src/common/store/user-list/reposts/types'
 import { open as openOverflowMenu } from 'common/store/ui/mobile-overflow-menu/slice'
 import { isEqual } from 'lodash'
+import { useSelector } from 'react-redux'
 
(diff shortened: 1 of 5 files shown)


packages/mobile/src/components/lineup-tile/TrackTile.tsx
```diff
@@ -78,6 +80,11 @@ const TrackTileComponent = ({
   const dispatchWeb = useDispatchWeb()
   const navigation = useNavigation()
   const currentUserId = useSelectorWeb(getUserId)
+  const playingUid = useSelector(getPlayingUid)
```
sliptype, line 83: This selector will cause every track tile to rerender when the playingUid changes. Using the selector that was in `LineupTile` prevents this! 

``` 
const isPlayingUid = useSelector(
    (state: AppState) => getPlayingUid(state) === uid
  )
```

sliptype (approved):
Good stuff! Just one comment about a selector change

packages/mobile/src/components/lineup-tile/TrackTile.tsx
Kyle-Shanks: Right, good point. Will update 👌

PR #1189, opened 2022-04-09
[AUD-1845]  Configure deep linking

sliptype:
### Description

* Configures mapping so deep links go to correct screens

### Dragons

* Need to run in release configuration to test links when the app is not already running. 
* Came across some logic for handling recovery params, not sure if we already have this implemented in the new app. Will need to double check on this

### How Has This Been Tested?

iOS sim

TestFlight QA

### DIFF
--- a/packages/mobile/src/components/navigation-container/NavigationContainer.tsx
+++ b/packages/mobile/src/components/navigation-container/NavigationContainer.tsx
@@ -1,10 +1,14 @@
 import { ReactNode, useContext } from 'react'
 
 import {
+  getStateFromPath,
   LinkingOptions,
   NavigationContainer as RNNavigationContainer
 } from '@react-navigation/native'
+import { getAccountUser } from 'audius-client/src/common/store/account/selectors'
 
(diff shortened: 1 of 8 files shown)


packages/mobile/src/components/navigation-container/NavigationContainer.tsx
```diff
@@ -14,85 +18,98 @@ import { navigationThemes } from './navigationThemes'
...
+                  }
+                },
+                favorites: {
+                  screens: {
+                    Favorites: 'favorites'
+                  }
+                },
+                profile: {
+                  screens: {
+                    UserProfile: 'profile'
```
raymondjacobson, line 86: profile allows /profile/tracks, profile/albums, etc. FYI!

raymondjacobson (approved):

packages/mobile/src/components/navigation-container/NavigationContainer.tsx
sliptype: Good catch, just got that situated as well!

PR #1191, opened 2022-04-11
[AUD-1723] Collectibles tab improvement

sliptype:
### Description

* Performance seems good and didn't run into any crashes on accounts with large numbers of collectibles (FerrariJetpack, Mike Shinoda, etc)
* Removes the wrapping tile because the bottom edge was being rendered above the bottom bar at all times. Made the header it's own tile at the top. This changes the look slightly but I think it still makes sense:
image

### Dragons
N/A

### How Has This Been Tested?

iOS sim

TestFlight QA


dylanjeffers (approved):
sweet all makes sense

PR #1195, opened 2022-04-11
[AUD-1855] Improve drawer performance

sliptype:
### Description

* Removes the debounce from the state sync. This was added to help with lag when navigating between screens via the bottom tabs, but other performance improvements have made it unnecessary
* Still seems to be a slight delay with the Share drawer sometimes, but there is not much rendering going on so if anything it's probably some small latency with the message bridge. Think it should be ok for now

### Dragons

Could potentially be a lot of rerenders again in places that I didn't check

### How Has This Been Tested?

iOS sim in dev and release configs

TestFlight QA


dylanjeffers (approved):
The end of an era! nice we still have the isEmpty check tho

PR #1197, opened 2022-04-12
[AUD-1884] Fix Lineup pull-to-refresh

dylanjeffers:
### Description

Fixes bug where passing onScroll to lineup overrides the onScroll implemented by the useOverflowHandlers in Section/Flat Lists.


sliptype (approved):
Good find! Dang I really broke everything with that change lol

PR #1198, opened 2022-04-12
Add back state sync debounce

sliptype:
### Description

* Adds back the state sync debounce at 50ms to fix lag when playing tracks

### Dragons

This is basically what we had before but instead of 200ms

### How Has This Been Tested?

iOS sim

TestFlight QA

### DIFF
--- a/packages/web/src/store/configureStore.ts
+++ b/packages/web/src/store/configureStore.ts
@@ -1,13 +1,15 @@
 import * as Sentry from '@sentry/browser'
 import { routerMiddleware, push as pushRoute } from 'connected-react-router'
-import { isEmpty, pick, pickBy } from 'lodash'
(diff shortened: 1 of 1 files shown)


packages/web/src/store/configureStore.ts
```diff
@@ -121,6 +123,22 @@ const middlewares = applyMiddleware(
...
+  () => {
+    if (!isEmpty(aggregateStateToSync)) {
+      postMessage({
+        type: MessageType.SYNC_COMMON_STATE,
+        state: aggregateStateToSync
+      })
+      aggregateStateToSync = null
+    }
+  },
+  50,
```
dylanjeffers, line 138: Worth making a constant just to fine tune in future?

dylanjeffers (approved):
And it's back! :)

PR #1201, opened 2022-04-12
[AUD-1884] Fix scrollTo bug on initial scroll

dylanjeffers:
### Description

Fixes bug where the initial scroll on a lineup would trigger a scrollTo(0) bug, which is needed when pull-to-refresh is triggered.


raymondjacobson (approved):

sliptype (approved):
Good solution!

PR #1205, opened 2022-04-12
[AUD-1859] Fix lock screen media controls when chromecasting

sliptype:
### Description

* Fixes issue where chromecast session was suspended when app was backgrounded
* Starts the chromecast media at the correct time to prevent the scrubber from bouncing between 2 values. (There are still some instances where it occasionally gets off by like 1 second, but I think this is due to the latency of starting the session)

### Dragons

Removes unused `openChromecastDialog` function

### How Has This Been Tested?
iOS device (lock screen controls don't work on simulator)

TestFlight QA

### DIFF
--- a/packages/mobile/ios/AudiusReactNative/AppDelegate.m
+++ b/packages/mobile/ios/AudiusReactNative/AppDelegate.m
@@ -80,6 +80,8 @@
   GCKCastOptions* options = [[GCKCastOptions alloc] initWithDiscoveryCriteria:criteria];
   // Allow our app to control chromecast volume
   options.physicalVolumeButtonsWillControlDeviceVolume = YES;
+  // Prevent backgrounding from suspending sessions
+  options.suspendSessionsWhenBackgrounded = NO;
   [GCKCastContext setSharedInstanceWithOptions:options];
 
   return YES;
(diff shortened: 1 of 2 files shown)


packages/mobile/ios/AudiusReactNative/AppDelegate.m
```diff
@@ -80,6 +80,8 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
   GCKCastOptions* options = [[GCKCastOptions alloc] initWithDiscoveryCriteria:criteria];
   // Allow our app to control chromecast volume
   options.physicalVolumeButtonsWillControlDeviceVolume = YES;
+  // Prevent backgrounding from suspending sessions
+  options.suspendSessionsWhenBackgrounded = NO;
```
raymondjacobson, line 84: lmao

packages/mobile/src/components/audio/GoogleCast.tsx
```diff
@@ -73,7 +73,7 @@ export const useChromecast = () => {
   // Load media when the cast connects
   useEffect(() => {
     if (castState === CastState.CONNECTED) {
-      loadCast(track)
+      loadCast(track, global.progress.currentTime ?? 0)
```
raymondjacobson, line 76: I think this make sense. In the past, we've had bugs where we assumed 0 time when it wasn't correct, but i think this is right

raymondjacobson (approved):

PR #1206, opened 2022-04-13
[AUD-1848] Improve app layout when PlayBar active

dylanjeffers:
### Description

Improves app layout when play bar is active by removing padding element on bottom-tab-bar which was visible behind play bar when springing down, replacing it with PlayBar aware lists which add extra padding as needed.


packages/mobile/src/components/core/PlayBarChin.tsx
```diff
@@ -0,0 +1,11 @@
+import { View } from 'react-native'
```
raymondjacobson, line 1: image

raymondjacobson (approved):
awesome

PR #1208, opened 2022-04-13
[AUD-1891] Fix Profile left header, back button

dylanjeffers:
### Description

Fixes profile left header spacing and back button navigation


raymondjacobson (approved):
🤗

PR #1209, opened 2022-04-13
[AUD-1889] Fix notification drawer swipes on tab screens

raymondjacobson:
### Description

Fixes 3 bugs having to do with swipe gestures:
- Clicking tabs doesn't properly respect stack navigation the way swiping does (e.g. clicking tab #3 and swiping back pops the stack instead of switches tabs)
- Navigating deeper into the stack from a tab screen no longer breaks the full screen swipe (this was because we wouldn't re-enable stack/drawer gestures after "pushing" stack -- leaving a tab screen)
- Drawer navigator can be correctly swiped on base tab screens (see video).
- Bonus bug -- fixes type error in useAppScreenOptions

The 3rd change is the biggest in the PR and it does have a major caveat:

react-native-gesture-handler will basically always take precedence over material top tabs. I actually think this is is *good* because the current behavior is non-deterministic. However, it just means that future usage of gesture handler needs to be a little bit context-aware of this patch. Hope others agree!!

Yay! Swiping is now very happy! ... on ios

### Dragons

### How Has This Been Tested?

Test device
* Feed => notifs => feed
* Trending week => notifs => week trending
* Trending week => trending month => notifs => trending month
* Trending week => trending month => track => trending month => trending week
* Explore => playlist => explore


dylanjeffers (approved):
bloody legend mate

PR #1212, opened 2022-04-13
Allow TrackTile click in the action button section

Kyle-Shanks:
### Description
Remove gesture responder capture to allow track tile to be clicked in the bottom part

### Dragons

Not sure if there was a reason we didn't allow this initially, or if any unintended issues arise from this change

### How Has This Been Tested?

Manually tested, click play works and the buttons still work 

N/A

### DIFF
--- a/packages/mobile/src/components/lineup-tile/LineupTileActionButtons.tsx
+++ b/packages/mobile/src/components/lineup-tile/LineupTileActionButtons.tsx
@@ -99,12 +99,7 @@ export const LineupTileActionButtons = ({
   )
 
   return (
-    <View
-      style={styles.bottomButtons}
-      // Capture touches to prevent from triggering play
-      onStartShouldSetResponder={() => true}
-      onTouchEnd={e => e.stopPropagation()}
-    >
+    <View style={styles.bottomButtons}>
       <View style={styles.leftButtons}>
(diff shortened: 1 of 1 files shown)


sliptype (approved):
So pressing the action buttons doesn't trigger a play anymore? Weird, I added this to fix that but maybe something changed

raymondjacobson:
> So pressing the action buttons doesn't trigger a play anymore? Weird, I added this to fix that but maybe something changed

that's what i thought

Kyle-Shanks:
> So pressing the action buttons doesn't trigger a play anymore? Weird, I added this to fix that but maybe something changed

Nope, doesn't trigger play/pause while testing 👍

dylanjeffers:
So one change from what we currently have, is that clicks in the ActionButtons region do now trigger play/pause, just not the buttons themselves. How do we feel about that? fwiw mobile app does allow play/pause in that region, so this change moves more in line with that

dylanjeffers:
> > So pressing the action buttons doesn't trigger a play anymore? Weird, I added this to fix that but maybe something changed
> 
> Nope, doesn't trigger play/pause while testing 👍

Yeah confirmed as well!

dylanjeffers (approved):

PR #1214, opened 2022-04-13
[AUD-1894] Remix Lineup fixes

sliptype:
### Description

* Adds "Remix of" UI
Screen Shot 2022-04-13 at 11 51 43 AM
* Fixes issue with Remixes screen lineup loading
* Fixes issue with track screen with no other tracks by artist lineup loading

### Dragons

Shuffled some things around on track screen, added a max of 6 on the track screen Lineup (this was only enforced at the component level for some reason)

### How Has This Been Tested?

iOS sim

TestFlight QA

### DIFF
--- a/packages/mobile/src/components/lineup/Lineup.tsx
+++ b/packages/mobile/src/components/lineup/Lineup.tsx
@@ -99,6 +99,7 @@ const styles = StyleSheet.create({
 
 type Section = {
   delineate: boolean
+  hasLeadingElement?: boolean
   title?: string
   data: Array<LineupItem | LoadingLineupItem>
 }
@@ -113,6 +114,7 @@ export const Lineup = ({
   disableTopTabScroll,
   isTrending,
   leadingElementId,
+  leadingElementDelineator,
   lineup,
   loadMore,
   header,
@@ -348,7 +350,7 @@ export const Lineup = ({
 
       return [
         { delineate: false, data: [artistPick] },
-        { delineate: true, data: restEntries }
+        { delineate: true, data: restEntries, hasLeadingElement: true }
       ]
     }
 
@@ -403,17 +405,20 @@ export const Lineup = ({
       ListFooterComponent={<View style={{ height: 16 }} />}
       onEndReached={handleLoadMore}
       onEndReachedThreshold={LOAD_MORE_THRESHOLD}
-      // TODO: Either style the refreshing indicator or
-      // roll our own
       onRefresh={refresh}
       refreshing={refreshing}
       sections={sections}
       stickySectionHeadersEnabled={false}
-      // TODO: figure out why this is causing duplicate ids
-      // keyExtractor={(item, index) => String(item.id + index)}
+      keyExtractor={(item, index) => `${item.id}  ${index}`}
       renderItem={renderItem}
       renderSectionHeader={({ section }) => {
-        return section.delineate ? <Delineator text={section.title} /> : null
+        if (section.delineate) {
+          if (section.hasLeadingElement && leadingElementDelineator) {
+            return leadingElementDelineator
+          }
+          return <Delineator text={section.title} />
+        }
+        return null
       }}
       listKey={listKey}
       scrollIndicatorInsets={{ right: Number.MIN_VALUE }}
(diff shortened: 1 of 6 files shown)


packages/mobile/src/components/lineup/Lineup.tsx
```diff
@@ -403,17 +409,20 @@ export const Lineup = ({
...
       onEndReachedThreshold={LOAD_MORE_THRESHOLD}
-      // TODO: Either style the refreshing indicator or
-      // roll our own
       onRefresh={refresh}
       refreshing={refreshing}
       sections={sections}
       stickySectionHeadersEnabled={false}
-      // TODO: figure out why this is causing duplicate ids
-      // keyExtractor={(item, index) => String(item.id + index)}
+      keyExtractor={(item, index) => `${item.id}  ${index}`}
```
dylanjeffers, line 416: do we want to use index here instead of just id?

packages/mobile/src/components/lineup/types.ts
```diff
@@ -41,6 +43,17 @@ export type LineupProps = {
    */
   disableTopTabScroll?: boolean
 
+  /**
+   * Optional payload to pass to the fetch action
+   */
+  fetchPayload?: any
```
dylanjeffers, line 49: no worries but could be good to have some typing, though now that i see it's use, not sure how you could

packages/mobile/src/screens/track-screen/TrackRemixesScreen.tsx
```diff
@@ -81,6 +81,7 @@ export const TrackRemixesScreen = () => {
       <Header text={messages.header} />
       <Lineup
         lineup={lineup}
+        fetchPayload={{ trackId: track?.track_id }}
```
dylanjeffers, line 84: ah i see so this is a way to avoid having to write a custom loadMore? smart

packages/mobile/src/screens/track-screen/TrackScreen.tsx
```diff
@@ -146,13 +151,74 @@ export const TrackScreen = () => {
...
+    (!remixParentTrackId && lineup.entries.length > 1)
+
+  const hasValidRemixParent =
+    !!remixParentTrackId &&
+    !!remixParentTrack &&
+    remixParentTrack.is_delete === false &&
+    !remixParentTrack.user?.is_deactivated
+
+  const moreByArtistTitle = showMoreByArtistTitle && (
+    <Text
```
dylanjeffers, line 179: lint issue i believe?

packages/web/src/common/store/pages/track/lineup/reducer.js
```diff
@@ -4,7 +4,8 @@ import { PREFIX } from 'common/store/pages/track/lineup/actions'
 
 const initialState = {
   ...initialLineupState,
-  prefix: PREFIX
+  prefix: PREFIX,
+  maxEntries: 6
```
dylanjeffers, line 8: nice

dylanjeffers (approved):
great work dude, maybe run a lint:fix on your side just to confirm?

packages/mobile/src/components/lineup/Lineup.tsx
sliptype: I think there could potentially be duplicates in a lineup, so added id just in case

packages/mobile/src/screens/track-screen/TrackRemixesScreen.tsx
sliptype: Yeah web has custom loadMore's everywhere but this is the first case I found that deviates from the pattern so passing the object made sense

dylanjeffers: i think there may also be a case for profile feed too, where we can use the payload instead of custom loadMore

packages/mobile/src/screens/track-screen/TrackScreen.tsx
sliptype: I don't think we have this in the eslint config yet but I updated this to a ternary

packages/mobile/src/screens/track-screen/TrackRemixesScreen.tsx
sliptype: Ohhh didn't know that was a thing

packages/mobile/src/screens/track-screen/TrackScreen.tsx
dylanjeffers: oh word, i was meaning less ternery but more just the jsx seemed funky, no worries tho

PR #1220, opened 2022-04-14
Fix track not playable bug

nicoback2:
### Description
Addresses https://linear.app/audius/issue/AUD-1799/when-a-lineup-on-a-track-page-hasnt-loaded-the-play-bar-plays-and  and https://linear.app/audius/issue/AUD-1523/track-not-playable-immediately-on-track-page-load

Fix bug where if you went to a track page from the feed and pressed "play" in the hero track tile before the "More by this artist" loaded, the play button would not work

Fix similar bug with same root cause on mobile - except instead of the "play" button not working, it would play/pause whatever was in the queue/lineup previously

### Dragons

### How Has This Been Tested?


packages/web/src/pages/track-page/store/lineups/tracks/sagas.ts
```diff
@@ -0,0 +1,97 @@
+import { call, select } from 'typed-redux-saga'
```
nicoback2, line 1: Typed this file - here is a more helpful diff:
https://quickdiff.net/?unique_id=F03981EC-D6E8-D14E-EEFD-B4D68A276E2C

packages/mobile/src/screens/track-screen/TrackScreen.tsx
```diff
@@ -95,7 +95,7 @@ const TrackScreenMainContent = ({
           track={track}
           user={user}
           uid={lineup?.entries?.[0]?.uid}
-          isLineupLoading={lineup.isMetadataLoading}
+          isLineupLoading={!lineup?.entries?.[0]}
```
Kyle-Shanks, line 98: Was there a benefit to doing it this way? would this cause an issue if the lineup is loaded, but empty?

sliptype (approved):
Looks good to me! Thanks for converting to typescript too 💪

packages/mobile/src/screens/track-screen/TrackScreen.tsx
nicoback2: I changed this because now we load the hero track first (separately from the rest of the lineup which is the "more by this artist") and put that in the lineup immediately, and then make the call to load the rest of the lineup which sets `isMetadataLoading` to true. I.e. if the track has loaded but the rest of the lineup has not, `lineup.isMetadataLoading` will still be false even though we're ready for the user to hit play.

`would this cause an issue if the lineup is loaded, but empty?` - in this case i believe you'd just get an error saying the track is unavailable

This matches the logic in web as well

audius-infra:
Preview this change https://demo.audius.co/aud-1523

packages/mobile/src/screens/track-screen/TrackScreen.tsx
Kyle-Shanks: yup, I read the rest of the PR and saw that this was fine 👌

PR #1221, opened 2022-04-14
[AUD-1899] Reduce notif rerenders on navigation

sliptype:
### Description

* Prevents notif rerenders on navigation by changing the NotificationsDrawerNavigationContext to not provide an object and wrapping NotificationScreen in memo

### Dragons

Could prevent the notifs screen from being render when it was before so could see some state out of sync

### How Has This Been Tested?

iOS sim

TestFlight

### DIFF
--- a/packages/mobile/src/screens/notifications-screen/NotificationsScreen.tsx
+++ b/packages/mobile/src/screens/notifications-screen/NotificationsScreen.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useContext, useEffect } from 'react'
+import { memo, useCallback, useContext, useEffect } from 'react'
 
 import { useDrawerStatus } from '@react-navigation/drawer'
 import { markAllAsViewed } from 'audius-client/src/common/store/notifications/actions'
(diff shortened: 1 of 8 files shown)


packages/mobile/src/screens/notifications-screen/NotificationsScreen.tsx
```diff
@@ -25,9 +25,9 @@ const styles = StyleSheet.create({
 /**
  * A component that renders a user's notifications
  */
-export const NotificationsScreen = () => {
+export const NotificationsScreen = memo(() => {
```
dylanjeffers, line 28: Just curious what is triggering this without memo? The handlers object changing in context?

dylanjeffers (approved):
Nice!

packages/mobile/src/screens/notifications-screen/NotificationsScreen.tsx
sliptype: No, I think the Drawer.Navigator `drawerContent` is just getting rerendered on navigation state changes, not sure why. But none of the props or the context value are changing so I added this memo to stop the rerender as high as possible

PR #1223, opened 2022-04-14
Improve profile social loading and cls

dylanjeffers:
### Description

Improves profile cls by rendering skeletons for social icons.


sliptype (approved):
Good idea

PR #1230, opened 2022-04-15
Fix swipe on first tab

sliptype:
### Description

* Fixes 2 issues with swiping from the first tab in a tab navigator:
  * Swiping perfectly horizontally wouldn't register as a swipe
  * Swiping to go back from a pushed stack screen didn't register every time

https://user-images.githubusercontent.com/19916043/163634521-6c93d63f-db92-4638-93a8-cacf5693a874.mov

### Dragons

Another patch, arbitrarily set `-1` and `1` values. Seems to work but I might have broken something else

### How Has This Been Tested?

iOS sim. swiping tabs, stack screens, notifications, etc

TestFlight QA

### DIFF
--- a/packages/mobile/patches/@react-navigation+drawer+6.3.1.patch
+++ b/packages/mobile/patches/@react-navigation+drawer+6.3.1.patch
@@ -1,8 +1,15 @@
 diff --git a/node_modules/@react-navigation/drawer/src/views/legacy/Drawer.tsx b/node_modules/@react-navigation/drawer/src/views/legacy/Drawer.tsx
-index bebfcbf..0211168 100644
+index bebfcbf..1fbd6f4 100644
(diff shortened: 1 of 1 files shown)


dylanjeffers (approved):
You're a wizard, stoked to try this out

packages/mobile/patches/@react-navigation+drawer+6.3.1.patch
```diff
@@ -13,6 +20,11 @@ index bebfcbf..0211168 100644
 +            isOpen
 +              ? [-SWIPE_DISTANCE_MINIMUM, Number.MAX_SAFE_INTEGER]
 +              : [-1 * Number.MAX_SAFE_INTEGER, SWIPE_DISTANCE_MINIMUM]
++          }
++          failOffsetX={
```
raymondjacobson, line 24: nice

raymondjacobson (approved):

PR #1233, opened 2022-04-18
[AUD-1904] Use high res artwork on NowPlaying

sliptype:
### Description

* Uses 1000x1000 art on NowPlaying

Before & After
Simulator Screen Shot - iPhone 13 - 2022-04-18 at 10 51 54
Simulator Screen Shot - iPhone 13 - 2022-04-18 at 10 51 31

### Dragons

Slow load times for images in NowPlaying

### How Has This Been Tested?

iOS sim

TestFlight QA


Kyle-Shanks (approved):
ezpz

PR #1235, opened 2022-04-18
[AUD-1927] Vertically align user badges on now playing

sliptype:
### Description

* Vertically align user badges on now playing
Simulator Screen Shot - iPhone 13 - 2022-04-18 at 14 18 38

### Dragons

Possible layout changes on different sized screens

### How Has This Been Tested?

iOS sim

TestFlight QA


dylanjeffers (approved):
Nice

PR #1236, opened 2022-04-18
Fix behavior of track screen play button

sliptype:
### Description

* Fixes behavior of TrackScreen button to match that of CollectionScreen

### Dragons

Changing logic slightly, but it's actually more consistent with the rest of the app (including Lineup)

### How Has This Been Tested?

iOS sim, playing/pausing from track screen when another track is already playing. Compared to working CollectionScreen

TestFlight

### DIFF
--- a/packages/mobile/src/screens/track-screen/TrackScreenDetailsTile.tsx
+++ b/packages/mobile/src/screens/track-screen/TrackScreenDetailsTile.tsx
@@ -49,7 +49,7 @@ import { useDispatchWeb } from 'app/hooks/useDispatchWeb'
 import { useNavigation } from 'app/hooks/useNavigation'
 import { useSelectorWeb } from 'app/hooks/useSelectorWeb'
 import { useTrackCoverArt } from 'app/hooks/useTrackCoverArt'
-import { getPlaying, getPlayingUid, getTrack } from 'app/store/audio/selectors'
+import { getPlaying, getPlayingUid } from 'app/store/audio/selectors'
 import { flexRowCentered, makeStyles } from 'app/styles'
 import { make, track as record } from 'app/utils/analytics'
 import { moodMap } from 'app/utils/moods'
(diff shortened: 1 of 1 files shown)


packages/mobile/src/screens/track-screen/TrackScreenDetailsTile.tsx
```diff
@@ -71,6 +71,16 @@ type TrackScreenDetailsTileProps = {
   isLineupLoading: boolean
 }
 
+const recordPlay = (id, play = true) => {
```
dylanjeffers, line 74: just wondering if this could be used elsewhere?

packages/mobile/src/screens/track-screen/TrackScreenDetailsTile.tsx
```diff
@@ -132,8 +142,8 @@ export const TrackScreenDetailsTile = ({
   const currentUserId = useSelectorWeb(getUserId)
   const dispatchWeb = useDispatchWeb()
   const playingUid = useSelector(getPlayingUid)
-  const queueTrack = useSelector(getTrack)
   const isPlaying = useSelector(getPlaying)
+  const isPlayingUid = playingUid === uid
```
dylanjeffers, line 146: so correct me if im wrong, so this is less strict as previously, where instead of seeing it's the same track and it's part of a queue, just if its the same track?

dylanjeffers (approved):
looks great, in terms of using that record function, just a callout maybe for the future to consolidate our analytics

packages/mobile/src/screens/track-screen/TrackScreenDetailsTile.tsx
sliptype: It should be the same because `queueTrack` before was just using the `getTrack` selector which is the playing track. We were only using that to do comparisons on the uid anyway

packages/mobile/src/screens/track-screen/TrackScreenDetailsTile.tsx
sliptype: It can! CollectionScreenDetailsTile and Lineup would be good candidates, I'll make a follow up PR

packages/mobile/src/screens/track-screen/TrackScreenDetailsTile.tsx
dylanjeffers: ah i see, thanks for clarification

PR #1237, opened 2022-04-18
Update lineup saga delay up to 100ms to avoid flakeyness

Kyle-Shanks:
### Description
Updated lineup saga delay up to 100ms

### Dragons

N/A

### How Has This Been Tested?

Manually tested. Appears to load fine

Will monitor for flakeness in the future and update if there are still issues


sliptype (approved):

audius-infra:
Preview this change https://demo.audius.co/kj-Update-lineup-saga-delay-again

dylanjeffers (approved):

PR #1238, opened 2022-04-18
[AUD-1920] Fix TrendingRewardsDrawer navigation

dylanjeffers:
### Description

Fixes TrendingRewardsDrawer navigation by replacing pushRouteWeb with useNavigation
- Cleans up styles
- Replaces ButtonWithArrow with core button


sliptype (approved):
Good catch! lgtm

PR #1239, opened 2022-04-18
[AUD-1921] Don't render ArtistRecommendations when empty

dylanjeffers:
### Description

We shouldn't render ArtistRecommendations when there are no artists to recommend


sliptype (approved):
🔥

PR #1241, opened 2022-04-18
[AUD-1930] Avoid showing content that gets a 403 (stop-gap)

raymondjacobson:
### Description

When we get Unauthorized (403) from the getCID route: https://github.com/AudiusProject/audius-protocol/pull/2951, avoid showing the content.

Note this is a stop gap solution in favor of a broader fix that indexes this data and makes it available to discovery nodes: https://linear.app/audius/project/better-track-unavailability-ui-c676176f9b93

### Dragons

### How Has This Been Tested?

Locally vs. production data on tracks getting 403s.

### DIFF
--- a/packages/web/src/common/store/cache/tracks/sagas.js
+++ b/packages/web/src/common/store/cache/tracks/sagas.js
@@ -62,19 +62,35 @@ function* fetchSegment(metadata) {
   return yield call(fetchCID, cid, gateways, /* cache */ false)
 }
 
+// TODO(AUD-1837) -- we should not rely on this logic anymore of fetching first
+// segments, particularly to flag unauthorized content, but it should probably
+// just be removed altogether since first segment fetch is usually fast.
 function* fetchFirstSegments(entries) {
   // Segments aren't part of the critical path so let them resolve later.
   try {
     const firstSegments = yield all(
       entries.map(e => call(fetchSegment, e.metadata))
     )
+
     yield put(
       cacheActions.update(
         Kind.TRACKS,
-        firstSegments.map((s, i) => ({
-          id: entries[i].id,
-          metadata: { _first_segment: s }
-        }))
+        firstSegments.map((s, i) => {
+          if (s === 'Unauthorized') {
+            return {
+              id: entries[i].id,
+              metadata: {
+                is_delete: true,
+                _blocked: true,
+                _marked_deleted: true
+              }
+            }
+          }
+          return {
+            id: entries[i].id,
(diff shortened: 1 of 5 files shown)


audius-infra:
Preview this change https://demo.audius.co/rj-aud-1930

sliptype (approved):
Looks good to me! The cache merge customization is kinda hacky but definitely makes sense for now

raymondjacobson:
> Looks good to me! The cache merge customization is kinda hacky but definitely makes sense for now

Yeah I very much dislike it. we honestly should try to remove it all after the SOL switch.

packages/web/src/common/store/cache/tracks/sagas.js
```diff
@@ -62,19 +62,35 @@ function* fetchSegment(metadata) {
...
-        firstSegments.map((s, i) => ({
-          id: entries[i].id,
-          metadata: { _first_segment: s }
-        }))
+        firstSegments.map((s, i) => {
+          if (s === 'Unauthorized') {
+            return {
+              id: entries[i].id,
+              metadata: {
+                is_delete: true,
```
SidSethi, line 83: idk the related code but guessing this is to conform with the existing client pattern but we would lose info here by conflating unauthorized as deleted

thoughts on instead passing an is_unauthorized or similar flag and handling that in UI in addition to is_delete?

also did you mean is_deleted haha

SidSethi (commented):
the backend/sagas code seems fine but i def do not know what the reducer / surrounding code is doing so am hesitant to approve

packages/web/src/common/store/cache/tracks/sagas.js
raymondjacobson: yes, that, i want to leave for the actual project to build this correctly. I don't love us adding more client-only derived fields, would prefer we stick to the behavior we have with `_blocked` (which was for this purpose already).

did not mean is_deleted (i wish): go ask this guy https://github.com/AudiusProject/audius-protocol-old/commit/2cb4db96f8eb0efa6ce029ae19cc62b8f7eb47c6#diff-e07b1f24355146ec1d8d2622de7d65a1ac36022aa52444cc5eb16a3e09213d17R31

SidSethi: LMAO

SidSethi: and yep makes sense i was a bit confused since it seems like _blocked already should cover this
ok cool lol

PR #1245, opened 2022-04-21
[AUD-1931] Fix notification UserLists

sliptype:
### Description

* Fixes notification user lists by setting refresh state based on the length of the user list

### Dragons

Could impact loading/refresh logic of the user lists

### How Has This Been Tested?

iOS simulator

TestFlight


dylanjeffers (approved):
Good stuff

PR #1246, opened 2022-04-21
[C-403] Fix lineup with artist pick

sliptype:
### Description

* Instead of slicing the artist pick out of the response, set it to null and keep track of the null count to avoid messing up the offset

### Dragons

This could break lineups

### How Has This Been Tested?

Locally against prod, web and ios

Michael will confirm fix

### DIFF
--- a/packages/mobile/src/components/lineup/Lineup.tsx
+++ b/packages/mobile/src/components/lineup/Lineup.tsx
@@ -155,10 +155,10 @@ export const Lineup = ({
   const countOrDefault = count !== undefined ? count : MAX_TILES_COUNT
 
   const handleLoadMore = useCallback(() => {
-    const { deleted, entries, hasMore, page, status } = lineup
+    const { deleted, nullCount, entries, hasMore, page, status } = lineup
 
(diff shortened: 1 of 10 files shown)


jowlee (approved):
lgtm

jowlee:
would also like to check that the lineup ends and doesn't try to load forever

audius-infra:
Preview this change https://demo.audius.co/sk-fix-profile-lineup

sliptype:
> would also like to check that the lineup ends and doesn't try to load forever

Good call, there was some weird infinite load behavior at the end of the lineup. I ended up fixing the problem in a different way, it's not very elegant but  I didn't want to use `containsDeleted` because we are no longer using it anywhere else and updated our endpoints to account for deleted items. 

Essentially needed to keep track of "deleted" and "removed because of artist pick" separately, I did this by setting the artist pick item as null and keeping track of a `nullCount`. This fixes on both web and mobile

dylanjeffers (approved):
tricky for sure! makes sense though

Issue #1127, opened 2022-03-25
Clicking on the "Profile" button in Android crashes the whole app.


Ashvith:
Title says it all. App crashes when I try to access the profile options on the bottom bar.

Elandig:
I'm experiencing the same issue. Any kind of profile opened immediately crashes the app.

sliptype:
Hey! We weren't immediately able to reproduce the issue. Can you make sure that the Audius app is fully updated?

Also could you let me know which device and version of android you are running?

Ashvith:
> Hey! We weren't immediately able to reproduce the issue. Can you make sure that the Audius app is fully updated?
> 
> Also could you let me know which device and version of android you are running?

Android version 11, device name is realme 5 pro.

subutex:
Hi, I've the same issue since some weeks ago. The Audius app version is 1.1.66 installed on Redmi Note 9 Pro running on Android 11. Everything is updated on my phone so far.

subutex:
This is very annoying because I can't access my profile anymore on my smartphone, I can only access my songs via notifications when someone forward or like them. And when I click on my song I can't share it on Twitter. When I click on the Twitter share button, Twitter app does not open. Could you please have a look at those issues ? Thank you in advance.

sliptype:
This is fixed in version 1.1.95, please upgrade and it should work!

Thanks for flagging 🙏

subutex:
Great, it works now ! Thank you 😊

PR #1247, opened 2022-04-25
[C-405] Fix legacy android build profile freeze

dylanjeffers:
### Description

Fixes app breaking bug on legacy android app. Issue was due to linear gradient breaking when not receiving at least 2 colors.


sliptype (approved):
YESSSS 🙏

#1127

PR #1249, opened 2022-04-26
[C-366] Remove antd Spin component and replace with LoadingSpinner

Kyle-Shanks:
### Description

[C-366]

Update all places that use the antd spin component to use the LoadingSpinner component

### Dragons

There may be some places where the loading is not the exact same size or position

### How Has This Been Tested?

Manually tested

N/a


sliptype (approved):
Good stuff!

PR #1251, opened 2022-04-26
Fix "More by artist" not updating correctly after track play bug fix

nicoback2:
### Description
On mobile, the "more by this artist" on a track did not load correctly after the [track play bug fix](https://github.com/AudiusProject/audius-client/pull/1220) was merged.
This was because:
-The track play bug fix PR added [this line](https://github.com/AudiusProject/audius-client/blob/5c02a70614949030c3690aa6ea886a0fe2e61cf8/packages/web/src/pages/track-page/store/sagas.js#L155-L156) to set the track ID in the track page store once it was found (in order to match lineup/queue entries corresponding a track page)
-Upon navigating to a new track page, the track ID was still that of the previously visited track since it is not reset on mobile. This caused `getTrack` to return the old track due to its [logical structure](https://github.com/AudiusProject/audius-client/blob/5c02a70614949030c3690aa6ea886a0fe2e61cf8/packages/web/src/common/store/pages/track/selectors.ts#L15-L26), whereas before it would return the correct track because the track ID was null and the track permalink was correct

Fix:
Instead of comparing `track.permalink` to the new path, compare `trackPermalink` which is always up to date
This is still a bandaid fix since we should really be resetting the track ID once we know we've navigated off a track page (which I believe is what web does). I tried doing (see `<-- ADDED THIS` comment):
```
  fetchTracks = (params: NonNullable<TrackRouteParams>) => {
    const { track } = this.props
    const { slug, trackId, handle } = params

    // Go to feed if the track is deleted
    if (track && track.track_id === trackId) {
      if (track._marked_deleted) {
        this.props.goToRoute(FEED_PAGE)
        return
      }
    }
    this.props.reset()
    if (trackId) {
      this.props.setTrackId(trackId)
    } else { // <<-- ADDED THIS ELSE STMT
      this.props.setTrackId(null)
    }
    if (slug && handle) {
      this.props.setTrackPermalink(`/${handle}/${slug}`)
    }
    this.props.fetchTrack(trackId, slug || '', handle || '', !!(slug && handle))
    if (handle) {
      this.setState({ ownerHandle: handle })
    }
  }
``` 
but this created another bug :/. The fix in this PR should suffice for now.

### Dragons

### How Has This Been Tested?
-Test loading many track pages on web and mobile
-Test renaming a track on web and make sure the behavior matches what's in prod/staging


nicoback2:
See this for changes: https://github.com/AudiusProject/audius-client/pull/1251/commits/5c02a70614949030c3690aa6ea886a0fe2e61cf8

sliptype (approved):
Ah good find! I think this is fine for now, but it really makes me think this line is where the reset of the track id happens: https://github.com/AudiusProject/audius-client/blob/main/packages/web/src/pages/track-page/TrackPageProvider.tsx#L199

I'm pretty sure I tried getting rid of the conditional around it and it was still broken but maybe worth one more check

nicoback2:
> Ah good find! I think this is fine for now, but it really makes me think this line is where the reset of the track id happens: https://github.com/AudiusProject/audius-client/blob/main/packages/web/src/pages/track-page/TrackPageProvider.tsx#L199
> 
> I'm pretty sure I tried getting rid of the conditional around it and it was still broken but maybe worth one more check

Yeah I’ll try it out! I was worried about removing the conditional since I didn’t really understand why it was there in the first place, I.e. if removing it would break something else

nicoback2:
> Ah good find! I think this is fine for now, but it really makes me think this line is where the reset of the track id happens: https://github.com/AudiusProject/audius-client/blob/main/packages/web/src/pages/track-page/TrackPageProvider.tsx#L199
> 
> I'm pretty sure I tried getting rid of the conditional around it and it was still broken but maybe worth one more check

Yeah I’ll try it out! I was worried about removing the conditional since I didn’t really understand why it was there in the first place, I.e. if removing it would break something else

Kyle-Shanks (approved):

PR #1252, opened 2022-04-26
Fix handling unknown AAO error

piazzatron:
### Description

Wasn't handling this error properly, causing numerous retries

### Dragons

Tested on stage

### How Has This Been Tested?


packages/web/src/pages/audio-rewards-page/store/sagas.ts
```diff
@@ -280,6 +280,7 @@ function* claimChallengeRewardAsync(
             yield put(claimChallengeRewardAlreadyClaimed())
             break
           case FailureReason.BLOCKED:
+          case FailureReason.AAO_ATTESTATION_UNKNOWN_RESPONSE:
```
rickyrombo, line 283: Maybe add this to the below case? It's not always a "block"

rickyrombo (approved):

PR #1253, opened 2022-04-26
Upgrade libs to 1.2.112, use core entrypoint

dylanjeffers:
### Description

Upgrades libs to latest, which includes large typescript refactor. Along with this we can start using the new dist/core entrypoint for constants and other values that need to be used separately from the libs main entrypoint (since that is dynamically imported)

### DIFF
--- a/packages/web/package-lock.json
+++ b/packages/web/package-lock.json
@@ -92,9 +92,9 @@
       }
     },
     "@audius/libs": {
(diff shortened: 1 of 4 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-upgrade-libs-core

sliptype (approved):

PR #1254, opened 2022-04-27
[C-417] Fix default profile cover photo

dylanjeffers:
### Description

Fixes case where default profile cover photo was not rendering properly


sliptype (approved):
Looks good!

PR #1260, opened 2022-04-27
[C-412] Set title alignment

sliptype:
### Description

* Centers the title in the header

### Dragons

n/a

### How Has This Been Tested?

android and ios

testflight / playstore


Kyle-Shanks (approved):
ezpz

PR #1262, opened 2022-04-27
[C427] Fix malformed js calls error

sliptype:
### Description

* On android a "Malformed js calls" error was crashing the app because the lineup `limit` and `offset` were `NaN` because `deleted` and `nullCount` were undefined. Defaulting to 0 fixes the issue

### Dragons

No dragons but this was a pain to debug. If we see this in the future, this stackoverflow answer helped a lot: https://stackoverflow.com/a/67100389/5705462

### How Has This Been Tested?

Android and ios

TestFlight / playstore


Kyle-Shanks (approved):
Nice fix 👍

PR #1265, opened 2022-04-27
[C-414] Fix size of PlayButton on android

sliptype:
### Description

* Adds resizeMode='cover' to the play button to fix the sizing on android. Not entirely sure why it's only needed on the play button and not other animated buttons but some context here: https://github.com/lottie-react-native/lottie-react-native/issues/428

### Dragons

Sizing of play button could be off

### How Has This Been Tested?

Tested on android and ios

TestFlight / playstore


Kyle-Shanks (approved):
nice

PR #1267, opened 2022-04-28
[C-421] Fix play button state on android

sliptype:
### Description

* Fixes play button state on android by adding a key to the LottieView and upgrading lottie
* Reduces rerenders of playbar by moving tracking state into tracking bar

### Dragons

Animations could be impacted, but things looked good

### How Has This Been Tested?
Android and ios

TestFlight / playstore


Kyle-Shanks (approved):
nice cleanup!

PR #1275, opened 2022-05-02
[C-422] Add pull to refresh on android

sliptype:
### Description

* Adds pull to refresh on android

Unfortunately, android does not support "overscrolling" or scrolling past the ends of a scrollview (contrary to what this prob would lead you to believe https://reactnative.dev/docs/scrollview#overscrollmode-android). So there is no way to use the same refresh indicator as we do on ios. Instead, we can use the native android one which looks a lot better than on ios

The only strange part of this is on the collapsible tabs on the profile screen. The indicator shows at the top of the list instead of at the top of the screen, which looks a little weird when scrolling down while the indicator is visible. But I think probably the best we can do for now because when shown at the top, the header covers the indicator and there seems to be no way of bringing the indicator to the front

https://user-images.githubusercontent.com/19916043/166322742-0f9d12cb-1326-4343-a613-b367fc51c1e8.mp4

### Dragons

Possible pull to refresh bugs

### How Has This Been Tested?

android and ios

Testflight / playstore

### DIFF
--- a/packages/mobile/src/components/core/FlatList.tsx
+++ b/packages/mobile/src/components/core/FlatList.tsx
@@ -4,9 +4,13 @@ import {
   Animated,
   FlatList as RNFlatList,
   FlatListProps as RNFlatListProps,
+  Platform,
+  RefreshControl,
   View
 } from 'react-native'
 import { useCollapsibleScene } from 'react-native-collapsible-tab-view'
+
+import { useThemeColors } from 'app/utils/theme'
 
 import { CollapsibleTabNavigatorContext } from '../top-tab-bar'
 
@@ -21,8 +25,9 @@ type CollapsibleFlatListProps = {
 
 const CollapsibleFlatList = (props: CollapsibleFlatListProps) => {
   const { sceneName, ...other } = props
-  const { onRefresh } = other
+  const { refreshing, onRefresh } = other
   const scrollPropsAndRef = useCollapsibleScene(sceneName)
+  const { neutral } = useThemeColors()
   return (
     <View>
       {onRefresh ? <PullToRefresh /> : null}
@@ -34,6 +39,16 @@ const CollapsibleFlatList = (props: CollapsibleFlatListProps) => {
           scrollPropsAndRef.onScroll,
           props.onScroll
         )}
+        refreshControl={
+          Platform.OS === 'ios' ? undefined : (
+            <RefreshControl
+              progressViewOffset={scrollPropsAndRef.progressViewOffset}
+              refreshing={!!refreshing}
+              onRefresh={onRefresh ?? undefined}
+              colors={[neutral]}
+            />
+          )
+        }
       />
     </View>
   )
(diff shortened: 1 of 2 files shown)


dylanjeffers (approved):
Looks great on lists at the top of the screen! regarding your issue with the profile... is there any way to add a refreshcontrol to a normal view and trigger manually depending on scroll or something?

packages/mobile/src/components/core/FlatList.tsx
```diff
@@ -34,6 +39,16 @@ const CollapsibleFlatList = (props: CollapsibleFlatListProps) => {
           scrollPropsAndRef.onScroll,
           props.onScroll
         )}
+        refreshControl={
+          Platform.OS === 'ios' ? undefined : (
+            <RefreshControl
+              progressViewOffset={scrollPropsAndRef.progressViewOffset}
+              refreshing={!!refreshing}
+              onRefresh={onRefresh ?? undefined}
```
dylanjeffers, line 47: i believe as long as onRefresh is undefined coming down don't need the ??. Is it null or something sometimes?

sliptype: Yeah it has a `null` in there so typescript was angry

packages/mobile/src/components/core/FlatList.tsx
```diff
@@ -74,6 +90,15 @@ const AnimatedFlatList = forwardRef<RNFlatList, FlatListProps>(
         <Animated.FlatList
           {...other}
           scrollToOverflowEnabled
+          refreshControl={
+            Platform.OS === 'ios' ? undefined : (
+              <RefreshControl
+                refreshing={!!isRefreshing}
+                onRefresh={onRefresh ?? undefined}
+                colors={[neutral]}
+              />
```
dylanjeffers, line 99: kinda a nit, but maybe we make this a base component that all the lists can use? handles color, maybe also platform etc?

sliptype:
> Looks great on lists at the top of the screen! regarding your issue with the profile... is there any way to add a refreshcontrol to a normal view and trigger manually depending on scroll or something?

Unfortunately there is no way to add RefreshControl outside of a scroll view, we would have to roll our own which would probably require designs because currently the refresh indicators aren't displayed on top of other components (except the cover photo). I think what I have should be good for a first android release and we can maybe do another pass at it!

packages/mobile/src/components/core/FlatList.tsx
sliptype: That was my first approach, but for some reason having anything other than `RefreshControl` at the top level of this value breaks the entire flatlist. Might have to do with how it's interfacing with native components. This code is necessarily WET

dylanjeffers: oh woww interesting! well thanks for the context

dylanjeffers:
> > Looks great on lists at the top of the screen! regarding your issue with the profile... is there any way to add a refreshcontrol to a normal view and trigger manually depending on scroll or something?
> 
> Unfortunately there is no way to add RefreshControl outside of a scroll view, we would have to roll our own which would probably require designs because currently the refresh indicators aren't displayed on top of other components (except the cover photo). I think what I have should be good for a first android release and we can maybe do another pass at it!

Thank you for helping me realize all the limitations with this api, this looks like a great first step

PR #1277, opened 2022-05-03
[C-413] Fix PlayBar offset on Android

sliptype:
### Description

* Fixes the positioning of the PlayBar on android. The problem was that we were using `window` dimensions when `screen` dimensions are more accurate on android. With screen dimensions we also need to account for the status bar and navigation bar. But it results in perfect positioning of the PlayBar:
Screenshot_20220503-103808

### Dragons

Had to pull in a dependency to accurately get the navbar height. I did a quick audit of the code and it seems safe, but is not published on npm so I pinned it to a commit on github

### How Has This Been Tested?

Android and iOS (multiple screen sizes)

TestFlight / playstore

### DIFF
--- a/packages/mobile/ios/Podfile.lock
+++ b/packages/mobile/ios/Podfile.lock
@@ -180,7 +180,7 @@ PODS:
     - GoogleUtilities/Logger
   - libevent (2.1.12)
   - lottie-ios (3.2.3)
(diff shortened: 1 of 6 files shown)


Kyle-Shanks (approved):
🤘

PR #1278, opened 2022-05-03
OAuth login form setup

nicoback2:
### Description
Set up OAuth page and login form (will close the other PR which had the OAuth route in the public site
)
### Dragons

This is a safe change since nothing in the existing app is affected.

### How Has This Been Tested?


sliptype (approved):
Looks good! I still think it's not ideal to load the whole app in the oauth popup, maybe once the sdk is in a good place we can split this out into a separate app?

nicoback2:
> Looks good! I still think it's not ideal to load the whole app in the oauth popup, maybe once the sdk is in a good place we can split this out into a separate app?

yeah sounds good!

PR #1282, opened 2022-05-03
[C-363] Remove antd progress component and use stems ProgressBar component

Kyle-Shanks:
### Description

[C-363]

* Replace antd progress component with stems progress component
* Delete unused ProgressBar component

### Dragons

Does not have the active flashing animation, but can be added if the stems component is updated

### How Has This Been Tested?

Manually tested

N/A


audius-infra:
Preview this change https://demo.audius.co/kj-Remove-antd-progress-component

sliptype (approved):
I guess we are losing the error state here but that's probably fine

Kyle-Shanks:
> I guess we are losing the error state here but that's probably fine

Yea, for now. It should still have copy letting the user know that an error happened and we can update the ProgressBar component in stems to handle more states if need be

raymondjacobson:
Wait, why are we removing the error state? did we check in w/ julian? we should avoid changing UI with this migration

audius-infra:
Preview this change https://demo.audius.co/kj-Remove-antd-progress-component

PR #1283, opened 2022-05-04
[PAY-168] Redesign rest of notifications

dylanjeffers:
### Description

Redesigns the rest of the notifications
- RemixCosign 
- RemixCreate
-  Favorite 
- TierChange 
- Milestone
-  Challenge
-  UserSubscription
-  TrendingTrack

DuranDylan Co-signed your Remix of sample-3s remix
SILVER TIER UNLOCKED
You've earned 1 SAUDIO for completing this challenge!
DeeJay followed you
sample-3s remix by DeeJay


audius-infra:
Preview this change https://demo.audius.co/dj-pay-168-redesign-rest-of-notifications

packages/web/src/components/notification/Notifications/FavoriteNotification.tsx
```diff
@@ -0,0 +1,48 @@
...
+  return (
+    <NotificationTile notification={notification} onClick={handleClick}>
+      <NotificationHeader icon={<IconFavorite />}>
+        <UserProfileList users={users} />
+      </NotificationHeader>
+      <NotificationBody>
+        <UserNameLink user={firstUser} notification={notification} />
+        {otherUsersCount > 0 ? messages.others(otherUsersCount) : null}
+        {messages.reposted}
+        {entityType.toLowerCase()}
```
rickyrombo, line 41: nit: These all seem like part of one "message", can we have some method or something return a string with all of these templated together for clarity?

packages/web/src/components/notification/Notifications/UserNameLink.module.css
```diff
@@ -1,5 +1,5 @@
-.root {
-  margin-right: 4px;
+.rootLeftMargin {
```
rickyrombo, line 1: to support RTL, should this be `margin-inline-start`?

packages/web/src/components/notification/Notifications/RemixCreateNotification.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  return (
+    <NotificationTile notification={notification}>
+      <NotificationHeader icon={<IconRemix />}>
+        <NotificationTitle>{messages.title}</NotificationTitle>
+      </NotificationHeader>
+      <NotificationBody>
+        <EntityLink entity={entity} entityType={entityType} />
+        <span>{messages.by}</span>
+        <UserNameLink user={user} notification={notification} leftMargin />
```
rickyrombo, line 43: Feels like the previous span should have a `margin-inline-end` vs having a prop for left margin in `UserNameLink`?

rickyrombo (approved):
These look so pretty!! 😍

packages/web/src/components/notification/Notifications/FavoriteNotification.tsx
dylanjeffers: yeah for sure! i was trying something out by putting it all in as react components, you think it's confusing ultimately? i guess it's hard to know that that is literally all just text huh? where it get's a lil tricky is when there is text that is also a link + tooltip, then i couldn't really template that in. thoughts?

packages/web/src/components/notification/Notifications/UserNameLink.module.css
dylanjeffers: ohh that is genius great catch

packages/web/src/components/notification/Notifications/RemixCreateNotification.tsx
dylanjeffers: completely agree, was wigging out on how to do this, styling def aint my strongsuit

packages/web/src/components/notification/Notifications/FavoriteNotification.tsx
rickyrombo: Yeah that's true... Generally I think having it all be one "message" makes it easier in the future for other translations to make it grammatically correct, otherwise the grammar is already baked in by the ordering of the rendered strings.

Just looked up https://react.i18next.com/latest/trans-component and maybe it's a non issue? Feel free to ignore for now and we can revisit when doing i18n

PR #1288, opened 2022-05-05
[PAY-158] Add TipSentNotification

dylanjeffers:
### Description

Adds TipSentNotification
A Your Tip Was Sent!

### DIFF
--- a/packages/web/src/components/notification/Notifications/TipSentNotification.tsx
+++ b/packages/web/src/components/notification/Notifications/TipSentNotification.tsx
@@ -1,0 +1,48 @@
+import React from 'react'
+
+import { TipSent } from 'common/store/notifications/types'
+
+import { AudioText } from './AudioText'
+import { NotificationBody } from './NotificationBody'
+import { NotificationFooter } from './NotificationFooter'
+import { NotificationHeader } from './NotificationHeader'
+import { NotificationTile } from './NotificationTile'
+import { NotificationTitle } from './NotificationTitle'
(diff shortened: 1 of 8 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-pay-158-tip-sent-notification

packages/web/src/components/notification/Notifications/TipSentNotification.tsx
```diff
@@ -0,0 +1,48 @@
...
+  return (
+    <NotificationTile notification={notification}>
+      <NotificationHeader icon={<IconTip />}>
+        <NotificationTitle>{messages.title}</NotificationTitle>
+      </NotificationHeader>
+      <NotificationBody className={styles.body}>
+        <ProfilePicture className={styles.profilePicture} user={user} />
+        <span>
+          {messages.sent} <AudioText value={value} /> {messages.to}{' '}
+          <UserNameLink user={user} notification={notification} addMargin />
```
rickyrombo, line 41: Not a big fan of `addMargin` - it also looks like it's not necessary anymore. Now that the `ArtistPopover` is a span, it should keep the whitespace?

Think maybe we can remove this?

rickyrombo (approved):

packages/web/src/components/notification/Notifications/TipSentNotification.tsx
dylanjeffers: yes down for sure, ahh that is why the whitespace was messed up, will do

PR #1290, opened 2022-05-06
Revert ArtistPopover zIndex value to its previous value

Kyle-Shanks:
### Description
Small issue with the zindex value of the artist popover where the popup would show up behind modals.
Reverting to its previous value of 20000 which seems like overkill, but it works

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A


dylanjeffers (approved):
Lol let's goo

audius-infra:
Preview this change https://demo.audius.co/kj-Update-artist-popover-z-index

PR #1293, opened 2022-05-06
c-456 Fix pending tx every session load

nicoback2:
### Description
Before:
There was a pending USER transaction on every app session load if the user had any playlist folders. This was because we compute playlists that are in the user's account but missing in their library - and if there are any, we update the playlist library. If the user had a folder with at least one playlist, the missing playlists were computed incorrectly - which caused the `update profile` operation to be called on every load.

### Dragons

### How Has This Been Tested?

### DIFF
--- a/packages/web/src/store/playlist-library/sagas.ts
+++ b/packages/web/src/store/playlist-library/sagas.ts
@@ -27,6 +27,7 @@ import * as cacheActions from 'common/store/cache/actions'
 import {
   containsTempPlaylist,
   extractTempPlaylistsFromLibrary,
+  getPlaylistsNotInLibrary,
   removePlaylistLibraryDuplicates,
   replaceTempWithResolvedPlaylists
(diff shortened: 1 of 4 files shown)


audius-infra:
Preview this change https://demo.audius.co/nkang--fix-confirmer-init

packages/web/src/store/playlist-library/sagas.ts
```diff
@@ -71,8 +72,10 @@ function* watchUpdatePlaylistLibrary() {
   yield takeEvery(update.type, function* updatePlaylistLibrary(
     action: ReturnType<typeof update>
   ) {
```
sliptype, line 74: Should we keep these logs?

sliptype (approved):
Nice work!

audius-infra:
Preview this change https://demo.audius.co/nkang--fix-confirmer-init

PR #1300, opened 2022-05-10
[C-462] Fix legacy image loading for native

dylanjeffers:
### Description

Fixes case where legacy images don't load properly in native.
- Cleans up redundant extra fetch (libs already handles this logic)

### How Has This Been Tested?

Legacy images load on native, and still load on desktop/mobile web


raymondjacobson (approved):
Big simplification 👍  
I assume you tested the audius profile on web too right?

PR #1301, opened 2022-05-10
[C-424] Fix profile screen bio links on android

sliptype:
### Description
* Fixes the positioning and functionality of bio links on android
Screenshot_20220510-141739
* Fixes issue without text cut off due to `numberOfLines=2`

### Dragons

This was already pretty fragile code and now it's even more fragile. Have to use `measureInWindow` on the refs of the links to get the position instead of `onLayout` because `onLayout` doesn't return correct values on android

### How Has This Been Tested?

iOS and android

TestFlight, play store

### DIFF
--- a/packages/mobile/android/app/src/main/AndroidManifest.xml
+++ b/packages/mobile/android/app/src/main/AndroidManifest.xml
@@ -101,4 +101,19 @@
 
     </application>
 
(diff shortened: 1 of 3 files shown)


Kyle-Shanks (approved):
👍

PR #1303, opened 2022-05-11
[C-475] Fix RemixesScreen header text alignment

dylanjeffers:
### Description

Fixes issue where long titles mess up RemixesScreen text wrapping

Before:
367B05C3-5FD2-481C-81D7-1B77BC47CAA0
After: 
image

Learned that Pressable has a bug where it doesn't align well with text, the suggested workaround is to just use the onPress on Text components since we don't care about onPresIn, onPressOut: https://stackoverflow.com/questions/66590167/vertically-align-pressable-inside-a-text-component

### DIFF
--- a/packages/mobile/package-lock.json
+++ b/packages/mobile/package-lock.json
@@ -6306,8 +6306,8 @@
     "audius-client": {
       "version": "file:../web",
       "requires": {
-        "@audius/libs": "1.2.112",
-        "@audius/stems": "0.3.28",
+        "@audius/libs": "1.2.115",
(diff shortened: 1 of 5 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-c-475-fix-remixes-screen-header-text

audius-infra:
Preview this change https://demo.audius.co/dj-c-475-fix-remixes-screen-header-text

sliptype (approved):
Nice work!

Interesting about Pressable vs Text, although I think eventually we might want active styling on everything pressable

PR #1308, opened 2022-05-11
[PAY-176] Implement NotificationPanel redesign

dylanjeffers:
### Description

Implements new NotificationPanel with new redesigned notifications

image


audius-infra:
Preview this change https://demo.audius.co/dj-pay-176-finalize-desktop-notification-redesign

audius-infra:
Preview this change https://demo.audius.co/dj-pay-176-finalize-desktop-notification-redesign

packages/web/src/components/notification/Notifications/Notification.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const entity = useSelector((state: CommonState) =>
+    getNotificationEntity(state, notificationProp)
+  )
+
+  const entities = useSelector((state: CommonState) =>
+    getNotificationEntities(state, notificationProp)
+  )
+
+  // Based on how notification types are defined, we need to cast like this.
+  // In the future we should select user/users/entity/entities in each notif.
```
rickyrombo, line 56: +1. I think ideally the NotificationPanel can use some mapping of NotificationType => Notification element and avoid the wrapper in the future

packages/web/src/components/notification/Notifications/NotificationTile.tsx
```diff
@@ -12,11 +19,24 @@ type NotificationTileProps = {
...
 export const NotificationTile = (props: NotificationTileProps) => {
-  const { notification, onClick, children, disabled } = props
+  const { notification, onClick, children, disabled, disableClosePanel } = props
   const { isRead } = notification
+  const dispatch = useDispatch()
+
+  const handleClick: MouseEventHandler = useCallback(
+    event => {
+      onClick?.(event)
+      if (!disableClosePanel) {
```
rickyrombo, line 34: This works, but after thinking about it, what if we don't close the `NotificationsPanel` on clicks, but rather on navigations? The goal is to close it whenever the user navigates away, right? I guess we'd need both as we probably want to close it on Twitter modal popups? I'm a bit uninformed here for sure.

rickyrombo (approved):

rickyrombo (commented):
One more thing - did we check electron? Saw the panel had some `isElectron` flag

packages/web/src/components/notification/Notifications/Notification.tsx
dylanjeffers: yeah the mapping def makes sense! im not sure if the wrapping is smart enough to figure out that if the type is x, then the component is y, and the prop shape is z, though, if that makes sense?

rickyrombo: I'm thinking in the future these notifs will all have the same props structure so we just pass it in the same to each one, but maybe not?

packages/web/src/components/notification/Notifications/NotificationTile.tsx
dylanjeffers: yeah this is good thinking for sure! yeah do we want to close in on twitter? we don't currently which i think makes sense. so def agree with the idea that it should be on navigates, how would we do that for the complex interaction of `click follow notif with multiple follows -> opens user list modal -> click on a user goes to profile and closes notification?` just have an event listener for url changes in panel?

dylanjeffers:
> One more thing - did we check electron? Saw the panel had some `isElectron` flag

that is a good call, i noticed the isElectron was never used in the original code, so looked safe to remove

packages/web/src/components/notification/Notifications/NotificationTile.tsx
rickyrombo: Ah if we don't currently then no - I'm a little hazy on the ins/outs of when we close it currently sorry! All the route changes should be going through the store, right? Maybe a saga listens for route pushes and put an action to hide notif panel? Seems like a misuse of sagas, but maybe it's fine 😁  Could also be a reducer but i think we slice those up so not sure a notif panel reducer would see the route actions

dylanjeffers: no worries youre thinking is spot on. ill think about things more and see if i come to a sensible way of handling this!

packages/web/src/components/notification/Notifications/Notification.tsx
dylanjeffers: they will be diff in the sense that they have BaseNotif, but then other things like value vs reaction vs whatevs i think!

rickyrombo: Hmm I'm thinking like:

NotificationPanel.tsx:
```jsx
// ....
       notifications.map(notification =>
           <notificationTypeMap[notification.type] key={notification.id} {...notification} />
```

Is that what you're thinking?

dylanjeffers: yeah the main issue with that with ts though is that `Notification` is a discriminated union, so we need to get the component from the type, and also merge in the notification with the corresponding type, that's why the switch case works here cause it aligns the notification type so the component + notification type checks correctly

rickyrombo: Makes sense, gotta appease the typescript engine. Also makes it easier to reason about anyway probably. I'm ok with switch cases, the only slight concern for me is the intermediate React component. I feel like I remember that causing wrapping <divs> and/or having perf implications, but it's been a while and I'm probably wrong and it's not an issue anymore or never was...

dylanjeffers: yeah i def hear you on the extra component... will be thinking of a way to make typescript happy while also avoiding a wrapper

PR #1311, opened 2022-05-12
C-408 Send OAuth read-only response

nicoback2:
### Description
Form JWT containing user's profile information to send back to the redirect URI.

### Dragons

### How Has This Been Tested?

### DIFF
--- a/packages/web/src/pages/oauth-login-page/OAuthLoginPage.tsx
+++ b/packages/web/src/pages/oauth-login-page/OAuthLoginPage.tsx
@@ -84,7 +84,18 @@ export const OAuthLoginPage = () => {
       return
     }
     const timestamp = Math.round(new Date().getTime() / 1000)
-    const response = { email, state, iat: timestamp }
+    const response = {
+      userId: account?.user_id,
+      email,
+      name: account?.name,
+      handle: account?.handle,
+      verified: account?.is_verified,
+      // TODO(nkang): Get profile pic URL
+      // imageURL: account?._profile_picture_sizes,
+      state,
+      sub: account?.user_id,
+      iat: timestamp
+    }
     const header = base64url.encode(
       JSON.stringify({ typ: 'JWT', alg: 'keccak256' })
     )
@@ -118,6 +129,7 @@ export const OAuthLoginPage = () => {
       signInResponse.user.name
     ) {
       // Success - perform Oauth authorization
+      await authAndRedirect()
     } else if (
       (!signInResponse.error &&
         signInResponse.user &&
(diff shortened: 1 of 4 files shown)


packages/web/src/pages/oauth-login-page/OAuthLoginPage.tsx
```diff
@@ -1,27 +1,119 @@
...
-  const { search } = useLocation()
-  const { scope, api_key, state, redirect } = queryString.parse(search)
+  const { search, hash } = useLocation()
+  const isRedirectValidRef = useRef<undefined | boolean>(undefined)
+  const { scope, state, redirect_uri } = queryString.parse(search)
+  const parsedRedirectUri = useMemo(() => {
+    if (redirect_uri && typeof redirect_uri === 'string') {
+      let res: URL
+      try {
+        res = new URL(redirect_uri)
```
sliptype, line 23: Can we just return here?
```suggestion
        return new URL(redirect_uri)
```

packages/web/src/pages/oauth-login-page/OAuthLoginPage.tsx
```diff
@@ -1,27 +1,119 @@
...
+    parsedRedirectUri == null ? null : parsedRedirectUri.host
+  const { token } = queryString.parse(hash)
   const account = useSelector(getAccountUser)
   const isLoggedIn = Boolean(account)
-  const hasMetaMask = !!(window as any).web3
   const [email, setEmail] = useState('')
   const [password, setPassword] = useState('')
   const [submitError, setSubmitError] = useState<string | null>(null)
 
+  useEffect(() => {
```
sliptype, line 40: What's the reasoning behind making `isRedirectValid` a ref?

You could use `useMemo` here I believe:

```suggestion
  const isRedirectValid = useMemo(() => {
```

packages/web/src/pages/oauth-login-page/OAuthLoginPage.tsx
```diff
@@ -49,30 +142,51 @@ export const OAuthLoginPage = () => {
...
-      // window.location.reload()
-    } catch (err) {
-      console.error({ err })
+  const authAndRedirect = async () => {
+    const jwt = await formOAuthResponse()
+    if (jwt == null) {
+      return
+    }
+    const statePart = state != null ? `state=${state}&` : ''
+    const fragment = `#${statePart}token=${jwt}`
```
sliptype, line 151: Why put the token in the hash rather than as a query param?

sliptype (approved):
This is looking great! The redirect validation is definitely interesting bc we aren't registering the url, but good enough for now

packages/web/src/pages/oauth-login-page/OAuthLoginPage.tsx
nicoback2: No real reason! I'll change it so it's more grokkable

audius-infra:
Preview this change https://demo.audius.co/nkang--oauth-response

PR #1312, opened 2022-05-12
[PAY-186] Address desktop notification qa

dylanjeffers:
### Description

Fixes various notification qa issues:

1. Fix markAsRead
2. Fix AnnouncementNotification
3. Enable UserListModal for repost/follow notifs with more than 9 users


audius-infra:
Preview this change https://demo.audius.co/dj-pay-186-desktop-notification-qa

packages/web/src/components/notification/Notifications/Notification.tsx
```diff
@@ -41,7 +42,7 @@ export const Notification = (props: NotificationProps) => {
   )
 
   const users = useSelector((state: CommonState) =>
-    getNotificationUsers(state, notificationProp, USER_LENGTH_LIMIT)
+    getNotificationUsers(state, notificationProp, USER_LENGTH_LIMIT + 3)
```
rickyrombo, line 45: Why +3?

packages/web/src/components/notification/Notifications/FollowNotification.tsx
```diff
@@ -37,10 +37,10 @@ type FollowNotificationProps = {
 
 export const FollowNotification = (props: FollowNotificationProps) => {
   const { notification } = props
-  const { users, timeLabel, isRead } = notification
-  const [firstUser, ...otherUsers] = users
-  const otherUsersCount = otherUsers.length
-  const isMultiUser = users.length > 1
+  const { users, userIds, timeLabel, isRead } = notification
+  const [firstUser] = users
```
rickyrombo, line 41: This is cool - we don't need `otherUsers` anymore!

rickyrombo (approved):
Looks really good!!

packages/web/src/components/notification/Notifications/Notification.tsx
dylanjeffers: Not great catch, that was a test

packages/web/src/components/notification/Notifications/FollowNotification.tsx
dylanjeffers: Yeah realized that couldn't be relied on since we only pull 8 users max, so need the ids from the actual notif to know the true number of users

rickyrombo: Wait but should it be `const otherUsersCount = userIds.length - 1` below? Won't this double count the first user?

dylanjeffers: Yup nice catch

PR #1318, opened 2022-05-16
Fix typecheck by adding missing notif type

sliptype:
### Description

* Fixes type check by adding missing `TipReceived` notif type to mappings

will cherry pick onto release branch

### Dragons

### How Has This Been Tested?


sliptype:
> Ah great catch! I wonder if there is a way to typecheck both projects on commits what you think?

Oh I see, this happened because of the change in the common type. Yeah I think we should be able to add the typecheck to `.hooks/shared` and it will run on all the packages regardless of which files changed

PR #1319, opened 2022-05-16
Remove CI steps for old way of handling AASA file

sliptype:
### Description

* A couple of build steps are failing because they are looking for the old location of the AASA file. I believe these steps are no longer necessary because the AASA file is part of the build in the public folder now
* Updates ci config to trigger both web and mobile flows when only ci config has changed

### Dragons

Possibly breaking AASA

### How Has This Been Tested?

Waiting for CI steps to pass

Verify that CI passes and universal links continue to function


raymondjacobson (approved):
Thanks!

PR #1323, opened 2022-05-17
[C-516] Fix typecheck in ci, address existing errors

dylanjeffers:
### Description

- Ensures typecheck is run on web ci run
- Fixes various existing type errors for web
- Fixes various existing type errors for native
- Fixes type issue with react-use (due to us using typescript v3, i have a ticket to upgrade, but it requires many type changes in the sagas)

### Dragons

For some reason a few type issues appeared in native. It seems to be trying to typecheck audius-client/src/components, i think due to some potential circular logic where a store file imports a component file. will look into this more. ideally this all goes away when we move web out of mobile soon though.

### DIFF
--- a/packages/web/src/common/store/tipping/slice.ts
+++ b/packages/web/src/common/store/tipping/slice.ts
@@ -7,6 +7,11 @@ import { User } from 'common/models/User'
 import { BNWei } from 'common/models/Wallet'
 import { TippingState } from 'common/store/tipping/types'
 
+type Thing = PayloadAction<{
+  senderUserId: ID
+  receiverUserId: ID
+}>
+
 const initialState: TippingState = {
   supporters: {},
   supporting: {},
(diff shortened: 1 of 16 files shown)


packages/web/src/common/store/tipping/slice.ts
```diff
@@ -7,6 +7,11 @@ import { User } from 'common/models/User'
 import { BNWei } from 'common/models/Wallet'
 import { TippingState } from 'common/store/tipping/types'
 
+type Thing = PayloadAction<{
+  senderUserId: ID
+  receiverUserId: ID
+}>
```
piazzatron, line 13: ???

piazzatron (approved):

packages/web/src/common/store/tipping/slice.ts
dylanjeffers: lol good catch, was using it to figure out the og error, removing

PR #1325, opened 2022-05-18
[PAY-212] Add ErrorBoundary to NotificationTile

dylanjeffers:
### Description

Adds ErrorBoundary to NotificationTile so issues with notification tiles don't break the whole page. This is necessary because we are still not strictly typing notification types, so there are cases where we expect users to be present when they aren't. Hoping to address this after native notifications, since the notification types need to be changed wholesale!


audius-infra:
Preview this change https://demo.audius.co/dj-pay-212-error-boundary-notification-tile

piazzatron (approved):

audius-infra:
Preview this change https://demo.audius.co/dj-pay-212-error-boundary-notification-tile

PR #1332, opened 2022-05-19
[C-521] Improve native notification performance

dylanjeffers:
### Description

Improves native notification perf by replacing react-native-shadow-2 with core Tile component. It definitely cuts down on the choppy frames when the notif drawer slides in.
Refactors complementary components to use modern styling system, named exports, and naming conventions


packages/mobile/src/screens/notifications-screen/NotificationsScreen.tsx
```diff
@@ -1,53 +1,38 @@
...
+    backgroundColor: palette.background
   }
-})
+}))
 
-/**
- * A component that renders a user's notifications
- */
-export const NotificationsScreen = memo(() => {
+export const NotificationsScreen = () => {
```
sliptype, line 20: I did add the memo at some point to prevent the notifications screen from rerendering on bottom tab navigation. Can you double check to see if that is still happening?

sliptype (approved):
Awesome! Yeah some of that notification styling was using the extra old methods

packages/mobile/src/screens/notifications-screen/NotificationsScreen.tsx
dylanjeffers: ahh good call, yeah will check

dylanjeffers: re-added and added comment!

PR #1339, opened 2022-05-20
[C-528] Fix nft playlist deep link

dylanjeffers:
### Description

Fixes issue where clicking on an nft-playlist deep link opens the track screen and breaks.
Also fixes case where audius.co links weren't picked up


raymondjacobson (approved):
do we want to have some sort of abstract mapping
{ regex : destination } to make this easier ot add to?

sliptype (approved):
Nice!

dylanjeffers:
> do we want to have some sort of abstract mapping { regex : destination } to make this easier ot add to?

yeah absolutely agree, adding a ticket for it

PR #1340, opened 2022-05-21
[C-528] Fix useSelectorWeb return type when using isEqual

dylanjeffers:
### Description

Noticed that lodash's `isEqual` type is `(a: any, b: any) => boolean` which resulted in `any` types for selected values using `useSelectorWeb` with `isEqual` equalityFn, ie
```
const thing = useSelectorWeb(getThing, isEqual)
```
results in `thing: any` which is bad!

- Re-exports a typed `isEqual` that is compatible with useSelectorWeb
- Updates usages
- Fixes types errors that resulted from implicitly passing around `any`
- Removes explicit typings trying to combat the original `any`


packages/mobile/src/store/search/types.ts
```diff
@@ -28,8 +28,8 @@ export type SearchUser = UserMultihash &
     blocknumber: number
     wallet: string
     created_at: string
-    current_user_followee_follow_count: null
-    does_current_user_follow: null
+    current_user_followee_follow_count: boolean
```
sliptype, line 31: Should this be a number?

sliptype (approved):
Great catch! Thanks

packages/mobile/src/store/search/types.ts
dylanjeffers: oh yeah def, nice

PR #1344, opened 2022-05-23
[C-537] Add button corner prop, text inherit color

dylanjeffers:
### Description

- Adds `corner` prop to button
- Adds `inherit` option for `color` prop for text

image

### Dragons

is `corner` the prop type we want here? the goal is to have a "variant" type option for pill vs rounded buttons.

### DIFF
--- a/packages/mobile/src/components/core/Button.tsx
+++ b/packages/mobile/src/components/core/Button.tsx
@@ -317,8 +317,6 @@ export const Button = (props: ButtonProps) => {
 
   const PressableComponent = url ? Link : Pressable
 
-  console.log(styles.root)
-
   return (
     <View
       style={rootHeightRef.current ? { height: rootHeightRef.current } : null}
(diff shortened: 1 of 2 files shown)


packages/mobile/src/components/core/Button.tsx
```diff
@@ -293,6 +317,8 @@ export const Button = (props: ButtonProps) => {
 
   const PressableComponent = url ? Link : Pressable
 
+  console.log(styles.root)
```
sliptype, line 320: log

sliptype (approved):
Do we want to support different types of corners for each variant? Or should `pill` just be a variant

dylanjeffers:
Yeah that's what I'm not sure about. Julian mentioned in the figma comment that there is a rounded and pill style, so felt like a different category like size, ya know?

nicoback2:
`corners: 'pill' | 'rounded'` LGTM if that's how Julian classifies it, though I think you could also just do `pill?: true | false` if the only corner variants are going to be normal (rounded) and pill forever

nicoback2 (approved):

dylanjeffers:
> `corners: 'pill' | 'rounded'` LGTM if that's how Julian classifies it, though I think you could also just do `pill?: true | false` if the only corner variants are going to be normal (rounded) and pill forever

yeah i can def see pill: true/false, also not sure about "corners" as the prop, so maybe if have another variant like squared, or transparent, or shadowed, maybe update prop name to reflect that type of direction....

PR #1351, opened 2022-05-24
[PAY-239] Add native supporter/supporting user lists

dylanjeffers:
### Description

Adds native supporter/supporting user lists
Adds supporting/supporter sections to user-list-item

image


packages/mobile/src/screens/user-list-screen/Tip.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export const Tip = (props: TipProps) => {
+  const { amount } = props
+  const styles = useStyles()
+  const { neutralLight4 } = useThemeColors()
+
+  return (
+    <View style={styles.root}>
+      <IconTip fill={neutralLight4} height={14} width={14} />
+      <Text color='neutralLight4'>
```
rickyrombo, line 28: Does it make sense to wrap like this or should we just have `color='neutralLight4'` twice

audius-infra:
Preview this change https://demo.audius.co/dj-pay-239-supporter-user-lists

packages/mobile/src/screens/user-list-screen/UserListItem.tsx
```diff
@@ -63,11 +66,12 @@ const useStyles = makeStyles(({ spacing, palette }) => ({
 }))
 
 type UserListItemProps = {
+  tag: string
```
rickyrombo, line 69: nit: made the same comment in @sddioulde 's PR - `tag` feels overloaded as a term, is there something else we could use? Also, should this maybe be an enum?

rickyrombo (approved):

packages/mobile/src/screens/user-list-screen/UserListItem.tsx
dylanjeffers: yes to all the above, ideally we have a better prop name, and then also define these tags in a common file. im down to look into that moving forward, just don't want to slow down this work for that large refactor? https://linear.app/audius/issue/C-541/consolidate-user-list-tags-and-rename-props

packages/mobile/src/screens/user-list-screen/Tip.tsx
dylanjeffers: i was trying something new out, but agreed it seems quite redundant here!

audius-infra:
Preview this change https://demo.audius.co/dj-pay-239-supporter-user-lists

PR #1352, opened 2022-05-24
Add tip tile on feed page

sddioulde:
### Description
Add recent tip tile on feed page based on logged in user's followees.
Click on X button to dismiss tile.
Click on tip button to go through tipping flow.

Screen Shot 2022-05-24 at 3 34 16 PM

### Dragons

n/a

### How Has This Been Tested?

local against stage

no extra monitoring beyond tip analytics


audius-infra:
Preview this change https://demo.audius.co/sd-tip-feed

packages/web/src/components/tipping/feed-tip-tile/FeedTipTile.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+  const handleClick = useCallback(() => {
+    dispatch(beginTip({ user }))
+  }, [dispatch, user])
+
+  return (
+    <div>
+      <Button
+        className={styles.sendTipButton}
+        // todo: move to stems or see if button design
```
sddioulde, line 83: captured in linear

packages/web/src/store/tipping/sagas.ts
```diff
@@ -223,6 +230,9 @@ function* fetchSupportingForUserAsync({
     return
   }
 
+  // todo: need to also get whether logged in user is supporting this user
```
sddioulde, line 233: this is what we talked about re fetching support data given sender and receiver. i will follow up

packages/web/src/store/tipping/sagas.ts
```diff
@@ -254,10 +263,125 @@ function* fetchSupportingForUserAsync({
...
+}: {
+  payload: { userId: ID }
+  type: string
+}) {
+  const encodedUserId = encodeHashId(userId)
+  if (!encodedUserId) {
+    return
+  }
+
+  // todo: need to also get whether this user is supported by logged in user
```
sddioulde, line 277: same as above

packages/web/src/store/tipping/utils.ts
```diff
@@ -0,0 +1,187 @@
...
+  const storage = getRecentTipsStorage()
+  return storage ? storage.minSlot : null
+}
+
+export const getDismissedSlotForRecentTips = () => {
+  const storage = getRecentTipsStorage()
+  return storage ? storage.dismissed : null
+}
+
+export const checkRecentTip = ({
```
sddioulde, line 37: want to refactor this, feels like there is room to simplify / reduce lines

audius-infra:
Preview this change https://demo.audius.co/sd-tip-feed

packages/web/src/pages/feed-page/components/desktop/FeedPageContent.tsx
```diff
@@ -108,6 +108,7 @@ const FeedPageContent = ({
           }
           endOfLineup={<EndOfLineup key='endOfLineup' />}
           key='feed'
+          isFeed={true}
```
dylanjeffers, line 111: btw just `isFeed` is preferred for booleans

packages/web/src/services/audius-backend/Tipping.ts
```diff
@@ -58,3 +61,71 @@ export const fetchSupporters = async ({
...
+  receiverIsVerified?: boolean
+  currentUserFollows?: 'sender' | 'receiver' | 'sender_or_receiver'
+  uniqueBy?: 'sender' | 'receiver'
+  minSlot?: number
+  maxSlot?: number
+  txSignatures?: string[]
+}
+type UserTipOmitIds = 'sender_id' | 'receiver_id' | 'followee_supporter_ids'
+type UserTipResponse = Omit<UserTip, UserTipOmitIds> & {
+  sender: any
```
dylanjeffers, line 79: just wondering why cant be string or smthg?

packages/web/src/services/audius-backend/Tipping.ts
```diff
@@ -58,3 +61,71 @@ export const fetchSupporters = async ({
...
+  minSlot,
+  maxSlot,
+  txSignatures
+}: UserTipRequest): Promise<UserTipResponse[]> => {
+  try {
+    await waitForLibsInit()
+    const response = await libs().discoveryProvider._makeRequest({
+      endpoint: `/v1/full/tips`,
+      queryParams: {
+        user_id: userId,
```
dylanjeffers, line 101: would it make sense to create queryParams const and then use it here and also for console.error?

packages/web/src/store/tipping/sagas.ts
```diff
@@ -254,10 +263,125 @@ function* fetchSupportingForUserAsync({
...
+        amount,
+        sender_id: senderId,
+        receiver_id: receiverId,
+        followee_supporter_ids: followeeSupporterIds,
+        slot,
+        created_at,
+        tx_signature
+      }
+    })
+    .filter((userTip): userTip is UserTip => !!userTip)
```
dylanjeffers, line 354: nice

packages/web/src/store/tipping/utils.ts
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  /**
+   * The list only comprises of recent tips.
+   * Sort them tips by least recent to parse through oldest tips first.
+   */
+  const sortedTips = recentTips.sort((tip1, tip2) => tip1.slot - tip2.slot)
+
+  /** Return oldest of the recent tips if no local storage */
+  if (!window.localStorage) {
```
dylanjeffers, line 55: i thinkk we are good, but just wanted to double check this won't run on native yeah?

dylanjeffers (approved):
wow tons of great stuff. from code perspective things look good to me, but may be good to get someone familiar with the api and state to review as wel

packages/web/src/services/audius-backend/Tipping.ts
```diff
@@ -58,3 +61,71 @@ export const fetchSupporters = async ({
...
+  receiverIsVerified,
+  currentUserFollows,
+  uniqueBy,
+  minSlot,
+  maxSlot,
+  txSignatures
+}: UserTipRequest): Promise<UserTipResponse[]> => {
+  try {
+    await waitForLibsInit()
+    const response = await libs().discoveryProvider._makeRequest({
```
piazzatron, line 98: Any reason we're calling `_makeRequest` directly? This probably should live in APIClient with the rest of our DN interface. Plus that would mean we don't have to expose all of these query params which we're not currently using to the rest of the client code

packages/web/src/services/audius-backend/Tipping.ts
```diff
@@ -1,11 +1,14 @@
...
+import { Supporter, Supporting, UserTip } from 'common/models/Tipping'
 import { waitForLibsInit } from 'services/audius-backend/eagerLoadUtils'
 
 export const TIPPING_SUPPORT_DEFAULT_LIMIT = 25
 
 // @ts-ignore
 const libs = () => window.audiusLibs
 
+// todo: I think we should revisit our UserMetadata, User, APIUser models
+// seems a bit off now and kind of confusing, reason why I have any types below
```
piazzatron, line 10: Not sure exactly what we mean here - imo when we're checking in todos they should be fairly concrete, ideally things we can track as issues

packages/web/src/store/tipping/utils.ts
```diff
@@ -0,0 +1,187 @@
...
+  if (!value) {
+    return null
+  }
+
+  return JSON.parse(value) as RecentTipsStorage
+}
+
+export const getMinSlotForRecentTips = () => {
+  const storage = getRecentTipsStorage()
+  return storage ? storage.minSlot : null
```
piazzatron, line 29: could do `storage?.minSlot ?? null` for these and below

packages/web/src/store/tipping/utils.ts
```diff
@@ -0,0 +1,187 @@
...
+  userId: ID
+  recentTips: UserTip[]
+}) => {
+  if (recentTips.length === 0) {
+    return null
+  }
+
+  /**
+   * The list only comprises of recent tips.
+   * Sort them tips by least recent to parse through oldest tips first.
```
piazzatron, line 50: typo `them tips`

packages/web/src/store/tipping/utils.ts
piazzatron, line 37: nit: naming feels a bit unclear, could be something like `findRecentTipToDisplay`?

packages/web/src/store/tipping/utils.ts
```diff
@@ -0,0 +1,187 @@
...
+   * Also set local storage values.
+   */
+  if (!storage) {
+    const oldestValidTip = sortedTips[0]
+    const newStorage = {
+      minSlot: oldestValidTip.slot,
+      dismissed: false,
+      lastDismissalTimestamp: null
+    }
+    window.localStorage.setItem(RECENT_TIPS_KEY, JSON.stringify(newStorage))
```
piazzatron, line 71: We could have a setRecentTipsStorage method to complement our getRecentTipsStorage one, which would handle stringifying + setting w the correct key. Would help clean this up a good bit!

packages/web/src/store/tipping/utils.ts
```diff
@@ -0,0 +1,187 @@
...
+   * more recent than dismissed tip.
+   * If not found, then try to find tip more recent
+   * than dismissed tip.
+   * If not found, then try to find tip as recent as that
+   * which was dismissed if the last dismissal is too old.
+   * Otherwise, there is nothing to return.
+   *
+   * Also update local storage if some tip is returned.
+   */
+  if (storage.dismissed) {
```
piazzatron, line 87: Is there any chance that minSlot doesn't exist for some reason? We should probably sanity check ourselves here and return null, just in case

packages/web/src/services/audius-backend/Tipping.ts
sddioulde: because it's a user-like object

packages/web/src/store/tipping/utils.ts
```diff
@@ -0,0 +1,187 @@
...
+}) => {
+  if (recentTips.length === 0) {
+    return null
+  }
+
+  /**
+   * The list only comprises of recent tips.
+   * Sort them tips by least recent to parse through oldest tips first.
+   */
+  const sortedTips = recentTips.sort((tip1, tip2) => tip1.slot - tip2.slot)
```
piazzatron, line 52: good idea w this, makes this a lot simpler

packages/web/src/services/audius-backend/Tipping.ts
sddioulde: yup makes sense

packages/web/src/store/tipping/utils.ts
```diff
@@ -0,0 +1,187 @@
...
+      return oldestValidTip
+    }
+
+    if (
+      storage.lastDismissalTimestamp &&
+      Date.now() - storage.lastDismissalTimestamp >
+        FEED_TIP_DISMISSAL_TIME_LIMIT
+    ) {
+      const oldestValidTip = sortedTips.find(
+        tip => tip.slot === storage.minSlot
```
piazzatron, line 118: Should this be `>=` instead of `===`? That way we don't have to have an exact match; if a user has the browser closed for a long time, chances are the stored minSlot won't be represented in the current tips

packages/web/src/store/tipping/utils.ts
```diff
@@ -0,0 +1,187 @@
...
+   * first try to find user's own tip that is
+   * more recent than dismissed tip.
+   * If not found, then try to find tip more recent
+   * than dismissed tip.
+   * If not found, then try to find tip as recent as that
+   * which was dismissed if the last dismissal is too old.
+   * Otherwise, there is nothing to return.
+   *
+   * Also update local storage if some tip is returned.
+   */
```
piazzatron, line 86: good looks including these comments - makes this much easier to grok

packages/web/src/store/tipping/utils.ts
sddioulde: are you asking if the native mobile code can reach this flow? i think the answer is yes because the tipping sagas calls this function when fetching recent tips on the feed page

packages/web/src/store/tipping/utils.ts
```diff
@@ -0,0 +1,187 @@
...
+    const newStorage = {
+      minSlot: oldestValidTip.slot,
+      dismissed: false,
+      lastDismissalTimestamp: null
+    }
+    window.localStorage.setItem(RECENT_TIPS_KEY, JSON.stringify(newStorage))
+    return oldestValidTip
+  }
+
+  return null
```
piazzatron, line 168: Hmm, this is the only part of the logic that I don't think is quite correct – pretty sure if haven't dismissed the tile but there are no newer tips than we saw last time, we should just show the newest thing (rather than showing nothing)?

packages/web/src/store/tipping/utils.ts
piazzatron: unless intentional, which I sort of like

packages/web/src/store/tipping/utils.ts
piazzatron: If we wanted to refactor, LMK if I'm missing something (very possible), but I think we could do something like:
- In all cases, do your two checks for own tip > minSlot, then fall back to any tip > minSlot. 
- If the above is null, then logic branch on whether we're in dismissed mode or not

packages/web/src/common/store/tipping/slice.ts
```diff
@@ -22,7 +22,10 @@ const initialState: TippingState = {
     user: null,
     amount: new BN('0') as BNWei,
     error: null
-  }
+  },
+  recentTips: [],
+  recentTip: null,
```
piazzatron, line 27: Wasn't clear what the difference between these two fields was based on the names - maybe something like `recentTips` and `tipToDisplay` would help?

packages/web/src/store/tipping/utils.ts
dylanjeffers: oh yeah nvm this is okay since this code will run in the webview on native, nvm!!

packages/web/src/common/store/wallet/selectors.ts
```diff
@@ -1,19 +1,30 @@
...
 import { CommonState } from 'common/store'
 import { Nullable } from 'common/utils/typeUtils'
 import { stringWeiToBN } from 'common/utils/wallet'
 
-export const getAccountBalance = (state: CommonState): Nullable<BNWei> => {
-  const balance = state.wallet.balance
-  if (!balance) return null
-  return stringWeiToBN(balance)
+// Previously, the getAccountBalance selector would return different
+// values (although shallowly the same) because of the return of a
```
piazzatron, line 9: interesting find 👀. Wonder how many other places are affected by this...

nit: probably want to say something like "numerically the same" - shallow compare means checking reference equality, so in this case they wouldn't be shallowly the same

piazzatron (approved):
Looks great! That algorithm was fun :)

packages/web/src/utils/constants.ts
```diff
@@ -9,3 +9,5 @@ export const TIPPING_TOP_RANK_THRESHOLD = 5
 export const MAX_PROFILE_TOP_SUPPORTERS = 5
 export const MAX_PROFILE_SUPPORTING_TILES = 3
 export const MAX_ARTIST_HOVER_TOP_SUPPORTING = 7
+export const NUM_FEED_TIPPERS_DISPLAYED = 2
+export const FEED_TIP_DISMISSAL_TIME_LIMIT = 30 * 24 * 60 * 60 // 30 days
```
sliptype, line 13: Should this be in milliseconds instead of seconds?

sliptype (approved):

packages/web/src/utils/constants.ts
sddioulde: yup!

packages/web/src/store/tipping/utils.ts
sddioulde: this should not even really be reachable. the check is for `>=` so if a tile had already existed before, its tip should always exists even if there is no new tip after it, which means we won't reach this scenario

packages/web/src/services/audius-backend/Tipping.ts
sddioulde: hmm did we not want to have more separated and domain specific files for api requests?

audius-infra:
Preview this change https://demo.audius.co/sd-tip-feed

packages/web/src/store/tipping/utils.ts
sddioulde: there is an ever-so-slight diff with that algo, but i dont think it matters significantly. will go with it

packages/web/src/store/tipping/utils.ts
sddioulde: lol. it was a typo

audius-infra:
Preview this change https://demo.audius.co/sd-tip-feed

audius-infra:
Preview this change https://demo.audius.co/sd-tip-feed

PR #1355, opened 2022-05-25
aud-1172 show lofi in trending genre pickers

csjiang:
### Description

Previously, the Lo-Fi genre was filtered out of trending genres due to being a new genre. At this point we have enough music to show, so this change reveals lofi music in trending.

### Dragons

N/A

### How Has This Been Tested?

ran client locally against stage

N/A


audius-infra:
Preview this change https://demo.audius.co/AUD-1172_show_lofi_genre_in_trending

PR #1357, opened 2022-05-25
Add new waitForUserRemoteConfig for feature flags that depend on user

sddioulde:
### Description

Add new waitForUserRemoteConfig for feature flags that depend on user

### Dragons

n/a

### How Has This Been Tested?

local against stage and tested the tipping feature flag returns true on profile sagas; it used to always be false.

### DIFF
--- a/packages/web/src/common/services/remote-config/remote-config.ts
+++ b/packages/web/src/common/services/remote-config/remote-config.ts
@@ -23,6 +23,8 @@ import {
 import { Nullable } from 'common/utils/typeUtils'
 import { uuid } from 'common/utils/uid'
 
+export const USER_ID_AVAILABLE_EVENT = 'USER_ID_AVAILABLE_EVENT'
+
 // Constants
 // All optimizely feature keys are lowercase_snake
 const REMOTE_CONFIG_FEATURE_KEY = 'remote_config'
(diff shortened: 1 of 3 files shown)


packages/web/src/common/services/remote-config/remote-config.ts
```diff
@@ -218,6 +218,33 @@ export const remoteConfig = <
...
+    await new Promise<void>((resolve, reject) => {
+      let numRetries = 6
+      const interval = setInterval(() => {
+        if (state.userId) {
+          onClientReady(resolve)
+          clearInterval(interval)
+        }
+        if (--numRetries === 0) {
+          clearInterval(interval)
+          reject(new Error('Remote config user id is not available.'))
```
sddioulde, line 242: should we reject? or just let it be so that calling places (e.g. profile sagas) don't have to try/catch.
wondering because the already existing waitForRemoteConfig does not reject

sddioulde:
open to other implementation suggestions

sddioulde:
ideally, i'd want some sort of event listening here. like when user is available we trigger an event, and listen to that event in this waitForUserRemoteConfig

audius-infra:
Preview this change https://demo.audius.co/sd-fix-remote-config-user

packages/web/src/common/services/remote-config/remote-config.ts
```diff
@@ -218,6 +220,29 @@ export const remoteConfig = <
...
+  /**
+   * Need this function for feature flags that depend on user id.
+   * This is because the waitForRemoteConfig does not ensure that
+   * user id is available before its promise resolution, meaning
+   * that it will sometimes return false for a feature flag
+   * that is enabled and supposed to return true.
+   */
+  const waitForUserRemoteConfig = async () => {
+    if (state.userId) {
+      await new Promise<void>(resolve => onClientReady(resolve))
```
dylanjeffers, line 232: im not that familiar with promises, but can we just do `return onClientReady()` since async always returns a promise anyway?

packages/web/src/store/account/sagas.js
```diff
@@ -184,8 +185,12 @@ export function* fetchAccountAsync(action) {
     return
   }
 
-  // Set account ID in remote-config provider
+  // Set account ID and let remote-config provider
+  // know that the user id is available
   remoteConfigInstance.setUserId(account.user_id)
+  const event = new CustomEvent(USER_ID_AVAILABLE_EVENT)
```
dylanjeffers, line 191: sweet

dylanjeffers (approved):
this impl loooks great, just curious if we dont need the extra new Promise?

audius-infra:
Preview this change https://demo.audius.co/sd-fix-remote-config-user

packages/web/src/common/services/remote-config/remote-config.ts
sddioulde: hmm, don't think that'll work because onClientReady needs the callback, which comes from the promise and lets us know we're done

PR #1359, opened 2022-05-25
[PAY-247] Add TipArtistModal and basic navigation

dylanjeffers:
### Description

Adds TipArtistModal and basic modal navigation
This is the first native modal that is correctly using react-navigation and not redux modal state

https://user-images.githubusercontent.com/8230000/170354779-ac89d197-e142-40cd-9e53-2200c7bacdaf.mov


rickyrombo (approved):
Amazing! The navigation seems kinda magical

PR #1360, opened 2022-05-25
Rename SearchInput TextInput

dylanjeffers:
### Description

Renames SearchInput to TextInput to make it more obvious to devs that we should be using the core TextInput as a replacement for RNTextInput whenever possible


sliptype (approved):
Thanks!

PR #1361, opened 2022-05-25
[C-546] Fix USER_ID cohort feature flags on native

sliptype:
### Description

* The user id was never being set on the optimizely client on native, resulting in feature flags with the user_id cohort to always return false. Fixed by setting the user id on the `signed_in` message 

### Dragons

n/a

### How Has This Been Tested?

Confirmed that the TikTok feature flag returns true


dylanjeffers (approved):
great find!

PR #1367, opened 2022-05-26
Add Amplitude events

nicoback2:
### Description
Add tracking for Oauth!

### Dragons

### How Has This Been Tested?

### DIFF
--- a/packages/web/src/common/models/Analytics.ts
+++ b/packages/web/src/common/models/Analytics.ts
@@ -62,6 +62,12 @@ export enum Name {
   TIKTOK_COMPLETE_SHARE_SOUND = 'TikTok: Complete Share Sound',
   TIKTOK_SHARE_SOUND_ERROR = 'TikTok: Share Sound Error',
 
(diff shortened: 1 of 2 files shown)


packages/web/src/common/models/Analytics.ts
```diff
@@ -1175,6 +1181,28 @@ type SocialProofError = {
   error: string
 }
 
+type AudiusOauthStart = {
+  eventName: Name.AUDIUS_OAUTH_START
+  redirectUriParam: string | string[]
```
dylanjeffers, line 1186: Just wondering how it could be string[]?

dylanjeffers (approved):
Great changes

packages/web/src/common/models/Analytics.ts
nicoback2: i think it'd be if you did something like `?foo=bar&foo=qux` or `?foo[]=bar&foo[]=qux`

PR #1369, opened 2022-05-26
[PAY-261] Refactor tipping slice amount type from BNWei to StringAudio

dylanjeffers:
### Description

Changes tipping slice amount type from BNWei to StringAudio, which fixes case where `dispatchWeb` incorrectly serializes BNWei (by calling .toString()), resulting in different mobile vs web behavior. Moving to StringWei helps since that is directly serializable across apps.

Also changes StringWei type to be a string alias since it's unclear how these types are actually `string & {_brand: string}` maybe in the future we can change the stringaudio data to properly match these types though.

### Dragons

Changing the underlying type may have messed up the tipping flow, so lets def check this works before merging

### How Has This Been Tested?

Types pass, but a local run-through is needed

### DIFF
--- a/packages/web/src/components/tipping/tip-audio/TipSent.tsx
+++ b/packages/web/src/components/tipping/tip-audio/TipSent.tsx
@@ -7,7 +7,7 @@ import { useSelector } from 'common/hooks/useSelector'
 import { Name } from 'common/models/Analytics'
 import { getAccountUser } from 'common/store/account/selectors'
 import { getSendTipData } from 'common/store/tipping/selectors'
-import { formatWei } from 'common/utils/wallet'
+import { formatNumberCommas } from 'common/utils/formatUtil'
 import { useRecord, make } from 'store/analytics/actions'
 import { openTwitterLink } from 'utils/tweet'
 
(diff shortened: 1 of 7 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-pay-261-refactor-tipping-amount-type

packages/web/src/components/tipping/tip-audio/TipSent.tsx
```diff
@@ -31,15 +31,9 @@ export const TipSent = () => {
 
   const handleShareClick = useCallback(() => {
     if (account && recipient) {
-      let recipientAndAmount = `${recipient.name} ${formatWei(
-        sendAmount,
-        true
-      )}`
+      let recipientAndAmount = `${recipient.name} ${formatWei}`
```
sddioulde, line 34: i think you meant ${sendAmount} instead of the function ${formatWei}

sddioulde: plus, we should format the sendAmount itself

sddioulde (commented):
looks good, pending comment. i also think we should handle formatting in this PR if possible

sddioulde (approved):

PR #1375, opened 2022-05-27
Add native notification error boundary

dylanjeffers:
### Description

Adds native notification error boundary to prevent weird notification type edge cases from breaking app. This is temporary until we properly type all notifications to prevent run time errors.


sliptype (approved):
Nice!

Wonder if we'll ever get a `useCatch` hook or something

PR #1376, opened 2022-05-27
[PAY-265] Add native profile screen redesign

dylanjeffers:
### Description

Adds native profile redesign
Improves TextButton, Link, and other core components

https://user-images.githubusercontent.com/8230000/170801382-86d8222f-35a0-4018-917b-502daaa7d14f.mov


sddioulde (approved):
looks great! and easy to follow.

packages/mobile/src/screens/profile-screen/ProfileHeaderV2/ProfileSocials.tsx
```diff
@@ -0,0 +1,92 @@
...
+    instagram_handle,
+    tiktok_handle
+  } = useSelectProfile([
+    'user_id',
+    'twitter_handle',
+    'instagram_handle',
+    'tiktok_handle'
+  ])
+
+  const socialsCount = [twitter_handle, instagram_handle, tiktok_handle].filter(
```
piazzatron, line 50: could filter.Boolean, idk if cleaner 🤷‍♂️

packages/mobile/src/screens/profile-screen/ProfileHeaderV2/SocialLink.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const styles = useStyles()
+  const { neutral } = useThemeColors()
+
+  if (text === undefined) {
+    return <Skeleton style={styles.skeleton} />
+  }
+
+  if (text === null) {
+    return null
+  }
```
piazzatron, line 62: This is kinda interesting - maybe worth an inline comment in the props delineating the difference btwn null and undefined?

packages/mobile/src/screens/profile-screen/ProfileHeaderV2/SocialsAndSites.tsx
```diff
@@ -0,0 +1,60 @@
...
+  siteText: {
+    color: palette.neutral,
+    fontSize: typography.fontSize.small,
+    fontFamily: typography.fontByWeight.medium,
+    letterSpacing: 0.5
+  }
+}))
+
+const prependProtocol = (url?: string) =>
+  !url?.match(/^https?:\/\//i) ? `https://${url}` : url
```
piazzatron, line 32: Worth a comment imo

packages/mobile/src/screens/profile-screen/ProfileScreen.tsx
```diff
@@ -157,37 +124,16 @@ export const ProfileScreen = () => {
...
-          )}
-          {profileActionButton}
-        </View>
-      </>
+  const renderHeader = useCallback(() => {
+    if (!isLoaded) return null
+    return isTippingEnabled ? (
+      <ProfileHeaderV2 scrollY={scrollY} />
+    ) : (
+      <ProfileHeader scrollY={scrollY} />
```
piazzatron, line 133: beautiful

packages/mobile/src/screens/profile-screen/utils.ts
```diff
@@ -64,6 +63,6 @@ export const useShouldShowCollectiblesTab = () => {
  * trims both leading and trailing newlines
  * @param {string} str
  */
-export const squashNewLines = (str: Nullable<string>) => {
-  return str ? str.replace(/\n\s*\n\s*\n/g, '\n\n').trim() : str
+export const squashNewLines = (str?: string) => {
+  return str ? str.replace(/\n\s*\n\s*\n/g, '\n\n').trim() : ''
```
piazzatron, line 67: We're sure this change doesn't affect anything elsewhere, right?

piazzatron (approved):
Love how you broke up all the Header components 😊

packages/mobile/src/screens/profile-screen/ProfileHeaderV2/ProfileSocials.tsx
dylanjeffers: yeah good call, changing

packages/mobile/src/screens/profile-screen/ProfileHeaderV2/SocialLink.tsx
dylanjeffers: yeah that is a great callout, it's a convenience of the api that we can rely on undefined as 'LOADING' lol

packages/mobile/src/screens/profile-screen/utils.ts
dylanjeffers: good q i think not, since any caller that originally used it had to pass a string, this just helps the one case where bio may be undefined so caller doesn't need to ?? ''

PR #1377, opened 2022-05-28
[C-552] Fix expandible-bio link render issue

dylanjeffers:
### Description

Fixes issue where expandible bio links were rendering on top of one another: 
Before: 

After: 
image
image

The render flow that seems to fix this is:

1. Render the bio with `Text` component
2. Check that it's size is large enough to show `expand` button
  - If it's not large enough, replace `Text` with `Hyperlink` component
  - If it is large enough, keep rendering the `Text` component with 32 height when not expanded (setting height on `Hyperlink` is causing the issue i believe), and replace `Text` with `Hyperlink` when expanded

### Dragons

The one not perfect case is when the bio is expandable, but closed. in that case we still render the text element, so any links on the first two lines of bio wont be clickable till the bio is expanded.


sliptype (approved):
Nice fix! this code keeps getting more ridiculous lol. ~What happens when the first line of the bio is a link?~ nvm just read in the description

raymondjacobson (approved):
code lgtm! i'm not expert here though :) 

wow!

PR #1381, opened 2022-05-30
Fix slider handle position when re-entering from background

dylanjeffers:
### Description

Fixes bug where now-playing scrubber is incorrectly positioned when coming back to app, esp after a new song starts.

### Dragons

The code here in general is a bit complex, and this adds yet another useEffect. there are cases where this may be a duplicate call with another useEffect when mediaKey changes. From what i can tell this is okay, though ideally we look at all thees effects and consider ways to make them simpler/work together.

### DIFF
--- a/packages/mobile/src/components/scrubber/Slider.tsx
+++ b/packages/mobile/src/components/scrubber/Slider.tsx
@@ -1,5 +1,6 @@
 import { memo, useCallback, useEffect, useRef, useState } from 'react'
 
+import { useAppState } from '@react-native-community/hooks'
 import {
   StyleSheet,
   View,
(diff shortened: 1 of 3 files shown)


packages/mobile/src/components/scrubber/Slider.tsx
```diff
@@ -282,6 +283,24 @@ export const Slider = memo(
       }
     }, [isPlaying, play, pause, mediaKey, durationRef])
 
+    const appState = useAppState()
+
+    /**
+     * Ensures the slider handle's position is correctly updated when app
+     * becomes active.
```
raymondjacobson, line 290: This looks right! we do similar in stems on window focus
https://github.com/AudiusProject/stems/blob/master/src/components/Scrubber/hooks.ts#L88

raymondjacobson (approved):

PR #1382, opened 2022-05-30
Prevent favorites tab from resetting on navigation

dylanjeffers:
### Description

Fixes bug where favorites tab always resets causing bad user experience

Before: 

https://user-images.githubusercontent.com/8230000/171063496-4f817ce5-6304-4af6-8cea-4e08d8f33c59.MP4

After:

https://user-images.githubusercontent.com/8230000/171063688-004a2420-18be-4c60-ae03-d025db18f124.mov

### Dragons

Just to be safe, we prevent clear only on native

### DIFF
--- a/packages/mobile/src/screens/favorites-screen/TracksTab.tsx
+++ b/packages/mobile/src/screens/favorites-screen/TracksTab.tsx
@@ -109,7 +109,11 @@ export const TracksTab = () => {
   )
 
   return (
-    <WithLoader loading={savedTracksStatus === Status.LOADING}>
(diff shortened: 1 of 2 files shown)


raymondjacobson (approved):
should we do this on web too?

audius-infra:
Preview this change https://demo.audius.co/dj-fix-favorites-tab-reset

dylanjeffers:
> should we do this on web too?

i dont see why not, but was focused on native with this. will check web and see if it's chill removing

PR #1384, opened 2022-05-30
Fix casting for android, setting for ios

dylanjeffers:
### Description

Fixes pretty bad bug where now-playing drawer always shows airplay for android devices, always leading to fatal error when clicked. Users can't switch to chromecast either since we dont show the option in settings.

Also fixes bug where ios cast setting doesn't default to the actually selected cast method.


raymondjacobson (approved):
yikes

PR #1385, opened 2022-05-30
Fix android crash when restoring from background

dylanjeffers:
### Description

Fixes cases where android app crashes when restoring from background
sentry: https://sentry.io/organizations/audius/issues/3272855032/?query=is%3Aunresolved+release.version%3A1.1.114&statsPeriod=14d
issue: https://github.com/software-mansion/react-native-screens/issues/17#issuecomment-989623502
fix: https://reactnavigation.org/docs/getting-started/#installing-dependencies-into-a-bare-react-native-project

### Dragons

Not sure if this affects the chromecast context call


raymondjacobson (approved):
deep

PR #1388, opened 2022-05-31
Display latest tip support state optimistically

sddioulde:
### Description

Optimistically update the tipping store state to show new supporting / supporter data after a tip is successfully sent, since re-fetching right away may not come back with new data as indexing may take a bit of time to process the new tip.

### Dragons

There will be weird edge cases e.g. in user list modals which fetch the users and will get the backend state instead of the optimistic state.
Should we always do a merge of backend <> optimistic state when setting the data in store?

### How Has This Been Tested?

locally against staging

n/a


audius-infra:
Preview this change https://demo.audius.co/sd-optimistic-support

rickyrombo (approved):

packages/web/src/store/tipping/sagas.ts
```diff
@@ -48,10 +55,94 @@ import {
...
+  receiver
+}: {
+  amountBN: BNWei
+  sender: User
+  receiver: User
+}) {
+  /**
+   * Get supporting map for sender.
+   */
+  const supportersMap = yield* select(getSupporters)
```
sddioulde, line 116: same as above

audius-infra:
Preview this change https://demo.audius.co/sd-optimistic-support

audius-infra:
Preview this change https://demo.audius.co/sd-optimistic-support

packages/web/src/common/store/tipping/selectors.ts
```diff
@@ -31,3 +40,143 @@ export const getTipToDisplay = (state: CommonState) =>
...
+        supportingOverrides[userId]
+      ) as unknown) as ID[]
+      for (const receiverId of receiverIds) {
+        const shouldOverrideReceiver =
+          !supporting[userId][receiverId] ||
+          stringWeiToBN(supporting[userId][receiverId].amount).lt(
+            stringWeiToBN(supportingOverrides[userId][receiverId].amount)
+          )
+        if (shouldOverrideReceiver) {
+          mergedMap = {
```
rickyrombo, line 83: I don't think we need to spread assign here, we can just set `mergedMap[userId][receiverId] = supportingOverrides[userId][receiverId]` right?

packages/web/src/common/store/tipping/selectors.ts
```diff
@@ -31,3 +40,143 @@ export const getTipToDisplay = (state: CommonState) =>
...
+  /**
+   * Merge the default and override maps.
+   */
+  const userIds = (Object.keys(supportingOverrides) as unknown) as ID[]
+  for (const userId of userIds) {
+    // If the supporting map for a given user id exists in the overrides
+    // but not in the default, copy the override map into the merged map.
+    const shouldOverrideMap = !supporting[userId]
+    if (shouldOverrideMap) {
+      mergedMap = {
```
rickyrombo, line 62: same here, I don't think we need a spread assign?

packages/web/src/common/store/tipping/selectors.ts
```diff
@@ -31,3 +40,143 @@ export const getTipToDisplay = (state: CommonState) =>
...
+    }
+  }
+
+  /**
+   * Re-rank everything based on newly merged map.
+   */
+  let result: SupportingMap = {}
+  const mergedUserIds = (Object.keys(mergedMap) as unknown) as ID[]
+  for (const userId of mergedUserIds) {
+    result = {
```
rickyrombo, line 101: I don't think this needs to be spread assigned?

packages/web/src/common/store/tipping/selectors.ts
```diff
@@ -31,3 +40,143 @@ export const getTipToDisplay = (state: CommonState) =>
...
+  }
+
+  return result
+}
+export const getOptimisticSupporting = createSelector(
+  getOptimisticSupportingBase,
+  result => result
+)
+
+const getOptimisticSupportersBase = (state: CommonState) => {
```
rickyrombo, line 114: These two functions look very similar, can we break out their individual pieces into reusable functions? Might have to use generics to make typescript happy?

packages/web/src/common/store/tipping/selectors.ts
```diff
@@ -31,3 +40,143 @@ export const getTipToDisplay = (state: CommonState) =>
...
+  const { supporters, supportersOverrides } = state.tipping
+
+  /**
+   * Copy the supporters map into the eventually-merged map.
+   */
+  let mergedMap: SupportersMap = {}
+  Object.assign(mergedMap, supporters)
```
rickyrombo, line 121: I think we can do `const mergedMap = { ...supporters }` right?

audius-infra:
Preview this change https://demo.audius.co/sd-optimistic-support

rickyrombo (commented):
`getOptimisticSupporting`, `overrideSupportingForUser`, and `optimisticallyUpdateSupporting` all have very similar names... I wonder if we can make things more clear?

Maybe `optimisticallyUpdateSupporting` could be in the `selectors.ts` to keep things colocated and encapsulated, since I think that's the only place it's used? Also could be called `rerankSupportingMapForUser` since that's what it does?

There's also a lot of opportunity to reuse code between supporter/supporting flows.

packages/web/src/components/tipping/feed-tip-tile/FeedTipTile.module.css
```diff
@@ -52,7 +52,6 @@
   text-decoration-color: var(--primary-light-2);
 }
 .badge {
-  width: 12px;
```
rickyrombo, line 55: nit: I know it's annoying to swap branches etc, but would have done this in a separate branch so that this fix doesn't get blocked on other things

packages/web/src/common/store/tipping/selectors.ts
sddioulde: on runtime when not using a spread assign, i got something like this
`Uncaught TypeError: Cannot assign to read only property '' of object '#<Object>`

rickyrombo: Hmm, would be good to not spread assign as I think that's extra expensive. We should figure out the typescript for this

audius-infra:
Preview this change https://demo.audius.co/sd-optimistic-support

audius-infra:
Preview this change https://demo.audius.co/sd-optimistic-support

audius-infra:
Preview this change https://demo.audius.co/sd-optimistic-support

audius-infra:
Preview this change https://demo.audius.co/sd-optimistic-support

audius-infra:
Preview this change https://demo.audius.co/sd-optimistic-support

audius-infra:
Preview this change https://demo.audius.co/sd-optimistic-support

PR #1393, opened 2022-06-01
Fix badges stacked up

sddioulde:
### Description

They were stacked instead of inline

### Dragons

n/a

### How Has This Been Tested?

locally against staging

n/a


audius-infra:
Preview this change https://demo.audius.co/sd-fix-badges

PR #1397, opened 2022-06-03
[PAY-281] Fetch current user tip amount if not already loaded when tipping

rickyrombo:
### Description

Ensures that the tip amount suggested to become top supporter takes into account the current tipped amount.

### Dragons

Decided to have it _not_ cache the result as the cache would serve very little utility, as sending repeat tips to the same artists warrants a refetch anyway.

### How Has This Been Tested?

Tested manually locally against staging accounts using stage DN1.

N/A


audius-infra:
Preview this change https://demo.audius.co/mjp-tip-more

PR #1404, opened 2022-06-06
Update scrollbar in UserListModal

rickyrombo:
### Description

Somehow this fixes PAY-218 even though that has to do with mutuals which uses a different modal entirely. ~~Not really sure how!~~ See below:

I added some console logs to both `simplebar/dist/simplebar-core.esm.js` and `simplebar-react-legacy/node_modules/simplebar/dist/simplebar.esm.js` and verified that _both_ of them were getting bound to the simplebar prior to this PR for the mutuals modal. After the change, only the former was bound. That was intriguing! It lined up with the usage of the legacy one being removed via the PR.

Looking into the source of the legacy one to see why it was being bound to the unrelated component, I noticed these lines in `simplebar-react-legacy/node_modules/simplebar/src/simplebar.js`:
```js
  static initHtmlApi() {
    this.initDOMLoadedElements = this.initDOMLoadedElements.bind(this);

    // MutationObserver is IE11+
    if (typeof MutationObserver !== 'undefined') {
      // Mutation observer to observe dynamically added elements
      this.globalObserver = new MutationObserver(mutations => {
        mutations.forEach(mutation => {
          Array.prototype.forEach.call(mutation.addedNodes, addedNode => {
            if (addedNode.nodeType === 1) {
              if (addedNode.hasAttribute('data-simplebar')) {
                !addedNode.SimpleBar &&
                  new SimpleBar(addedNode, SimpleBar.getElOptions(addedNode));
              } else {
                Array.prototype.forEach.call(
                  addedNode.querySelectorAll('[data-simplebar]'),
                  el => {
                    !el.SimpleBar &&
                      new SimpleBar(el, SimpleBar.getElOptions(el));
                  }
                );
              }
            }
          });

          Array.prototype.forEach.call(mutation.removedNodes, removedNode => {
            if (removedNode.nodeType === 1) {
              if (removedNode.hasAttribute('data-simplebar')) {
                removedNode.SimpleBar && removedNode.SimpleBar.unMount();
              } else {
                Array.prototype.forEach.call(
                  removedNode.querySelectorAll('[data-simplebar]'),
                  el => {
                    el.SimpleBar && el.SimpleBar.unMount();
                  }
                );
              }
            }
          });
        });
      });

      this.globalObserver.observe(document, { childList: true, subtree: true });
    }
```

SimpleBar _watches the whole DOM tree for mutations_! Notice the `document.querySelectorAll('[data-simplebar]')` - it's getting _all_ elements with that attribute. For existing SimpleBars created with the new version, the check to see if it's already initialized as a SimpleBar is `false` because the new version of SimpleBar doesn't track them with a `SimpleBar` property, but with a set of instances that it maintains.  Thus the legacy version reinitializes on that `SimpleBar`, causing a fight between the different values the two obtain. The legacy version fights for the offset being 15px to the right, the new version (rightfully) has a 0px offset. Whichever one modifies it second  is the one that sticks.

With my change, the old simplebar doesn't get mounted and thus the listeners don't fire. However, the FollowPage still has the legacy simplebar and might be susceptible to introducing this bug, though I imagine that it unmounts it after signup so it might not be an issue?

### Dragons

~~I'm not actually sure why this works.~~

~~I've bisected the change and it's definitely not the CSS, the mutuals modal definitely doesn't use this file, and keeping the import I removed doesn't make it repro either. I have no idea what fixes this for mutuals.~~

### How Has This Been Tested?

Tested TopSupporters, Supporting, Favorites, Reposts, and Mutuals user lists to make sure they look good and scroll fine.

N/A?


nicoback2 (approved):
Lgtm! Just want to confirm that you tested the infinite scroll and it still worked?

audius-infra:
Preview this change https://demo.audius.co/mjp-user-list-scrollbar

PR #1412, opened 2022-06-07
[PAY-224] Dismiss Popover when navigating away

rickyrombo:
### Description

Dismisses the artist hover tile when an interaction within it causes a modal to pop or the page to navigate.
- *Note:* The page navigation one is redundant, as it was being dismissed anyway due to its parent being unmounted. Added support anyway for consistency

### Dragons

- Required some particularly gritty debugging and also requires https://github.com/AudiusProject/stems/pull/136
  - Before the stems change, the hover tile would get unmounted before the click event would even fire since the modal (its parent) would dismiss via a listener to mousedown
  - Putting the modal `useOutsideClick` on the click event makes it part of the same event bubbling, thus the outside click happens _after_ the supporting tile is clicked, so the order of what happens if the hover tile is from a UserListModal looks like:
    - ArtistSupporting handles click, dispatches action making user list visible
    - UserListModal handles click from `useOutsideClick`, dispatches action making user list invisible
  - To remedy this, I had to put the ArtistSupporting dispatch inside a `setTimeout` (using a microtask via a promise didn't work for some reason? I don't understand why) so that the dispatch that makes modals visible happens _after_ all modals have already dismissed.

### How Has This Been Tested?

Tested locally on Chrome web against stage

N/A

### DIFF
--- a/packages/web/src/components/artist/ArtistPopover.tsx
+++ b/packages/web/src/components/artist/ArtistPopover.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, ReactNode } from 'react'
+import React, { useCallback, ReactNode, useState } from 'react'
 
 import Popover from 'antd/lib/popover'
 import cn from 'classnames'
@@ -47,6 +47,7 @@ export const ArtistPopover = ({
   mouseEnterDelay = 0.5,
   component: Component = 'div'
 }: ArtistPopoverProps) => {
+  const [isPopupVisible, setIsPopupVisible] = useState(false)
   const creator = useSelector((state: CommonState) =>
     getUser(state, { handle: handle.toLowerCase() })
   )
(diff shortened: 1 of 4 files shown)


packages/web/src/components/artist/ArtistPopover.tsx
```diff
@@ -74,7 +75,12 @@ export const ArtistPopover = ({
 
   const content =
     creator && userId !== creator.user_id ? (
-      <ArtistCard artist={creator} />
+      <ArtistCard
+        artist={creator}
+        onNavigateAway={() => {
```
dylanjeffers, line 80: Would we just want this to be onClick possibly? Like I assume whenever the user actually clicks on the popever, we want to close it?

dylanjeffers (approved):
Looks sweet just small callout re clicking

sddioulde (commented):
looks good to me, but wondering if you tested with the same user

like hovering on a track artist tile, then clicking on the artist supporting section, then closing the list modal, then re-hovering on that same artist.

i remember seeing a weirdness like that when i as trying to work on this prior.

packages/web/src/components/artist/ArtistPopover.tsx
rickyrombo: Hmm interesting product implications of that - if they just click on the cover photo or some other non-actionable spot, I assumed we wouldn't want to dismiss? It would for sure clean this up a ton though

sddioulde (approved):
tested it! nice

PR #1414, opened 2022-06-08
Kj import stems into monorepo

Kyle-Shanks:
### Description

* Imported stems into the monorepo using lerna
* Update the readme to include the link to stems

### Dragons

This is a huge change, but it appears to have just ported all of the code over. Not sure if other changes need to be made in order to support the additional package so pls let me know if that is the case.

### How Has This Been Tested?

Imported, npm installed, build, typechecked, and ran storybook and it loaded fine 👍 

N/A


nicoback2 (approved):
woooo! this LGTM but sebastian should prob take a look as well before merging

dylanjeffers (commented):
Just wondering, was this able to be merged in preserving the commit history? i believe @sliptype is familiar with how to do that if not

sliptype (approved):
This all looks good to me! Yeah @dylanjeffers we used `npx lerna import --preserve-commit  ../stems`

dylanjeffers:
So sweet nicee

PR #1428, opened 2022-06-09
Udpate profile support and mutuals tiles and and content transitions

sddioulde:
### Description

Udpate profile support and mutuals tiles and and content transitions based on:
- On Hover, let’s make all of these mutuals/supporters/supporting tiles grow by 1.01 (instead of 1.005)
- let’s change the active state on tiles from .995 to .99
- And let’s keep duration of .18s for the tile
- lets do .07s for the purple (both the text and the arrow icon)

### Dragons

n/a

### How Has This Been Tested?

local dapp against stage

n/a


audius-infra:
Preview this change https://demo.audius.co/sd-transition

rickyrombo (approved):

PR #1429, opened 2022-06-09
[PAY-302] Update tipping analytics for web and mobile

sddioulde:
### Description

- Separate web vs mobile tips analytics by adding a device param
- Include source param for profile vs feed
- Add analytic for dismissing the feed tip tile

### Dragons

n/a

### How Has This Been Tested?

local against stage

amplitude


piazzatron (approved):

piazzatron:
🔥

PR #1432, opened 2022-06-09
C-570 Update vertical tooltip styles to avoid arrow overlap

Kyle-Shanks:
### Description
* Update position styles for vertical tooltips to avoid overlap

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A


sliptype (approved):
Does this work for tooltips containing multiple lines of text? Or is that not a thing

Kyle-Shanks:
> Does this work for tooltips containing multiple lines of text? Or is that not a thing

I think the only one that we have is the profile completion one, but I assume it works for that

PR #1434, opened 2022-06-09
Use correct number for total followee follow count

sddioulde:
### Description

Use the followee follow count prop instead of number of users for profile mutuals.

### Dragons

n/a

### How Has This Been Tested?

local dapp against stage

n/a


piazzatron (approved):

audius-infra:
Preview this change https://demo.audius.co/sd-mutuals-count

PR #1437, opened 2022-06-09
[PAY-315] Fix mutuals profile pics loading

sliptype:
### Description

Wouldn't you know it, mutable state strikes again!

I noticed that even with the cache update action commented out the pictures were still loading. It turns out that the `user._profile_picture_sizes` was updating the user object and the cache update action wasn't causing the `getProfile` selector to run again.

Changing this to an immutable update causes the cache update to actually rerender the profile page. 

### Dragons

I was having a hard time reproing the initial issue, so I added some long delays on the picture fetching to recreate the issue. Would love if someone who can repro the initial issue could double check this fix!

### How Has This Been Tested?

Locally against prod

### DIFF
--- a/packages/web/src/common/store/cache/users/sagas.js
+++ b/packages/web/src/common/store/cache/users/sagas.js
@@ -253,7 +253,7 @@ function* watchFetchProfilePicture() {
     inProgress.add(key)
 
     try {
-      let user = yield select(getUser, { id: userId })
+      const user = yield select(getUser, { id: userId })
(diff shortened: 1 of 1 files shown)


audius-infra:
Preview this change https://demo.audius.co/sk-pay-315-mutuals-pics-fix

sddioulde (approved):
looks good thx. checking out this branch to attempt repro

rickyrombo (approved):

sddioulde:
cant repro fwiw. so we may be good

PR #1440, opened 2022-06-09
[PAY-323] Don't dismiss modal when clicking the follow button

rickyrombo:
### Description

After adding the event listener to the modal user list rows, clicking the follow button shouldn't propagate the click to the row.

### Dragons

### How Has This Been Tested?

Tested against stage (Mac Chrome)

### DIFF
--- a/packages/web/src/components/follow-button/FollowButton.js
+++ b/packages/web/src/components/follow-button/FollowButton.js
@@ -29,13 +29,22 @@ const FollowButton = props => {
     [styles.medium]: props.size === 'medium',
     [styles.small]: props.size === 'small'
   }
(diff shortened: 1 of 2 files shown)


raymondjacobson (approved):
Good guarded change. Just curious if you think there's ever a world in which we would not want to stop propagation? I think probably always stop..?

rickyrombo:
Was just thinking that - if you think it's safe, I'll change it to stop propagation on default instead of the prop

raymondjacobson:
> @raymondjacobson Was just thinking that - if you think it's safe, I'll change it to stop propagation on default instead of the prop

maybe this for hotfix as is. then soak the always on

PR #1443, opened 2022-06-10
[PAY-324] Add +N UI on mobile profile picture list component

sddioulde:
### Description

Add +N UI on mobile profile picture list component

_(the first two images do not represent the accurate number of profile pictures for notififications and profile top supporters that would trigger the limit for the +N UI, it's only for illustration purposes)_
Simulator Screen Shot - iPhone 13 - 2022-06-10 at 11 16 55
Simulator Screen Shot - iPhone 13 - 2022-06-10 at 11 17 02
Simulator Screen Shot - iPhone 13 - 2022-06-10 at 11 17 09

### Dragons

make sure that the correct number of profile pictures is displayed on different components that use the profile picture list

### How Has This Been Tested?

local mobile vs local dapp vs stage

n/a

### DIFF
--- a/packages/mobile/src/screens/notifications-screen/Notification/ProfilePictureList.tsx
+++ b/packages/mobile/src/screens/notifications-screen/Notification/ProfilePictureList.tsx
@@ -1,21 +1,51 @@
 import { User } from 'audius-client/src/common/models/User'
-import { StyleProp, View, ViewStyle } from 'react-native'
+import { formatCount } from 'audius-client/src/common/utils/formatUtil'
+import { StyleProp, View, ViewStyle, Text } from 'react-native'
 
 import { makeStyles } from 'app/styles'
 
 import { ProfilePicture } from './ProfilePicture'
 
-const useStyles = makeStyles(({ spacing }) => ({
-  root: {
-    flexDirection: 'row'
-  },
-  image: {
-    marginRight: spacing(-2)
-  }
-}))
+const USER_LENGTH_LIMIT = 9
+
+/**
+ * Not all profile picture lists have the same profile picture size.
+ * Some components pass in the dimensions (width and height) while others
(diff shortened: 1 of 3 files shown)


packages/mobile/src/screens/notifications-screen/Notification/ProfilePictureList.tsx
```diff
@@ -1,21 +1,51 @@
...
+const useStyles = makeStyles(
+  ({ spacing, palette, typography }, { imageDimensions }) => ({
+    root: {
+      flexDirection: 'row'
+    },
+    image: {
+      marginRight: spacing(-2)
+    },
+    imageExtraRoot: {
+      display: 'flex',
```
dylanjeffers, line 29: default is flex!

packages/mobile/src/screens/notifications-screen/Notification/ProfilePictureList.tsx
```diff
@@ -26,20 +56,49 @@ type ProfilePictureListProps = {
...
+  const {
+    users,
+    totalUserCount = users.length,
+    limit = USER_LENGTH_LIMIT,
+    style,
+    navigationType,
+    interactive,
+    imageStyles
+  } = props
+  const styles = useStyles({
```
dylanjeffers, line 68: def a small optimization, which is that every re-render will cause useStyles to get called, we can prevent that by useMemo'ing the useStyles params, not the cleanest, but best we have for now...

dylanjeffers (approved):
looks perfect!

packages/mobile/src/screens/notifications-screen/Notification/ProfilePictureList.tsx
sddioulde: cool thx

PR #1450, opened 2022-06-10
[C-576] Fix NotificationUsersScreen title, large list

dylanjeffers:
### Description

Fixes two issues with native follow notifications with 8+ followers
- Fixes incorrect title count by using `userIds` instead of `users` (which is limited to 8)
- Fixes user-list loading by renaming the tag correctly


raymondjacobson (approved):
BLESS

PR #1452, opened 2022-06-10
[PAY-325] Fix user profile picture list extra logic and consolidate profile tiles

sddioulde:
### Description

- fix logic for when to display the +N logic in web user profile picture list.
- consolidate profile top supporters and mutuals tiles
- extract profile nav title component and use in supporting list section

Screen Shot 2022-06-10 at 5 50 39 PM

### Dragons

n/a

### How Has This Been Tested?

local dapp against stage

n/a


packages/web/src/components/tipping/support/SupportingList.tsx
```diff
@@ -62,11 +63,10 @@ export const SupportingList = () => {
 
   return profile && rankedSupportingList.length > 0 ? (
     <div className={styles.container}>
-      <div className={styles.titleContainer}>
-        <IconTip className={styles.tipIcon} />
-        <span className={styles.titleText}>{messages.supporting}</span>
-        <span className={styles.line} />
-      </div>
+      <ProfileNavTitle
```
sddioulde, line 66: maybe different name for this since it may be used outside profile one day.
...also... it having the name ProfileNavTitle but being used in ProfilePictureListTile seems a bit weird. down for suggestions

audius-infra:
Preview this change https://demo.audius.co/sd-fix-user-profile-picture-list-extra-logic

packages/web/src/components/tipping/support/SupportingList.tsx
piazzatron: IMO it's fine to tie this with Profile for now – maybe something like `ProfilePageNavSectionTitle`?

piazzatron (approved):
LGTM aside from small comments - thx for consolidating these ❤️

PR #1454, opened 2022-06-13
Update web dependency on stems to use the local package

Kyle-Shanks:
### Description
same same but different but still same

### Dragons

n/a

### How Has This Been Tested?

n/a

n/a


sliptype (approved):

sliptype (approved):

sliptype (changes requested):
Wait why are there 33k additions lol

Kyle-Shanks:
> Wait why are there 33k additions lol

It added the entire dependency list for the stems repo, which is like 32,500 lines. (I think the same thing happen for the client dep in the mobile-client lock file)
the other changes were some sha1 -> sha512 changes

dylanjeffers (changes requested):
i actually think there is a better way to do this with `lerna bootstrap` happy to take this on cause it's a lil tricky with `npm build` and stuff

sliptype:
> i actually think there is a better way to do this with `lerna bootstrap` happy to take this on cause it's a lil tricky with `npm build` and stuff

We discussed this a little, one thing to call out is that we initially opted not to use lerna bootstrap because react native doesn't support symlinks. But I did try using bootstrap while keeping the web dependency in mobile local and things seemed to work

Also we will need to add lerna bootstrap to the build steps in CI

raymondjacobson:
Looks like we can close this PR b/c https://github.com/AudiusProject/audius-client/pull/1542 is in the works! @Kyle-Shanks

Kyle-Shanks:
Closed in favor of #1542

PR #1455, opened 2022-06-13
[C-582] Upgrade web to react v17

dylanjeffers:
### Description

Upgrades audius-client web react to v17. Removes unneeded references to React.

### Dragons

The tricky part with this upgrade is that we still need react v16 for audius-client native web view, since `react-nil` only works with v16. Since v17 has no feature changes this is safe since our code is compatible with both versions. The main tricky bit is adding `react16` dependency and using webpack import aliasing for native.


packages/web/src/components/action-drawer/ActionDrawer.tsx
```diff
@@ -1,4 +1,5 @@
-import React, { MouseEventHandler, ReactNode } from 'react'
+import { MouseEventHandler, ReactNode } from 'react'
+import * as React from 'react'
```
sliptype, line 2: Would be nice to change the `* as React`s to named imports too!

sliptype (approved):
Beautiful! interesting that using react 16 on native works without the react import in every file. But I guess that's because we are only importing things from common (no components)?

dylanjeffers:
> Beautiful! interesting that using react 16 on native works without the react import in every file. But I guess that's because we are only importing things from common (no components)?

Good questions, so native is also 17, so imports from common are covered. The part that is surprising is web view being v16 still working, but latest v16 can actually deal with no react import too. What was technically holding us back for client was ts below 4, but thought this a good chance to get stems client and native all on same react version

packages/web/src/components/action-drawer/ActionDrawer.tsx
dylanjeffers: yeah down to do this for sure ^^

PR #1460, opened 2022-06-13
[C-591] Upgrade stems react to v17

dylanjeffers:
### Description

Upgrades stems react to v17, removes extraneous react imports


sliptype (approved):
Awesome! Was gonna call out the react-jsx thing but it looks like you added it in the web branch too

PR #1462, opened 2022-06-14
Fix Modals/Popups closing on open

rickyrombo:
### Description

Ran into another Modal/Popup issue, again the modals were closing on open. I believe this is due to the handler of the click event setting the Modal to visible _before_ the click event bubbles to the document, so by the time the event bubbles to the document the listener is ready. Uses a `setTimeout` to ensure that the handler is only added _after_ any currently bubbling event is finished bubbling.

### Dragons

Modals/Popups are very widespread in usage.

### How Has This Been Tested?

Tested locally on Chrome Mac against stage

N/A

### DIFF
--- a/packages/stems/src/hooks/useClickOutside.ts
+++ b/packages/stems/src/hooks/useClickOutside.ts
@@ -35,7 +35,10 @@ export const useClickOutside = (
     }
 
     if (isVisible) {
-      document.addEventListener('click', handleClick)
+      // Don't attach the listener until all the current events are finished bubbling
+      setTimeout(() => {
+        document.addEventListener('click', handleClick)
+      }, 0)
     }
     return () => document.removeEventListener('click', handleClick)
   }, [onClick, ignoreClick, isVisible])


sliptype (approved):
Interesting!

We probably want to update the stems version and publish it too

packages/stems/src/hooks/useClickOutside.ts
```diff
@@ -35,7 +35,10 @@ export const useClickOutside = (
     }
 
     if (isVisible) {
-      document.addEventListener('click', handleClick)
+      // Don't attach the listener until all the current events are finished bubbling
+      setTimeout(() => {
```
piazzatron, line 39: Could use setImmediate?

piazzatron (approved):

packages/stems/src/hooks/useClickOutside.ts
rickyrombo: https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate doesn't seem standard or supported

PR #1470, opened 2022-06-15
[C-360] Flesh out the stories for the current stems components a bit

Kyle-Shanks:
### Description
Update the stories for the current stems components where needed.

### Dragons

N/a

### How Has This Been Tested?

Manually tested

N/a


sliptype (approved):
Sweet, thanks for doing this!

PR #1472, opened 2022-06-15
[PAY-339] [PAY-327][PAY-326] [PAY-274] Fix some QA issues

sddioulde:
### Description

- Fix jitter in artist hover tile

Screen Shot 2022-06-15 at 1 40 34 PM

^ reason why the height is 102px

- Do not force fetching users from source when fetching supporting for a user
- Remove force retrieve from source for tipping users
- Optimistically update sender supporting count and receiver supporter count if appropriate
- Add spacing in reward notif text
- Add delay on artist hover tile in list modal row

### Dragons

### How Has This Been Tested?

local dapp against stage:
- did not notice any jittering
- the unwanted modal refresh not happening anymore
- sending a tip updating the supporting count on the profile page

n/a

### DIFF
--- a/packages/web/src/store/tipping/sagas.ts
+++ b/packages/web/src/store/tipping/sagas.ts
@@ -332,7 +332,7 @@ function* refreshSupportAsync({
       )
     ]
 
-    yield call(fetchUsers, userIds, new Set(), true)
+    yield call(fetchUsers, userIds)
 
     const supportingForSenderMap: Record<string, Supporting> = {}
     supportingForSenderList.forEach(supporting => {
@@ -625,7 +625,7 @@ function* fetchRecentTipsAsync(action: ReturnType<typeof fetchRecentTips>) {
         ...tipToDisplay.followee_supporter_ids
       ])
     ]
-    yield call(fetchUsers, userIds, new Set(), true)
+    yield call(fetchUsers, userIds)
 
     /**
      * We need to get supporting data for logged in user and
--- a/packages/web/src/store/tipping/sagas.ts
+++ b/packages/web/src/store/tipping/sagas.ts
@@ -404,7 +404,7 @@ function* fetchSupportingForUserAsync({
     decodeHashId(supporting.receiver.id)
   )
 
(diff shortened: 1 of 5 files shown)


audius-infra:
Preview this change https://demo.audius.co/sd-qa

packages/web/src/store/tipping/sagas.ts
```diff
@@ -139,6 +158,23 @@ function* overrideSupportersForUser({
   const supportersMap = yield* select(getOptimisticSupporters)
   const supportersForReceiver = supportersMap[receiver.user_id] ?? {}
 
+  /**
+   * If sender was not previously supporting receiver, then
+   * optimistically increment the sender's supporting_count
```
rickyrombo, line 163: Not: optimistically update the _receiver's_ supporter count?

piazzatron, line 163: nit: supporter count

packages/web/src/store/tipping/sagas.ts
```diff
@@ -97,6 +99,23 @@ function* overrideSupportingForUser({
   const supportingMap = yield* select(getOptimisticSupporting)
   const supportingForSender = supportingMap[sender.user_id] ?? {}
 
+  /**
+   * If sender was not previously supporting receiver, then
```
piazzatron, line 103: nit: code seems to say, if receiver was not previously supporting sender?

piazzatron (approved):
Code looks good!!

rickyrombo (approved):

packages/web/src/store/tipping/sagas.ts
rickyrombo: @piazzatron I don't think so - `supportingForSender` is a list of who the sender supports, and it's checking if the receiver is there

piazzatron: u right

sddioulde: yup what marcus said

PR #1476, opened 2022-06-16
[C-558] Add FeatureFlagOverrideModal

Kyle-Shanks:
### Description
* Add FeatureFlagOverrideModal (Opens with 'F' in dev mode) 
* Update the useFlag hook to respect the overrides

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A

### DIFF
--- a/packages/web/src/common/hooks/useFeatureFlag.ts
+++ b/packages/web/src/common/hooks/useFeatureFlag.ts
@@ -12,25 +12,45 @@ import { getAccountUser } from 'common/store/account/selectors'
 import { isRemoteConfigLoaded } from 'common/store/remote-config/selectors'
 import { StateWithRemoteConfig } from 'common/store/remote-config/slice'
 
+export const FEATURE_FLAG_OVERRIDE_KEY = 'FeatureFlagOverride'
+
+export type OverrideSetting = 'enabled' | 'disabled' | null
+
 /**
  * Hooks into updates for a given feature flag.
  * Returns both `isLoaded` and `isEnabled` for more granular control
  * @param flag
  */
-export const createUseFeatureFlagHook = <State extends StateWithRemoteConfig>(
+export const createUseFeatureFlagHook = <State extends StateWithRemoteConfig>({
+  remoteConfigInstance,
+  getLocalStorageItem,
+  setLocalStorageItem
+}: {
   remoteConfigInstance: RemoteConfigInstance
-) => (flag: FeatureFlags) => {
+  getLocalStorageItem?: (key: string) => string | null
+  setLocalStorageItem?: (key: string, value: string | null) => void
+}) => (flag: FeatureFlags) => {
+  const overrideKey = `${FEATURE_FLAG_OVERRIDE_KEY}:${flag}`
   const configLoaded = useSelector((state: State) =>
     isRemoteConfigLoaded<State>(state)
   )
   const userIdFlag = flagCohortType[flag] === FeatureFlagCohortType.USER_ID
   const hasAccount = useSelector(getAccountUser)
   const shouldRecompute = userIdFlag ? hasAccount : true
+  const setOverride = (value: OverrideSetting) => {
+    setLocalStorageItem?.(overrideKey, value)
+  }
   const isEnabled = useMemo(
-    () => remoteConfigInstance.getFeatureEnabled(flag),
+    () => {
+      const override = getLocalStorageItem?.(overrideKey) as OverrideSetting
+      if (override === 'enabled') return true
+      if (override === 'disabled') return false
+
+      return remoteConfigInstance.getFeatureEnabled(flag)
+    },
     // We want configLoaded and shouldRecompute to trigger refreshes of the memo
     // eslint-disable-next-line
     [flag, configLoaded, shouldRecompute]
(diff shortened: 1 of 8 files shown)


packages/web/src/components/feature-flag-override-modal/FeatureFlagOverrideModal.tsx
```diff
@@ -0,0 +1,192 @@
...
+const setOverrideSetting = (flag: FeatureFlags, val: OverrideSetting) => {
+  const flagKey = `${FEATURE_FLAG_OVERRIDE_KEY}:${flag}`
+  if (val === null) return localStorage.removeItem(flagKey)
+  localStorage.setItem(flagKey, val)
+}
+
+export const FeatureFlagOverrideModal = () => {
+  const [isOpen, setIsOpen] = useModalState('FeatureFlagOverride')
+
+  const [challengeRewardsUiSetting, setChallengeRewardsUiSetting] = useState(
```
sliptype, line 39: Is there any way we can map over the the FeatureFlag enum to create the initial state & info map? Would be ideal not to have to change another file when adding/removing flags.

Something like this maybe:

```ts
const [flagOverrideState, setFlagOverrideState] = useState(
  mapValues(FeatureFlags, (flag) => getOverrideSetting(flag))
)
```

packages/web/src/components/feature-flag-override-modal/FeatureFlagOverrideModal.tsx
```diff
@@ -0,0 +1,192 @@
...
+      setter: setTippingSetting
+    }
+  }
+
+  const openModal = () => setIsOpen(true)
+  const closeModal = () => setIsOpen(false)
+
+  useEffect(() => {
+    window.addEventListener('keypress', (e: KeyboardEvent) => {
+      if (e.ctrlKey && e.key === 'f') isOpen ? closeModal() : openModal()
```
sliptype, line 150: Kinda worried about this being too accessible for the general public. I think we should do it like the discovery node selection popup does: https://github.com/AudiusProject/audius-client/blob/main/packages/web/src/components/discovery-node-selection/DiscoveryNodeSelection.tsx#L13

That requires the dev mode local storage item to be set for the hotkey to work

sliptype (approved):
Looks awesome! Had a couple ideas

packages/web/src/common/hooks/useFeatureFlag.ts
```diff
@@ -20,17 +24,28 @@ import { StateWithRemoteConfig } from 'common/store/remote-config/slice'
...
 ) => (flag: FeatureFlags) => {
+  const overrideKey = `${FEATURE_FLAG_OVERRIDE_KEY}:${flag}`
   const configLoaded = useSelector((state: State) =>
     isRemoteConfigLoaded<State>(state)
   )
   const userIdFlag = flagCohortType[flag] === FeatureFlagCohortType.USER_ID
   const hasAccount = useSelector(getAccountUser)
   const shouldRecompute = userIdFlag ? hasAccount : true
+  const setOverride = (val: OverrideSetting) => {
+    if (val === null) return localStorage.removeItem(overrideKey)
```
raymondjacobson, line 35: We don't want to depend directly on localstorage here in common right? Can we use the state context stuff? Or else this really shouldn't live in common.

packages/web/src/components/feature-flag-override-modal/FeatureFlagOverrideModal.tsx
raymondjacobson: +1

packages/web/src/components/feature-flag-override-modal/FeatureFlagOverrideModal.tsx
Kyle-Shanks: i was looking for a way to do this. I'll try this out and see if I can get it to work 👍

packages/web/src/common/hooks/useFeatureFlag.ts
sliptype: Oh good call. Yeah the `remoteConfigInstance` passed into here already has `getFeatureFlagSessionId` and `setFeatureFlagSessionId`. We could change those to simply be `getLocalStorageItem` and `setLocalStorageItem` and then expose/use those in this hook. @Kyle-Shanks lmk if you wanna sync on this so I can explain better

audius-infra:
Preview this change https://demo.audius.co/kj-Add-feature-flag-toggler

raymondjacobson (approved):
Screen Shot 2022-06-16 at 4 39 02 PM

Code LGTM! Default is a bit confusing, b/c it makes it a bit hard to tell off the bat what i do have enabled and what i don't, but im not sure if there's a better solution. Maybe automatically putting Enabled or Disabled based on what the default is and then a "Reset default" button at the bottom?

audius-infra:
Preview this change https://demo.audius.co/kj-Add-feature-flag-toggler

audius-infra:
Preview this change https://demo.audius.co/kj-Add-feature-flag-toggler

sliptype (approved):
Nice! Excited to use this 🥳

PR #1479, opened 2022-06-17
[PAY-351] Prevent re-toasting for the same new disbursement

raymondjacobson:
### Description

Was seeing a bug where I was repeatedly getting toasted for new disbursements.
https://www.loom.com/share/8b1b2036e0524452a217b832a082241e

I realize we don't dedupe this anyway. There may be a more canonical way to do this, lmk if you have thoughts. This appears to fix it, but would only toast once per session per challenge id. Maybe that's ok?

### Dragons

### How Has This Been Tested?

Locally vs. staging. Can share creds for this acct if useful.


audius-infra:
Preview this change https://demo.audius.co/rj-pay-351

PR #1483, opened 2022-06-21
[PAY-350][PAY-355] Mobile: Supporting Tile Redesign

rickyrombo:
### Description

https://www.figma.com/file/Q67LDHvSoc5KMLwU5esC24/%F0%9F%92%B8-Tipping?node-id=2628%3A36127

See image below

Note:
- Removes hardcoded `maxWidth` of the name contents, opting to use `flexShrink` instead to trigger ellipses
- Renames style references (eg. `backgroundImage` and `gradient` vs `supporterInfo` and `supporterInfoRoot`) mostly for my own benefit but can go back to old "root" convention if wanted
- Also fixes [PAY-355](https://linear.app/audius/issue/PAY-355/supporting-tiles-are-clipping-before-the-edges-of-the-device)

### Dragons
- Had to use negative margins on the rank text to achieve correct padding, as otherwise the container left space for ascenders and descenders in the font which aren't necessary for numbers (we only need cap height)
- Cheesed the size of the trophy icon to match Figma, since the Figma has the icon using a slightly smaller width/height than the container it's in and on mobile it seems it goes to the edges
- The badge icons seem larger than on Figma
- It looks terrible on my Android emulator, but that might be my emulator being junk (was having other problems too) - will try to run on device today
- Uses some janky/hacky negative margins to fix PAY-355

### How Has This Been Tested?

iPhone 13 iOS 15.0 Simulator
image

Google Pixel 4 API 29 Android Emulator
image

### DIFF
--- a/packages/mobile/src/screens/profile-screen/ProfileHeaderV2/SupportingTile.tsx
+++ b/packages/mobile/src/screens/profile-screen/ProfileHeaderV2/SupportingTile.tsx
@@ -57,9 +57,9 @@ const useStyles = makeStyles(({ spacing, palette }) => ({
     paddingVertical: spacing(0.5),
     paddingHorizontal: spacing(1.125),
     borderWidth: 1,
-    borderColor: palette.neutralLight8,
+    borderColor: palette.staticNeutralLight8,
     borderRadius: spacing(4),
-    backgroundColor: palette.white
+    backgroundColor: palette.staticWhite
   },
   rankNumberSymbol: {
     marginStart: spacing(1),
--- a/packages/mobile/src/screens/profile-screen/ProfileHeaderV2/SupportingTile.tsx
+++ b/packages/mobile/src/screens/profile-screen/ProfileHeaderV2/SupportingTile.tsx
@@ -8,7 +8,6 @@ import { profilePage } from 'audius-client/src/utils/route'
 import { ImageBackground, StyleProp, View, ViewStyle } from 'react-native'
 import LinearGradient from 'react-native-linear-gradient'
 
-import IconTip from 'app/assets/images/iconTip.svg'
 import IconTrophy from 'app/assets/images/iconTrophy.svg'
 import { Text, Tile } from 'app/components/core'
 import { ProfilePicture } from 'app/components/user'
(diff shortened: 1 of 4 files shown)


packages/mobile/src/screens/profile-screen/ProfileHeaderV2/SupportingTile.tsx
```diff
@@ -20,47 +19,59 @@ import { makeStyles } from 'app/styles'
...
-    borderTopRightRadius: 8,
+  backgroundImage: {
+    borderRadius: 8,
     overflow: 'hidden'
   },
-  supporterInfoRoot: {
+  gradient: {
+    flexDirection: 'column',
+    width: 220,
+    height: 88,
```
dylanjeffers, line 35: Column is default btw

packages/mobile/src/screens/profile-screen/ProfileHeaderV2/SupportingTile.tsx
```diff
@@ -20,47 +19,59 @@ import { makeStyles } from 'app/styles'
...
   rankText: {
-    textTransform: 'uppercase'
+    textTransform: 'uppercase',
+    marginTop: spacing(-1),
+    marginBottom: spacing(-1)
   },
   nameText: {
     color: palette.staticWhite,
-    maxWidth: spacing(32)
+    flexShrink: 1
```
dylanjeffers, line 74: Nice

packages/mobile/src/screens/profile-screen/ProfileHeaderV2/SupportingTile.tsx
```diff
@@ -97,48 +108,58 @@ export const SupportingTile = (props: SupportingTileProps) => {
...
-
   return user ? (
     <Tile style={[styles.root, style]} onPress={handlePress}>
       <ImageBackground
-        style={styles.supporterInfo}
+        style={styles.backgroundImage}
         source={{ uri: coverPhoto }}
       >
         <LinearGradient
-          colors={['#0000004D', '#0000001A']}
```
dylanjeffers, line 118: Would love to learn how you made these changes with linear gradient sometime

dylanjeffers (approved):
Excellent work, love that it was all pretty chill changes overall

packages/mobile/src/screens/profile-screen/ProfileHeaderV2/SupportingTile.tsx
rickyrombo: Just made it match the new figma:

```css
background: linear-gradient(180deg, rgba(0, 0, 0, 0.1) 50%, rgba(0, 0, 0, 0.3) 100%), url(.jpg);
```
Convert rgba to hex for each of those, and use the same angle. I believe the stops are defaulted how we want them, but if not I could add `locations=[0.5,1]`

piazzatron (approved):

PR #1485, opened 2022-06-21
[PAY-342][PAY-344] Fix user list modal issues

sddioulde:
### Description

- Closing the modal should not reset state and be in loading state before hiding the modal.
- Switching between user list modals should not lead to wrong content

https://user-images.githubusercontent.com/9600175/174846627-c271c3a3-b323-4bbb-8074-ad2350fa724d.mov

### Dragons

we now reset the list modal state on initial load rather than on exit, but should not really cause issues.

### How Has This Been Tested?

local dapp against stage. also check out the loom video

n/a

### DIFF
--- a/packages/web/src/components/user-list/UserList.tsx
+++ b/packages/web/src/components/user-list/UserList.tsx
@@ -1,4 +1,4 @@
-import { useEffect } from 'react'
+import { useEffect, useState } from 'react'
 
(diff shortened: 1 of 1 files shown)


audius-infra:
Preview this change https://demo.audius.co/sd-qa

piazzatron (approved):

PR #1487, opened 2022-06-21
Add feature flag helper for flag overrides

Kyle-Shanks:
### Description
Adding feature flag helper for the areas that cant use the hook to maintain the override logic

### Dragons

Should this file live somewhere else?
Should we build in the waitForUserRemoteConfig logic or keep that separate?
Should the components that are using this be updated to use the useFlag hook?

### How Has This Been Tested?

Manually tested

N/A


audius-infra:
Preview this change https://demo.audius.co/kj-Add-feature-flag-helper-for-overrides

PR #1491, opened 2022-06-22
[PAY-46] Fix ProgressBar color in matrix mode

dharit-tan:
### Description

ProgressBar style was arranged differently than default and dark
modes: page-header-gradient used hardcoded values instead of
page-header-gradient-color-1 and -2 vars. Changed it to match
the other themes.

ProgressBar is the only component that uses
page-header-gradient-color-1 and -2 directly so it was the only
one that displayed the bug. Other components use
page-header-gradient. 

### Dragons

None

### How Has This Been Tested?

Manual testing on local web client. Set localStorage theme to matrix.

Screen Shot 2022-06-21 at 5 35 45 PM

### DIFF
--- a/packages/web/src/utils/theme/matrix.ts
+++ b/packages/web/src/utils/theme/matrix.ts
@@ -33,10 +33,10 @@ const theme = {
   '--white': '#1F211F',
   '--darkmode-static-white': 'var(--white)',
 
+  '--page-header-gradient-color-1': '#4FF069',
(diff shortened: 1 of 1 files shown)


rickyrombo (approved):

audius-infra:
Preview this change https://demo.audius.co/rt-PAY-46

sddioulde (approved):

PR #1492, opened 2022-06-22
Close user list modals when a nested popover triggers a navigation

rickyrombo:
### Description

Pass through a handler to trigger the modal to close if a nested popover navigates.

It looks like this works in part because the handler that triggers the modal to reappear will trigger after the event is finished bubbling, so the hide happens first:
https://github.com/AudiusProject/audius-client/blob/9ab24da187ff7b54745e62fc61ac95cc526691af/packages/web/src/components/artist/ArtistSupporting.tsx#L95-L99

This was the piece I was wary of, it may have been missing the last time we attempted this route.

### Dragons

- Are there other modals that have popovers inside?

### How Has This Been Tested?

Tested locally against stage.

Supporting -> Supporting works

Please help me test vigorously!

### DIFF
--- a/packages/web/src/components/artist/ArtistChip.tsx
+++ b/packages/web/src/components/artist/ArtistChip.tsx
@@ -28,17 +28,24 @@ type ArtistIdentifierProps = {
   handle: string
   showPopover: boolean
   popoverMount?: MountPlacement
+  onNavigateAway?: () => void
 } & ComponentPropsWithoutRef<'div'>
 const ArtistIdentifier = ({
   userId,
   name,
   handle,
   showPopover,
-  popoverMount
+  popoverMount,
+  onNavigateAway
 }: ArtistIdentifierProps) => {
   return showPopover ? (
     <div>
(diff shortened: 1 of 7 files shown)


sliptype (approved):
Looks good, yeah I don't recall if we had that setTimeout before. But also could you double check artist popups from the notification panel, I remember that being an issue

raymondjacobson (approved):
Tests:
1. open reposts modal, close reposts modal
2. open reposts modal, click user name => goes to profile page
3. open followers modal, click row => goes to profile page
4. open followers modal, hover for popover, click cover photo in popover => no action
5. open followers modal, hover for popover, click name => opens profile page
6. open tipper modal, hover for popover, click on support => new modal opens
7. open followers modal, hover for popover, click follow => modal stays open as well as popover


LGTM!

rickyrombo:
> Looks good, yeah I don't recall if we had that setTimeout before. But also could you double check artist popups from the notification panel, I remember that being an issue

Ah yes, this was the doozy timesuck 😢

rickyrombo:
@sliptype / @raymondjacobson - merging, but definitely test on staging and we can follow up with more changes tomorrow

PR #1501, opened 2022-06-22
Pass entity through to milestone notif

raymondjacobson:
### Description

As far as I understand, entity was never part of the notification itself, which comes from the makeGetAllNotifications selector and we have a separate selector to pull the entity.

This may not be the full solution, but this unblocks local build! @isaacsolo  & I discovered this

### Dragons

### How Has This Been Tested?

Locally vs. staging w/ a listens milestone notification


dylanjeffers (approved):
Oh yeah totally nice catch. Improving the incorrect types for notifs soon as well to prevent other issues like this

raymondjacobson:
> Oh yeah totally nice catch. Improving the incorrect types for notifs soon as well to prevent other issues like this

💜

PR #1503, opened 2022-06-23
[PAY-353] Mobile: Update user list items

rickyrombo:
### Description

- Updates user list items to have equally sized metric fonts, icons
- Updates the profile image to be 72x72 (with a 1px border instead of 2px)
- Aligns the profile image to the top

### How Has This Been Tested?

iOS sim: iPhone 13 iOS 15.0

(See Linear for before pic)

After:

image
image


dylanjeffers (approved):
Thanks for pixel perfecting!

piazzatron (approved):

PR #1508, opened 2022-06-23
Upgrade CRA and webpack to v5

dylanjeffers:
### Description

Attempts to upgrade CRA and webpack to v5

### Dragons

Was able to mostly polyfil what we need, so the app doesn't have any strict import errors. The app itself doesn't load currently with an error in the console about express.js 

### How Has This Been Tested?


packages/web/package.json
```diff
@@ -5,10 +5,16 @@
   "author": "Audius",
   "version": "1.1.10",
   "private": true,
+  "browser": {
```
sliptype, line 8: I'm not 100% sure this will work. For the sdk rollup we used an "ignore" plugin for rollup that replaces given modules with an empty module. And with metro in react native we mocked modules with empty manually:

https://github.com/AudiusProject/audius-client/blob/main/packages/mobile/metro.config.js#L58

dylanjeffers: yeah i think youre right, it wasnt helping anyway, just throwing things at the wall! yeah i did end up just doing that for wepack

sliptype: I believe the browser field is only relevant when the package is a dependency of something else. So if we were importing `audius-client` somewhere and bundling that

sliptype:
Is this working now?

dylanjeffers:
> Is this working now?

yes, wrapping things up into a reviewable state today, thanks for patience

packages/web/craco.config.ts
```diff
@@ -1,65 +1,70 @@
...
-
-      addBeforeLoader(webpackConfig, loaderByName('file-loader'), wasmLoader)
-
-      return webpackConfig
     }
   },
-  // Disabling for now while we upgrade eslint and improve our config
   eslint: {
     enable: false
+  },
```
dylanjeffers, line 66: on by default, will remove this line

packages/web/package.json
```diff
@@ -33,6 +34,7 @@
     "dayjs": "1.10.7",
     "electron-log": "3.0.9",
     "electron-updater": "4.3.5",
```
dylanjeffers, line 36: dont need this, removing

packages/web/package.json
```diff
@@ -8,7 +8,8 @@
   "dependencies": {
     "@audius/sdk": "0.0.4",
     "@audius/stems": "1.1.12",
-    "@craco/craco": "6.4.3",
+    "@certusone/wormhole-sdk": "^0.4.4",
```
dylanjeffers, line 11: also removing

packages/web/package.json
```diff
@@ -99,8 +101,10 @@
     "simplebar-react": "2.1.0",
     "simplebar-react-legacy": "npm:simplebar-react@1.2.3",
     "tar": "6.1.11",
+    "tls": "0.0.1",
     "typed-redux-saga": "1.3.1",
     "typesafe-actions": "5.1.0",
+    "util": "^0.12.4",
```
dylanjeffers, line 107: can move to devDeps

packages/web/package.json
```diff
@@ -99,8 +101,10 @@
     "simplebar-react": "2.1.0",
     "simplebar-react-legacy": "npm:simplebar-react@1.2.3",
     "tar": "6.1.11",
+    "tls": "0.0.1",
```
dylanjeffers, line 104: can move to devDeps

packages/web/src/index.tsx
```diff
@@ -7,6 +7,7 @@ import './index.css'
 // Unsure why importing this component first would change that, but it appears to
 // when running in dev mode.
 import Root from './root'
+// import { Root } from './root2'
```
dylanjeffers, line 10: removing

packages/web/src/root2.tsx
```diff
@@ -0,0 +1,13 @@
+import { useEffect, useRef } from 'react'
```
dylanjeffers, line 1: removing

packages/web/package.json
```diff
@@ -113,7 +117,7 @@
     "configure-local-env": "node ./scripts/configureLocalEnv.js",
     "configure-local-env-cloud": "node ./scripts/configureLocalEnv.js --remote-host",
     "publish-scripts": "./scripts/publishScripts.sh",
-    "start": "craco start",
+    "start": "craco --max-old-space-size=8076 start",
```
raymondjacobson, line 120: did we see OOM on CI? or locally?

packages/web/package.json
```diff
@@ -216,15 +219,20 @@
     "@types/redux-saga": "^0.10.5",
     "@types/redux-sentry-middleware": "^0.2.1",
     "@types/resize-observer-browser": "^0.1.7",
-    "@types/webpack": "4.41.32",
+    "@types/webpack": "5.28.0",
     "@types/webpack-bundle-analyzer": "4.4.1",
     "@typescript-eslint/eslint-plugin": "5.3.0",
     "@typescript-eslint/parser": "5.3.0",
     "abort-controller": "3.0.0",
+    "assert": "^2.0.0",
```
raymondjacobson, line 227: some deps to pin here pls!

packages/web/src/audio/AudioStream.test.js
```diff
@@ -172,8 +172,8 @@ describe('stop', () => {
     audioStream.stop()
 
     expect(pause).toBeCalled()
-    setImmediate(() => {
+    setTimeout(() => {
```
raymondjacobson, line 175: do we lose support for setImmediate?

packages/web/src/root2.tsx
raymondjacobson, line 1: reminder to rm

raymondjacobson (approved):
code looks good.

one weird thing i'm noticing is that when running locally, i see the recaptcha thing show up while the page is loading:
Screen Shot 2022-07-05 at 4 36 56 PM

It should be always hidden though: https://github.com/AudiusProject/audius-client/blob/main/packages/web/public/index.html#L183

Very odd. Was trying to repro on a full build, but didn't get that working yet

audius-infra:
Preview this change https://demo.audius.co/cra-webpack-v5

audius-infra:
Preview this change https://demo.audius.co/cra-webpack-v5

audius-infra:
Preview this change https://demo.audius.co/cra-webpack-v5

PR #1514, opened 2022-06-24
[PAY-369][PAY-349][PAY-331][PAY-376][PAY-389][PAY-219] Fix more tipping QA issues

sddioulde:
### Description

Fix more tipping QA issues:

- fix mobile notification text wrapping
- add margin between go back button and chevron in confirm send tip screen
- vertically center send tip button text on mobile feed tip tile
- fix tipping conversion screen, error state, sending state, and converting state
- give profile bio links active state on mobile

### Dragons

n/a

### How Has This Been Tested?

- local dapp and local mobile app against stage.
- also tested eth<>spl audio conversion, which failed a few times and eventually succeeded when i kept retrying. the UI showed the expected states during sending, conversion, error, and success.

n/a

### DIFF
--- a/packages/mobile/src/components/feed-tip-tile/SendTipButton.tsx
+++ b/packages/mobile/src/components/feed-tip-tile/SendTipButton.tsx
@@ -19,6 +19,7 @@ const useStyles = makeStyles(({ spacing, palette, typography }) => ({
     marginTop: spacing(4)
   },
   sendTipButtonTitleContainer: {
+    paddingTop: spacing(0.5),
     display: 'flex',
     flexDirection: 'row',
     justifyContent: 'center',
--- a/packages/web/src/components/tipping/tip-audio/TipAudio.module.css
+++ b/packages/web/src/components/tipping/tip-audio/TipAudio.module.css
@@ -350,7 +350,7 @@
 .button:hover {
   background: var(--primary-light-1) !important;
 }
-.button:active {
+.button:not(.disabled):active {
   background: var(--primary-dark-1) !important;
   transform: perspective(1px) scale3d(0.99, 0.99, 0.99);
 }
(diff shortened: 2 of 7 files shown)


audius-infra:
Preview this change https://demo.audius.co/sd-qa-mobile

packages/web/src/components/tipping/tip-audio/TipAudio.module.css
```diff
@@ -259,6 +259,9 @@
 .confirmButton.disabled {
   width: 186px;
```
rickyrombo, line 260: Maybe instead of adding that new class we can just make this wider - looks like it's also supposed to be 251px per figma?

packages/mobile/src/components/feed-tip-tile/SendTipButton.tsx
```diff
@@ -19,6 +19,7 @@ const useStyles = makeStyles(({ spacing, palette, typography }) => ({
     marginTop: spacing(4)
   },
   sendTipButtonTitleContainer: {
+    paddingTop: spacing(0.5),
```
rickyrombo, line 22: It's hard to see while reviewing, but curious why this fixes the vertical centering? Looks like the flex should have already centered it - what was throwing it off?

rickyrombo (approved):

packages/mobile/src/components/feed-tip-tile/SendTipButton.tsx
sddioulde: good callout... i dont like this change... but flex didnt center the content perfectly. i added a background color on the icon and the text to see if they were the culprit but they didn't seem to be. somehow the centering is just a bit off (by about 2px)

packages/web/src/components/tipping/tip-audio/TipAudio.module.css
sddioulde: made it 252px as that's even and divisible by 4. and yes looks like the figma also update the basic confirm buttpn to be as wide and the error confirm button. good catch thx

audius-infra:
Preview this change https://demo.audius.co/sd-qa-mobile

PR #1525, opened 2022-06-27
Fix challenge confetti re-render

piazzatron:
### Description

Can be reviewed per commit to cut down on some of the type fixing noise.

- Set state.userChallenges to be a mapping of challengeId -> specifier -> user_challenge, and then adapt the selectors and calling code as necessary

### Dragons

None

### How Has This Been Tested?

Tested challenge completion locally. Also no longer see confetti spam

### Feature Flags ###

### DIFF
--- a/packages/web/src/common/store/pages/audio-rewards/selectors.ts
+++ b/packages/web/src/common/store/pages/audio-rewards/selectors.ts
@@ -1,3 +1,5 @@
+import { createSelector } from 'reselect'
+
 import { ChallengeRewardID } from 'common/models/AudioRewards'
 import { CommonState } from 'common/store'
 
@@ -7,8 +9,23 @@ export const getTrendingRewardsModalType = (state: CommonState) =>
 export const getChallengeRewardsModalType = (state: CommonState) =>
   state.pages.audioRewards.challengeRewardsModalType
 
-export const getUserChallenges = (state: CommonState) =>
+export const getUserChallengeSpecifierMap = (state: CommonState) =>
   state.pages.audioRewards.userChallenges
+
+// Returns just a single challenge per challengeId
+export const getUserChallenges = createSelector(
+  [getUserChallengeSpecifierMap],
+  challenges => {
+    return Object.values(challenges).reduce((acc, cur) => {
+      const challenge = Object.values(cur)[0]
+      if (!challenge) return acc // Shouldn't happen
+      return {
+        ...acc,
+        [challenge.challenge_id]: challenge
+      }
+    }, {})
+  }
+)
 
 export const getUndisbursedUserChallenges = (state: CommonState) =>
   state.pages.audioRewards.undisbursedChallenges.filter(challenge => {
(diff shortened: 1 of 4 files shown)


packages/web/src/common/store/pages/audio-rewards/selectors.ts
```diff
@@ -7,9 +9,24 @@ export const getTrendingRewardsModalType = (state: CommonState) =>
 export const getChallengeRewardsModalType = (state: CommonState) =>
   state.pages.audioRewards.challengeRewardsModalType
 
-export const getUserChallenges = (state: CommonState) =>
+export const getUserChallengeSpecifierMap = (state: CommonState) =>
   state.pages.audioRewards.userChallenges
 
+// Returns just a single challenge per challengeId
+export const getUserChallenges = createSelector(
```
rickyrombo, line 16: Should this even exist? Seems like it could lead to confusion or bugs in the future. If we're keeping it, what if it returns the _list_ of challenges for that ID and let the consumer decide how to prioritize? Also curious if we should be more sophisticated than taking the arbitrary first challenge?

packages/web/src/common/store/pages/audio-rewards/slice.ts
```diff
@@ -128,9 +133,13 @@ const slice = createSlice({
       }>
     ) => {
       const { challengeId, specifiers } = action.payload
-      const userChallenge = state.userChallenges[challengeId]
+
+      const userChallenge = Object.values(
+        state.userChallenges[challengeId] || {}
+      )[0]
```
rickyrombo, line 139: `userChallenge` can be undefined, right?

rickyrombo (approved):

packages/web/src/common/store/pages/audio-rewards/selectors.ts
piazzatron: Yeah... in my initial stab at this I rewrote this selector return multiple challenges per challenge_id, but it ends up being quite the sweeping code change to have to change the call site every single time we call this selector. Most of the time we call this wanting to know some property about the challenge itself (like the amount, the type, etc), which is the same for every instance of the challenge.

On top of all of this, it really feels like trending challenges are an edge case - it's the only challenge you can complete multiple times, since aggregate challenges get rolled up in the response from DN. Down to revisit this if the state modeling doesn't take us where we want it to go, but don't want to spend too much time for this edge case

packages/web/src/common/store/pages/audio-rewards/slice.ts
piazzatron: I'm surprirsed TS didn't catch that...

packages/web/src/common/store/pages/audio-rewards/selectors.ts
rickyrombo: Makes sense - maybe we can defensively remove specifier specific info from this selector at some point in the future instead and make it clear not to use this for those cases?

rickyrombo (changes requested):
fix tests please!

rickyrombo (approved):

PR #1527, opened 2022-06-28
[PAY-397] Add v2 user-list headers

dylanjeffers:
### Description

Adds v2 user-list headers which were missing for non-tipping user-lists


piazzatron (approved):
Sweet!!!

PR #1528, opened 2022-06-28
[PAY-398] Improve reaction perf when drawer closed

dylanjeffers:
### Description

Further improves reaction perf by pausing animation when drawer is closed


piazzatron (approved):
Great find

PR #1532, opened 2022-06-28
[C-618] Remove unnecessary calls to onClickArtistName in ArtistChip to avoid multiple route changes

Kyle-Shanks:
### Description

Removing other click handlers within ArtistChip as the call is handled by the parent div

### Dragons

Added saliou and piazza as assignees in case these were added to cover some specific case

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A


rickyrombo (approved):
Looks good to me!

audius-infra:
Preview this change https://demo.audius.co/kj-Fix-Artist-Chip-navigation-calls

raymondjacobson (approved):

PR #1533, opened 2022-06-28
[PAY-400] Fix twitter button repeat share

piazzatron:
### Description

This one requires some explanation!

**Problem**: The share to Twitter button doesn't let you click it twice

**Reason**:
The flow here was:
- Button starts in idle state, we haven't fetched social handles
- We click it, set button to loading, and wait for twitterHandle !== null before setting loading = succeeded

This doesn't work, because initially a user's social handles are undefined (unfetched), and in the saga layer once we fetch, if they are still empty, the saga sets them to null. But the Twitter button thinks that null means still fetching, so it never moves to success state.

**Solution**
- Loading state should be represented by `undefined` (unfetched), not null. Added some comments to make this more explicit.
- Also stop propagation on desktop, to prevent modal panel from closing after share
- Also refactor this into a hook to share behavior on desktop and mobile
- In the refactored hook, was able to drop two useEffect and use a regular conditional instead, making the flower easier to grok

### Dragons

None

### How Has This Been Tested?

Tested on web, testing on mobile

### Feature Flags ###

### DIFF
--- a/packages/web/src/components/notification/Notification/components/TwitterShareButton.tsx
+++ b/packages/web/src/components/notification/Notification/components/TwitterShareButton.tsx
@@ -1,8 +1,9 @@
-import { useCallback, useEffect, useState } from 'react'
+import { MouseEventHandler, useCallback } from 'react'
 
 import { useDispatch, useSelector } from 'react-redux'
 
 import { ReactComponent as IconTwitterBird } from 'assets/img/iconTwitterBird.svg'
+import { useTwitterButtonStatus } from 'common/hooks/useTwitterButtonStatus'
 import { CommonState } from 'common/store'
 import { fetchUserSocials } from 'common/store/cache/users/actions'
 import { getUser } from 'common/store/cache/users/selectors'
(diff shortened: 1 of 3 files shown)


packages/web/src/components/notification/Notification/components/TwitterShareButton.tsx
```diff
@@ -41,52 +40,51 @@ export const TwitterShareButton = (props: TwitterShareButtonProps) => {
...
-      const twitterData = other.shareData(handle)
-      if (twitterData) {
-        const { shareText, analytics } = twitterData
-        openTwitterLink(url, shareText)
-        // @ts-ignore issues with record type
-        record(analytics)
-        setShareTwitterStatus('idle')
+  const handleClick: MouseEventHandler<HTMLButtonElement> = useCallback(
+    e => {
+      e.stopPropagation()
```
rickyrombo, line 57: 🙌  nice

rickyrombo (approved):

audius-infra:
Preview this change https://demo.audius.co/piazz-fix-twitter-share

dylanjeffers (approved):
Looks solid, nice fixes and improvements

PR #1537, opened 2022-06-29
[C-410] Automate creation of release branch

sliptype:
### Description

All details outlined here: https://www.notion.so/audiusproject/Automated-Client-Release-Process-7fa3a660fccd4fc787dd16a2beee852d

### Dragons

### How Has This Been Tested?

### Feature Flags ###


.circleci/config.yml
```diff
@@ -32,8 +32,37 @@ jobs:
...
+    docker:
+      - image: circleci/node:14.18
+    steps:
+      - checkout
+      - run:
+          name: Bump version
+          command: |
+            git config --global user.email "audius-infra@audius.co"
+            git config --global user.name "audius-infra"
+            PREV_VERSION=$(grep -m1 version ./lerna.json | awk -F: '{ print $2 }' | sed 's/[", ]//g')
```
joaquincasares, line 47: My vote would be to use `jq` here, but this works!

If you can send me a sample lerna.json I can write the `jq` command. :)

joaquincasares (approved):

.circleci/config.yml
sliptype: Oh cool, I pulled this from an existing command but would love to make it more elegant! The `lerna.json` is super simple:

```
{
  "packages": [
    "packages/*"
  ],
  "version": "1.1.11"
}
```

joaquincasares: Ah perfect!

I would rewrite your line like this:

```
PREV_VERSION=$(jq -r .version ./lerna.json)
```

`jq .` shows the whole json, `jq .version` only shows the version key, `jq -r` strips the quotes using raw formatting, and the input can either be a filename or piped directly to `jq` via `| jq`.

I hope that helps!

audius-infra:
Preview this change https://demo.audius.co/sk-autogen-release

.circleci/config.yml
joaquincasares: Here are the related docs if you ever want to go heavy into `jq` (it takes TIME and you quickly forget things :sweat_smile: ) or ever want to do something similar for yaml using `yq`:

* https://stedolan.github.io/jq/manual/
* https://mikefarah.gitbook.io/yq/

They're definitely nice tools when trying to extract data. :nerd_face:

.circleci/src/jobs/@root-jobs.yml
```diff
@@ -0,0 +1,29 @@
...
+  docker:
+    - image: circleci/node:14.18
+  steps:
+    - checkout
+    - run:
+        name: npm Auth
+        command: echo "//registry.npmjs.org/:_authToken=$NPM_AUTH_TOKEN" > .npmrc
+    - run:
+        name: Publish packages
+        command: npx lerna publish from-git --yes
```
raymondjacobson, line 13: any way we can depend on not npx to do this? like from node modules or something like that?

raymondjacobson (approved):

.circleci/config.yml
dmanjunath: sorry don't mean to butt in but aren't jq and yq not included by default? also up to @sliptype but i feel like this might just be a bit over optimization to try to install and use it for such a small syntactic gain 🙂

.circleci/src/jobs/@root-jobs.yml
sliptype: Oh yeah there should be an `npm i` step first so we aren't making a network call with npx. But I think it's still ok to use npx once the packages are installed bc it will just use the local one

.circleci/config.yml
joaquincasares: I'm not sure if `jq` is included by default, but it *should* be since it's a goto like `grep` or `awk` when messing with json files. If not, `jq` is [easily installable](https://stedolan.github.io/jq/download/).

We also don't need `yq`. Sebastian sounded excited about a more elegant solution, so I was just sharing `yq` for his own future reference.

Indeed, as I mentioned in my original post, the current solution works and usage of `jq` is optional (even if it is the ideal choice when interacting with json files).

sliptype: I'll see if it's included in the image we are running! Installing it seems like overkill in this situation

joaquincasares: +1 on both counts! :)

.circleci/src/jobs/@root-jobs.yml
raymondjacobson: cool, sgtm!

audius-infra:
Preview this change https://demo.audius.co/sk-autogen-release

PR #1538, opened 2022-06-30
[PAY-304] Adjust feed tip tile for small screens

sddioulde:
### Description

Show all text in one line.
If screen width is small enough, then do not include the name in the send tip button.

https://user-images.githubusercontent.com/9600175/177311967-1eefe645-8e0c-4aa0-a2fa-48e3bf2d03dd.mp4

### Dragons

### How Has This Been Tested?

Local dapp against stage

n/a

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/sd-small-screens

packages/web/src/components/tipping/feed-tip-tile/FeedTipTile.tsx
```diff
@@ -188,19 +188,8 @@ const MIN_USERS_TO_BUTTONS_MARGIN = 8
...
-    setContainerPosition(node.getBoundingClientRect())
-  }, [])
-
-  const [buttonsPosition, setButtonsPosition] = useState<Nullable<DOMRect>>(
-    null
-  )
-  const buttonsCallbackRef = useCallback((node: HTMLDivElement) => {
-    setButtonsPosition(node.getBoundingClientRect())
-  }, [])
+  const { position: containerPosition, ref: containerRef } = usePosition()
```
sddioulde, line 191: doubt: can we call the same hook, as I'm doing here, with no arguments in the hook, and return diff objects? when i tested on the client, it works, meaning I did not get the same object twice. but wondering if it's safe... could we possibly get the same object twice here in these 2 `usePosition()` hook calls?

audius-infra:
Preview this change https://demo.audius.co/sd-small-screens

packages/web/src/components/tipping/feed-tip-tile/FeedTipTile.tsx
```diff
@@ -170,9 +177,30 @@ const DismissTipButton = () => {
...
+ * When this is smaller than the margin we want to allow, we use the short
+ * version of the button which does not include the name.
+ */
+const MIN_USERS_TO_BUTTONS_MARGIN = 8
+
 export const FeedTipTile = () => {
   const isTippingEnabled = getFeatureEnabled(FeatureFlags.TIPPING_ENABLED)
 
+  const { position: containerPosition, ref: containerRef } = usePosition()
+  const { position: buttonsPosition, ref: buttonsRef } = usePosition()
```
raymondjacobson, line 192: Hello! drive by comment. we do this elsewhere using media queries from JS. I think more performant and potentially cleaner:

https://github.com/AudiusProject/audius-client/blob/5492cb20da8dee8f3aa0d70fa8c37dca062010a0/packages/stems/src/components/Button/hooks.ts#L22

(see behavior when collapsing UI on the track page)

sddioulde: interesting. thanks will look

audius-infra:
Preview this change https://demo.audius.co/sd-small-screens

packages/stems/src/hooks/useMediaQueryListener.ts
```diff
@@ -0,0 +1,27 @@
+import { useEffect, useState, useCallback } from 'react'
+
+export const useMediaQueryListener = (mediaQuery: string) => {
+  // Stores whether there is a match
+  const [isMatch, setIsMatch] = useState(false)
+  console.log('here in med que lis', isMatch)
```
piazzatron, line 6: log

packages/web/src/utils/constants.ts
```diff
@@ -12,3 +12,9 @@ export const MAX_ARTIST_HOVER_TOP_SUPPORTING = 7
 export const NUM_FEED_TIPPERS_DISPLAYED = 2
 export const RECENT_TIPS_KEY = 'recent-tips'
 export const FEED_TIP_DISMISSAL_TIME_LIMIT = 30 * 24 * 60 * 60 * 1000 // 30 days
+
+/**
+ * When the screen is smaller than this width, we use the short
+ * version of the button which does not include the name.
+ */
+export const MAX_WIDTH_FOR_SHORT_TIP_BUTTON = 884
```
piazzatron, line 23: IMO probably fine to have this colocated with `FeedTipTile.tsx` since it's a pretty specific UI concern for that class

piazzatron (approved):
Looks great!

PR #1541, opened 2022-07-01
[PAY-383] Created Web UI For First Tip Challenge

dharit-tan:
### Description
Created web UI for first tip challenge including tile on /audio and
modal. Action button links to /explore/heavy-rotation.

### Dragons
Using a one-off component to create an image out of IconGoldBadge in config.tsx, which should probably be refactored as it's the same code as in TipAudioModal. Ideally all the badges should be exported from a common file. Happy to do this but I think it merits its own ticket.

### How Has This Been Tested?

Local web client.

Screen Shot 2022-06-30 at 10 15 37 PM
Screen Shot 2022-06-30 at 10 15 32 PM


audius-infra:
Preview this change https://demo.audius.co/rt-first-tip-ui-web

rickyrombo (approved):

piazzatron (approved):
Looks great! Def supportive of creating a linear issue for exporting the badge components from a central place

PR #1542, opened 2022-07-01
[C-649] Use lerna bootstrap

dylanjeffers:
### Description

Replaces local npm file references with lerna bootstrap
- Updates docs
- Updates ci scripts
- simplifies init scripts, simply run `npm i` and our postinstall handles the rest

> Note the huge diff in mobile package-lock is due to not needing audius-client in it's package-lock anymore


raymondjacobson (commented):
Code looks great, but I believe CI is breaking because the `dist` build from the init step for stems isn't being persisted to the web job (i ssh'd into the web job and ls-d into node_modules/@audius/stems and there was no dist)

dylanjeffers:
> Code looks great, but I believe CI is breaking because the `dist` build from the init step for stems isn't being persisted to the web job (i ssh'd into the web job and ls-d into node_modules/@audius/stems and there was no dist)

Yeah that's a good call, thinking we maybe add the install and build step in the setup script, and then copy over node mods too, not sure... but anyway can add the stems dist at least

dylanjeffers:
> > Code looks great, but I believe CI is breaking because the `dist` build from the init step for stems isn't being persisted to the web job (i ssh'd into the web job and ls-d into node_modules/@audius/stems and there was no dist)
> 
> Yeah that's a good call, thinking we maybe add the install and build step in the setup script, and then copy over node mods too, not sure... but anyway can add the stems dist, or just rebuild it

sliptype (approved):
Looks great!

Does this work on local mobile dev? Bc react-native doesn't support symlinks which is why we weren't using `lerna bootstrap` initially

audius-infra:
Preview this change https://demo.audius.co/dj-c-649-lerna-bootstrap

.circleci/src/jobs/@web-jobs.yml
```diff
@@ -49,6 +43,8 @@ web-init:
         root: ./
         paths:
           - packages/web/node_modules
+          - packages/stems/node_modules
+          - packages/stems/dist
```
raymondjacobson, line 47: SICK

audius-infra:
Preview this change https://demo.audius.co/dj-c-649-lerna-bootstrap

PR #1566, opened 2022-07-09
Fix jitter

sddioulde:
### Description

The non-empty supporting container must've changed height. So the empty container is updated to match the height and prevent the jitter.

### Dragons

n/a

### How Has This Been Tested?

local dapp vs stage

n/a

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/sd-fix-jitter

raymondjacobson (approved):

PR #1569, opened 2022-07-11
Add ssh key for audius-infra write access

sliptype:
### Description

Auto version bump failed on `main` because it is a protected branch and CircleCI wasn't using an ssh key for the audius-infra account. This adds the ssh key and prevents the post install script from running because we only need the root level dependencies in this ci run

### Dragons

### How Has This Been Tested?

Tested on a custom protected branch with the same settings as `main`

Release branch should be properly generated on friday

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/sk-circleci-ssh

PR #1570, opened 2022-07-11
[C-409] Add eslint-config-audius, use in packages

dylanjeffers:
### Description

- Adds eslint-config-audius, migrating the common rules across all our projects into one place
- Installs this config in each project
- Moves all eslint packages to root package.json to save space, and also ensure versions are the same.
- Adds lint task at root to easily check if any package has issue with existing config.

### Dragons

- It's a big architectural change, but fairly safe given we are getting the same lint output
- One will likely need to restart vscode (or reset the linting plugin) for it to correctly resolve deps and new eslint configs

### How Has This Been Tested?

npm run lint from the root passes.


packages/eslint-config-audius/src/index.js
```diff
@@ -0,0 +1,107 @@
...
+  parserOptions: {
+    ecmaFeatures: {
+      jsx: true,
+    },
+    ecmaVersion: 2018,
+    sourceType: "module",
+  },
+  plugins: ["react", "react-hooks", "@typescript-eslint", "jest", "import"],
+  rules: {
+    "@typescript-eslint/explicit-function-return-type": "off",
```
sliptype, line 29: Now we can start enabling these rules one by one!

packages/eslint-config-audius/src/index.js
```diff
@@ -0,0 +1,107 @@
...
+  },
+  parser: "@typescript-eslint/parser",
+  parserOptions: {
+    ecmaFeatures: {
+      jsx: true,
+    },
+    ecmaVersion: 2018,
+    sourceType: "module",
+  },
+  plugins: ["react", "react-hooks", "@typescript-eslint", "jest", "import"],
```
sliptype, line 27: If we wanna apply this same config to our services too, it might make sense so split out the react stuff into `audius-react` or something. But we can do that in a future pr!

sliptype (approved):
Yesssss, thank you for tackling this. This is gonna make life so much better

audius-infra:
Preview this change https://demo.audius.co/dj-c-409-add-eslint-config-audius

packages/eslint-config-audius/src/index.js
dylanjeffers: yes 1000%

PR #1573, opened 2022-07-11
[PAY-399] Fix reaction notification selection bug

dylanjeffers:
### Description

Fixes issue where quickly swiping on notification reactions results in multiple reactions being selected at the same time. This also caused the entire notification drawer to freeze since reaction gesture handler was blocking all other gestures, yikes.

### DIFF
--- a/packages/mobile/src/screens/notifications-screen/Reaction/ReactionList.tsx
+++ b/packages/mobile/src/screens/notifications-screen/Reaction/ReactionList.tsx
@@ -7,9 +7,19 @@ import {
 import { Nullable } from 'audius-client/src/common/utils/typeUtils'
 import { View, PanResponderGestureState, PanResponder } from 'react-native'
 
+import { makeStyles } from 'app/styles'
+
 import { NotificationsDrawerNavigationContext } from '../NotificationsDrawerNavigationContext'
 
 import { reactionMap } from './reactions'
+
+const useStyles = makeStyles(() => ({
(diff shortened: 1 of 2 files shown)


packages/mobile/src/screens/notifications-screen/Reaction/ReactionList.tsx
```diff
@@ -60,37 +73,37 @@ export const ReactionList = (props: ReactionListProps) => {
...
         handleGesture(e, gestureState)
       },
-      onPanResponderMove: handleGesture,
-      onPanResponderRelease: () => {
-        onChange(interactingRef.current)
-        interactingRef.current = null
-        setInteracting(null)
-        setGesturesDisabled?.(false)
+      onPanResponderMove: (...args) => {
+        handleGesture(...args)
```
piazzatron, line 90: just curious, howcome we're spreading these args now?

piazzatron (approved):
LGTM! Only note is that this gesture system is p complex and could probably use a bit of top level documentation explaining the purpose of the various callbacks and state we have to track, etc

packages/mobile/src/screens/notifications-screen/Reaction/ReactionList.tsx
dylanjeffers: oh nice catch, dont need too, just an artifact of calling a console.log in there

PR #1576, opened 2022-07-12
[CON-234] Add write quorum feature flag

theoilie:
### Description
Add write quorum feature flag: `WRITE_QUORUM_ENABLED`. This flag can be set to true to tell libs/SDK to pass the `Enforce-Write-Quorum` header. If Content Nodes see this header set to true/false when data is written (track/playlist upload or user metadata change), then they'll wait for the metadata to sync to a secondary node before returning a 200 response.

### Dragons
The flag defaults to false, so negative impact should be minimal. Ideally https://github.com/AudiusProject/audius-protocol/pull/3408 will be released before this PR so the client doesn't make calls to libs with an extra param that the older version of libs doesn't yet recognize.

### How Has This Been Tested? 
Use libs changes from https://github.com/AudiusProject/audius-protocol/pull/3408 (link local libs if the PR isn't merged). Search for `Enforce-Write-Quorum` header in Network tab for metadata endpoints when uploading a track or changing user profile.

Screenshot of /tracks/metadata passing the header when uploading a track:
Screen Shot 2022-07-11 at 11 42 24 PM

Screenshot of /audius_users/metadata passing the header when updating profile pic:
Screen Shot 2022-07-11 at 11 44 05 PM

Any increase in errors uploading should make us suspect write quorum. Search logs for "Failed to reach 2/3 write quorum" and turn off the ENABLE_WRITE_QUORUM optimizely flag if necessary.

### Feature Flags ###

Passes writeQuorumEnabled boolean from new WRITE_QUORUM_ENABLED feature flag to the libs usages that support write quorum:

- User.updateCreator
- User.upgradeToCreator
- Track.uploadTrack
- File.uploadImage
- Libs sanityCheck (because it updates a user's replica set, which uses write quorum to update the user's metadata)

See https://github.com/AudiusProject/audius-protocol/pull/3408 for libs/SDK functions that accept the `writeQuorumEnabled` flag.

### DIFF
--- a/packages/web/README.md
+++ b/packages/web/README.md
@@ -39,9 +39,9 @@ npm run start:<environment>
 npm run build:<environment>
 ```
 
-### Using local linked version of @audius/libs
+### Using local linked version of @audius/libs (@audius/sdk)
 
-To run audius-client/web with a local version of @audius/libs run:
+To run audius-client/web with a local version of @audius/sdk run:
 
 - In a separate terminal for `audius-protocol/libs`
 ```bash
@@ -49,11 +49,11 @@ npm link
 npm run dev
 ```
 
-This will link @audius/libs, and also run it it dev mode, which will transpiles changes and triggers a hot-reload in web.
+This will link @audius/sdk, and also run it it dev mode, which will transpiles changes and triggers a hot-reload in web.
 
-- In terminal for `audius-client/web`
+- In terminal for `audius-client/packages/web`
 ```bash
-npm link @audius/libs
+npm link @audius/sdk
 ```
 
 Then start the application using the start/build commands above.
--- a/packages/web/src/services/AudiusBackend.js
+++ b/packages/web/src/services/AudiusBackend.js
@@ -285,14 +285,6 @@ const fetchImageCID = async (cid, creatorNodeGateways = [], cache = true) => {
 class AudiusBackend {
   static currentDiscoveryProvider = null
   static didSelectDiscoveryProviderListeners = []
-  static writeQuorumEnabled = flagDefaults[FeatureFlags.WRITE_QUORUM_ENABLED]
-
-  // eslint-disable-next-line
-  static {
-    this.writeQuorumEnabled = getFeatureEnabled(
-      FeatureFlags.WRITE_QUORUM_ENABLED
-    )
-  }
 
   static addDiscoveryProviderSelectionListener(listener) {
     AudiusBackend.didSelectDiscoveryProviderListeners.push(listener)
@@ -406,7 +398,10 @@ class AudiusBackend {
         skipRollover: getRemoteVar(BooleanKeys.SKIP_ROLLOVER_NODES_SANITY_CHECK)
       }
       const sanityChecks = new SanityChecks(audiusLibs, sanityCheckOptions)
-      await sanityChecks.run(null, this.writeQuorumEnabled)
+      const writeQuorumEnabled = getFeatureEnabled(
+        FeatureFlags.WRITE_QUORUM_ENABLED
+      )
+      await sanityChecks.run(null, writeQuorumEnabled)
     } catch (e) {
       console.error(`Sanity checks failed: ${e}`)
     }
@@ -1011,22 +1006,28 @@ class AudiusBackend {
    * @param {string} newCreatorNodeEndpoint will follow the structure 'cn1,cn2,cn3'
    */
   static async upgradeToCreator(newCreatorNodeEndpoint) {
+    const writeQuorumEnabled = getFeatureEnabled(
+      FeatureFlags.WRITE_QUORUM_ENABLED
+    )
     return audiusLibs.User.upgradeToCreator(
       USER_NODE,
       newCreatorNodeEndpoint,
-      this.writeQuorumEnabled
+      writeQuorumEnabled
     )
   }
 
   // Uploads a single track
   // Returns { trackId, error, phase }
   static async uploadTrack(trackFile, coverArtFile, metadata, onProgress) {
+    const writeQuorumEnabled = getFeatureEnabled(
+      FeatureFlags.WRITE_QUORUM_ENABLED
+    )
     return await audiusLibs.Track.uploadTrack(
       trackFile,
       coverArtFile,
       metadata,
       onProgress,
-      this.writeQuorumEnabled
+      writeQuorumEnabled
     )
   }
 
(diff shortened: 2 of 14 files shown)


theoilie:
@raymondjacobson I started by trying to pass the WRITE_QUORUM_ENABLED flag from upstream through the sagas but realized it might be ok to just pass it to libs directly in AudiusBackend. so this isn't ready for review yet and has a mix both ways – could you please take a glance and lmk if it's fine to just do everything directly in AudiusBackend so I can remove passing the flag around in the sagas?

raymondjacobson:
> 

yes 100% for doing directly in audius backend!

you can also pass it into the libs ctor in AudiusBackend in 1 go and then plumb it through on the libs side with a `if (this.writeQuorumEnabled)` kinda thing

audius-infra:
Preview this change https://demo.audius.co/theo-write-quorum

audius-infra:
Preview this change https://demo.audius.co/theo-write-quorum

theoilie:
> > 
> 
> yes 100% for doing directly in audius backend!
> 
> you can also pass it into the libs ctor in AudiusBackend in 1 go and then plumb it through on the libs side with a `if (this.writeQuorumEnabled)` kinda thing

awesome I moved everything to just use AudiusBackend. one note about using the constructor -- it seems to short-circuit and return `null` because `client` isn't ready yet as per this line. so now it sends the actual boolean when not using the static init block, but this is what I was doing before that didn't work (first debug prints false, second prints null):

```js
static writeQuorumEnabled = flagDefaults[FeatureFlags.WRITE_QUORUM_ENABLED]

// eslint-disable-next-line
static {
  console.log(`theo client writeQuorumEnabled before static init: ${AudiusBackend.writeQuorumEnabled}`)
  AudiusBackend.writeQuorumEnabled = getFeatureEnabled(
    FeatureFlags.WRITE_QUORUM_ENABLED
  )
  console.log(`theo client writeQuorumEnabled after static init: ${AudiusBackend.writeQuorumEnabled}`)
}
```

theoilie:
@raymondjacobson I requested your review since you have context on write quorum but please feel free to reassign if someone else is a more appropriate reviewer :)

packages/web/README.md
```diff
@@ -39,21 +39,21 @@ npm run start:<environment>
...
 ```
 
-This will link @audius/libs, and also run it it dev mode, which will transpiles changes and triggers a hot-reload in web.
+This will link @audius/sdk, and also run it it dev mode, which will transpiles changes and triggers a hot-reload in web.
 
-- In terminal for `audius-client/web`
+- In terminal for `audius-client/packages/web`
 ```bash
-npm link @audius/libs
+npm link @audius/sdk
```
raymondjacobson, line 56: ty!

packages/web/src/services/AudiusBackend.js
```diff
@@ -396,7 +397,10 @@ class AudiusBackend {
         skipRollover: getRemoteVar(BooleanKeys.SKIP_ROLLOVER_NODES_SANITY_CHECK)
       }
       const sanityChecks = new SanityChecks(audiusLibs, sanityCheckOptions)
-      await sanityChecks.run()
+      const writeQuorumEnabled = getFeatureEnabled(
+        FeatureFlags.WRITE_QUORUM_ENABLED
+      )
+      await sanityChecks.run(null, writeQuorumEnabled)
```
raymondjacobson, line 403: good eye on this

raymondjacobson: hopefully one day we can delete all this "sanity check" code

packages/web/src/services/AudiusBackend.js
```diff
@@ -1047,7 +1066,10 @@ class AudiusBackend {
...
 
   static async uploadImage(file) {
-    return audiusLibs.File.uploadImage(file)
+    const writeQuorumEnabled = getFeatureEnabled(
+      FeatureFlags.WRITE_QUORUM_ENABLED
+    )
+    return audiusLibs.File.uploadImage(file, writeQuorumEnabled)
   }
 
   static async updateTrack(trackId, metadata) {
```
raymondjacobson, line 1075: what about updateTrack / updateUser / etc.?

raymondjacobson (approved):
LGTM minus i think maybe missing a couple spots?

raymondjacobson:
> > > 
> > 
> > 
> > yes 100% for doing directly in audius backend!
> > you can also pass it into the libs ctor in AudiusBackend in 1 go and then plumb it through on the libs side with a `if (this.writeQuorumEnabled)` kinda thing
> 
> awesome I moved everything to just use AudiusBackend. one note about using the constructor -- it seems to short-circuit and return `null` because `client` isn't ready yet as per this line. so now it sends the actual boolean when not using the static init block, but this is what I was doing before that didn't work (first debug prints false, second prints null):
> 
> ```js
> static writeQuorumEnabled = flagDefaults[FeatureFlags.WRITE_QUORUM_ENABLED]
> 
> // eslint-disable-next-line
> static {
>   console.log(`theo client writeQuorumEnabled before static init: ${AudiusBackend.writeQuorumEnabled}`)
>   AudiusBackend.writeQuorumEnabled = getFeatureEnabled(
>     FeatureFlags.WRITE_QUORUM_ENABLED
>   )
>   console.log(`theo client writeQuorumEnabled after static init: ${AudiusBackend.writeQuorumEnabled}`)
> }
> ```

Possible that we want to call `waitForRemoteConfig`?  I was actually thinking of using the libs constructor and doing things there. I have a slight preference for that b/c it will keep the logic confined a bit to only the "sdk" package.

`new AudiusLibs(configCreatorNode: { writeQuorumEnabled })`

and then inside the libs File.x, etc. we can just use that value

audius-infra:
Preview this change https://demo.audius.co/theo-write-quorum

raymondjacobson:
See this similar flag: https://github.com/AudiusProject/audius-client/blob/main/packages/web/src/services/AudiusBackend.js#L499

theoilie:
ah gotcha, thanks @raymondjacobson ! will look at changing that tomorrow. gotta get the CN+libs PR through first anyway

dmanjunath:
@theoilie agreed with @raymondjacobson that we should set it in the constructor in sdk. another suggestion is could we just pass the header in the sdk to all content node requests and just let cn handle the awaits or whether it should call on that side? that way we don't couple client/sdk with quorum logic that only needs to live in content node

theoilie:
> @theoilie agreed with @raymondjacobson that we should set it in the constructor in sdk. another suggestion is could we just pass the header in the sdk to all content node requests and just let cn handle the awaits or whether it should call on that side? that way we don't couple client/sdk with quorum logic that only needs to live in content node

makes sense yeah that sounds like a good way to decouple. I just made that change in the libs+CN PR -- https://github.com/AudiusProject/audius-protocol/pull/3408. also updated this PR to pass in via the libs constructor. much simpler now 😃

dmanjunath (approved):
Awesome, so much simpler! Thanks for making these changes

PR #1578, opened 2022-07-12
[C-674] Upgrade eslint, prettier, and plugins to latest

dylanjeffers:
### Description

- Upgrades our linting plugins to latest
- runs lint:fix + manual fixes for codebase


packages/eslint-config-audius/src/index.js
```diff
@@ -4,13 +4,11 @@ module.exports = {
     es6: true,
   },
   extends: [
-    "standard",
```
sliptype, line 7: Hmmm, why aren't these needed anymore?

sliptype (approved):
I'm just gonna assume everything is good 😅

packages/eslint-config-audius/src/index.js
dylanjeffers: yeah good catch.... will re-run with it added, that was meant to be there!

audius-infra:
Preview this change https://demo.audius.co/dj-c-674-upgrade-eslint

PR #1586, opened 2022-07-13
[PAY-425] Fix optimistic trophy icon for supporting

rickyrombo:
### Description

The `getOptimisticSupporters` selector re-ranks supporters based on the amounts, ensuring the ranking of supporters in the "Top Supporters" modal is correct. However, it doesn't do the same for the _supporting_, which is set to -1 when applied optimistically. Rather than rewrite this to do reranking at the saga level, I added the `getOptimisticSupporters` selector to the `getOptimisticSupporting` selector as a dependency. That way we'll have the updated rankings already computed on the supporter side and can map the rankings over to the supporting side.

Additionally, includes a bugfix where the profile image/name weren't at the bottom of the tile for non-top supporters on web

### Dragons

This might slightly affect perf, as we now have to loop over all entries in the entire supportingMap to check rankings. Reselect should prevent recomputations

### How Has This Been Tested?

Tested against staging

### Feature Flags ###

### DIFF
--- a/packages/web/src/common/store/tipping/selectors.ts
+++ b/packages/web/src/common/store/tipping/selectors.ts
@@ -77,16 +77,6 @@ const mergeMaps = <
   return mergedMap
 }
 
-export const getOptimisticSupporting = createSelector(
-  getSupporting,
-  getSupportingOverrides,
-  (supporting, supportingOverrides) =>
-    mergeMaps<SupportingMap>({
-      map: supporting,
-      mapOverrides: supportingOverrides
-    })
-)
-
 export const rerankSupportersMapForUser = (
   supportersForUser: SupportersMapForUser
 ) => {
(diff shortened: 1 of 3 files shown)


packages/web/src/common/store/tipping/selectors.ts
```diff
@@ -162,6 +152,43 @@ export const getOptimisticSupporters = createSelector(
...
+     */
+    const result: SupportingMap = {}
+    const mergedUserIds = Object.keys(mergedMap) as unknown as ID[]
+    for (const userId of mergedUserIds) {
+      result[userId] = {}
+      const supportingUserIds = Object.keys(
+        mergedMap[userId]
+      ) as unknown as ID[]
+      for (const supportingUserId of supportingUserIds) {
+        result[userId][supportingUserId] = {
```
rickyrombo, line 179: I don't think i need this reducer-like immutability here, do I? I should be able to just change rank directly on `mergedMap` right? Testing now...

audius-infra:
Preview this change https://demo.audius.co/mjp-trophy-bug

piazzatron (approved):
Clever 🤓

PR #1587, opened 2022-07-14
[PAY-393] Fix active state animation on mobile SupportingTile

piazzatron:
### Description

This animation was too aggressive - scaling down to default 0.97 looks bad for large tiles.

### Dragons

None

### How Has This Been Tested?

Device

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/screens/profile-screen/ProfileHeaderV2/SupportingList.tsx
+++ b/packages/mobile/src/screens/profile-screen/ProfileHeaderV2/SupportingList.tsx
@@ -81,6 +81,7 @@ export const SupportingList = () => {
       <SupportingTile
         style={styles.singleSupporterTile}
         supporting={supportingSorted[0]}
+        onPressScaleTo={0.985}
       />
     )
   }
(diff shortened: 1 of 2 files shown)


packages/mobile/src/screens/profile-screen/ProfileHeaderV2/SupportingList.tsx
```diff
@@ -81,6 +81,7 @@ export const SupportingList = () => {
       <SupportingTile
         style={styles.singleSupporterTile}
         supporting={supportingSorted[0]}
+        onPressScaleTo={0.985}
```
dylanjeffers, line 84: even if the the prop `scaleTo` is not clear for tile, i think it's best to use the same prop names for things that are the same type of component. so since SupportingTile is a tile, then `scaleTo` would be preferred by me to keep api consistent. You can also update SupportingTile props in a potentially cleaner way by doing

```
type SupportingTileProps = TileProps & {...}
```

dylanjeffers (commented):
thoughts re api

PR #1588, opened 2022-07-14
[C-427] Fix matrix mode color for dismiss feed tip tile

dylanjeffers:
### Description

- Fixes matrix mode color for native feed-tip-tile
- Adds environment variable for mobile configs, and enables matrix mode on staging for all users (thinking this is a helpful feature for beta testing)
- Replaces IconClose with IconRemove in mobile and web, since we essentially had duplicate icons
- Uses `IconButton` for web close button for better a11y and use of stems
- Adds hover and active scale animations for close button to better align with design system. (we should honestly move all that to IconButton eventually)

### Dragons

Got to make sure matrix is only avail on staging with this change, which i think is the case since it's checking for 'staging', which prod config doesn't have


packages/mobile/.env.dev.tmpl
```diff
@@ -1,3 +1,4 @@
+ENVIRONMENT=staging
```
rickyrombo, line 1: This doesn't feel right - shouldn't staging be separate from dev?

packages/mobile/src/screens/settings-screen/AppearanceSettingsRow.tsx
```diff
@@ -14,6 +15,8 @@ import { SettingsRow } from './SettingsRow'
 import { SettingsRowContent } from './SettingsRowContent'
 import { SettingsRowDescription } from './SettingsRowDescription'
 
+const isStaging = Config.ENVIRONMENT === 'staging'
```
rickyrombo, line 18: Maybe instead we can check for `Config.ENVIRONMENT !=== 'production'` or something?

packages/web/.env/.env.dev
```diff
@@ -1,4 +1,4 @@
-REACT_APP_ENVIRONMENT=
+REACT_APP_ENVIRONMENT=staging
```
rickyrombo, line 1: Does this get overridden by configureLocalDev or no?

rickyrombo (approved):

rickyrombo:
I like the idea wrt the matrix mode being enabled on staging - though I think it might also work to give all of us tons of $AUDIO via the spout instead to exercise that enable matrix mode path

dylanjeffers:
> I like the idea wrt the matrix mode being enabled on staging - though I think it might also work to give all of us tons of $AUDIO via the spout instead to exercise that enable matrix mode path

Agreed with that, would help with a bunch of other things too

audius-infra:
Preview this change https://demo.audius.co/dj-c-427-fix-matrix-mode-close-feed-tip-tile

PR #1591, opened 2022-07-14
[C-679] Fix Probers

sliptype:
### Description

`serve` got a major version update so the path we were referencing no longer existed. Opted to use `npx` to keep breaking changes from preventing it from running

### Dragons

### How Has This Been Tested?

### Feature Flags ###


dylanjeffers (approved):
changes make sense! i wonder if we just install serve in dependencies in the future too. can look into it when i upgrade cypress anyway

PR #1596, opened 2022-07-14
[PAY-424] Remove search icon in user-lists

dylanjeffers:
### Description

Removes search icon in user-lists, since it makes it look like the user-list is searchable

### DIFF
--- a/packages/mobile/src/screens/user-list-screen/UserListScreen.tsx
+++ b/packages/mobile/src/screens/user-list-screen/UserListScreen.tsx
@@ -31,7 +31,8 @@ export const UserListScreen = (props: UserListScreenProps) => {
     <Screen
       variant='white'
       title={isTippingEnabled ? undefined : title}
-      headerTitle={isTippingEnabled ? headerTitle : undefined}>
+      headerTitle={isTippingEnabled ? headerTitle : undefined}
+      topbarRight={null}>
       {children}
     </Screen>
   )


dharit-tan (approved):

packages/mobile/src/screens/user-list-screen/UserListScreen.tsx
```diff
@@ -31,7 +31,8 @@ export const UserListScreen = (props: UserListScreenProps) => {
     <Screen
       variant='white'
       title={isTippingEnabled ? undefined : title}
-      headerTitle={isTippingEnabled ? headerTitle : undefined}>
+      headerTitle={isTippingEnabled ? headerTitle : undefined}
+      topbarRight={null}>
```
dharit-tan, line 35: just curious - where did topbarRight default to the search icon in the first place? can't seem to find it..

dylanjeffers: yeah that's a good point, it's super hard to find where this is first off, and secondly it prob doesn't make sense to have search as a default thing anyway...

dylanjeffers: its here tho: https://github.com/AudiusProject/audius-client/blob/main/packages/mobile/src/screens/app-screen/useAppScreenOptions.tsx

PR #1599, opened 2022-07-15
[PAY-432] Freeze balance briefly after sending a tip

piazzatron:
### Description

There is a bug (exercised by @dharit-tan) where after sending a tip and optimistically updating the balance, some balance polling rewrites the balance before Solana has returns the correct value. This is easily seen by sending a tip and going to the artist-dashboard, which triggers a polling interval.

This definitely isn't the *best* way to handle this – the proper way would be to look up the tx referenced by the signature returned from identity relay, store the slot, and then when fetching balance, compare the stored slot to the current on chain slot. But too much effort for a small bug like this.

### Dragons

Not really.

### How Has This Been Tested?

Tested on stage

### Feature Flags ###

### DIFF
--- a/packages/web/src/common/store/wallet/selectors.ts
+++ b/packages/web/src/common/store/wallet/selectors.ts
@@ -19,14 +19,16 @@ export const getAccountBalance = createSelector(
   (balance) => (balance ? stringWeiToBN(balance) : null)
 )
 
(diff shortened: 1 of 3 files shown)


audius-infra:
Preview this change https://demo.audius.co/piazz-freeze-balance-refresh-after-tip

rickyrombo (approved):
Lovely hack <3

PR #1624, opened 2022-07-19
[PAY-453] Update feed-tip-tile sender styles

dylanjeffers:
### Description

Replaces opacity with underline for feed-tip-tile sender styles


sddioulde (approved):
lg thx!

PR #1626, opened 2022-07-19
[PAY-452] Fetch audio balance on audio/tip-screen

dylanjeffers:
### Description

Fetches $AUDIO balance when navigating to audio/tip-screens, and when returning to app with one of these screens open


rickyrombo (approved):

piazzatron (approved):
senddddddd

PR #1635, opened 2022-07-20
[C-705] Fix mac desktop build

sliptype:
### Description

Installs `dmg-license` on the fly because it is an optional dependency due to only being installable on mac. I didn't want to prevent devs from using non mac systems, hence installing the package in ci

### Dragons

### How Has This Been Tested?

### Feature Flags ###


dylanjeffers (approved):
awesome makes sense, so installing it fails in windows and linux? or just a way to prevent node_modules from getting to large? only thought is that if it is kinda large, might be faster to just include in node_mods and rely on our caching?

audius-infra:
Preview this change https://demo.audius.co/sk-c-705-fix-mac-desktop-build

sliptype:
> awesome makes sense, so installing it fails in windows and linux? or just a way to prevent node_modules from getting to large? only thought is that if it is kinda large, might be faster to just include in node_mods and rely on our caching?

According to this thread it fails on other platforms: https://github.com/electron-userland/electron-builder/issues/6489

But also I don't think we are specifying `--no-optional` anywhere so not sure why it isn't installed, maybe a lerna thing?

dylanjeffers:
> > awesome makes sense, so installing it fails in windows and linux? or just a way to prevent node_modules from getting to large? only thought is that if it is kinda large, might be faster to just include in node_mods and rely on our caching?
> 
> According to this thread it fails on other platforms: [electron-userland/electron-builder#6489](https://github.com/electron-userland/electron-builder/issues/6489)
> 
> But also I don't think we are specifying `--no-optional` anywhere so not sure why it isn't installed, maybe a lerna thing?

word, weird, well good fix

PR #1639, opened 2022-07-21
[C-639] Update track list item title to display the deleted by artist text better

Kyle-Shanks:
### Description
Updated the "Deleted By Artist" text to always be displayed in the track list item titles

### Dragons

There was some funkiness getting the sizing to work out properly, but I think I have it dialed in here

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A

### Before:
Screen Shot 2022-07-21 at 4 40 48 PM

### After:
Screen Shot 2022-07-21 at 4 41 16 PM


dylanjeffers (approved):
would love a before and after!

PR #1645, opened 2022-07-22
[C-709] Fix native audio tier

dylanjeffers:
### Description

Native audio tiers where showing current tier silver and gold


Kyle-Shanks (approved):
Nice 😅

PR #1646, opened 2022-07-23
[C-584] Add others link to web notifications

dylanjeffers:
### Description

- Add others link to web notifications to highlight there are more people
- Also fixes favorite notification text that said repost

image

### DIFF
--- a/packages/web/src/components/notification/Notification/FavoriteNotification.tsx
+++ b/packages/web/src/components/notification/Notification/FavoriteNotification.tsx
@@ -23,7 +23,7 @@ import { IconFavorite } from './components/icons'
 import { entityToUserListEntity } from './utils'
 
 const messages = {
-  reposted: ' favorited your '
+  favorited: ' favorited your '
 }
 
 type FavoriteNotificationProps = {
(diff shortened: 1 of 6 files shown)


sliptype (approved):
Looks great!

More of a design thing but the copy in the "1 other" case feels a lil clunky. Like "x and 1 other followed you" could maybe be better as "x and y followed you". But I guess we want it to be obvious where pressing the notification tile will take you

packages/web/src/components/notification/Notification/FavoriteNotification.tsx
```diff
@@ -16,14 +16,14 @@ import { NotificationBody } from './components/NotificationBody'
...
 import { UserNameLink } from './components/UserNameLink'
 import { UserProfilePictureList } from './components/UserProfilePictureList'
 import { IconFavorite } from './components/icons'
-import { entityToUserListEntity, formatOthersCount } from './utils'
+import { entityToUserListEntity } from './utils'
 
 const messages = {
-  others: formatOthersCount,
-  reposted: ' reposted your '
+  reposted: ' favorited your '
```
Kyle-Shanks, line 26: reposted

PR #1653, opened 2022-07-27
[PAY-244][PAY-245] Update tip modal header and extract pill button

sddioulde:
### Description

Update stems components:
- update ModalHeader in stems to take styles for the dismiss button
- update tip audio modal to use ModalHeader
- add new PillButton

The new PillButton can be used for future components. It takes a type prop that will change the hover and active state styles accordingly (currently only primary and secondary types).

### Dragons

n/a

### How Has This Been Tested?

local against stage

Screen Shot 2022-07-27 at 6 30 52 PM
Screen Shot 2022-07-27 at 6 31 02 PM
Screen Shot 2022-07-27 at 6 31 43 PM

=============

Screen Shot 2022-07-27 at 5 52 13 PM
Screen Shot 2022-07-27 at 5 52 19 PM
Screen Shot 2022-07-27 at 5 52 22 PM
Screen Shot 2022-07-27 at 5 23 00 PM
Screen Shot 2022-07-27 at 5 23 06 PM
Screen Shot 2022-07-27 at 5 23 30 PM

n/a

### Feature Flags ###

none

### DIFF
--- a/packages/stems/src/components/PillButton/PillButton.module.css
+++ b/packages/stems/src/components/PillButton/PillButton.module.css
@@ -1,0 +1,42 @@
+.button {
+  height: unset !important;
+  padding: 10px 0;
+  border-radius: 100px;
+  border: 1px solid var(--neutral-light-4);
+  background: var(--white);
+}
+.buttonText {
+  display: flex;
+  align-items: center;
+  text-transform: uppercase;
+  font-size: var(--font-xs) !important;
+  color: var(--neutral) !important;
+  letter-spacing: unset !important;
+}
+.button:hover {
+  transform: scale3d(1.01, 1.01, 1.01) !important;
+  border: unset !important;
+}
+.button:active {
+  transform: scale3d(0.99, 0.99, 0.99) !important;
+}
+.button:hover .buttonText,
+.button:active .buttonText {
+  color: var(--white) !important;
+}
+.button.primary:hover,
+.button.primary:hover .buttonText {
+  background: var(--primary) !important;
+}
+.button.primary:active,
+.button.primary:active .buttonText {
+  background: var(--primary-dark-2) !important;
+}
+.button.secondary:hover,
+.button.secondary:hover .buttonText {
+  background: var(--secondary-light-2) !important;
+}
+.button.secondary:active,
+.button.secondary:active .buttonText {
+  background: var(--secondary) !important;
+}
--- a/packages/stems/src/components/PillButton/PillButton.tsx
+++ b/packages/stems/src/components/PillButton/PillButton.tsx
@@ -1,0 +1,29 @@
+import cn from 'classnames'
+
+import { Button, Type as ButtonType } from 'components/Button'
+
+import styles from './PillButton.module.css'
+import { PillButtonProps, Type as PillButtonType } from './types'
+
+const TYPE_STYLE_MAP = {
+  [PillButtonType.PRIMARY]: styles.primary,
+  [PillButtonType.SECONDARY]: styles.secondary
+}
+export const PillButton = (props: PillButtonProps) => {
+  const {
+    text,
+    onClick,
+    type = PillButtonType.PRIMARY,
+    className,
+    textClassName
+  } = props
+  return (
+    <Button
+      className={cn(styles.button, TYPE_STYLE_MAP[type], className)}
+      textClassName={cn(styles.buttonText, textClassName)}
+      type={ButtonType.COMMON}
+      text={text}
+      onClick={onClick}
+    />
+  )
+}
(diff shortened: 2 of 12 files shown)


packages/stems/src/components/PillButton/PillButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+    onClick,
+    type = PillButtonType.PRIMARY,
+    className,
+    textClassName
+  } = props
+  return (
+    <Button
+      className={cn(styles.button, TYPE_STYLE_MAP[type], className)}
+      textClassName={cn(styles.buttonText, textClassName)}
+      type={ButtonType.COMMON}
```
dylanjeffers, line 24: love that we can reuse `type` here. just a headsup, i am wanting us to move away from the `type` prop since it is overloaded with the actual button `type` prop which is `submit | button` etc. so to avoid having that with pill button, can we use `variant` and just pass that into button's `type` for now?

dylanjeffers: also, we probably want to pass the rest of button props through yeah? so we get id, data-testid, etc. i like doing
```typescript
const {className, textClassName, ...other} = props
<Button
   className={...}
   textClassName={...}
   {...other}
/>

File: packages/stems/src/components/PillButton/PillButton.module.css [hunk 2]
```diff
@@ -0,0 +1,42 @@
+.button {
+  height: unset !important;
```
rickyrombo, line 2: Using `!important` is very bad smell, are there other ways we could accomplish this? 😨

packages/stems/src/components/PillButton/PillButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+export const PillButton = (props: PillButtonProps) => {
+  const {
+    text,
+    onClick,
+    type = PillButtonType.PRIMARY,
+    className,
+    textClassName
+  } = props
+  return (
+    <Button
```
rickyrombo, line 21: How hard would it be to make a `PillRadioButton` that uses a `<input type="radio">` under the hood but keeps all the styles :eyes:

packages/stems/src/components/PillButton/types.ts
```diff
@@ -0,0 +1,18 @@
+import { HTMLAttributes } from 'react'
+// import { HTMLAttributes, ReactNode } from 'react'
+import * as React from 'react'
+
+export enum Type {
+  PRIMARY = 'primary',
+  SECONDARY = 'secondary'
+}
+
+export type PillButtonProps = HTMLAttributes<HTMLButtonElement> & {
```
dylanjeffers, line 10: Can we extend on Button, so 
```typescript
type PillButtonProps = ButtonProps
```
or if we are gonna do the variant change (which now makes me think lets not do it and just do it with button eventually:

```typescript
type PillButtonProps = ButtonProps & {
  variant: PillVariant
}
```

packages/web/src/components/tipping/feed-tip-tile/FeedTipTile.tsx
```diff
@@ -115,38 +115,36 @@ type SendTipToButtonProps = {
...
-                badgeSize={12}
-                inline
-              />
-            </div>
-          )
-        }
-        onClick={handleClick}
-      />
-    </div>
+    <PillButton
```
dylanjeffers, line 142: love it!

packages/web/src/components/tipping/tip-audio/TipAudioModal.tsx
```diff
@@ -36,42 +36,22 @@ const GoldBadgeIconImage = () => (
...
+const titleMessagesMap: { [key in TippingSendStatus]?: string } = {
+  SEND: messages.sendATip,
+  CONFIRM: messages.confirm,
+  SENDING: messages.sending,
+  CONVERTING: messages.holdOn,
+  ERROR: messages.confirm,
+  SUCCESS: messages.tipSent
+}
+
+const titleIconsMap = {
```
dylanjeffers, line 48: clean

packages/web/src/components/tipping/tip-audio/TipAudioModal.tsx
```diff
@@ -140,20 +120,26 @@ export const TipAudioModal = () => {
...
-      showDismissButton={
-        sendStatus !== 'SENDING' && sendStatus !== 'CONVERTING'
-      }
       dismissOnClickOutside={
         sendStatus !== 'SENDING' && sendStatus !== 'CONVERTING'
-      }
-      contentHorizontalPadding={24}
-      useGradientTitle={false}>
+      }>
+      <ModalHeader
```
dylanjeffers, line 130: also great update

dylanjeffers (approved):
awesome awesome work. main comments are related to api for pill etc

PR #1668, opened 2022-08-03
INF-210 Add Slack alerts to CircleCI

joaquincasares:
### Description

Copy the audius-protocol implementation of Slack alerts into audius-client.

### Dragons

See comments.

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,18 +10,163 @@ commands:
       - run:
           name: Combine package-lock.json files to single file
           command: ls -d -- packages/*/*-lock.json | awk -F packages '{printf "\"packages%s\" ", $2}' | xargs cat > << parameters.filename >>
+  slack-basic:
+    parameters:
+      event:
+        description: when this notification will fire. options are fail, pass, always.
+        type: string
+        default: always
+      branch_pattern:
+        description: branch_pattern arg for slack/notify
+        type: string
+        default: .+
+      channel:
+        description: Slack channel to send message to
+        type: string
+        default: $SLACK_DEFAULT_CHANNEL
+      text:
+        description: Markdown formatted Slack message
+        type: string
+        default: "Job started. :crossed_fingers:"
+      slack_mentions_user:
+        description: Used for CircleCI @mentions
+        type: string
+        default: ""
+    steps:
+      - slack/notify:
+          channel: << parameters.channel >>
+          branch_pattern: << parameters.branch_pattern >>
+          custom: |
+            {
+              "blocks": [
+                {
+                  "type": "header",
+                  "text": {
+                    "type": "plain_text",
+                    "text": "<< parameters.text >>",
+                    "emoji": true
+                  }
+                },
+                {
+                  "type": "section",
+                  "fields": [
+                    {
+                      "type": "mrkdwn",
+                      "text": "*Job*: ${CIRCLE_JOB}"
+                    }
+                  ]
+                },
+                {
+                  "type": "section",
+                  "fields": [
+                    {
+                      "type": "mrkdwn",
+                      "text": "*Project*: $CIRCLE_PROJECT_REPONAME"
+                    },
+                    {
+                      "type": "mrkdwn",
+                      "text": "*Branch*: $CIRCLE_BRANCH"
+                    },
+                    {
+                      "type": "mrkdwn",
+                      "text": "*Author*: $CIRCLE_USERNAME"
+                    },
+                    {
+                      "type": "mrkdwn",
+                      "text": "*Mentions*: << parameters.slack_mentions_user >>"
+                    }
+                  ]
+                },
+                {
+                  "type": "actions",
+                  "elements": [
+                    {
+                      "type": "button",
+                      "text": {
+                        "type": "plain_text",
+                        "text": "View Job"
+                      },
+                      "url": "${CIRCLE_BUILD_URL}"
+                    }
+                  ]
+                }
+              ]
+            }
+          event: << parameters.event >>
+  slack-fail:
+    parameters:
+      event:
+        description: when this notification will fire. options are fail, pass, always.
+        type: string
+        default: fail
+      branch_pattern:
+        description: branch_pattern arg for slack/notify
+        type: string
+        default: .+
+      channel:
+        description: Slack channel to send message to
+        type: string
+        default: $SLACK_DEFAULT_CHANNEL
+      text:
+        description: Markdown formatted Slack message
+        type: string
+        default: "Job Failed. :red_circle:"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -10,18 +10,163 @@ commands:
...
+        default: "Job started. :crossed_fingers:"
+      slack_mentions_user:
+        description: Used for CircleCI @mentions
+        type: string
+        default: ""
+    steps:
+      - slack/notify:
+          channel: << parameters.channel >>
+          branch_pattern: << parameters.branch_pattern >>
+          custom: |
```
joaquincasares, line 39: We remove the OVERLY large CircleCI logo:
* https://github.com/CircleCI-Public/slack-orb/blob/master/src/message_templates/basic_success_1.json#L43

We also consolidate the Mentions field into an existing section to minimize vertical height noise.

.circleci/config.yml
```diff
@@ -10,18 +10,163 @@ commands:
...
+                }
+              ]
+            }
+          event: << parameters.event >>
+  slack-fail:
+    parameters:
+      event:
+        description: when this notification will fire. options are fail, pass, always.
+        type: string
+        default: fail
```
joaquincasares, line 101: Default to `fail` since that's usually when we'll use this.

.circleci/config.yml
```diff
@@ -10,18 +10,163 @@ commands:
...
+        type: string
+        default: always
+      branch_pattern:
+        description: branch_pattern arg for slack/notify
+        type: string
+        default: .+
+      channel:
+        description: Slack channel to send message to
+        type: string
+        default: $SLACK_DEFAULT_CHANNEL
```
joaquincasares, line 26: This is set via the `slack-secrets` context as a required envar: 

* https://github.com/CircleCI-Public/slack-orb/wiki/Setup#4-create-a-context-on-circleci

.circleci/config.yml
```diff
@@ -10,18 +10,163 @@ commands:
...
+      text:
+        description: Markdown formatted Slack message
+        type: string
+        default: "Job started. :crossed_fingers:"
```
joaquincasares, line 30: This is self-deprecating humor. Feel free to make it more positive. 🤓 

I'll follow suit. :D

.circleci/config.yml
```diff
@@ -10,18 +10,163 @@ commands:
       - run:
           name: Combine package-lock.json files to single file
           command: ls -d -- packages/*/*-lock.json | awk -F packages '{printf "\"packages%s\" ", $2}' | xargs cat > << parameters.filename >>
+  slack-basic:
```
joaquincasares, line 13: Probably better to go in it's own commands.yml, but I'll keep an eye on how you do this so I can do the same in audius-protocol. 

Side note: Perhaps we can have a call soon so you can walk me through how you organized your ymls? I noticed `generate-config` today. Do you think it's worth separating out our config.yml within `audius-protocol` today?

.circleci/config.yml
```diff
@@ -10,18 +10,163 @@ commands:
...
+          branch_pattern: << parameters.branch_pattern >>
+          channel: << parameters.channel >>
+          text: << parameters.text >>
+          slack_mentions_user: << parameters.slack_mentions_user >>
+  slack-success:
+    parameters:
+      event:
+        description: when this notification will fire. options are fail, pass, always.
+        type: string
+        default: pass
```
joaquincasares, line 130: Default to `pass`.

.circleci/config.yml
```diff
@@ -87,6 +232,10 @@ jobs:
       - add_ssh_keys:
           fingerprints:
             - "a4:99:2e:fa:80:31:dc:23:25:7b:a5:70:3f:08:5f:fa"
+      - slack-basic:
+          event: always
+          branch_pattern: master
+          slack_mentions_user: << pipeline.parameters.slack_mentions_user >>
```
joaquincasares, line 238: Within `audius-protocol` I'm passing this around as a job parameter as well.

This seemed easier and perhaps a better style to call `pipeline.parameters` directly?

.circleci/config.yml
```diff
@@ -105,6 +254,12 @@ jobs:
             VERSION=$(jq -r .version ./lerna.json)
             git checkout -b release-v${VERSION}
             git push -u origin release-v${VERSION}
+      - slack-fail:
+          branch_pattern: master
+          slack_mentions_user: << pipeline.parameters.slack_mentions_user >>
+      - slack-success:
+          branch_pattern: master
+          slack_mentions_user: << pipeline.parameters.slack_mentions_user >>
```
joaquincasares, line 262: If the previous `run` command fails, we'll trigger only `slack-fail`, else we'll only trigger `slack-success`.

.circleci/config.yml
```diff
@@ -161,3 +316,5 @@ workflows:
         - equal: ["generate-release", << pipeline.schedule.name >>]
     jobs:
       - generate-release
+          context:
+            - slack-secrets
```
joaquincasares, line 320: We're using the same context as audius-protocol which defaults the channel to #eng-deploy-announcements using the CircleCI user I created.

audius-infra:
Preview this change https://demo.audius.co/jc-inf-210.client

.circleci/config.yml
joaquincasares: Not sure why this line is causing build errors when it's similar to audius-protocol:

* https://github.com/AudiusProject/audius-protocol/blob/89e0db2520df13187b809488c4b3aa723a2a7361/.circleci/config.yml#L1017

sliptype: I think the context needs to be at the workflow level instead of the job level

.circleci/config.yml
```diff
@@ -105,6 +254,12 @@ jobs:
             VERSION=$(jq -r .version ./lerna.json)
             git checkout -b release-v${VERSION}
             git push -u origin release-v${VERSION}
+      - slack-fail:
+          branch_pattern: master
```
sliptype, line 258:
```suggestion
          branch_pattern: main
```

.circleci/config.yml
sliptype: Yeah, seems cleaner!

.circleci/config.yml
sliptype: Yup I think we can put this in the `.circleci/commands/commands.yml` file. Let's chat today about how this is structured, at a high level it's using circleci's dynamic configuration feature: https://circleci.com/docs/dynamic-config

I would say yes it makes sense to break up the `audius-protocol` config as well. It's pretty big already as one file

sliptype (approved):

PR #1670, opened 2022-08-03
Make audiusBackend configurable for different environments

sliptype:
### Description

* Updates `audiusBackend` from using env vars to accepting parameters for configuration values. This allows us to use `config` in react native instead of process.env. Also refactors audiusBackend into a function instead of a static class
* Creates `audiusBackendInstance`. Updates all references in web from `AudiusBackend` -> `audiusBackendInstance`. Once we have a corresponding `audiusBackendInstance` in native, we will probably have to pass these instances into the relevant sagas, etc.
* Migrates all dependencies of `audiusBackend` to common in preparation for moving `audiusBackend` to common itself
* Removes most instances of `window` in `audiusBackend`, there are still a couple but I believe they are limited to claiming rewards. We can address these when we build out screens

In following prs will:
* Move `audiusBackend` into @audius/common
* Do a pass of `audiusBackend` with the new libs types

### Dragons

Pretty big structural changes, smoke tested locally but definitely want to do some through testing once on stage

### How Has This Been Tested?

### Feature Flags ###


packages/web/src/services/AudiusBackend.ts
```diff
@@ -246,123 +173,253 @@ export const fetchCID = async (
...
+      const timeoutMs =
+        getRemoteVar(IntKeys.IMAGE_QUICK_FETCH_TIMEOUT_MS) ?? undefined
+      const timeout = setTimeout(() => {
+        preloadImageTimer.end(start)
+        resolve(false)
+      }, timeoutMs)
 
-    if (cache) CIDCache.add(cid, url)
+      // Avoid garbage collection by keeping a few images in an in-mem array
+      const image = new Image()
```
dylanjeffers, line 360: this compat with native?

packages/web/src/services/AudiusBackend.ts
```diff
@@ -246,123 +173,253 @@ export const fetchCID = async (
...
+    await waitForLibsInit()
+    // Else, race fetching of the image from all gateways & return the image url blob
+    try {
+      const image = await audiusLibs.File.fetchCID(
+        cid,
+        creatorNodeGateways,
+        () => {}
+      )
+
+      const url = nativeMobile
```
dylanjeffers, line 409: nice

packages/web/src/services/AudiusBackend.ts
```diff
@@ -1911,24 +1925,24 @@ class AudiusBackend {
...
-    const host = AudiusBackend._getHostUrl()
+    const host = _getHostUrl()
     return audiusLibs.Account.generateRecoveryLink({ host })
   }
 
-  static _getHostUrl() {
-    return NATIVE_MOBILE && process.env.REACT_APP_ENVIRONMENT === 'production'
-      ? AUDIUS_ORIGIN
+  async function _getHostUrl() {
+    return nativeMobile && process.env.REACT_APP_ENVIRONMENT === 'production'
```
dylanjeffers, line 1940: i kinda forget if process exits like this in native mobile, seemingly REACT_APP wouldn't since thats a CRA thing?

dylanjeffers (approved):
awesome stuff! so just wondering if the goal is to move audius-backend to common after this?

packages/web/src/services/AudiusBackend.ts
sliptype: Oooo I don't think so, good catch

Let me think about how to address this

packages/web/src/services/AudiusBackend.ts
sliptype: Yeah I just missed this one, updated now!

nicoback2 (approved):
this is awesome!

PR #1677, opened 2022-08-05
[C-780] Update AudiusBackend to work in native context

dylanjeffers:
### Description

Updates a few AudiusBackend internals that were not native compliant.
- Added `getLibs` method to inject the libs code, since native needs a different libs entrypoint
- Moved `withEagerOption` and `waitForAudiusInit` as arguments, since both of these rely on web code that breaks native. (Native will just stub these out since libs is ready to go on load)
- Moved `fetchCID` to arguments, since it relied on `waitForAudiusInit`, which is now a param. this kinda makes sense since fetchCID will likely need a native implementation anyway

### Dragons

Merge this after the legacy type updates pr in audius-protocol, since ts doesn't think legacy has types with latest @audius/sdk


sliptype (approved):
Oh my god it's coming together!!!

audius-infra:
Preview this change https://demo.audius.co/dj-c-780-update-audius-backend-to-work-in-native-context

PR #1683, opened 2022-08-05
[C-782] Catch fetch errors in AudiusBackend

sliptype:
### Description

We weren't awaiting the fetch promises and there was no `.catch` so errors were not being caught, causing the app to crash. Added `await` for each fetch call, ideally we would refactor so there are no chained `then`s also but that isn't necessary right away

### Dragons

Because we are catching more errors now this should be fairly safe, the impact being that something that did cause the app to crash before no longer does so

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/web/src/services/AudiusBackend.ts
+++ b/packages/web/src/services/AudiusBackend.ts
@@ -2104,7 +2104,7 @@ export const audiusBackend = ({
     if (!account) return
     try {
       const { data, signature } = await signData()
-      return fetch(`${identityServiceUrl}/notifications/all`, {
+      return await fetch(`${identityServiceUrl}/notifications/all`, {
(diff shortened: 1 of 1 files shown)


audius-infra:
Preview this change https://demo.audius.co/sk-c-782-fix-error-handling

raymondjacobson (approved):
sensible way to do this change

raymondjacobson:
I assume you tested by throwing or blocking a network req.

sliptype:
> I assume you tested by throwing or blocking a network req.

Yup, exactly!

sliptype:
Might also fix https://linear.app/audius/issue/C-728/app-instantly-goes-to-error-page-if-identity-is-down

PR #1684, opened 2022-08-05
[C-781] Add minimum threshold to common password check

Kyle-Shanks:
### Description
Add a minimum threshold to the common password check to make it less aggressive

### Dragons

N/A

### How Has This Been Tested?

Manually tested

Will watch to see if the check is still too aggressive

### Feature Flags ###

N/A

### DIFF
--- a/packages/web/src/utils/commonPasswordCheck.ts
+++ b/packages/web/src/utils/commonPasswordCheck.ts
@@ -2,6 +2,8 @@ const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms))
 
 const HIBP_URL = 'https://api.pwnedpasswords.com/range/'
 
+const API_MIN_MATCH_COUNT = 20
+
 export const commonPasswordCheck = async (
   password: string
 ): Promise<boolean> => {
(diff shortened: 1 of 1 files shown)


packages/web/src/utils/commonPasswordCheck.ts
```diff
@@ -23,8 +25,18 @@ export const commonPasswordCheck = async (
   if (result) {
     // @ts-ignore
     const text = (await result?.text()) as string
-    const map = text.split(/\s+/g).map((s) => s.slice(0, s.indexOf(':')))
-    return map.includes(hash.slice(5))
+    const hashArr = text.split(/\s+/g).map((s) => s.slice(0, s.indexOf(':')))
+
+    // If there is no match, return false
+    if (!hashArr.includes(hash.slice(5))) {
```
sliptype, line 31: Should we be checking for the full hash instead of just the first 5 characters? Here and on line 35

sliptype (approved):
Nice! My other thought is that it could be more performant to only traverse the `text` once and create an object of { hash: value }. Then could check if the entry exists in the object and what the value is in constant time

audius-infra:
Preview this change https://demo.audius.co/kj-Add-min-threshold-to-common-password-check

packages/web/src/utils/commonPasswordCheck.ts
Kyle-Shanks: opposite. this is slicing off the first 5 chars and checking the rest. this is just because of how the API returns the data

sliptype: Ah gotcha

PR #1685, opened 2022-08-05
[C-774] Move audius-backend to common, add native libs-instance

dylanjeffers:
### Description

- Moves AudiusBackend to audius-client common
- Add mobile audius-libs-instance,
- Add monitoringCallbacks to config since they rely on custom optimizely and analytics
- Add react-native-url-polyfill for libs monitoring callbacks (uses `new URL(...)`)

### How Has This Been Tested?

Able to make api requests through sdk in native


packages/mobile/src/services/audius-backend-instance.ts
```diff
@@ -0,0 +1,83 @@
...
+    ethTokenBridgeAddress: Config.ETH_TOKEN_BRIDGE_ADDRESS,
+    solBridgeAddress: Config.SOL_BRIDGE_ADDRESS,
+    solTokenBridgeAddress: Config.SOL_TOKEN_BRIDGE_ADDRESS,
+    wormholeRpcHosts: Config.WORMHOLE_RPC_HOSTS
+  },
+  getLibs: async () => nativeLibs,
+  waitForLibsInit: async () => {},
+  withEagerOption: ({ normal }, ...args) => {
+    if (audiusLibs) {
+      console.log('eager option working?', audiusLibs)
```
sliptype, line 79: Log, but probably fine for now

packages/mobile/src/services/audius-backend-instance.ts
```diff
@@ -0,0 +1,83 @@
...
+ * audiusBackend initialized for a mobile environment
+ */
+export const audiusBackendInstance = audiusBackend({
+  claimDistributionContractAddress: Config.CLAIM_DISTRIBUTION_CONTRACT_ADDRESS,
+  ethOwnerWallet: Config.ETH_OWNER_WALLET,
+  ethProviderUrls: (Config.ETH_PROVIDER_URL || '').split(','),
+  ethRegistryAddress: Config.ETH_REGISTRY_ADDRESS,
+  ethTokenAddress: Config.ETH_TOKEN_ADDRESS,
+  fetchCID: async () => ({}),
+  getFeatureEnabled: () => false,
```
sliptype, line 24: We should be able to import remoteConfigInstance and get the feature flags that way, but stub is probably fine for now too

sliptype (approved):
Freaking amazing dude

packages/mobile/src/services/audius-backend-instance.ts
dylanjeffers: yeah agreed, so weirdly feature flags are broken on native when (when i modify metro to resolve native node_modules first, which we want ideally)

sliptype: Interesting, does that have anything to do with the hooks or more likely the optimizely sdk?

dylanjeffers: still investigating, testing to see if its related to polyfills, wether we load the web vs native optimizely code etc...

dylanjeffers: yay figured it out, added.

audius-infra:
Preview this change https://demo.audius.co/dj-c-774-common-audius-backend

PR #1687, opened 2022-08-08
[C-796] Refactor AudiusAPIClient to work in native context

dylanjeffers:
### Description

Breaks apart audius-backend, audius-api-client, and the main store into common implementations. This prepares us to start integrating the first sagas into mobile!


packages/mobile/src/store/backend/sagas.ts
```diff
@@ -0,0 +1,16 @@
+import { call, put } from 'typed-redux-saga'
+
+import { audiusBackendInstance } from 'app/services/audius-backend-instance'
```
nicoback2, line 3: i thought this would only work for web?

nicoback2 (approved):
LGTM besides one q

audius-infra:
Preview this change https://demo.audius.co/dj-c-796-add-common-store-to-native

packages/mobile/src/store/backend/sagas.ts
dylanjeffers: so this audius-backend-instance is mobile compatible! but you're right that we shouldnt copy backend/sagas.ts here, and instead use the one in common (coming in a pr shortly)

audius-infra:
Preview this change https://demo.audius.co/dj-c-796-add-common-store-to-native

PR #1688, opened 2022-08-08
[C-799] Add saga context, refactor backend state

dylanjeffers:
### Description

Adds saga context, and puts audiusApi, audiusBackendInstance, and remoteConfigInstance in it. This is needed to allow sagas to run on mobile, since we will provide our own version of these instances.

- Updates backend state to ts, uses context instead of instance imports, and moves to common since it can actually be used by mobile now


audius-infra:
Preview this change https://demo.audius.co/dj-c-799-add-saga-context

packages/web/src/common/store/storeContext.ts
```diff
@@ -0,0 +1,12 @@
+import { RemoteConfigInstance } from '@audius/common'
+
+import { AudiusAPIClient } from 'common/services/audius-api-client'
+import { AudiusBackend } from 'common/services/audius-backend'
+
+export type CommonStoreContext = {
+  getLocalStorageItem: (key: string) => Promise<string | null>
+  setLocalStorageItem: (key: string, value: string) => Promise<void>
+  remoteConfigInstance: RemoteConfigInstance
+  audiusBackendInstance: AudiusBackend
```
sliptype, line 10: 👌

packages/web/src/store/configureStore.ts
```diff
@@ -114,7 +115,10 @@ const sentryMiddleware = createSentryMiddleware(Sentry, {
   stateTransformer: statePruner
 })
 
-const sagaMiddleware = createSagaMiddleware({ onError: onSagaError })
+const sagaMiddleware = createSagaMiddleware({
+  onError: onSagaError,
+  context: webStoreContext
```
sliptype, line 120: Interesting, we weren't even using the context on web before?

sliptype: Oh just manually on specific sagas I guess

sliptype (approved):
Looks great!

PR #1690, opened 2022-08-08
[C-800] Add native saga context and update common sagas

dylanjeffers:
### Description

- Implements native saga context
- Updates existing common sagas to pull from context instead of passed as args


audius-infra:
Preview this change https://demo.audius.co/dj-c-800-add-context-and-update-common-sagas

packages/mobile/src/store/sagas.ts
```diff
@@ -1,15 +1,13 @@
 import remoteConfig from 'audius-client/src/common/store/remote-config/sagas'
-import { all, fork } from 'typed-redux-saga/macro'
-
-import { remoteConfigInstance } from 'app/services/remote-config/remote-config-instance'
+import { all, fork } from 'typed-redux-saga'
```
sliptype, line 2: Why do we not need the `/macro` here?

packages/web/src/common/store/remote-config/sagas.ts
```diff
@@ -1,27 +1,29 @@
...
-  function* watchRemoteConfigLoad() {
-    // Emit event when provider is ready
-    const chan = eventChannel((emitter) => {
-      remoteConfigInstance.onClientReady(() => {
-        emitter(CLIENT_READY_EVENT)
-        emitter(END)
-      })
-      return () => {}
+function* watchRemoteConfigLoad() {
+  const remoteConfigInstance = yield* getContext('remoteConfigInstance')
```
sliptype, line 11: It might be nice to do some context helper functions like `getRemoteConfigInstance()` or something

packages/web/src/store/sagas.ts
```diff
@@ -136,7 +135,7 @@ export default function* rootSaga() {
     walletSagas(),
 
     // Cast
-    castSagas(webStoreContext),
+    castSagas(storeContext),
```
sliptype, line 138: We could probably also refactor cast sagas to use the context

sliptype (approved):
🔥

packages/mobile/src/store/sagas.ts
dylanjeffers: there is weird behavior with /macro when using across web and mobile, so for consistency rn, removing macro usages until we move everything to common

packages/web/src/common/store/remote-config/sagas.ts
dylanjeffers: yeah very down to figure out a less verbose pattern here

packages/web/src/store/sagas.ts
dylanjeffers: def!

PR #1691, opened 2022-08-08
[C-803] Add common state and backend sagas to native

dylanjeffers:
### Description

- Adds common state (reducers) to mobile root reducer
- Adds common backend sagas to native, leveraging new saga context
- Refactors FingerprintClient to work with web and native, and adds to context


audius-infra:
Preview this change https://demo.audius.co/dj-c-803-add-common-state-backend-sagas

packages/mobile/src/store/store.ts
```diff
@@ -57,7 +58,8 @@ const createRootReducer = () =>
     remoteConfig,
     search,
     signon,
-    web
+    web,
+    ...commonReducers()
```
sliptype, line 62: 👀

sliptype (approved):
poggers

packages/mobile/src/store/store.ts
dylanjeffers: yeahh... so it technically works, obvs nothing gets updated since sagas aren't live though. do we want to include this big change now you think? or slowly add common reducers?

sliptype: I think it's fine to have all of them now

PR #1698, opened 2022-08-09
[C-806] Prevent multiple overflow menus at a time

sddioulde:
### Description

Prevent multiple overflow menus at a time.

The issue was that there was an event.stopPropagation as the click event from the the kebab menu in the track/playlist was bubbling up. This was there so that the underlying track does not play/pause on kebab menu click. However, because the click event stopped bubbling up, the logic for the click outside of already open menus would not trigger, leaving those menus open.
This PR does not stop the click event from bubbling up to the document, and it prevents the play/pause track if the click target is from the kebab menu.

### Dragons

Other menu overflow behavior should still work as expected.

### How Has This Been Tested?

local vs stage

https://www.loom.com/share/034b9e3fb52e4ac09fdd2b975cd2ad08

n/a

### Feature Flags ###

none

### DIFF
--- a/packages/web/src/components/track/helpers.ts
+++ b/packages/web/src/components/track/helpers.ts
@@ -85,12 +85,3 @@ export const getUserWithFallback = (user: User | null) => {
     }
   )
 }
-
-export const isDescendantElementOf = (
-  descendant: any,
-  ancestor: Nullable<HTMLElement>
-) => {
-  const descendantElement =
-    descendant instanceof Element ? (descendant as Element) : null
-  return ancestor?.contains(descendantElement)
-}
--- a/packages/web/src/components/track/helpers.ts
+++ b/packages/web/src/components/track/helpers.ts
@@ -1,4 +1,10 @@
-import { Collection, FieldVisibility, Track, User } from '@audius/common'
+import {
+  Nullable,
+  Collection,
+  FieldVisibility,
+  Track,
+  User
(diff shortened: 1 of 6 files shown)


sddioulde:
have to do it for the playlist tile too

audius-infra:
Preview this change https://demo.audius.co/sd-multiple-menus

sddioulde:
done for the playlist tiles too.

feels like it can be refactored.

audius-infra:
Preview this change https://demo.audius.co/sd-multiple-menus

audius-infra:
Preview this change https://demo.audius.co/sd-multiple-menus

packages/web/src/components/track/helpers.ts
```diff
@@ -79,3 +85,12 @@ export const getUserWithFallback = (user: User | null) => {
     }
   )
 }
+
+export const isDescendantElementOf = (
```
raymondjacobson, line 89: https://github.com/AudiusProject/audius-client/blob/e42bf86d7e6ed496e0bb1ffbd03743c75083dc40/packages/web/src/utils/domUtils.ts#L1

mind putting this here or trying to re-use?

raymondjacobson (approved):
bugfix lgtm, but definitely this stuff is hairy

audius-infra:
Preview this change https://demo.audius.co/sd-multiple-menus

PR #1704, opened 2022-08-10
[C-773] Prevent showing file explorer on pressing enter

sddioulde:
### Description

Prevent showing file explorer on pressing enter and file open button not on focus.
We still have the ability to not only show the explorer on button click, but also on pressing enter/space as long as the button is on focus.

### Dragons

n/a

### How Has This Been Tested?

local against stage

n/a

### Feature Flags ###

none


audius-infra:
Preview this change https://demo.audius.co/sd-signup-enter

raymondjacobson (approved):

PR #1706, opened 2022-08-10
[C-801] Add tag search text to mobile

dylanjeffers:
### Description

- Adds the tag text to tagged search in mobile.
- Adds `Tag` component, and uses it track screen and tag search

image


raymondjacobson (approved):
LGTM, but i suspect this will change post design b/c of baseline alignment & also maybe want to make it look more like web's version

TY for doing this!!

PR #1708, opened 2022-08-10
[C-808] Don't show unnecessarily show profile option on overflow menu

dylanjeffers:
### Description

Fixes case where artist's track's overflow menu shows option to go to artist page, which doesn't make sense since we are already on the artist page

### Dragons

Definitely a bit hacky.

### DIFF
--- a/packages/mobile/src/components/lineup-tile/TrackTile.tsx
+++ b/packages/mobile/src/components/lineup-tile/TrackTile.tsx
@@ -93,6 +93,10 @@ const TrackTileComponent = ({
     track_id
   } = track
 
+  const currentScreen = navigation.getState().history?.[0]
+  // @ts-expect-error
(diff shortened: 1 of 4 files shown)


packages/mobile/src/components/lineup-tile/TrackTile.tsx
```diff
@@ -93,6 +93,10 @@ const TrackTileComponent = ({
     track_id
   } = track
 
+  const currentScreen = navigation.getState().history?.[0]
+  // @ts-expect-error
```
raymondjacobson, line 97: how come?

raymondjacobson (approved):

packages/mobile/src/components/lineup-tile/TrackTile.tsx
dylanjeffers: weirdly, history returns undefined[]... what do you think?

PR #1711, opened 2022-08-12
[C-820] Move apiClient into saga context, Add WalletClient to context

dylanjeffers:
### Description

- Moves apiClient to saga context
- Adds walletClient to context


packages/web/src/pages/following-page/sagas.ts
```diff
@@ -21,12 +20,8 @@ const provider = createUserListProvider<User>({
     limit,
     offset,
     entityId,
-    currentUserId
```
nicoback2, line 24: was the type here deleted intentionally?

packages/mobile/src/store/storeContext.ts
```diff
@@ -5,12 +5,14 @@ import { apiClient } from 'app/services/audius-api-client'
...
 import { fingerprintClient } from 'app/services/fingerprint'
 import { remoteConfigInstance } from 'app/services/remote-config'
+import { walletClient } from 'app/services/wallet-client'
 
 export const storeContext: CommonStoreContext = {
   getLocalStorageItem: async (key) => AsyncStorage.getItem(key),
   setLocalStorageItem: async (key, value) => AsyncStorage.setItem(key, value),
   remoteConfigInstance,
   audiusBackendInstance,
   apiClient,
```
nicoback2, line 15: just curious, why are the first two suffixed with Instance and the rest isn't?

nicoback2 (approved):

packages/mobile/src/store/storeContext.ts
sliptype: We started following that pattern with remoteConfigInstance, but I would be down to remove `Instance` or add it to the other ones. Either way

dylanjeffers: yeah its a good call, im just going off existing naming conventions. in the future would be down to rename all this to something more consistent, do you have thoughts? maybe remove the "instance" since that's just a lot of extra chars?

sliptype (approved):

packages/mobile/src/store/storeContext.ts
dylanjeffers: oh clutch ahah we were typing the sameee thing <3

PR #1717, opened 2022-08-15
[C-815] Fix electron update

sliptype:
### Description

* Use `app.getPath('appData')` instead of `app.getAppPath` to store/retrieve the update builds because `app.getAppPath` can't apparently be modified when the app is built using electron-build (it's packaged into a single file)
* On initial build registration, conditionally register the build located in the app data directory (if an update has been previously applied)
* Update logic of replacing builds on update because they are now stored in a different location
> For some reason the `packages/web` path was not being accounted for, did something change the structure of the s3 bucket?
* Add `appDataPath` helper

### Dragons

I would like to test this on windows. Does anyone have a windows machine to test on?

Also need prevent auto updater from downloading full update if only a patch version is required (still think this is happening), will address in separate pr

### How Has This Been Tested?

Tested via `npm run electron:stage` and `npm run dist:stage`

### Feature Flags ###

### DIFF
--- a/packages/web/src/electron.js
+++ b/packages/web/src/electron.js
@@ -31,6 +31,10 @@ const Environment = Object.freeze({
 const args = process.argv.slice(2)
 const appVersion = require('../package.json').version
 
+const appDataPath = (relativePath) => {
+  return path.resolve(app.getPath('appData'), app.getName(), relativePath)
+}
+
 let appEnvironment,
   buildName,
   localhostPort,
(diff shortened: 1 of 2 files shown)


packages/web/src/electron.js
```diff
@@ -431,34 +436,25 @@ ipcMain.on('update', async (event, arg) => {
 })
 
 ipcMain.on('web-update', async (event, arg) => {
-  if (
-    fs.existsSync(path.resolve(app.getAppPath(), `web-update/${buildName}`))
-  ) {
+  const buildPath = appDataPath(`web-update/packages/web/${buildName}`)
```
sliptype, line 439: Not sure why `packages/web` wasn't needed before?

audius-infra:
Preview this change https://demo.audius.co/sk-fix-electron-update

packages/web/src/electron.js
Kyle-Shanks: 🤯

Kyle-Shanks (approved):

raymondjacobson (approved):

PR #1718, opened 2022-08-16
[C-832] Add local-storage service to saga context

dylanjeffers:
### Description

Adds local-storage service to saga context.
Updates getAccount usages to wait until account is loaded (since it is now async) before requesting it

### Dragons

There may be subtle places where we are meaning to request data with an authenticated user, but instead are doing the public fetch. There may be other places that are expecting an account right away even though it is still loading. I believe I have covered the majority of places 

### How Has This Been Tested?

Running staging and prod and going through the major flows.

Looking for any issues in sentry that result from this change, as well as responding to any internal feedback of loading issues.


packages/web/src/common/store/backend/sagas.ts
```diff
@@ -78,7 +78,7 @@ export function* setupBackend() {
   const audiusBackendInstance = yield* getContext('audiusBackendInstance')
 
   // Init APICLient
-  yield* call(() => apiClient.init())
+  apiClient.init()
```
sliptype, line 81: Why remove the yield here when we made `init` async?

packages/web/src/common/services/local-storage/LocalStorage.ts
```diff
@@ -0,0 +1,76 @@
...
+  getItem: (key: string) => Promise<string | null> | string | null
+  setItem: (key: string, value: string) => Promise<void> | void
+  removeItem: (key: string) => Promise<void> | void
+}
+
+type LocalStorageConfig = {
+  localStorage: LocalStorageType
+}
+
+export class LocalStorage {
```
sliptype, line 21: Did you make this a class instead of a function for consistency with the other services?

packages/web/src/utils/sagaHelpers.js
```diff
@@ -104,6 +105,15 @@ export function* doEvery(millis, fn, times = null) {
   return chan
 }
 
+export function* waitForAccount() {
+  yield call(
+    waitForValue,
+    (state) => state.account.status,
+    null,
+    (status) => status !== Status.LOADING
```
sliptype, line 113: Nice

sliptype (approved):
Looks good to me!

packages/web/src/common/store/backend/sagas.ts
dylanjeffers: oh yeah i can bring that back, i actually didnt make it async since it was messing up a lot of places and for very little upside, which is just to init the eager discovery provider, which is only a web thing anyway.

packages/web/src/common/services/local-storage/LocalStorage.ts
dylanjeffers: yeah i guess so, happy to make a function if we think it makes more sense though.

packages/web/src/utils/sagaHelpers.js
dylanjeffers: yeah this felt good, thank the lord, lol

audius-infra:
Preview this change https://demo.audius.co/dj-c-832-make-local-storage-service-native-compatible

PR #1719, opened 2022-08-16
[C-838] Fix desktop update banner flow

raymondjacobson:
### Description

Redid this PR from scratch. @sliptype / @Kyle-Shanks could use extra scrutiny here.

I think there are still two issues with the desktop update flow that prevent "web app" updates from working:

1. we trigger an electron banner update and a web banner update at the same time
2. we are using some state variable `showWebUpdate` to determine whether to show the web update banner, and i'm pretty sure that never gets set to true (it's `showWebUpdateBanner`), so `#1` isn't actually an issue we observe

These changes are set up in a way such though that when we download the electron changes in the background, on restarting, the user will still get the full electron update.

### Dragons

### How Has This Been Tested?

Will test after merge to main via the staging desktop app

### Feature Flags ###

### DIFF
--- a/packages/web/src/pages/App.js
+++ b/packages/web/src/pages/App.js
@@ -209,10 +209,13 @@ class App extends Component {
     showCTABanner: false,
     showWeb3ErrorBanner: null,
 
-    showUpdateAppBanner: false,
+    // A patch version update of the web app is available
     showWebUpdateBanner: false,
-    showRequiresUpdate: false,
+    // A version update of the web app is required
     showRequiresWebUpdate: false,
+    // A minor version update of the entire electron app is required
+    showRequiresUpdate: false,
+
     isUpdating: false,
 
     initialPage: true,
@@ -255,7 +258,6 @@ class App extends Component {
       // We downloaded an update, the user can safely restart
       this.ipc.on('updateDownloaded', (event, arg) => {
         console.info('updateDownload', event, arg)
-        this.setState({ showUpdateAppBanner: true })
       })
 
       this.ipc.on('updateDownloadProgress', (event, arg) => {
@@ -437,9 +439,6 @@ class App extends Component {
   }
 
   acceptUpdateApp = () => {
-    if (this.state.showUpdateAppBanner) {
-      this.dismissUpdateAppBanner()
-    }
     this.setState({ isUpdating: true })
     this.ipc.send('update')
   }
@@ -452,10 +451,6 @@ class App extends Component {
     }
     this.setState({ isUpdating: true })
     this.ipc.send('web-update')
-  }
-
-  dismissUpdateAppBanner = () => {
-    this.setState({ showUpdateAppBanner: false })
   }
 
   dismissUpdateWebAppBanner = () => {
@@ -487,8 +482,7 @@ class App extends Component {
 
     const {
       showCTABanner,
-      showUpdateAppBanner,
-      showWebUpdate,
+      showWebUpdateBanner,
       showWeb3ErrorBanner,
       isUpdating,
       showRequiresUpdate,
@@ -517,7 +511,7 @@ class App extends Component {
       )
 
     const showBanner =
-      showCTABanner || showUpdateAppBanner || showWeb3ErrorBanner
+      showCTABanner || showWeb3ErrorBanner || showWebUpdateBanner
     if (this.headerGutterRef.current) {
       if (showBanner) {
         this.headerGutterRef.current.classList.add(styles.bannerMargin)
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://demo.audius.co/rj-c-838

sliptype (approved):
Kinda confused on how this is working. Doesn't this conditional conflict with the one we have in the handler? https://github.com/AudiusProject/audius-client/blob/main/packages/web/src/pages/App.js#L287

Are we losing the forced major version updates with this?

raymondjacobson:
> Kinda confused on how this is working. Doesn't this conditional conflict with the one we have in the handler? https://github.com/AudiusProject/audius-client/blob/main/packages/web/src/pages/App.js#L287
> 
> Are we losing the forced major version updates with this?

Based on the code in this change--

If we are running 1.0.1 and 1.0.2 exists, electron will recognize an update, but because major and minor are the same, will not send an `updateAvailable` message to web. When quitting & restarting the app, the update will be installed. The separate webUpdateAvailable code will run though and trigger a patch version refresh update.

If we are running 1.0.1 and 1.1.0 exists, electron will recognize an update, and because minor version is different, it will send an `updateAvailable` message to web. Web will then see the `updateAvailable` and see `semver.minor(currentVersion) < semver.minor(version)` and pop up the message that forces an update.

sliptype:
> > Kinda confused on how this is working. Doesn't this conditional conflict with the one we have in the handler? https://github.com/AudiusProject/audius-client/blob/main/packages/web/src/pages/App.js#L287
> > Are we losing the forced major version updates with this?
> 
> Based on the code in this change--
> 
> If we are running 1.0.1 and 1.0.2 exists, electron will recognize an update, but because major and minor are the same, will not send an `updateAvailable` message to web. When quitting & restarting the app, the update will be installed. The separate webUpdateAvailable code will run though and trigger a patch version refresh update.
> 
> If we are running 1.0.1 and 1.1.0 exists, electron will recognize an update, and because minor version is different, it will send an `updateAvailable` message to web. Web will then see the `updateAvailable` and see `semver.minor(currentVersion) < semver.minor(version)` and pop up the message that forces an update.

But doesn't the handler in App.tsx already only show the banner if it's a minor or major version update? Won't auto-updater still download the patch version in the background anyway? Maybe I'm still not quite understanding, lets chat in our 1 on 1

raymondjacobson:
> > > Kinda confused on how this is working. Doesn't this conditional conflict with the one we have in the handler? https://github.com/AudiusProject/audius-client/blob/main/packages/web/src/pages/App.js#L287
> > > Are we losing the forced major version updates with this?
> > 
> > 
> > Based on the code in this change--
> > If we are running 1.0.1 and 1.0.2 exists, electron will recognize an update, but because major and minor are the same, will not send an `updateAvailable` message to web. When quitting & restarting the app, the update will be installed. The separate webUpdateAvailable code will run though and trigger a patch version refresh update.
> > If we are running 1.0.1 and 1.1.0 exists, electron will recognize an update, and because minor version is different, it will send an `updateAvailable` message to web. Web will then see the `updateAvailable` and see `semver.minor(currentVersion) < semver.minor(version)` and pop up the message that forces an update.
> 
> But doesn't the handler in App.tsx already only show the banner if it's a minor or major version update? Won't auto-updater still download the patch version in the background anyway? Maybe I'm still not quite understanding, lets chat in our 1 on 1

Redid this PR

audius-infra:
Preview this change https://demo.audius.co/rj-c-838

raymondjacobson:
> This mostly makes sense but we are still setting `showWebUpdate` to true:
> 
> https://github.com/AudiusProject/audius-client/blob/39fb657d7873c221fb3cef11887f9b8909230829/packages/web/src/pages/App.js#L283

Yeah great catch. That's how it was working before

audius-infra:
Preview this change https://demo.audius.co/rj-c-838

Kyle-Shanks (approved):

packages/web/src/pages/App.js
```diff
@@ -437,9 +439,6 @@ class App extends Component {
   }
 
   acceptUpdateApp = () => {
-    if (this.state.showUpdateAppBanner) {
-      this.dismissUpdateAppBanner()
-    }
```
amendelsohn, line 442: Is this no longer necessary because the app is going to restart on update and not show the banner after?

amendelsohn (approved):

packages/web/src/pages/App.js
raymondjacobson: Basically there are two flows for upgrading desktop:
* An update to entire electron app (you need a new chromium basically)
* An update to just the static JS for the SPA
KJ added the latter recently, which lets us basically not worry about showing a banner for the former and letting auto update on quit take care of that behind the scenes. There's still banner that shows that will get dismissed in the below method (acceptWebUpdate) but that's for the latter SPA refresh

PR #1720, opened 2022-08-16
[PAY-431] Slide in supporters after loading

dharit-tan:
### Description
Add slide-in animation for supporters on user profile page after supporters load. Mobile-only.

https://user-images.githubusercontent.com/3893871/184787256-fdcbbc9f-571a-447a-a585-d7cfcb71a5be.mp4

### How Has This Been Tested?

iOS simulator
Tested on Android hardware


dylanjeffers (approved):
awesome work! cool to know we went with LayoutAnimation, just smoother overall basically?

dharit-tan:
> awesome work! cool to know we went with LayoutAnimation, just smoother overall basically?

Yup way smoother and paired with the boolean ref it only renders once.

packages/mobile/src/screens/profile-screen/ProfileHeader/TopSupporters.tsx
```diff
@@ -97,6 +98,14 @@ export const TopSupporters = () => {
     })
   }, [navigation, user_id])
 
+  useLayoutEffect(() => {
+    // Prevent multiple re-renders from rankedSupporters changing
+    if (rankedSupporters.length > 0 && !hasRankedSupportersLoaded.current) {
```
piazzatron, line 103: nitty nit: could just do `rankedSupporters.length` instead of `rankedSupporters.length > 0`

packages/mobile/src/screens/profile-screen/ProfileHeader/TopSupporters.tsx
```diff
@@ -97,6 +98,14 @@ export const TopSupporters = () => {
     })
   }, [navigation, user_id])
 
+  useLayoutEffect(() => {
```
piazzatron, line 101: Let's add a comment here explaining this is for controlling a loading animation!

piazzatron (approved):
This animation looks *great*, and super clean solution!!

If you wanted, could take it to the next level by pulling out a hook that combines both the ref and the useLayoutEffect (`useLoadingAnimation`) or something. Even though it wouldn't be used across components, it can be nice to do this bc the function call name makes it clear what's going on in the component w/o adding comments ("Oh I see, we're doing loading animation stuff now"), *and* it colocates the useRef with the useLayoutEffect so it's clear they're related

dharit-tan:
Removing the LayoutAnimation flag for android here because we're putting it in App.tsx: https://github.com/AudiusProject/audius-client/pull/1754

packages/mobile/src/hooks/useLoadingAnimation.ts
```diff
@@ -0,0 +1,17 @@
+import { useRef, useLayoutEffect } from 'react'
```
piazzatron, line 1: For hooks that we're unlikely to reuse, best to just include it in the single component that uses it imo rather than a separate file!

piazzatron (approved):

PR #1721, opened 2022-08-16
[Pay-516] Show see more only if profile header should be expansible on mobile.

sddioulde:
### Description

conditions to show see more on mobile profile:
- has a badge tier
- has mutuals
- is supporting (has tipped) other users
- has socials
- has long bio description

### Dragons

nah

### How Has This Been Tested?

local mobile against local dapp against stage

n/a

### Feature Flags ###

none

### DIFF
--- a/packages/mobile/src/screens/profile-screen/ProfileHeader/Bio.tsx
+++ b/packages/mobile/src/screens/profile-screen/ProfileHeader/Bio.tsx
@@ -20,8 +20,8 @@ const useStyles = makeStyles(({ typography, palette, spacing }) => ({
 }))
 
 type BioProps = TextProps & {
-  isExpansible: boolean
-  setIsExpansible: (isExpansible: boolean) => void
+  isExpansible?: boolean
+  setIsExpansible?: (isExpansible: boolean) => void
 }
 
 export const Bio = (props: BioProps) => {
@@ -37,7 +37,7 @@ export const Bio = (props: BioProps) => {
       <View pointerEvents='none'>
         <Text
           onTextLayout={(e) => {
-            if (e.nativeEvent.lines.length > MAX_BIO_LINES) {
+            if (setIsExpansible && e.nativeEvent.lines.length > MAX_BIO_LINES) {
               setIsExpansible(true)
             }
           }}
--- a/packages/mobile/src/screens/profile-screen/ProfileHeader/Bio.tsx
+++ b/packages/mobile/src/screens/profile-screen/ProfileHeader/Bio.tsx
@@ -7,6 +7,8 @@ import { makeStyles } from 'app/styles'
 import { useSelectProfile } from '../selectors'
 import { squashNewLines } from '../utils'
 
+const MAX_BIO_LINES = 2
+
 const useStyles = makeStyles(({ typography, palette, spacing }) => ({
   root: {
     marginBottom: spacing(2)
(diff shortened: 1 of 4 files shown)


packages/mobile/src/screens/profile-screen/ProfileHeader/Bio.tsx
```diff
@@ -17,20 +17,44 @@ const useStyles = makeStyles(({ typography, palette, spacing }) => ({
...
+      }
+    : {}
+
   if (numberOfLines)
     return (
       <View pointerEvents='none'>
-        <Text variant='body' style={styles.root} {...props}>
+        <Text
+          onTextLayout={(e) => {
+            if (e.nativeEvent.lines.length > 2) {
```
sddioulde, line 50: i'll make 2 a cosntant

packages/mobile/src/screens/profile-screen/ProfileHeader/ProfileHeader.tsx
```diff
@@ -46,11 +47,80 @@ type ProfileHeaderProps = {
...
+  ])
+  const hasMultipleSocials =
+    [website, donation, twitterHandle, instagramHandle, tikTokHandle].filter(
+      Boolean
+    ).length > 1
+  const isSupporting = supportingCount > 0
   const [isExpanded, setIsExpanded] = useToggle(false)
+  const [isExpansible, setIsExpansible] = useState(false)
+
+  console.log(
```
sddioulde, line 85: will remove

sddioulde:
wondering if we should consolidate the uses of useSelectProfileRoot and useSelectProfile in the ProfileHeader component...

packages/mobile/src/screens/profile-screen/ProfileHeader/ProfileHeader.tsx
```diff
@@ -46,11 +47,66 @@ type ProfileHeaderProps = {
...
+  ])
+  const hasMultipleSocials =
+    [website, donation, twitterHandle, instagramHandle, tikTokHandle].filter(
+      Boolean
+    ).length > 1
+  const isSupporting = supportingCount > 0
   const [isExpanded, setIsExpanded] = useToggle(false)
+  const [isExpansible, setIsExpansible] = useState(false)
+
+  /**
```
dylanjeffers, line 85: I like the comment, but wondering if the code is clear enough? like hasTier || hasMutuals || has.... seems quite nice to me!

dylanjeffers: maybe create a "shouldExpand" const before the useEffect, and just pass that into dep array and check in if statement?

packages/mobile/src/screens/profile-screen/ProfileHeader/Bio.tsx
```diff
@@ -17,20 +19,44 @@ const useStyles = makeStyles(({ typography, palette, spacing }) => ({
...
+      }
+    : {}
+
   if (numberOfLines)
     return (
       <View pointerEvents='none'>
-        <Text variant='body' style={styles.root} {...props}>
+        <Text
+          onTextLayout={(e) => {
+            if (e.nativeEvent.lines.length > MAX_BIO_LINES) {
```
dylanjeffers, line 52: oh that is sooo clutch that gives you lines length, wow great find

packages/mobile/src/screens/profile-screen/ProfileHeader/Bio.tsx
```diff
@@ -17,20 +19,44 @@ const useStyles = makeStyles(({ typography, palette, spacing }) => ({
...
+    ? {
+        ...props,
+        // only set number of lines after we determine that the text should be truncated
+        // this allows us to let the parent component know whether we have met one of
+        // the conditions to make the bio section expansible.
+        numberOfLines: isExpansible ? numberOfLines : undefined,
+        isExpansible: undefined, // Bio prop but not Text prop
+        setIsExpansible: undefined // Bio prop but not Text prop
+      }
+    : {}
```
dylanjeffers, line 45: what about something like
```tsx
const { numberOfLines, isExpansible, setIsExpansible, ...other} = props
....
<Text 
  onTextLayout={...}
  {...other}
  numberOfLines={isExpansible && numberOfLines ? numberOfLines : undefined} 
/>
```

packages/mobile/src/screens/profile-screen/ProfileHeader/ProfileHeader.tsx
dylanjeffers: one last thing, definitely not something we need to do, but could create an `<ExpandableHeaderDetails />` component that handles all the logic written here, which will nicely encapsulate this complex logic away from the root component, which can just focus on getting the core components rendering. its also a slight perf boost since rerenders related to that section don't trigger full rerender on the whole header

dylanjeffers (approved):
Awesome work on this, would be down to work through some of the suggestions, so lmk :)

dylanjeffers:
> wondering if we should consolidate the uses of useSelectProfileRoot and useSelectProfile in the ProfileHeader component...

The core idea there is that one of them returns nullable profile, and the other assumes it's there, which is the case when the root does actually render, maybe another reason to create an `ExpandibleHeaderDetails` component?

packages/mobile/src/screens/profile-screen/ProfileHeader/Bio.tsx
sddioulde: much better!!

packages/mobile/src/screens/profile-screen/ProfileHeader/ProfileHeader.tsx
sddioulde: yeah considered the shouldExpand var and went back and forth with myself. but yeah overall agree with the abovesss and will address

PR #1724, opened 2022-08-16
[PAY-536] Short-circuits reward claim retries on AAO error

dharit-tan:
### Description
If an AAO error code is encountered, then the user must have been rejected by the AAO and we do not need to retry the rewards claim.

### Dragons

NA

### How Has This Been Tested?

Tested on local client against staging.


audius-infra:
Preview this change https://demo.audius.co/rt-PAY-536

sddioulde (approved):

PR #1726, opened 2022-08-16
Fix wallet-client usages

dylanjeffers:
### Description

Fixes wallet-client usages, using redux-saga context arrays


sliptype (approved):
Why do we have to use the `[context, fn]` syntax for walletClient but not for audiusBackendInstance for example?

dylanjeffers:
> Why do we have to use the `[context, fn]` syntax for walletClient but not for audiusBackendInstance for example?

class vs function returning object. maybe another case for using functions like you said :)

audius-infra:
Preview this change https://demo.audius.co/dj-fix-wallet-client-usages

PR #1727, opened 2022-08-16
[C-841] Put getFeatureEnabled in storeContext

sliptype:
### Description

Moves `getFeatureEnabled` to storeContext in preparation for notification sagas

### Dragons

### How Has This Been Tested?

### Feature Flags ###


packages/mobile/src/store/storeContext.ts
```diff
@@ -4,12 +4,16 @@ import type { CommonStoreContext } from 'audius-client/src/common/store'
...
+import {
+  getFeatureEnabled,
+  remoteConfigInstance
+} from 'app/services/remote-config'
 import { walletClient } from 'app/services/wallet-client'
 
 export const storeContext: CommonStoreContext = {
   getLocalStorageItem: async (key) => AsyncStorage.getItem(key),
   setLocalStorageItem: async (key, value) => AsyncStorage.setItem(key, value),
+  getFeatureEnabled,
```
dylanjeffers, line 16: so can we use remoteConfigInstance to call getFeatureEnabled, or are they separate things?

packages/web/src/components/notification/store/sagas.ts
```diff
@@ -663,10 +665,6 @@ export function* markAllNotificationsViewed() {
   const audiusBackendInstance = yield* getContext('audiusBackendInstance')
   yield* call(waitForBackendSetup)
   yield* call(audiusBackendInstance.markAllNotificationAsViewed)
-  if (NATIVE_MOBILE) {
```
dylanjeffers, line 666: cool this is going away now! is it safe to do in this intermediate phase?

dylanjeffers (approved):
looks great, isn't there also the `getFeatureEnabled` in the audius-backend-instance? we using the same one there and here? Also just wanted to double check regarding the NATIVE_MOBILE removals!

packages/web/src/components/notification/store/sagas.ts
sliptype: No this shouldn't have been part of this commit, my b

packages/mobile/src/store/storeContext.ts
sliptype: They are separate things right now. The `getFeatureFlag` helper checks local storage for the feature flag overrides, but I think I can consolidate it

sliptype: Actually we do need both, the one in remoteConfig is used by the feature flag modal to show the default state. The helper is used by actual features to check if they're enabled accounting for overrides. I could return the one from audiusBackendInstance out and update the usages to use that one, but it seems like it makes more sense at the top level of the storeContext. wdyt?

dylanjeffers: yeah im good with top level, but was just wondering why it seemed like we had duplicate things. maybe when everything is sorted, we take a loot at context and see if we can consolidate? things like getLocalStorageItem vs localStorage, is another example

audius-infra:
Preview this change https://demo.audius.co/sk-c-841-getfeatureenabled-storecontext

PR #1729, opened 2022-08-16
[PAY-497] Fix native mobile install challenge

piazzatron:
### Description
There is a bug where if you sign up on web and then load your account into mobile, we don't mark the mobile install challenge as claimable because your is_mobile user_events haven't propagated to DN yet. So optimistically mark this challenge as claimable if mobile.

I was a bit on the fence about where to put this check - at first I was going to put it in userChallengeOverrides in the slice, which required adding a new action, and then felt that dispatching an action was really indirect compared to just putting this check in the obvious place, even though it makes this code more tightly coupled to a specific challenge now. 

### Dragons

None

### How Has This Been Tested?

I haven't tested this yet :) 

### Feature Flags ###


rickyrombo (approved):

audius-infra:
Preview this change https://demo.audius.co/piazz-fix-claim-mobile-challenge

PR #1735, opened 2022-08-17
[PAY-531] Fix become first supporter bug

sddioulde:
### Description

This bug is new because it's now possible to tip a user without going to their profile and without clicking from the feed tip tile. The bug is because the receiver's supporters weren't fetched and so the UI thinks there is no supporter for the user.
So the fix is to fetch that user i.e. the receiver's supporters.

=====

I think this change will make sure that the account and receiver's necessary supporting / supporters exist, so if there is another UI flow that goes into tipping in the future, we'd still be okay.

https://user-images.githubusercontent.com/9600175/185279845-c84e3b50-af10-4df7-9660-8913f2be4ce0.mov

### Dragons

### How Has This Been Tested?

mobile against local dapp pointing to prod

n/a

### Feature Flags ###

none

### DIFF
--- a/packages/mobile/src/screens/tip-artist-screen/SendTipScreen.tsx
+++ b/packages/mobile/src/screens/tip-artist-screen/SendTipScreen.tsx
@@ -10,7 +10,8 @@ import {
 } from 'audius-client/src/common/store/tipping/selectors'
 import {
   sendTip,
-  fetchUserSupporter
+  fetchUserSupporter,
+  refreshSupport
 } from 'audius-client/src/common/store/tipping/slice'
 import { getAccountBalance } from 'audius-client/src/common/store/wallet/selectors'
 import { getBalance } from 'audius-client/src/common/store/wallet/slice'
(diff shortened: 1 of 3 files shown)


sddioulde:
What i really want, is dispatch the action within the hook so that consuming components dont have to all be touched. for now only the SendTipScreen component cares, but if we decided to directly open the tip modal for web, then the SendTip component would also need to be updated.
Want to keep trying but will timebox as this fix should be merged soon.

packages/mobile/src/screens/tip-artist-screen/SendTipScreen.tsx
```diff
@@ -90,6 +92,17 @@ export const SendTipScreen = () => {
     }
   }, [shouldFetchUserSupporter, account, receiver, dispatchWeb])
 
+  useEffect(() => {
+    if (shouldFetchSupportersForReceiver && account && receiver) {
+      dispatchWeb(
```
sddioulde, line 97: What i really want, is dispatch the action within the hook so that consuming components dont have to all be touched. for now only the SendTipScreen component cares, but if we decided to directly open the tip modal for web, then the SendTip component would also need to be updated.
Want to keep trying but will timebox as this fix should be merged soon.

audius-infra:
Preview this change https://demo.audius.co/sd-pay-531

dharit-tan (approved):

PR #1736, opened 2022-08-18
[C-856] Use solanaweb3 in native, improve account flow

dylanjeffers:
### Description

- Adds SolanaWeb3Manager to native, preventing the missing `getUserBank` call
- Adds react-native-get-random-values, needed by @solana/web3.js native
- Updates other account saga utilities to work in native. The flow now ends with an actual user in the account state


packages/mobile/index.js
```diff
@@ -4,6 +4,7 @@ import { AppRegistry, LogBox } from 'react-native'
 import { name as appName } from './app.json'
 
 require('node-libs-react-native/globals')
+require('react-native-get-random-values')
```
raymondjacobson, line 7: maybe comment on why this is necessary?

raymondjacobson (approved):

audius-infra:
Preview this change https://demo.audius.co/dj-c-856-use-solanaweb3-and-improve-account-flow

PR #1737, opened 2022-08-18
Fix eager discovery loader

raymondjacobson:
### Description

Eager loader is returning early if no cached value. Instead we should pull from the env vars set in the .env file.

Observed behavior is when you load localhost:3001/df/playlist/probers_playlist_do_not_delete-511 in incognito for the first time, it would crash b/c we can't initialize API client w/o some discovery node.

### Dragons

### How Has This Been Tested?

Locally w/ probers

### Feature Flags ###


dylanjeffers (approved):
sweet so basically try cached, if not get from env?

raymondjacobson:
> sweet so basically try cached, if not get from env?

Yep!

audius-infra:
Preview this change https://demo.audius.co/rj-fix-eager-loader

PR #1739, opened 2022-08-18
Get sign in/out working end to end

raymondjacobson:
### Description

Ended up struggling to test sign up stuff without sign out, so I went ahead in that direction. Hopefully didn't steal too much.

Curious for thoughts here. I believe this is "shippable" with the sagas on mobile enabled at this point. That might save us some pain of having everything be off & then on while we're working.

This is all a good learning exercise for me. I'm starting to get the feeling that we're doing an "in then back out" approach to things.
1. Mirror dispatch & fetch logic in core stuff like account/sign in (double dispatch/select)
2. Move all the "inner pages" onto the local redux store instead of common
3. Go back to the core stuff and remove the web dispatch/select stuff
Maybe this was already obvious to folks...

https://user-images.githubusercontent.com/2731362/185336740-0e9de7ab-45b7-446a-80d5-3b27745bad68.mp4

### Dragons

### How Has This Been Tested?

Local simulator

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/sign-out-confirmation-drawer/SignOutConfirmationDrawer.tsx
+++ b/packages/mobile/src/components/sign-out-confirmation-drawer/SignOutConfirmationDrawer.tsx
@@ -1,7 +1,8 @@
 import { useCallback } from 'react'
 
-import { signOut } from 'audius-client/src/common/store/sign-out/slice'
+import { signOut } from 'common/store/sign-out/slice'
 import { View } from 'react-native'
+import { useDispatch } from 'react-redux'
 
 import { Button, Text } from 'app/components/core'
 import { AppDrawer, useDrawerState } from 'app/components/drawer/AppDrawer'
@@ -37,16 +38,18 @@ const useStyles = makeStyles(({ spacing }) => ({
 export const SignOutConfirmationDrawer = () => {
   const styles = useStyles()
   const dispatchWeb = useDispatchWeb()
+  const dispatch = useDispatch()
   const { clearHistory } = useSearchHistory()
 
   const { onClose } = useDrawerState(MODAL_NAME)
 
   const handleSignOut = useCallback(() => {
-    dispatchWeb(signOut)
+    dispatch(signOut({}))
     // TODO: move to the sign-out saga when store migrated to react-native
+    dispatchWeb(signOut)
     clearHistory()
     onClose()
-  }, [dispatchWeb, clearHistory, onClose])
+  }, [dispatchWeb, dispatch, clearHistory, onClose])
 
   return (
     <AppDrawer modalName={MODAL_NAME} title={messages.drawerTitle}>
--- a/packages/mobile/src/screens/signon/SignOn.tsx
+++ b/packages/mobile/src/screens/signon/SignOn.tsx
@@ -649,7 +649,7 @@ const SignOn = ({ navigation }: SignOnProps) => {
                   // in case email is what was wrong with the credentials
                   setShowDefaultError(true)
                 }
-              } else if (emailIsAvailable && emailStatus === 'done') {
+              } else if (emailIsAvailable && emailStatus === 'success') {
                 dispatch(signOnActionsLegacy.signinFailedReset())
                 setIsWorking(false)
                 navigation.replace('CreatePassword', { email })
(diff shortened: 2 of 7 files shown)


audius-infra:
Preview this change https://demo.audius.co/rj-signinout

audius-infra:
Preview this change https://demo.audius.co/rj-signinout

packages/mobile/src/store/store.ts
```diff
@@ -67,7 +67,7 @@ const createRootReducer = () =>
 
 const sagaMiddleware = createSagaMiddleware({ context: storeContext })
 const middlewares = applyMiddleware(sagaMiddleware)
-const composeEnhancers = composeWithDevTools({ trace: true, traceLimit: 25 })
+const composeEnhancers = composeWithDevTools({ trace: true, traceLimit: 250 })
```
sliptype, line 70: Yes

packages/mobile/src/screens/signon/SignOn.tsx
```diff
@@ -336,14 +336,26 @@ const SignOn = ({ navigation }: SignOnProps) => {
...
+  const passwordFieldValue: { error: string } = useSelector(getPasswordField)
+  const emailFieldValue: {
+    error: '' | 'inUse' | 'characters'
+    status: 'success' | 'failure'
+  } = useSelector(getEmailField)
+  const accountUser = useSelector(getAccountUser)
+
+  const signedIn = signOnStatus === 'success'
+  const isSigninError = passwordFieldValue.error
+  const emailIsAvailable = emailFieldValue.error !== 'inUse'
```
sliptype, line 355: Is there a reason we aren't using selectors for these anymore? If so can we delete the selectors?

packages/mobile/src/store/sagas.ts
```diff
@@ -17,16 +18,19 @@ export default function* rootSaga() {
   const sagas = [
     // config
     ...backendSagas(),
-    // ...accountSagas(),
+    ...accountSagas(),
```
sliptype, line 21: Did you notice any worse perf with everything enabled like this?

sliptype (approved):
Looks good! your thinking makes sense, we are gonna have to have the "mirroring" until we drop the webview entirely and then can get rid of it

Also I'm on board with keeping the sagas enabled, but let's keep an eye on perf

packages/web/src/common/store/sign-out/sagas.ts
```diff
@@ -15,6 +16,7 @@ function* watchSignOut() {
   const audiusBackendInstance = yield* getContext('audiusBackendInstance')
   const localStorage = yield* getContext('localStorage')
   yield takeLatest(signOutAction.type, function* () {
+    yield put(resetAccount())
```
dylanjeffers, line 19: was this just a bug you were seeing overall?

dylanjeffers (approved):
awesome changes man, this is happening

packages/mobile/src/screens/signon/SignOn.tsx
raymondjacobson: these are derived values from the result of the selector

packages/mobile/src/store/sagas.ts
raymondjacobson: Across the app no. There's a slight hiccup when pressing the sign in button, but i think fix-forward makes sense for it. Can disable if it's a bad call once it's on the staging app

packages/mobile/src/components/sign-out-confirmation-drawer/SignOutConfirmationDrawer.tsx
```diff
@@ -37,16 +38,18 @@ const useStyles = makeStyles(({ spacing }) => ({
...
   const styles = useStyles()
   const dispatchWeb = useDispatchWeb()
+  const dispatch = useDispatch()
   const { clearHistory } = useSearchHistory()
 
   const { onClose } = useDrawerState(MODAL_NAME)
 
   const handleSignOut = useCallback(() => {
-    dispatchWeb(signOut)
+    dispatch(signOut({}))
```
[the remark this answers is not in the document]
raymondjacobson: ts was complaining b/c it's an action creator vs. some js garbo actions. both technically work, but this is more "correct" and I didn't want to change the other

packages/web/src/common/store/sign-out/sagas.ts
raymondjacobson: yeah, present on web too, but less impactful

audius-infra:
Preview this change https://demo.audius.co/rj-signinout

PR #1753, opened 2022-08-19
[C-862] Add env to saga context, fix eagerDiscprov in native

dylanjeffers:
### Description

Adds env to saga context, making it easier to work with process.env and Config
Fixes eagerDiscprov in native, which was using window.localStorage before

### DIFF
--- a/packages/mobile/src/services/env.ts
+++ b/packages/mobile/src/services/env.ts
@@ -1,0 +1,3 @@
+export const env = {
+  EAGER_DISCOVERY_NODES: process.env.REACT_APP_EAGER_DISCOVERY_NODES
+}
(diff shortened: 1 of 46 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-c-862-add-env-to-saga-context

packages/mobile/src/services/env.ts
```diff
@@ -0,0 +1,3 @@
+export const env = {
+  EAGER_DISCOVERY_NODES: process.env.REACT_APP_EAGER_DISCOVERY_NODES
+}
```
sliptype, line 3: It probably makes sense to refactor audiusBackend to use this too. But can do that later

packages/web/src/common/services/audius-api-client/AudiusAPIClient.ts
```diff
@@ -1421,7 +1429,9 @@ export class AudiusAPIClient {
     }
 
     // Set the state to the eager discprov
-    const eagerDiscprov = getEagerDiscprov()
+    console.log('aight its getting eager')
```
sliptype, line 1432: Logs

sliptype (approved):
Looks good!

PR #1764, opened 2022-08-22
[C-875] Remove blocklist

dylanjeffers:
### Description

Removes blocklist feature from client, which is no longer needed now because of trusted-notifier.


packages/web/src/common/services/audius-backend/AudiusBackend.ts
```diff
@@ -94,6 +94,7 @@ declare global {
     phantom: any
     solana: PhantomProvider
     web3: any
```
raymondjacobson, line 96: what even is lowercase web3? I feel like that doesn't exist actually

raymondjacobson (approved):
Thanks!

packages/web/src/common/services/audius-backend/AudiusBackend.ts
dylanjeffers: yeah agreed, this looks like a bug

packages/web/src/components/search-bar/store/sagas.js
```diff
@@ -35,16 +33,11 @@ export function* getSearchResults(searchText) {
 
   const { tracks, albums, playlists, users } = results
   const checkedUsers = users.filter((t) => !t.is_deactivated)
-  const checkedTracks = (yield call(
-    setTracksIsBlocked,
-    tracks,
-    audiusBackendInstance
-  )).filter((t) => !t.is_delete && !t._blocked && !t.user.is_deactivated)
```
sliptype, line 42: Do we still need to check is_delete and is_deactivated?

sliptype (approved):

audius-infra:
Preview this change https://demo.audius.co/dj-c-875-remove-blist

audius-infra:
Preview this change https://demo.audius.co/dj-c-875-remove-blist

PR #1767, opened 2022-08-22
Fix @audius/common external warnings, model imports, lint

dylanjeffers:
### Description

Adds a few random fixes i noticed while integrating with new common!

- Fixes remaining common external warnings
- Removes /index imports
- Uses absolute imports for models
- Updates lint

### DIFF
--- a/packages/common/rollup.config.js
+++ b/packages/common/rollup.config.js
@@ -24,7 +24,13 @@ export default [
 
     external: [
       ...Object.keys(pkg.dependencies),
-      ...Object.keys(pkg.devDependencies)
+      ...Object.keys(pkg.devDependencies),
+      ...Object.keys(pkg.peerDependencies),
+      'redux-saga/effects',
+      'events',
+      '@audius/sdk/dist/core',
+      'dayjs/plugin/timezone',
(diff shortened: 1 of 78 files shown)


sliptype (approved):
Cool! Why is the models abs path change possible now? Is the plan to use absolute paths for all the directories

dylanjeffers:
> Cool! Why is the models abs path change possible now? Is the plan to use absolute paths for all the directories

I think because we have node module resolve in tsconfig it just works. Basically there were already absolute model imports, so I just kept things consistent!

nicoback2:
ooh thank you so much!  @dylanjeffers  did you run typecheck in web and mobile? curious because when i was using absolute imports, it broke a bunch of things there.

dylanjeffers:
> ooh thank you so much! @dylanjeffers did you run typecheck in web and mobile? curious because when i was using absolute imports, it broke a bunch of things there which is why I changed a bunch to relative.

yeah i didnt.. checking it now though, yeah i think youre right. well at very least ill just keep the /index removal? will try to get to bottom though, cause there are existing absolute model imports there too, so not sure why those work :/

PR #1768, opened 2022-08-23
[C-858] Add track page sagas to native, Refactor TrackScreen

dylanjeffers:
### Description

- Adds track-page sagas to native
- Refactors TrackScreen to fetch tracks within native

### Dragons

- For now, useImageSize is not only using `useDispatch` breaking other places it's used, like profile. Thinking this is okay in the interim.
- Added `disableImagePreload` to mobile audius-backend, which simplifies fetching the track image. This seems like a decent approach, until we maybe want to add a native version of preload?
- Loading track page is slower perf right now since we can't rely on a cached version from the lineup screen. should be fixed as we continue to migrate, but just an fyi
- Separated trackPageLineupSagas from trackPageSagas since lineup requires work with player first

### DIFF
--- a/packages/mobile/src/screens/track-screen/TrackScreen.tsx
+++ b/packages/mobile/src/screens/track-screen/TrackScreen.tsx
@@ -1,11 +1,15 @@
+import { useEffect } from 'react'
+
 import {
   lineupSelectors,
   trackPageLineupActions,
+  trackPageActions,
   trackPageSelectors
 } from '@audius/common'
 import { trackRemixesPage } from 'audius-client/src/utils/route'
-import { omit } from 'lodash'
+import { useSelector } from 'common/hooks/useSelector'
 import { Text, View } from 'react-native'
+import { useDispatch } from 'react-redux'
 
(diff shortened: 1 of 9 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-c-858-reloaded-track-screen

packages/mobile/src/screens/track-screen/TrackScreen.tsx
```diff
@@ -51,20 +56,16 @@ export const TrackScreen = () => {
...
+  const dispatch = useDispatch()
 
   // params is incorrectly typed and can sometimes be undefined
-  const { searchTrack } = params ?? {}
-
-  const cachedTrack = useSelectorWeb(
-    (state) => getTrack(state, params),
-    // Omitting uneeded fields from the equality check because they are
-    // causing extra renders when added to the `track` object
-    (a, b) => {
```
sliptype, line 62: Yessss

sliptype (approved):
Looks great!

Yeah we need to rethink image fetching as a whole on native, we can make it much more performant

PR #1775, opened 2022-08-23
[C-890] Polyfill TextEncoder, fixes builds

dylanjeffers:
### Description

Fixes build issue by polyfilling TextEncoder, which is needed for newly added SolanaUtils.

### Dragons

This issue was not caught in development because the ReactNativeDebugger app injects it's own implementation of TextEncoder, which allows the app to run without a hitch. The issue only appears when you turn off Chrome Debugging (or obviously when run on a device from testflight)

### How Has This Been Tested?

Ran on device with Chrome Debugging disabled

### DIFF
--- a/packages/mobile/index.js
+++ b/packages/mobile/index.js
@@ -2,10 +2,11 @@ import 'react-native-gesture-handler'
 import { AppRegistry, LogBox } from 'react-native'
 
 import { name as appName } from './app.json'
+// Needed for TextEncoder to work correctly
 import 'text-encoding-polyfill'
 
 require('node-libs-react-native/globals')
-// needed for @solana/web3.js to run correctly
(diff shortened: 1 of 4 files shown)


sliptype (approved):
🔥

packages/mobile/index.js
```diff
@@ -2,6 +2,7 @@ import 'react-native-gesture-handler'
 import { AppRegistry, LogBox } from 'react-native'
 
 import { name as appName } from './app.json'
+import 'text-encoding-polyfill'
```
piazzatron, line 5: Should add a comment why we need this imo

piazzatron (approved):

packages/mobile/index.js
dylanjeffers: totes

PR #1776, opened 2022-08-23
[C-887] Host sourcemaps on s3, fixes cloudflare deploy

sliptype:
### Description

* Add ci to upload sourcemaps to `sourcemaps.audius.co` s3 bucket
* Remove sourcemaps from build deployed to cloudflare
* Update craco config to point at s3 bucket for sourcemaps
* Remove `@types/webpack` (included with webpack v5)
* Uses `cimg/python:3.7.12-node` image for cloudflare deploy steps because it has both node 14 and python

Was initially going to do the sourcemap move as a separate job but that required attaching and persisting the workspace which added unnecessary run time.

### Dragons

Right now all source maps are `sync`ed into the same directory in the bucket, this is never getting pruned so could get big/expensive. Can do another pass to create separate directories for the different builds, and overwrite existing files. But wanna merge this to unblock builds for now

### How Has This Been Tested?

Tested by deploying to staging and checking new sourcemaps work

### Feature Flags ###

### DIFF
--- a/packages/web/craco.config.ts
+++ b/packages/web/craco.config.ts
@@ -61,7 +61,7 @@ export default {
           }),
           new SourceMapDevToolPlugin({
             publicPath:
-              'https://s3.us-west-1.amazonaws.com/sourcemaps.audius.co',
+              'https://s3.us-west-1.amazonaws.com/sourcemaps.audius.co/',
             filename: '[file].map'
           })
         ],
--- a/packages/web/craco.config.ts
+++ b/packages/web/craco.config.ts
@@ -9,6 +9,8 @@ import {
 
 const isNative = process.env.REACT_APP_NATIVE_NAVIGATION_ENABLED === 'true'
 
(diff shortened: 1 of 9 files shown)


packages/web/craco.config.ts
```diff
@@ -52,6 +58,11 @@ export default {
           new ProvidePlugin({
             process: 'process/browser',
             Buffer: ['buffer', 'Buffer']
+          }),
+          new SourceMapDevToolPlugin({
+            publicPath:
+              'https://s3.us-west-1.amazonaws.com/sourcemaps.audius.co/',
```
piazzatron, line 64: nit: maybe pull out into a const?

audius-infra:
Preview this change https://demo.audius.co/sk-sourcemap-external

piazzatron (approved):
Glorious, thanks for handling this. My one thought is we should confirm that the sentry step still works at some point after this change, but not blocking merging this imo

PR #1779, opened 2022-08-24
[PAY-479] Check recipient wallet SOL before send

dharit-tan:
### Description
When transferring WAudio to an external wallet, throw an error if the recipient wouldn't be able to receive it, ie. if the recipient doesn't have an associated WAudio token account AND doesn't have a SOL balance to fund the creation of one.

### Dragons

I'm throwing an error in the selector for canRecipientReceiveWAudio on the assumption that if something is trying to access it from the incorrect state, then the caller probably depends on it in some incorrect way and so things should break. But open to suggestions!

Also in the selector and action, I am nesting the happy path but I think this is a case where it's necessary.

### How Has This Been Tested?

Tested on local browser.

https://user-images.githubusercontent.com/3893871/186984425-43bd5c4e-5d5f-4bac-ac90-d6788e4f5703.mov

https://user-images.githubusercontent.com/3893871/186984432-710f99b7-fd15-4e5e-a9f0-1979a9b96635.mov

### DIFF
--- a/packages/common/src/services/audius-backend/AudiusBackend.ts
+++ b/packages/common/src/services/audius-backend/AudiusBackend.ts
@@ -2916,13 +2916,10 @@ export const audiusBackend = ({
     return receipt
   }
 
-  /**
-   * Make a request to send solana wrapped audio
-   */
-  async function sendWAudioTokens(address: string, amount: BNWei) {
+  async function getAssociatedTokenAccountInfo(address: string) {
     await waitForLibsInit()
 
-    // Check when sending waudio if the user has a user bank acccount
+    // Check if the user has a user bank acccount
     let tokenAccountInfo =
       await audiusLibs.solanaWeb3Manager.getTokenAccountInfo(address)
     if (!tokenAccountInfo) {
@@ -2934,34 +2931,45 @@ export const audiusBackend = ({
         associatedTokenAccount.toString()
       )
 
-      // If it's not a valid token account, we need to make one first
-      if (!tokenAccountInfo) {
-        // We do not want to relay gas fees for this token account creation,
-        // so we ask the user to create one with phantom, showing an error
-        // if phantom is not found.
-        if (!window.phantom) {
-          return {
-            error:
-              'Recipient has no $AUDIO token account. Please install Phantom-Wallet to create one.'
-          }
-        }
-        if (!window.solana.isConnected) {
-          await window.solana.connect()
-        }
+    return tokenAccountInfo
+  }
 
-        const phantomWallet = window.solana.publicKey?.toString()
-        const tx = await getCreateAssociatedTokenAccountTransaction({
-          feePayerKey: SolanaUtils.newPublicKeyNullable(phantomWallet),
-          solanaWalletKey: SolanaUtils.newPublicKeyNullable(address),
-          mintKey: audiusLibs.solanaWeb3Manager.mintKey,
-          solanaTokenProgramKey: audiusLibs.solanaWeb3Manager.solanaTokenKey,
-          connection: audiusLibs.solanaWeb3Manager.connection
-        })
-        const { signature } = await window.solana.signAndSendTransaction(tx)
-        await audiusLibs.solanaWeb3Manager.connection.confirmTransaction(
-          signature
-        )
+  /**
+   * Make a request to send solana wrapped audio
+   */
+  async function sendWAudioTokens(address: string, amount: BNWei) {
+    await waitForLibsInit()
+
+    // Check when sending waudio if the user has a user bank acccount
+    const tokenAccountInfo = await getAssociatedTokenAccountInfo(address)
+
+    // If it's not a valid token account, we need to make one first
+    if (!tokenAccountInfo) {
+      // We do not want to relay gas fees for this token account creation,
+      // so we ask the user to create one with phantom, showing an error
+      // if phantom is not found.
+      if (!window.phantom) {
+        return {
+          error:
+            'Recipient has no $AUDIO token account. Please install Phantom-Wallet to create one.'
+        }
       }
+      if (!window.solana.isConnected) {
+        await window.solana.connect()
+      }
+
+      const phantomWallet = window.solana.publicKey?.toString()
+      const tx = await getCreateAssociatedTokenAccountTransaction({
+        feePayerKey: SolanaUtils.newPublicKeyNullable(phantomWallet),
+        solanaWalletKey: SolanaUtils.newPublicKeyNullable(address),
+        mintKey: audiusLibs.solanaWeb3Manager.mintKey,
+        solanaTokenProgramKey: audiusLibs.solanaWeb3Manager.solanaTokenKey,
+        connection: audiusLibs.solanaWeb3Manager.connection
+      })
+      const { signature } = await window.solana.signAndSendTransaction(tx)
+      await audiusLibs.solanaWeb3Manager.connection.confirmTransaction(
+        signature
+      )
     }
     return audiusLibs.solanaWeb3Manager.transferWAudio(address, amount)
   }
@@ -3016,6 +3024,17 @@ export const audiusBackend = ({
       return new BN('0')
     }
     return waudioBalance
+  }
+
+  /**
+   * Fetches the Sol balance for the given wallet address
+   * @param {string} The solana wallet address
+   * @returns {Promise<BNWei>}
+   */
+  async function getAddressSolBalance(address: string): Promise<BNWei> {
+    await waitForLibsInit()
+    const solBalance = await audiusLibs.solanaWeb3Manager.getSolBalance(address)
+    return solBalance
   }
 
   /**
@@ -3145,6 +3164,8 @@ export const audiusBackend = ({
     getAccount,
     getAddressTotalStakedBalance,
     getAddressWAudioBalance,
+    getAddressSolBalance,
+    getAssociatedTokenAccountInfo,
     getAllTracks,
     getArtistTracks,
     getAudiusLibs,
--- a/packages/common/src/services/audius-backend/AudiusBackend.ts
+++ b/packages/common/src/services/audius-backend/AudiusBackend.ts
@@ -2867,6 +2867,17 @@ export const audiusBackend = ({
   }
 
   /**
+   * Fetches the Sol balance for the given wallet address
+   * @param {string} The solana wallet address
+   * @returns {Promise<BNWei>}
+   */
+  async function getAddressSolBalance(address: string): Promise<BNWei> {
+    await waitForLibsInit()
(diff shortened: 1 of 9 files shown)


packages/common/src/store/pages/token-dashboard/selectors.ts
```diff
@@ -36,3 +36,16 @@ export const getHasAssociatedWallets = (state: CommonState) => {
...
+export const getCanRecipientReceiveWAudio = (state: CommonState) => {
+  if (
+    state.pages.tokenDashboard.modalState?.stage === 'SEND' &&
+    state.pages.tokenDashboard.modalState.flowState.stage ===
+      'AWAITING_CONFIRMATION'
+  ) {
+    return state.pages.tokenDashboard.modalState.flowState
+      .canRecipientReceiveWAudio
+  }
+  throw new Error(
```
rickyrombo, line 48: I don't think the selector should throw. We're also not handling this exception in the UI it doesn't look like.

Let's just pick a good default here?

packages/common/src/store/pages/token-dashboard/slice.ts
```diff
@@ -70,11 +70,30 @@ const slice = createSlice({
           stage: 'AWAITING_CONFIRMATION',
           amount,
           recipientWallet: wallet,
-          chain
+          chain,
+          canRecipientReceiveWAudio: true
```
rickyrombo, line 74: Why set this to true here? We don't know yet if it will succeed right?

packages/web/src/store/wallet/sagas.ts
```diff
@@ -217,6 +223,30 @@ function* fetchBalanceAsync() {
   }
 }
 
+/**
+ * Check if we can send WAudio to a recipient by checking if they already have
+ * an associated WAudio token account, or if they have enough SOL to create one.
+ */
+function* checkAssociatedTokenAccountOrSol() {
+  const walletClient = yield* getContext('walletClient')
+  const ret = yield* select(getSendData)
```
rickyrombo, line 232: We should be able to get the wallet address directly from the action (as an argument to this saga) without having to depend on the reducer already having run

packages/common/src/services/audius-backend/AudiusBackend.ts
```diff
@@ -3019,6 +3028,17 @@ export const audiusBackend = ({
     return waudioBalance
   }
 
+  /**
+   * Fetches the Sol balance for the given wallet address
+   * @param {string} The solana wallet address
+   * @returns {Promise<BNWei>}
+   */
+  async function getAddressSolBalance(address: string): Promise<BNWei> {
```
rickyrombo, line 3036: @piazzatron / @jowlee I wasn't aware of this `AudiusBackend.ts` / `WalletClient.ts` pattern before - should we continue it? Feels like `WalletClient.ts` could just call on libs itself?

rickyrombo: Also should we be using `WalletAddress` types here?

rickyrombo (commented):
Looking good! Only a few notes. 

Biggest thing is I don't think we should throw errors in selectors, I would never expect one to throw. If there's an error, it should be represented in the state that's returned. For this case, I'm not sure it's meaningful. Maybe we should just return whatever the state is for that member regardless of the stage?

audius-infra:
Preview this change https://demo.audius.co/rt-PAY-479

packages/common/src/store/pages/token-dashboard/selectors.ts
dharit-tan: Good point!

packages/web/src/store/wallet/sagas.ts
```diff
@@ -40,6 +46,10 @@ const errors = {
   rateLimitError: 'Please wait before trying again'
 }
 
+const ASSOCIATED_TOKEN_ACCOUNT_CREATION_FEE = new BN(
```
rickyrombo, line 49: nit: Technically the amount of SOL is slightly higher than this, but this should 80/20 it. If we want to be precise, there's an RPC method for getting the minimum rent for exemption given a byte size. I use it in the But Audio flow

audius-infra:
Preview this change https://demo.audius.co/rt-PAY-479

piazzatron (approved):
Looks great!! Good looks including the screen recording too 👍

packages/web/src/store/wallet/sagas.ts
dharit-tan: Nice!

dharit-tan:
@piazzatron @rickyrombo can you guys take another look at this plz? Ended up changing a few things - added a loading state since with the extra RPC call now there's enough of a lag that the error message was flickering in.

packages/web/src/store/wallet/sagas.ts
```diff
@@ -238,8 +234,12 @@ function* checkAssociatedTokenAccountOrSol(action: InputSendDataAction) {
...
   )
   if (!associatedTokenAccount) {
-    const balance = yield* call(() => walletClient.getWalletSolBalance(address))
-    if (balance.lt(ASSOCIATED_TOKEN_ACCOUNT_CREATION_FEE)) {
+    const balance: BNWei = yield* call(() =>
+      walletClient.getWalletSolBalance(address)
+    )
+    const minRentForATA = yield* call(getAssociatedTokenRentExemptionMinimum)
+    const minRentBN = new BN(minRentForATA)
+    if (balance.lt(minRentBN)) {
```
rickyrombo, line 246: Looks great 🎉

packages/common/src/services/audius-backend/eagerLoadUtils.ts
```diff
@@ -3,6 +3,25 @@ import { LocalStorage } from '../local-storage'
...
+
+/**
+ * Wait for the `LIBS_INITTED_EVENT` or pass through if there
+ * already exists a mounted `window.audiusLibs` object.
+ */
+export const waitForLibsInit = async () => {
+  // If libs is already defined, it has already loaded & initted
+  // so do nothing
+  // @ts-ignore
+  if (window.audiusLibs) return
```
rickyrombo, line 16: I don't think this can be in `common` since this code also will run on mobile which doesn't have a `window`

rickyrombo (commented):
Concerned about putting deps on web env in `common` - maybe this is why we had the wallet client/audius backend split?

packages/common/src/services/audius-backend/eagerLoadUtils.ts
dharit-tan: Hm how can I implement WaitForLibsInit in common then? Is there another way to access audiusLibs?

rickyrombo: Hmm i see, I think this is the reason that there was that indirection between wallet client and audius backend - the waitForLibs method gets injected into audius backend. I led you astray 😢

dharit-tan: Otherwise I can just revert to using AudiusBackend

rickyrombo: yeah I think back to backend makes sense sorry 😞

packages/common/src/store/pages/token-dashboard/selectors.ts
```diff
@@ -36,3 +36,24 @@ export const getHasAssociatedWallets = (state: CommonState) => {
...
   state.pages.tokenDashboard.associatedWallets.removeWallet
+export const getAwaitingConfirmationIsLoading = (state: CommonState) => {
+  if (
+    state.pages.tokenDashboard.modalState?.stage === 'SEND' &&
+    state.pages.tokenDashboard.modalState.flowState.stage ===
+      'AWAITING_CONFIRMATION'
+  ) {
+    return state.pages.tokenDashboard.modalState?.flowState.loading
+  }
+  return false
```
piazzatron, line 47: nit: could refactor this and the below selector as a single return
```
return state.pages.tokenDashboard.modalState?.stage === 'SEND' &&
    state.pages.tokenDashboard.modalState.flowState.stage ===
      'AWAITING_CONFIRMATION' && state.pages.tokenDashboard.modalState?.flowState.loading
```

packages/common/src/store/pages/token-dashboard/types.ts
```diff
@@ -14,6 +14,8 @@ type SendingState =
       amount: StringWei
       recipientWallet: string
       chain: Chain
+      canRecipientReceiveWAudio: boolean
+      loading: false
```
piazzatron, line 18: Thinking through the data modeling here - rather than modeling this as two separate booleans, where there are certain combos that don't make sense (canReceive = false, loading = true), and it's not super clear what loading refers to, I think we could simplify this by giving canReceiveWaudio 3 states: `'false' | 'loading' | 'true'`

audius-infra:
Preview this change https://demo.audius.co/rt-PAY-479

piazzatron (commented):
Generally don't want to only show spinners for a fraction of a second if we can avoid it – let's huddle and think of a way around this

audius-infra:
Preview this change https://demo.audius.co/rt-PAY-479

packages/web/src/pages/audio-rewards-page/components/SendInputConfirmation.tsx
```diff
@@ -40,6 +49,20 @@ const SendInputConfirmation = ({
   recipientAddress,
   onSend
 }: SendInputConfirmationProps) => {
+  const [hasOneSecondElapsed, setHasOneSecondElapsed] = useState(false)
```
piazzatron, line 52: nit: best if we don't name the variable after the duration itself, so we give ourselves some flexibility to change the duration as config w/o changing the code. Something like `hasLoadingDurationElapsed`, idk

packages/web/src/pages/audio-rewards-page/components/SendInputConfirmation.tsx
```diff
@@ -40,6 +49,20 @@ const SendInputConfirmation = ({
...
+  const canRecipientReceiveWAudio = useSelector(getCanRecipientReceiveWAudio)
+  const isLongLoading =
+    hasOneSecondElapsed && canRecipientReceiveWAudio === 'loading'
+
+  // State to help determine whether to show a loading spinner,
+  // for example if Solana is being slow
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      setHasOneSecondElapsed(true)
+    }, 1000)
```
piazzatron, line 62: nit: let's pull this magic # out into a const at the top of the file

piazzatron (approved):
Looks great! Only nits :), 🚢

audius-infra:
Preview this change https://demo.audius.co/rt-PAY-479

audius-infra:
Preview this change https://demo.audius.co/rt-PAY-479

PR #1780, opened 2022-08-24
[PAY-476] Display error when send to wallet fails

dharit-tan:
### Description
Adds a generic error message when sending to an external wallet fails.


audius-infra:
Preview this change https://demo.audius.co/rt-PAY-476

rickyrombo (approved):

PR #1784, opened 2022-08-24
[C-871] Migrate player state to common

dylanjeffers:
### Description

Migrates player redux state to common, moves actions, slice, and selectors


sliptype (approved):
Nice! All the type changes and refactors make sense

packages/common/src/store/player/slice.ts
```diff
@@ -1,20 +1,18 @@
-import { UID, ID, Collectible, Nullable } from '@audius/common'
 import { createSlice, PayloadAction } from '@reduxjs/toolkit'
 
-import NativeMobileAudio from 'audio/NativeMobileAudio'
+import { UID, ID, Collectible } from '../../models'
+import { Nullable } from '../../utils'
 
-import { AudioState } from './types'
+import { Audio } from './types'
```
nicoback2, line 6: just OOC, why the name change? Might make sense make the name more specific since it's exported at the top level of common

nicoback2 (approved):
awesome!

audius-infra:
Preview this change https://demo.audius.co/dj-c-871-move-player-queue-to-common

packages/common/src/store/player/slice.ts
dylanjeffers: yeah def agreed, updating it to `AudioPlayer`

PR #1786, opened 2022-08-24
Migrate queue selectors to common

dylanjeffers:
### Description

Migrates queue selectors now that player is in common


sliptype (approved):

audius-infra:
Preview this change https://demo.audius.co/dj-migrate-queue-selectors

PR #1795, opened 2022-08-25
[C-900] Move queue, lineup, feed-page sagas to common

dylanjeffers:
### Description

- Moves queue, lineup, and feed-page sagas to common to prepare for native feed-page
- Also needed for native queue/player

### Dragons

Trying out a new pattern where mobileSagas are unhooked from the "common" saga to avoid polluting the native usage of the common saga. Adding the mobileSaga at the end of the web/sagas file for compatibility.


packages/web/src/common/store/lineup/sagas.js
```diff
@@ -25,15 +25,37 @@ import {
...
-import { isMobile } from 'utils/clientUtil'
+import { getToQueue } from 'common/store/queue/sagas'
 
 const { getSource, getUid, getPositions } = queueSelectors
 const { getUid: getCurrentPlayerTrackUid } = playerSelectors
 const { getUsers } = cacheUsersSelectors
 const { getTrack, getTracks } = cacheTracksSelectors
 const { getCollection } = cacheCollectionsSelectors
 
+const IS_NATIVE_MOBILE = process.env.REACT_APP_NATIVE_MOBILE
```
dylanjeffers, line 36: note: duplicating here since it was defined in a non-common file imported by 90 other files

nicoback2: maybe put this in a comment? (the part about it being ignored in native)

nicoback2: sorry just thought about this more, you could also have the caller pass `isMobile` to the `fetchLineupMetadatasAsync` function? since thats the only place it's used

nicoback2 (approved):

sliptype (approved):

packages/web/src/common/store/lineup/sagas.js
dylanjeffers: yeah so caller is in this same file, `watchFetchLineupMetadata` which is just watching for the action to be dispatched, so i think the main way we can get env specific stuff here is to have the "isMobile" thing in the context. I went ahead and short-circuted this isMobile thing by adding the `isNativeMobile` context check which will ensure it never gets called with new system, but is compatible with old. what do you think?

audius-infra:
Preview this change https://demo.audius.co/dj-c-900-move-queue-lineup-feed-sagas

PR #1797, opened 2022-08-25
[C-901] Move audioPlayer to saga context

dylanjeffers:
### Description
By moving audioPlayer to sagaContext, we can begin development on a native audioPlayer. Without this, native will try to import hls and a bunch of other web only deps.

- Moves audioPlayer to saga context
- Adds AudioPlayer and NativeMobileAudio to web store context to maintain backwards compatibility
- Adds audioPlayer shim in native store just to avoid having an `undefined` check in all the usages in sagas/web components.u


audius-infra:
Preview this change https://demo.audius.co/dj-c-901-move-audio-player-to-saga-context

packages/web/src/services/audio-player/NativeMobileAudio.ts
```diff
@@ -10,13 +10,15 @@ import { generateM3U8Variants } from 'utils/hlsUtil'
...
-class NativeMobileAudio {
+export class NativeMobileAudio {
+  audio: any
   m3u8: string
   position: number
   duration: number
   // Whether or not the user has made a seek action.
   // We use this to make sure we don't read stale position values from the native layer.
   seekOverride: number | null
+  audioCtx = null
```
raymondjacobson, line 21: this doesn't carry any meaning in native land, but i guess this file goes away entirely anyway right?

raymondjacobson (approved):
makes sense!

packages/web/src/services/audio-player/NativeMobileAudio.ts
dylanjeffers: yeah just to get types to be happy, and yes goes away in native-reloaded

PR #1799, opened 2022-08-26
[C-903] Migrate feed to native store

dylanjeffers:
### Description

Migrates the feed displaying and loading to native store. So awesome it was this easy ultimately


sliptype (approved):
Let's gooo

PR #1805, opened 2022-08-26
[C-906] Migrate queue sagas and hls-utils to common

dylanjeffers:
### Description

Migrates queue sagas to common
Migrates hlsUtils to @audius/common

These are the final pieces to enable fully native audio


audius-infra:
Preview this change https://demo.audius.co/dj-c-906-migrate-queue-hls-to-common

packages/web/src/common/store/queue/mobileSagas.ts
```diff
@@ -72,7 +73,10 @@ function* getTrackInfo(id: ID, uid: UID) {
     : track.cover_art
 
   const m3u8Gateways = gateways.concat(PUBLIC_IPFS_GATEWAY)
-  const m3u8 = generateM3U8Variants(track.track_segments, [], m3u8Gateways)
+  const m3u8 = generateM3U8Variants({
```
raymondjacobson, line 76: don't we want to do this inside the component on the native side? or is that not necessary now?

raymondjacobson (approved):

packages/web/src/common/store/queue/mobileSagas.ts
dylanjeffers: yeah we do, but this is just for backwards compat on main!

raymondjacobson: ohhh i didn't recognize the diff base here. sounds great

sliptype (approved):
Awesome! thanks for converting to ts too

PR #1807, opened 2022-08-26
[C-872] Play audio in native

dylanjeffers:
### Description

Updates native app to play audio entirely using it's own store!
Removes tons of legacy message passing


raymondjacobson (approved):
really impressed that you were able to grok all this and get this out!!

packages/mobile/src/components/audio/Audio.tsx
```diff
@@ -227,7 +265,7 @@ const Audio = ({
           isAction: true
         })
       }
-    } else if (track && track.isDelete) {
+    } else if (track && track.is_delete) {
       if (webRef.current) {
         // Sync w/ isPlaying false to set player state in dapp and hide drawer
         postMessage(webRef.current, {
```
sliptype, line 271: Are we keeping this to minimize breakages for now?

packages/mobile/src/components/lineup/Lineup.tsx
```diff
@@ -229,43 +237,29 @@ export const Lineup = ({
...
-      isPlaying
-    }: {
-      uid: UID
-      id: ID
-      source: PlaybackSource
-      isPlayingUid: boolean
-      isPlaying: boolean
-    }) => {
+    ({ uid, id, source, isPlayingUid, isPlaying }: TogglePlayConfig) => {
       // setImmediate prevents this cpu-intensive callback from firing until
```
sliptype, line 241: Can remove comment here

sliptype (approved):
So spicy

packages/mobile/src/components/audio/Audio.tsx
dylanjeffers: just wanted to double check your meaning. the type previously was a track from the "syncQueue" state that i guess munges the track type to use camelCase for things, so this was typescript complaining since we are now using an actual track type.

sliptype: Sorry, this was unclear, I mean the `postMessage` back to the web view

dylanjeffers: oh great call, can def remove those

PR #1812, opened 2022-08-29
Fix lineup types in mobile build

dylanjeffers:
### Description

Case where our stricter mobile types were complaining about web code recently introduced


sliptype (approved):
Cool, hopefully we can get the lineup sagas ts at some point

dylanjeffers:
> Cool, hopefully we can get the lineup sagas ts at some point

yeah down to approach this as we integrate the other lineups

PR #1815, opened 2022-08-29
Fix track removal from playlist with entity manager

isaacsolo:
### Description

Entity manager enabled track removals were broken. The metadata_timestamp would not match the playlist indexing timestamp so no track would be removed. This logic will check if the playlist is enabled / has a metadata_timestamp then match based on that. If entity manager is disabled, it will use indexing timestamp instead.

### Dragons

### How Has This Been Tested?

Tested against staging.
Test track addition/removal with without cached state.
Tested legacy flow as well.

### Feature Flags ###
This change is behind the playlist_entity_manager_enabled flag.


audius-infra:
Preview this change https://demo.audius.co/is-fix-track-deletion-2

dylanjeffers (approved):
cool makes sense

PR #1816, opened 2022-08-29
[C-917] Add native trending screen

dylanjeffers:
### Description

Add trending page sagas to mobile
Replace all usages of useSelectorWeb and dispatchWeb


sliptype (approved):
ezpz

PR #1829, opened 2022-08-30
[C-934] Migrate user-list sagas to common

dylanjeffers:
### Description

Migrates user-list sagas to common. needed for native user-lists


audius-infra:
Preview this change https://demo.audius.co/dj-c-934-user-list-sagas

packages/web/src/common/store/user-list/supporting/sagas.ts
```diff
@@ -18,7 +18,7 @@ import {
 import { put, select } from 'typed-redux-saga'
 
 import { watchSupportingError } from 'common/store/user-list/supporting/errorSagas'
-import { createUserListProvider } from 'components/user-list/utils'
+import { createUserListProvider } from 'common/store/user-list/utils'
 import { fetchSupporting } from 'services/audius-backend/Tipping'
```
sliptype, line 22: I think Tipping is still dependent on the web waitForLibsInit

sliptype (approved):
Nice! Think we should fix the Tipping dep though

packages/web/src/common/store/user-list/supporting/sagas.ts
dylanjeffers: thanks, going to avoid this one until we hit tipping i think. should i not move this one over you think even?

sliptype: Ah, no it's probably fine!

PR #1831, opened 2022-08-31
[C-936] Remove WebView from native

dylanjeffers:
### Description

- Removes WebView from native
- Updates useSelectorWeb to throw error (this will help us convert the rest of the app)
- Fixes happy path to not throw useSelectorWeb error

Runs without the web app running!

### Dragons

- Splash screen behavior has been removed, so we will need to figure out best way to re-integrate

- Im halfway through removing the native-mobile-interface on web as well, so happy to add that to this pr, or do in a followup


packages/mobile/ios/AudiusReactNative/Info.plist
```diff
@@ -1,104 +1,104 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
raymondjacobson, line 1: what's up with the info.plist change?

packages/mobile/package.json
```diff
@@ -106,7 +106,6 @@
     "react-native-screens": "3.13.0",
     "react-native-scrypt": "1.2.1",
     "react-native-shadow-2": "6.0.1",
-    "react-native-static-server": "github:futurepress/react-native-static-server.git#034c04e7a2999ae11a799fb68266c94389fc61bf",
```
raymondjacobson, line 109: :feelsgood:

packages/mobile/src/hooks/useKeyboardListeners.ts
```diff
@@ -1,37 +0,0 @@
-import type { RefObject } from 'react'
```
raymondjacobson, line 1: we could have probably dropped this after the native UI. nice

raymondjacobson (approved):
im sure this will break many a thing, but all for merging it

packages/mobile/src/Drawers.tsx
```diff
@@ -117,7 +116,6 @@ export const Drawers = () => {
           drawer={Drawer}
         />
       ))}
-      <DiscordDrawer />
```
sliptype, line 120: Why remove this?

packages/mobile/ios/AudiusReactNative/Info.plist
dylanjeffers: yeah it's just my editor auto-formatting. i hade to update version so we didnt get the "updateRequired" stuff

dylanjeffers: but happy to revert the auto-format, and just modify the one line

raymondjacobson: this is fine w/ me

packages/mobile/src/hooks/useDispatchWeb.ts
```diff
@@ -1,18 +1,11 @@
...
-  const { webRef } = useContext(WebRefContext)
-  return useCallback(
-    (action: { type: string; [key: string]: any }) => {
-      if (webRef?.current) {
-        postMessage(webRef.current, { isAction: true, ...action })
-      }
-    },
-    [webRef]
-  )
+  return useCallback((_action: { type: string; [key: string]: any }) => {
```
sliptype, line 6: Nice 👍

packages/mobile/src/hooks/useUpdateRequired.ts
```diff
@@ -7,6 +7,7 @@ import { useRemoteVar } from 'app/hooks/useRemoteConfig'
 export const useUpdateRequired = () => {
   const { appVersion } = VersionNumber
   const minAppVersion = useRemoteVar(StringKeys.MIN_APP_VERSION)
+  console.log({ appVersion, minAppVersion })
```
sliptype, line 10: Needed?

packages/mobile/src/screens/signon/SignOn.tsx
```diff
@@ -339,8 +339,6 @@ const SignOn = ({ navigation }: SignOnProps) => {
 
   const isKeyboardOpen = useSelector(getIsKeyboardOpen)
 
-  // TODO: Remove when fully on react native store
-  const dappLoaded = useSelector(getDappLoaded)
   const lifecycleSignIn = useSelector(getIsSignedIn)
```
sliptype, line 342: We can probably get this from the account slice instead

sliptype (approved):
FREAKING SWEET!!!

Let's think through the splash screen stuff, if we can get the skeletons displaying immediately after the splash that would be ideal

packages/mobile/src/screens/signon/SignOn.tsx
dylanjeffers: so thinking we don't even need right? this screen will only load when "notSignedIn" is true in AppScreen, at which point we can immediately run the sign-on screen animations since we have libs ready to go right away?

sliptype: Oh yup makes sense

audius-infra:
Preview this change https://demo.audius.co/dj-c-936-remove-webview-from-native

PR #1835, opened 2022-08-31
[C-836] Fix enter on search

dylanjeffers:
### Description

Fixes case where pressing enter in search bar does not send user to search page


raymondjacobson (approved):

sliptype (approved):
Does pressing enter before the autocomplete results have loaded still work?

dylanjeffers:
> Does pressing enter before the autocomplete results have loaded still work?

great call, no it doesn't im just gonna remove that check entirely then

PR #1837, opened 2022-08-31
[PAY-507] Fix feed tip tile tip button resizing

dharit-tan:
### Description
Fixes feed tip tile "send tip" button resizing issue. Now also wraps the recipient info and "was tipped by" section.

### Dragons

Some jumpiness if the user resizes the window quickly, see the end of the video.

Now 
### How Has This Been Tested?

Tested on local web

https://user-images.githubusercontent.com/3893871/187829466-d366ffe3-5a9c-46c7-899f-3e8325e55848.mov

### DIFF
--- a/packages/web/src/components/tipping/feed-tip-tile/FeedTipTile.tsx
+++ b/packages/web/src/components/tipping/feed-tip-tile/FeedTipTile.tsx
@@ -1,4 +1,10 @@
-import { useCallback, useEffect } from 'react'
+import {
+  useState,
+  useCallback,
+  useEffect,
+  useRef,
+  MutableRefObject
+} from 'react'
 
 import {
   Name,
@@ -8,7 +14,7 @@ import {
   tippingSelectors,
   tippingActions
 } from '@audius/common'
-import { IconButton, PillButton, useMediaQueryListener } from '@audius/stems'
+import { IconButton, PillButton } from '@audius/stems'
 import { push as pushRoute } from 'connected-react-router'
 import { useDispatch, useSelector } from 'react-redux'
 
@@ -117,9 +123,20 @@ const Tippers = ({ tippers, receiver }: TippersProps) => {
 type SendTipButtonProps = {
   user: User
   hideName?: boolean
+  sendTipRef: MutableRefObject<HTMLDivElement | null>
+  firstLoadRef: MutableRefObject<boolean>
+  maxTipButtonWidth: MutableRefObject<number>
+  handleResize: () => void
 }
 
-const SendTipButton = ({ user, hideName = false }: SendTipButtonProps) => {
+const SendTipButton = ({
+  user,
+  hideName = false,
+  sendTipRef,
+  firstLoadRef,
+  maxTipButtonWidth,
+  handleResize
+}: SendTipButtonProps) => {
   const dispatch = useDispatch()
 
   const handleClick = useCallback(() => {
@@ -141,6 +158,17 @@ const SendTipButton = ({ user, hideName = false }: SendTipButtonProps) => {
         />
       </div>
     )
+
+  // On first load, save the max width of the send tip button
+  useEffect(() => {
+    if (!sendTipRef.current || !firstLoadRef.current) {
+      return
+    }
+    firstLoadRef.current = false
+    maxTipButtonWidth.current = sendTipRef.current.offsetWidth
+    // If the window is small on first load, downsize the button
+    handleResize()
+  })
 
   return (
     <PillButton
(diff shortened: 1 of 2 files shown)


packages/web/src/components/tipping/feed-tip-tile/FeedTipTile.tsx
```diff
@@ -218,15 +248,41 @@ export const FeedTipTile = () => {
     }
   }, [dispatch, usersMap, tipToDisplay])
 
+  const handleResize = () => {
```
rickyrombo, line 251: We have a handy dandy `useMeasure` hook we could potentially use instead of attaching to the window:

https://github.com/streamich/react-use/blob/master/docs/useMeasure.md

might make some of this easier

packages/web/src/components/tipping/feed-tip-tile/FeedTipTile.tsx
```diff
@@ -257,10 +313,14 @@ export const FeedTipTile = () => {
           receiver={usersMap[tipToDisplay.receiver_id]}
         />
       </div>
-      <div className={styles.buttons}>
+      <div className={styles.buttons} ref={sendTipRef}>
         <SendTipButton
           user={usersMap[tipToDisplay.receiver_id]}
-          hideName={useShortButtonFormat}
+          hideName={useShortButtonFormat.current}
```
rickyrombo, line 319: Instead of all the JS logic, can we add CSS media queries for various screen size breakpoints? That's typically how responsive ui is done afaik

audius-infra:
Preview this change https://demo.audius.co/rt-PAY-507

packages/web/src/components/tipping/feed-tip-tile/FeedTipTile.tsx
dharit-tan: I think that was how we were doing it before but that's where the problem came from. The resize really depends on the width of the other contents of the tile, not the screen size. There's probably a way we can determine whether to resize by incorporating the screen size, but I don't think that'll save us from having to do these calculations?

packages/web/src/components/tipping/feed-tip-tile/FeedTipTile.tsx
dharit-tan: Oo nice

packages/web/src/components/tipping/feed-tip-tile/FeedTipTile.tsx
rickyrombo: Hmm I see, we're not resizing based on window width alone, but based on widths of the button + the other content being >= the container?

dharit-tan: Yep. Tho I could use those numbers to calculate the window size at which we'd need to resize if that's preferred.

packages/web/src/components/tipping/feed-tip-tile/FeedTipTile.tsx
```diff
@@ -207,6 +227,16 @@ export const FeedTipTile = () => {
...
   )
 
+  const [hasShortButtonChanged, setHasShortButtonChanged] = useState(false)
+  const containerRef: MutableRefObject<HTMLDivElement | null> = useRef(null)
+  const containerLeftContentsRef: MutableRefObject<HTMLDivElement | null> =
+    useRef(null)
+  const sendTipRef: MutableRefObject<HTMLDivElement | null> = useRef(null)
+  const firstLoadRef = useRef(true)
+  const maxTipButtonWidth = useRef(0)
+  const useShortButtonFormat = useRef(false)
```
rickyrombo, line 237: I think this can probably be a state instead of ref? What's the reasoning for a ref here?

packages/web/src/components/tipping/feed-tip-tile/FeedTipTile.tsx
rickyrombo: I think I get it and what you have seems right...

Sorry for my confusion, I'm ramping up to speed here - there's two things happening from what I see:
1) The button is changing at the component level based on how much space it has - this is what you've solved here with measuring the size of the two elements and the container and passing a prop to the button. This makes sense because the goal is to change the button based on how long the names are, not how big the screen is - otherwise we could use CSS media queries here to hide the "to (username)" elements.
2) The content is overflowing when it shouldn't on small screens even after the smaller button - this is a "responsive" bug and is going to need some `flex-shrink`, `flex-wrap`, or CSS media queries. This is out of scope for the PR

rickyrombo (approved):

PR #1841, opened 2022-09-01
[PAY-591] Improve supporter count visibility

dharit-tan:
### Description
Adds text shadow to follower count, and adds a black view with 10% opacity on top of the last profile picture in supporters.
Before:
Screen Shot 2022-09-01 at 2 13 45 PM

After:
Screen Shot 2022-09-01 at 2 13 51 PM

### How Has This Been Tested?

Tested on ios simulator


rickyrombo (approved):
Nice!

PR #1847, opened 2022-09-01
Split up sign-out saga to web and native

dylanjeffers:
### Description

- Splits up sign-out sagas to web and native variant
- Removes native messaging related to signout
- Removes a bunch of strange sign-out calls in components, now we just call the signout action and let the sagas do their thing.


packages/web/src/store/sign-out/sagas.ts
```diff
@@ -0,0 +1,29 @@
...
+} from '@audius/common'
+import { takeLatest, put } from 'redux-saga/effects'
+
+import { make } from 'common/store/analytics/actions'
+import { signOut } from 'store/sign-out/signOut'
+const { resetAccount } = accountActions
+const { signOut: signOutAction } = signOutActions
+
+function* watchSignOut() {
+  const audiusBackendInstance = yield* getContext('audiusBackendInstance')
```
sliptype, line 15: Same here

sliptype (approved):
🔥🔥🔥

packages/web/src/pages/settings-page/components/desktop/SettingsPage.tsx
```diff
@@ -381,4 +389,10 @@ class SettingsPage extends Component<SettingsPageProps, SettingsPageState> {
   }
 }
 
-export default SettingsPage
+function mapDispatchToProps(dispatch: Dispatch) {
```
raymondjacobson, line 392: this should go in SettingsPageProvider to follow the existing pattern of connected "provider" components

raymondjacobson: also, this leaves out the mobile web version of this page right?

raymondjacobson (approved):

packages/mobile/src/store/sign-out/sagas.ts
```diff
@@ -0,0 +1,35 @@
...
+const { resetAccount } = accountActions
+const { signOut: signOutAction } = signOutActions
+
+function* watchSignOut() {
+  yield* takeLatest(signOutAction.type, function* () {
+    yield* put(resetAccount())
+    yield* call(disablePushNotifications)
+    yield* put(make(Name.SETTINGS_LOG_OUT, {}))
+
+    const localStorage = yield* getContext('localStorage')
```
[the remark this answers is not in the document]
dylanjeffers: oh yeah technically not!

packages/web/src/pages/settings-page/components/desktop/SettingsPage.tsx
dylanjeffers: ah yeah that makes sense, well mobile just using dispatch manually itself, ie they both handle it, but i can thread it through both to keep the pattern

dylanjeffers: oh actually they are different since mobile signs out in the confirmation drawer...

PR #1848, opened 2022-09-02
[C-955] Fully native settings screen

dylanjeffers:
### Description

- Updates settings/account-settings/listening history to use native store
- This is precursor work to updating theme/cast/notif settings to work correctly in web vs native context


sliptype (approved):
Nice!

PR #1851, opened 2022-09-02
[C-937] Refactor theme state, Implement native theme

dylanjeffers:
### Description

- Refactors theme state to slice
- Adds systemAppearance value (useful for native theme management)
- Splits web theme sagas into web and native variants, removing message passing classes
- Removes `ThemeContext` in react-native which was somewhat of an anti-pattern compared to our redux-saga approach
- Updates theme state usages
- Cleans up react-native-dark-mode usages


audius-infra:
Preview this change https://demo.audius.co/dj-c-937-native-theme

packages/mobile/src/components/navigation-container/NavigationContainer.tsx
```diff
@@ -11,24 +10,29 @@ import { useSelector } from 'react-redux'
...
-  const { theme, isSystemDarkMode } = useContext(ThemeContext)
+const NavigationContainer = (props: NavigationContainerProps) => {
+  const { children } = props
+  const theme = useSelector(getTheme)
+  const systemAppearance = useSelector(getSystemAppearance)
   const account = useSelector(getAccountUser)
 
   const navigationTheme =
-    theme === 'auto' ? (isSystemDarkMode ? 'dark' : 'default') : theme
+    theme === 'auto'
```
sliptype, line 31: Could we encapsulate this logic in a hook since we are using it in multiple places?

packages/mobile/src/store/theme/sagas.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+    eventEmitter.on('currentModeChanged', listener)
+  })
+
+  eventEmitter.removeListener('currentModeChanged', listener)
+
+  return systemAppearance
+}
+
+// Watches for changes in the native OS theme by
```
sliptype, line 26: I don't think this comment is accurate anymore

sliptype (approved):
Wow, super clean. nice work!

packages/mobile/src/components/navigation-container/NavigationContainer.tsx
dylanjeffers: yeah i like that, this one is slightly diff tho, since it's 'auto'? but yeah i hear you that this type of logic is in a lot of spots....

PR #1860, opened 2022-09-03
Fix social features and other sig-util features

dylanjeffers:
### Description

Fixes issue where metro.config munges up sig-util dependency versions.

### Dragons

These changes fix the issue, but we may not need all these verisons explicitly listed. I can go through and see the minimal change needed to unblock us.

TODO: we can remove all this when everything is migrated to @audius/common and we can simplify the metro.config node_module resolutions.


sliptype (approved):
Amazing!

packages/mobile/package.json
```diff
@@ -66,6 +66,11 @@
     "babel-plugin-transform-remove-console": "6.9.4",
     "bn.js": "5.2.0",
     "color": "3.2.1",
+    "eth-sig-util": "2.5.4",
+    "ethereumjs-abi": "0.6.8",
+    "ethereumjs-util": "^5.1.1",
```
raymondjacobson, line 71: can we pin these?

raymondjacobson (approved):
bae.

PR #1865, opened 2022-09-06
[C-879] Fix rewards page navigation

dylanjeffers:
### Description

Fixes issue where signed in users navigating to /audio page are redirected to /trending page

Essentially our `useRequiresAccount` is not waiting for user object before redirecting.


sliptype (approved):
Wow, nice fix!

sliptype:
We can probably cherry pick this onto the release, wdyt?

PR #1866, opened 2022-09-06
Clean up remote-config init handlers

sliptype:
### Description

* Makes `waitForUserRemoteConfig` platform agnostic by using EventEmitter instead of window events
* Updates `init` to be consistent with `setUserId` in using the EventEmitter

### Dragons

Removes the `USER_ID_AVAILABLE_EVENT` event, removed all references in the codebase though

### How Has This Been Tested?

Tested on web, ensured that both callbacks were being hit at the correct time and the Feature Flag override modal still works properly

### Feature Flags ###

### DIFF
--- a/packages/common/src/services/remote-config/remote-config.ts
+++ b/packages/common/src/services/remote-config/remote-config.ts
@@ -20,8 +20,6 @@ import {
   AllRemoteConfigKeys
 } from './types'
 
-export const USER_ID_AVAILABLE_EVENT = 'USER_ID_AVAILABLE_EVENT'
-
 // Constants
 // All optimizely feature keys are lowercase_snake
 const REMOTE_CONFIG_FEATURE_KEY = 'remote_config'
@@ -30,7 +28,6 @@ const REMOTE_CONFIG_FEATURE_KEY = 'remote_config'
 type State = {
   didInitialize: boolean
   id: Nullable<number>
-  initializationCallbacks: (() => void)[]
 }
 
 export type RemoteConfigOptions<Client> = {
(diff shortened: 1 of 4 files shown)


audius-infra:
Preview this change https://demo.audius.co/sk-remote-config-wait-for-user

packages/common/src/services/remote-config/remote-config.ts
```diff
@@ -82,37 +82,50 @@ export const remoteConfig = <
...
+      listenForUserId(handler)
+    }
+  }
 
   /**
    * Set the userId for calls to Optimizely.
    * Prior to calling, uses the ANONYMOUS_USER_ID constant.
-   * @param userId
    */
   function setUserId(userId: ID) {
```
piazzatron, line 126: nit: since we're touching this anyways, the comment about using ANONYMOUS_USER_ID lies - we use the session ID now

packages/common/src/services/remote-config/remote-config.ts
```diff
@@ -82,37 +82,50 @@ export const remoteConfig = <
...
+      emitter.addListener('init', handler)
     }
   }
 
-  // Use event emission to track setUser events
-  const emitter = new EventEmitter()
+  /**
+   * Register a callback for user ready.
+   */
+  function onUserReady(cb: () => void) {
```
piazzatron, line 110: nit: since I don't think we're consuming this function outside of remote-config, maybe move it out of the `api` section and make it private?

piazzatron (approved):
Looks good to me! Love seeing all the red

PR #1868, opened 2022-09-07
Remove duplicate create user bank on sign up

raymondjacobson:
### Description

Currently we create user banks twice on sign up most of the time because
1. createUserBank is performed async and might take some time
2. we call the onFetchAccount hook after sign up succeeds before the solana RPC would say that the user bank exists
3. we check if we need a user bank and try to create one. The relay then fails of course.

### Dragons

### How Has This Been Tested?

Locally vs. staging & dev

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/rj-remove-duplicate-create-user-bank

rickyrombo (approved):

PR #1869, opened 2022-09-07
[PAY-565] TokenValueInputV2 Component

rickyrombo:
### Description

Nobody asked for it but this component was bothering me so I made a replacement 😂 

- Adds a new `TokenValueInputV2` component to stems to replace `TokenValueInput`
  - **Fixes**:
    - `TokenValueInput` had a bug where because it parsed the input as a number it would lose precision for large numbers
    - `TokenValueInput` needed a lot of special casing for various styles since it used absolute positioning. The new version uses flexbox instead, making it actually aligned with the number in the text field and given the proper padding (see videos).
    - `TokenValueInput` lacked a proper label for accessibility. The new one enforces either `label` or `aria-label` are set
    - `TokenValueInput` didn't have a way to pass in a ref to the input element to autofocus it
    - `TokenValueInput` used to allow for non-numeric inputs, which is counter-intuitive given its name.
    - `TokenValueInput` used to allow for it to be a text-area, which is counter-intuitive given its name.
  - **New features**:
    - `TokenValueInputV2` automatically converts the input to a BN for the `onChange` handler
    - `TokenValueInputV2` supports a fixed decimal count (eg. use `decimals=8` for SPL $AUDIO)
  
 - Updates tip modal and on-ramp modal to use new component
 - Autofocuses the "Custom Amount" field in the on-ramp when the pill radio button gets selected

**Before:** Notice the numbers after a certain point are all 0s. Also the $AUDIO label doesn't align with the numbers.

https://user-images.githubusercontent.com/3690498/188813075-854409f2-ed54-4693-8b7d-55afa98887fe.mov

**After:** No precision loss and the $AUDIO label aligns well, even has some extra padding to make it cleaner!

https://user-images.githubusercontent.com/3690498/188813099-7c2a2b44-69ec-4c7f-af07-825e857cd1dc.mov

### Dragons

- Will require design pass
- Should we import common into stems? I'm using the `formatNumberCommas` method (slightly modified to allow decimals)

### How Has This Been Tested?

- Tested manually against stage

- N/A

### Feature Flags ###

- No feature flags


audius-infra:
Preview this change https://demo.audius.co/mjp-token-amount-input-v2

packages/stems/src/components/TokenValueInput/TokenValueInput.tsx
```diff
@@ -27,6 +27,9 @@ const formatValue = (value: string | undefined) => {
 
 const inputPaddingRight = 16
 
+/**
+ * @deprecated Use TokenValueInputV2 instead
+ */
```
raymondjacobson, line 32: nit, how do we feel about naming the new one TokenValueInput and renaming the old one instead?

packages/stems/src/components/TokenValueInputV2/TokenValueInputV2.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const ref = useRef<HTMLInputElement>(null)
+  return (
+    <TokenValueInputV2
+      {...args}
+      value={value}
+      inputRef={ref}
+      onChange={(value, valueBN) => {
+        setValue(value)
+        setValueBN(valueBN)
+        console.log({ value, valueBN: valueBN.toString() })
```
raymondjacobson, line 26: 🪵

packages/stems/src/components/TokenValueInputV2/TokenValueInputV2.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+import { useControlled } from 'hooks/useControlled'
+
+import styles from './TokenValueInputV2.module.css'
+import { TokenValueInputV2Props } from './types'
+
+/**
+ * Format a number to have commas
+ */
+const formatNumberCommas = (num: number | string) => {
```
raymondjacobson, line 15: do we want this in a common string util somewhere?

raymondjacobson (approved):

packages/stems/src/components/TokenValueInputV2/TokenValueInputV2.tsx
rickyrombo: It's in common already! Should I pull it in?

raymondjacobson: YUH

packages/stems/src/components/TokenValueInput/TokenValueInput.tsx
rickyrombo: What if I name the new one `TokenAmountInput` so there's no breaking change but it's still nicely named?

packages/stems/src/components/TokenValueInputV2/TokenValueInputV2.tsx
rickyrombo: Possible side effect - I'm changing it so that if there's a `.` then it includes a `.` - even if there's no decimal digits

raymondjacobson: I think that's probably ok! assuming this new behavior came with some design?

packages/stems/src/components/TokenValueInput/TokenValueInput.tsx
raymondjacobson: 👍 like that!

packages/stems/src/components/TokenValueInputV2/TokenValueInputV2.tsx
rickyrombo: Comes from not being able to type in a `.` without it. I think effectively this is the same?

raymondjacobson: 👍

audius-infra:
Preview this change https://demo.audius.co/mjp-token-amount-input-v2

audius-infra:
Preview this change https://demo.audius.co/mjp-token-amount-input-v2

packages/stems/src/components/TokenValueInputV2/TokenValueInputV2.tsx
rickyrombo: Eh, adding @audius/common to stems was a PITA. keeping this as-is for now until client folks can figure out the best way to integrate these.

audius-infra:
Preview this change https://demo.audius.co/mjp-token-amount-input-v2

PR #1874, opened 2022-09-08
Fix optimizely for local dev

piazzatron:
### Description

Previous change to default to stage didn't work because we would take the stage eth provider url instead of remote. This change checks which environment we're in, and uses that to determine which source to pull the eth provider from.

### Dragons

Doesn't seem like there are any other instances of remote config overriding remote dev env

### How Has This Been Tested?

Tested against stage + remote

### Feature Flags ###


raymondjacobson (approved):
im coo w/ dis. thx for comment.

wondering if possible to just have a dev optimizely env instead of the staging one? maybe that's not even worth the headache

audius-infra:
Preview this change https://demo.audius.co/piazz-fix-optimizely-local

packages/common/src/services/audius-backend/AudiusBackend.ts
```diff
@@ -717,8 +717,12 @@ export const audiusBackend = ({
   }
 
   function getEthWeb3Config() {
-    const providerUrls =
-      getRemoteVar(StringKeys.ETH_PROVIDER_URLS) || ethProviderUrls
+    // In a dev env, always ignore the remote var which is inherited from staging
+    const isDevelopment = process.env.REACT_APP_ENVIRONMENT === 'development'
```
dylanjeffers, line 721: So with Audius-backend we are trying to make it env agnostic. The pattern is to put this in the env.ts file in common, and pass env into Audius backend. Lmk if I can help here

dylanjeffers (changes requested):
Generally good, but let's get that car into env

packages/common/src/services/audius-backend/AudiusBackend.ts
raymondjacobson: ah good call. thx @dylanjeffers . i was not observant enough here

piazzatron:
> im coo w/ dis. thx for comment.
> 
> 
> 
> wondering if possible to just have a dev optimizely env instead of the staging one? maybe that's not even worth the headache

I think too painful to maintain consistent variables between optimizely envs - even keeping stage + prod in sync is annoying :|

packages/mobile/src/services/env.ts
```diff
@@ -1,3 +1,4 @@
 export const env = {
-  EAGER_DISCOVERY_NODES: process.env.REACT_APP_EAGER_DISCOVERY_NODES
```
dylanjeffers, line 2: really close! for mobile, its `Config.BLAH` where BLAH is the thing minus the REACT_APP, so `ENVIRONMENT: Config.ENVIRONMENT` but also don't trip much cause this is all fleshed out in the the native-reloaded branch actually

dylanjeffers (approved):
nicee

PR #1881, opened 2022-09-08
Dms

stereosteve:
### Description

Example POC code for dms.

Notes:

* might have to clear babel cache: `rm -rf node_modules/.cache` in `packages/web` before `npm i` in root.
* run `npm run start:stage`

open console... to send yourself a message:

* `myWallet = audiusLibs.hedgehog.wallet.getChecksumAddressString()`
* `dmSend(myWallet, 'hello self 1')`
* `dmGet()`

### Dragons

* case sensitivity of wallet is an issue... currently code will use the "checksum" mixed case version of wallets in places.  Things should still mostly work... but this could be a foot gun.  Will have to think about what the best option is here.
* ethereum [signing code](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/util/src/signature.ts#L34-L67) is all weird... have to re-add `04` byte prefix that they chop off.  I believe it will always be `04`
* You have to look up people's wallets manually atm... the first time the "public key recovery" can take several seconds depending on several factors... after initial lookup it will be fast.  I think there are some things we can do to make the initial lookup faster too.

See also PR for server side:

https://github.com/AudiusProject/audius-docker-compose/pull/68

### DIFF
--- a/packages/web/craco.config.ts
+++ b/packages/web/craco.config.ts
@@ -40,6 +40,12 @@ export default {
           rules: [
             ...(config.module?.rules ?? []),
             {
+              test: /\.m?js/,
+              resolve: {
+                fullySpecified: false
+              }
+            },
+            {
               test: /\.js$/,
               enforce: 'pre',
               use: ['source-map-loader']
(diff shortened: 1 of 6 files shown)


audius-infra:
Preview this change https://demo.audius.co/dms

PR #1891, opened 2022-09-09
[PAY-625] Playlist challenge optimistic claiming

dharit-tan:
### Description
Implements optimistic claiming for first-playlist challenge.

### How Has This Been Tested?
Tested on local mobile + web against staging.

### Feature Flags ###

Playlist challenge is already behind a feature flag.

### DIFF
--- a/packages/common/src/store/pages/audio-rewards/slice.ts
+++ b/packages/common/src/store/pages/audio-rewards/slice.ts
@@ -160,6 +160,13 @@ const slice = createSlice({
         delete userChallengeOverride.current_step_count
       }
     },
+    setPlaylistChallengeOptimisticCompleted: (state) => {
+      if (state.userChallengesOverrides['first-playlist'] !== undefined) {
+        state.userChallengesOverrides['first-playlist'].is_complete = true
+        return
+      }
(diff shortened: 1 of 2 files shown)


packages/common/src/store/pages/audio-rewards/slice.ts
```diff
@@ -277,7 +284,8 @@ export const {
   resetRewardClaimedToast,
   updateOptimisticListenStreak,
   setUserChallengeCurrentStepCount,
```
rickyrombo, line 286: thought just occurred to me - what happens when we set the step count to 1 for 1 step challenges 🤔  (probably ideal to have your action anyway but curious if that works)

rickyrombo (approved):

piazzatron (approved):

packages/common/src/store/pages/audio-rewards/slice.ts
dharit-tan: Good thought but doesn't seem to work for me.

audius-infra:
Preview this change https://demo.audius.co/rt-PAY-625

PR #1896, opened 2022-09-10
[C-1006] Improve initial load, hotfix missing web worker

dylanjeffers:
### Description
- Hotfixes breaking issue on testflight due to missing Worker which is injected with chrome debugger, classic....
- Adds splash-screen to initial load. there is still a long blank white section, but its better than rendering the sign-up screen


audius-infra:
Preview this change https://demo.audius.co/dj-c-tmp-fix-for-web-workers

raymondjacobson (approved):

PR #1909, opened 2022-09-13
[C-1026] Improve bottom-bar perf

dylanjeffers:
### Description

Improves bottom-bar performance by removing delay that we needed when we had  to deal with web-app syncing


sliptype (approved):
Oooo nice!

PR #1912, opened 2022-09-13
[C-1029, C-1036] Fix pull-to-refresh

dylanjeffers:
### Description

Fixes pull-to-refresh infinite loader
Also (potentially) improves api by building pull-to-refresh into the lineup, so consumers don't need to re-implement everything.


sliptype (approved):
This looks good, just wondering if/how this impacts profile screen pull to refresh?

dylanjeffers:
Oh yeah good q, will double check on that. It's should be backwards compatible tho

PR #1915, opened 2022-09-13
[C-1057] Fix blank edit-profile screen

dylanjeffers:
### Description

Fixes blank edit-profile screen. Weirdly updates are not happening consistently on staging, which im thinking may just be related to stage env, since it's also failing on web.


sliptype (approved):
Weird, so can we get rid of `getProfile` overall?

dylanjeffers:
> Weird, so can we get rid of `getProfile` overall?

yeah probably since it is so heavy with the rerenders

PR #1920, opened 2022-09-13
[C-1041] Fix infinite 404

sliptype:
### Description

* AudiusAPIClient was referencing `window.audiusLibs`, so when a 404 happened the recursion never ended because`audiusLibs` was always undefined

I was able to repro this by going to `hotelgaruda`'s profile on mobile

### Dragons

This does not enable eager load inside of AudiusBackend, but we are already doing eager load when AudiusAPIClient is used. Might be valuable to rethink this logic in general because if a request 404s, we keep retrying until AudiusLibs is initialized

Can enable eager load in AudiusBackend in the future

Can also dig into why we are requesting a deleted track in the first place

### How Has This Been Tested?

Tested both mobile and web, confirmed that 404 requests only fail until libs is initialized

### Feature Flags ###

### DIFF
--- a/packages/common/src/services/audius-api-client/AudiusAPIClient.ts
+++ b/packages/common/src/services/audius-api-client/AudiusAPIClient.ts
@@ -426,7 +426,7 @@ type GetUserSupporterArgs = {
 
 type AudiusAPIClientConfig = {
   audiusBackendInstance: AudiusBackend
-  audiusLibs?: AudiusLibs
+  getAudiusLibs: () => AudiusLibs
   overrideEndpoint?: string
   remoteConfigInstance: RemoteConfigInstance
   localStorage: LocalStorage
@@ -439,7 +439,7 @@ export class AudiusAPIClient {
   }
 
   audiusBackendInstance: AudiusBackend
-  audiusLibs?: AudiusLibs
+  getAudiusLibs: () => AudiusLibs
   overrideEndpoint?: string
   remoteConfigInstance: RemoteConfigInstance
   localStorage: LocalStorage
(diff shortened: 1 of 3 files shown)


packages/common/src/services/audius-api-client/AudiusAPIClient.ts
```diff
@@ -426,7 +426,7 @@ type GetUserSupporterArgs = {
 
 type AudiusAPIClientConfig = {
   audiusBackendInstance: AudiusBackend
-  audiusLibs?: AudiusLibs
+  getAudiusLibs: () => AudiusLibs
```
dylanjeffers, line 429: Nice

packages/common/src/services/audius-api-client/AudiusAPIClient.ts
```diff
@@ -1513,11 +1513,10 @@ export class AudiusAPIClient {
     }, {})
 
     const formattedPath = this._formatPath(pathType, path)
-    const audiusLibs =
-      this.audiusLibs ??
-      (this.initializationState.type === 'libs' && window.audiusLibs)
-    if (audiusLibs) {
-      const data = await audiusLibs.discoveryProvider._makeRequest(
+    const audiusLibs = this.getAudiusLibs()
```
dylanjeffers, line 1516: So in this case we aren't awaiting?

dylanjeffers (approved):
Great finds

packages/common/src/services/audius-api-client/AudiusAPIClient.ts
sliptype: I think the diff might look a little funky, but no we aren't awaiting libs because we want to make the request eagerly before it is initialized. `_makeRequest` is still awaited if audiusLibs exists

audius-infra:
Preview this change https://demo.audius.co/sk-fix-infinite-404

PR #1924, opened 2022-09-14
Fix account profile screen loading

sliptype:
### Description

#1916 made  it so the handle coming from params is used to fetch the profile, but on the user's own account the param value is `accountUser`. Need to check for this to allow the screen to load

### Dragons

We don't have anyone who actually has the handle `accountUser`, but I wonder if it would break?

### How Has This Been Tested?

### Feature Flags ###


dylanjeffers (approved):
Yup nice, I was dumb and thought it was just magically working :D

PR #1932, opened 2022-09-15
[C-1091] Improve selector performance

dylanjeffers:
### Description

Improves selector performance with a new library, [proxy-memoize](https://github.com/dai-shi/proxy-memoize). What this library does to improve our perf is to look at every distinct element in the state that we need for a specific selector, and strictly compare those for any changes. This is much better than what we currently have with reselect, which will trigger a re-render every time our cache changes for selectors that need to grab multiple items from it.

This is just a demo showing it's usage, and we should decide an approach. For collection-list-item, we went from ~80 rerenders to 5, so a 90% reduction! 

Moving forward some things to decide:
- Use this new `useProxySelector` hook for the selectors that require many things from the store
- Update some of our complex selectors directly in @audius/common, and see if it's possible to use the vanilla `useSelector` hook, or something similar.

### DIFF
--- a/packages/mobile/package.json
+++ b/packages/mobile/package.json
@@ -73,7 +73,7 @@
     "node-libs-react-native": "1.2.1",
     "numeral": "2.0.6",
     "path-dirname": "1.0.2",
-    "proxy-memoize": "^1.2.0",
+    "proxy-memoize": "1.2.0",
     "query-string": "7.0.1",
     "react": "17.0.2",
     "react-native": "0.66.3",
--- a/packages/mobile/package.json
+++ b/packages/mobile/package.json
@@ -73,6 +73,7 @@
     "node-libs-react-native": "1.2.1",
     "numeral": "2.0.6",
     "path-dirname": "1.0.2",
+    "proxy-memoize": "^1.2.0",
     "query-string": "7.0.1",
     "react": "17.0.2",
     "react-native": "0.66.3",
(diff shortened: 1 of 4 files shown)


packages/mobile/package.json
```diff
@@ -73,6 +73,7 @@
     "node-libs-react-native": "1.2.1",
     "numeral": "2.0.6",
     "path-dirname": "1.0.2",
+    "proxy-memoize": "^1.2.0",
```
raymondjacobson, line 76: pin

raymondjacobson (approved):
Curious if we want to use this approach or replace the selectors themselves at the "reselect" level. That way dev flow looks similar with useSelector still and the component doesn't need to think about it
https://github.com/dai-shi/proxy-memoize#reselect

packages/mobile/src/components/lineup-tile/CollectionTile.tsx
```diff
@@ -23,12 +23,17 @@ import {
   OverflowSource,
   mobileOverflowMenuUIActions,
   shareModalUIActions,
-  RepostType
+  RepostType,
+  Uid,
+  removeNullable
 } from '@audius/common'
+import { isEqual } from 'lodash'
```
sliptype, line 30: Unused import

sliptype (approved):

sliptype:
> Curious if we want to use this approach or replace the selectors themselves at the "reselect" level. That way dev flow looks similar with useSelector still and the component doesn't need to think about it https://github.com/dai-shi/proxy-memoize#reselect

We can do that in some cases, but because these ones need the `uid` argument we need to wrap it in a useCallback and useSelector anyway. So `useProxySelector` is just a shortcut for that

nicoback2 (approved):

PR #1938, opened 2022-09-16
[C-1023] [C-1039] Sync account entropy from webview and improve splashscreen

sliptype:
### Description

* Wait to set up backend until we either have account entropy or we've checked the webview for it in localStorage
  > This is written in a way where we can rip out `WebAppAccountSync` easily in the future
* Fix sign out by setting initial account status to `IDLE`
* Improve splash screen by layering it on top of the other screens and rendering null as soon as the animation is complete. The animation is triggered once the account status is known

### Dragons

### How Has This Been Tested?

Tested e2e by running the old webview version of the app locally, signing in, then upgrading to the new version and confirming that I was still signed in

### Feature Flags ###


dylanjeffers (approved):
Wow this is awesome work!

PR #1939, opened 2022-09-16
[C-296] Persist stack history across bottom tabs

dylanjeffers:
### Description

Trying out this new feature now that we don't need to have to deal with web navigation! Main thing to watch out for is perf going across bottom-tab bars


sliptype (approved):
Did you see a perf impact? Also I think we should call this out to Forrest before we merge just to confirm this is still behavior we want

dylanjeffers:
> Did you see a perf impact? Also I think we should call this out to Forrest before we merge just to confirm this is still behavior we want

No i didn't notice any major perf differences! and sure thing

PR #1957, opened 2022-09-20
[C-1114] Prevent track restart

sliptype:
### Description

This ended up being straightforward but I'm still not sure why the user reference equality breaks

* Refactor `Audio` to redux hooks
* Add equality function to `trackOwner` useSelector

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/audio/Audio.tsx
+++ b/packages/mobile/src/components/audio/Audio.tsx
@@ -1,6 +1,5 @@
 import { useState, useRef, useEffect, useCallback } from 'react'
 
-import type { CommonState } from '@audius/common'
 import {
   accountSelectors,
   cacheUsersSelectors,
@@ -17,8 +16,7 @@ import MusicControl from 'react-native-music-control'
 import { Command } from 'react-native-music-control/lib/types'
 import type { OnProgressData } from 'react-native-video'
 import Video from 'react-native-video'
-import { connect, useSelector } from 'react-redux'
-import type { Dispatch } from 'redux'
+import { useDispatch, useSelector } from 'react-redux'
 
 import { audiusBackendInstance } from 'app/services/audius-backend-instance'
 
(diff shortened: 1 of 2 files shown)


packages/mobile/src/components/audio/Audio.tsx
```diff
@@ -59,28 +57,36 @@ const styles = StyleSheet.create({
...
+export const Audio = () => {
+  const track = useSelector(getCurrentTrack)
+  const index = useSelector(getIndex)
+  const queueLength = useSelector(getLength)
+  const playing = useSelector(getPlaying)
+  const seek = useSelector(getSeek)
+  const repeatMode = useSelector(getRepeat)
+  const isShuffleOn = useSelector(getShuffle)
+  const shuffleIndex = useSelector(getShuffleIndex)
+  const trackOwner = useSelector(
```
dylanjeffers, line 69: Yeah a bit confusing there. And so useproxyselector also doesn't do anything? What does this mean... that the users cache is updating when going to a user that has reposted the track your playing? Or going to the user of the playing track?

dylanjeffers (approved):
Do we only use the track owner for creator-node-endpoint? Maybe we can just select the specific things we need to avoid a rerender when that user is updated some way in the app?

packages/mobile/src/components/audio/Audio.tsx
sliptype: Yeah the users cache is updating when going to the artist who posted the track. We are using it for the gateway and the name that displays on the music control. But let me try selecting those things specifically to avoid this

dylanjeffers: going to merge this in for now, and then we can figure out why profile's gateways are changing

sliptype: Cool, it has to do with the collectibles fetch. We are getting diffent metadata back from that call and updating the user in cache. Might have it figured out by Monday!

PR #1961, opened 2022-09-21
[C-1061] Remove filter skeletons from new tables

Kyle-Shanks:
### Description
Remove the extra skeletons from the new tables when filtering in a virtualized table

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A


packages/web/src/common/store/pages/saved/sagas.js
```diff
@@ -76,15 +76,18 @@ function* watchFetchSaves() {
...
+          .fill(0)
+          .map((_) => ({}))
 
         fullSaves.splice(offset, saves.length, ...saves)
 
-        yield put(actions.fetchSavesSucceeded(fullSaves))
+        yield put(
+          actions.fetchSavesSucceeded(newTablesEnabled ? fullSaves : saves)
+        )
+        if (limit && saves.length < limit) {
```
sliptype, line 88: Should this be like this to match the one below?

```suggestion
        if (limit > 0 && saves.length < limit) {
```

packages/web/src/pages/saved-page/SavedPageProvider.tsx
```diff
@@ -101,6 +101,7 @@ class SavedPage extends PureComponent<SavedPageProps, SavedPageState> {
   }, 300)
 
   handleFetchMoreSavedTracks = (offset: number, limit: number) => {
+    if (this.props.endFetchStatus) return
```
sliptype, line 104: "If endFetchStatus" seems kind of vague, maybe would be good to name it something like `hasReachedEnd` or `hasReachedEndOfFetchableItems` if we wanna be really verbose. Or we could add a comment here to explain it a bit

sliptype (approved):
Good stuff!

PR #1962, opened 2022-09-21
[C-1133] Prevent toasts from appearing when overflow menu is clicked

sliptype:
### Description

Because we are no longer doing `stopPropagation` on click in the `Card` component (https://github.com/AudiusProject/audius-client/commit/f3c02b659d40b15ed6628654fa0dc3bcc0d0186e) any card wrapped in a `Toast` component had toasts appearing when the card/menu was clicked. The only cases of this were on the SearchResults page

Fix is simply preventing the toasts from being trigger via the `firesOnClick` prop

Also uncovered a larger bug that already exists in prod: The toasts don't show up at all from the actions in the overflow menu on SearchResults. For example, clicking "share" copies the link to clipboard but there is no toast or visual feedback. https://linear.app/audius/issue/C-1138/no-tooltips-on-overflow-menu-options-on-search-results Can address this in a future cycle

### Dragons

This `Toast` pattern is gross and we should really standardize on using the `ToastContext`. Right now it only supports top level toasts but it would be easy to pass a ref into it to position a toast on a nested element

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/web/src/pages/search-page/components/desktop/SearchPageContent.js
+++ b/packages/web/src/pages/search-page/components/desktop/SearchPageContent.js
@@ -148,6 +148,7 @@ class SearchPageContent extends Component {
           open={cardToast[toastId] && cardToast[toastId].open}
           placement='bottom'
           fillParent={false}
+          firesOnClick={false}
         >
           <Card
             id={artist.user_id}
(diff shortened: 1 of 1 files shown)


audius-infra:
Preview this change https://demo.audius.co/sk-c-1133-toast-overflow

dylanjeffers (approved):
Nice fix and finds of other issues

PR #1963, opened 2022-09-21
[C-1090] Improve profile store layout

dylanjeffers:
### Description

- Improves profile store layout, by allowing multiple profile page states to be saved at once.
- This improves mobile profile navigation, since we don't have to reset profile state, which causes poor navigation performance and ui.
- Written in a backwards-compatible way, so web doesn't need to do anything different. Eventually we may want to apply this new system to web though.
- Updates lineup default state from .LOADING to .IDLE

### DIFF
--- a/packages/common/src/store/pages/profile/reducer.ts
+++ b/packages/common/src/store/pages/profile/reducer.ts
@@ -28,10 +28,10 @@ import {
 import { PREFIX as feedPrefix } from './lineups/feed/actions'
 import { PREFIX as tracksPrefix } from './lineups/tracks/actions'
 
-const initialState = {
+const initialProfileState = {
   handle: null,
   userId: null,
-  status: Status.LOADING,
+  status: Status.IDLE,
 
   isNotificationSubscribed: false,
   updating: false,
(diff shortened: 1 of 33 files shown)


packages/common/src/store/pages/profile/reducer.ts
```diff
@@ -43,86 +43,169 @@ const initialState = {
...
+
 const actionsMap = {
   [FETCH_PROFILE](state, action) {
-    if (action.fetchOnly) return state
+    const { fetchOnly, shouldSetLoading, handle, userId } = action
+    if (fetchOnly) return state
+
+    const profileState = state.entries[handle] ?? initialProfileState
 
     const newState = {
```
sliptype, line 60: Could rename this to `newEntry` for clarity

packages/common/src/store/pages/profile/reducer.ts
```diff
@@ -43,86 +43,169 @@ const initialState = {
...
+
+    const profileState = entries[profileHandle]
+
     return {
       ...state,
-      [action.followerGroup]: {
-        userIds:
-          state[action.followerGroup].userIds.concat(filteredAddedUserIds),
-        status: Status.SUCCESS
+      entries: {
```
sliptype, line 128: Starting to look like we should convert this to redux toolkit so that we can leverage immer

packages/mobile/src/screens/profile-screen/RepostsTab.tsx
```diff
@@ -1,34 +1,42 @@
...
 
 const { getProfileFeedLineup } = profilePageSelectors
-const { makeGetLineupMetadatas } = lineupSelectors
-
-const getUserFeedMetadatas = makeGetLineupMetadatas(getProfileFeedLineup)
 
 export const RepostsTab = () => {
-  const lineup = useSelector(getUserFeedMetadatas)
+  const { handle } = useSelectProfile(['handle'])
+  // const lineup = useSelector((state) => getProfileFeedLineup(state, handle))
```
sliptype, line 18: Extra comment

packages/web/src/pages/profile-page/ProfilePageProvider.tsx
```diff
@@ -146,9 +141,6 @@ class ProfilePage extends PureComponent<ProfilePageProps, ProfilePageState> {
         getPathname(this.props.location) !== getPathname(location) ||
         action === 'POP'
       ) {
-        this.props.resetProfile()
```
sliptype, line 149: So even on web we don't have to reset anymore?

sliptype (approved):
This is awesome! Super clean and makes sense

packages/common/src/store/pages/profile/reducer.ts
dylanjeffers: totes

packages/common/src/store/pages/profile/reducer.ts
dylanjeffers: yeah def agreed

packages/web/src/pages/profile-page/ProfilePageProvider.tsx
dylanjeffers: yeah cause the "currentUser" will go to that users state, so we don't get weird overlap!

sliptype: Hell yeah, so nice

audius-infra:
Preview this change https://demo.audius.co/dj-c-1090-improve-profile-store-layout

PR #1965, opened 2022-09-21
[C-978] Check for track repeat before advancing shuffle

amendelsohn:
### Description

We were failing to repeat a single track when shuffle was toggled on. This is because the logic to advance the shuffle index was happening before the repeat single track check in the reducer.

### Dragons
N/A

### How Has This Been Tested?
Manually. Unit test added to cover this behavior in the future, but the test itself couldn't be tested due to broken jest setup on main.

N/A

### Feature Flags ###
N/A


dylanjeffers (approved):
awesome fix!

PR #1967, opened 2022-09-21
[C-1137] Add debounce to change in NetworkInfo status

amendelsohn:
### Description

Prevent offline placeholder from flashing on and off when network status is changing rapidly.

We will wait 2 sec before making a call to `updateConnectivity` and only process the last call within the 5 second window.

### How Has This Been Tested?

iOS simulator

### DIFF
--- a/packages/mobile/src/utils/connectivity.ts
+++ b/packages/mobile/src/utils/connectivity.ts
@@ -29,6 +29,4 @@ const updateConnectivity = (state: NetInfoState) => {
   }
 }
 
-NetInfo.addEventListener(
-  debounce(updateConnectivity, 2000, { maxWait: 5000 })
-)
+NetInfo.addEventListener(debounce(updateConnectivity, 2000, { maxWait: 5000 }))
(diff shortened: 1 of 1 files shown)


packages/mobile/src/utils/connectivity.ts
```diff
@@ -17,12 +18,17 @@ export const checkConnectivity = (netInfo: NetInfoState | null) => {
...
   if (!newValue) {
     dispatch(reachabilityActions.setUnreachable())
   } else {
     dispatch(reachabilityActions.setReachable())
   }
-})
+}
+
+NetInfo.addEventListener(
+  debounce((state) => updateConnectivity(state), 2000, { maxWait: 5000 })
```
sliptype, line 33:
```suggestion
  debounce(updateConnectivity, 2000, { maxWait: 5000 })
```

sliptype (approved):
Nice!

nicoback2 (approved):

packages/mobile/src/utils/connectivity.ts
amendelsohn: Duh. Ty

PR #1970, opened 2022-09-21
[C-951] Fix "track not found" error in collection tiles

sliptype:
### Description

This is a fun one! Was occasionally getting a ton of "track not found" errors:
image

Turns out that there is a default limit of 100 on the `getTracks` endpoint https://github.com/AudiusProject/audius-protocol/blob/master/discovery-provider/src/queries/query_helpers.py#L40

We are using that endpoint to fetch all the tracks in collections in a lineup. We pass a big long list of ids and sometimes that list is longer than 100 👀. So we get into a state where collection tracks are only partially loaded on the lineup. 

Fixed by specifying the limit as the number of ids we are requesting. Would probably be more elegant to remove the limit on the `getTracks` endpoint, because why limit the response if we are passing a list of ids? But there is some extra complexity in that endpoint because it handles multiple cases. Also would be an api change but I'm open to thoughts here!

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/common/src/services/audius-api-client/AudiusAPIClient.ts
+++ b/packages/common/src/services/audius-api-client/AudiusAPIClient.ts
@@ -811,7 +811,8 @@ export class AudiusAPIClient {
     const encodedCurrentUserId = encodeHashId(currentUserId)
     const params = {
       id: encodedTrackIds,
-      user_id: encodedCurrentUserId || undefined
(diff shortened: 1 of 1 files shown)


dylanjeffers (approved):
oh wow, didn't think it would be here!

PR #1973, opened 2022-09-22
[Reloaded] Keep relevant search results in store c-1117

nicoback2:
### Description
-Improve search results screen performance
Before: When you switched tabs, it would refetch the results even if it was already fetched before
After: We keep and reuse old results for same query

Regular search:
earchy

Tag search:
earchy2

### Dragons

### How Has This Been Tested?
Tested in mobile and tested for regressions on web.

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/screens/search-results-screen/tabs/SearchResultsTab.tsx
+++ b/packages/mobile/src/screens/search-results-screen/tabs/SearchResultsTab.tsx
@@ -1,13 +1,11 @@
 import type { ReactNode } from 'react'
-import { useContext, useEffect, useState } from 'react'
 
-import { Status, searchResultsPageSelectors } from '@audius/common'
+import { searchResultsPageSelectors, Status } from '@audius/common'
 import { useSelector } from 'react-redux'
 
 import { WithLoader } from 'app/components/with-loader/WithLoader'
 
 import { EmptyResults } from '../EmptyResults'
-import { SearchFocusContext } from '../SearchFocusContext'
 const { getSearchStatus } = searchResultsPageSelectors
 
 type SearchResultsTabProps = {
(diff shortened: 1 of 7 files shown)


packages/mobile/src/screens/search-results-screen/tabs/SearchResultsTab.tsx
```diff
@@ -18,29 +16,12 @@ type SearchResultsTabProps = {
 
 export const SearchResultsTab = (props: SearchResultsTabProps) => {
   const { children, noResults, status } = props
-  const { isFocused } = useContext(SearchFocusContext)
   const searchStatus = useSelector(getSearchStatus)
-  const [isRefreshing, setIsRefreshing] = useState(true)
-
-  useEffect(() => {
-    if (!isFocused) {
-      // Prevents a large rerender in the middle of a stack navigation.
```
dylanjeffers, line 27: great catch here as well

packages/mobile/src/screens/search-results-screen/tabs/TracksTab.tsx
```diff
@@ -25,13 +24,6 @@ const getSearchTracksLineupMetadatas = makeGetLineupMetadatas(
 export const TracksTab = () => {
   const lineup = useSelector(getSearchTracksLineupMetadatas)
   const dispatch = useDispatch()
-  useFocusEffect(
-    useCallback(() => {
-      return () => {
-        dispatch(tracksActions.reset())
```
dylanjeffers, line 31: so we want to call reset?

dylanjeffers (approved):
woah this is all so sick! so does it work with desktop as well?

audius-infra:
Preview this change https://demo.audius.co/nkang--optimize-search

packages/mobile/src/screens/search-results-screen/tabs/TracksTab.tsx
nicoback2: We don't need to call it here because otherwise it'll clear the lineup even if you're moving between tabs (we want to keep it in case you come back). Tested thoroughly and there doesn't seem to be any issues with leaving this out altogether.

nicoback2:
> woah this is all so sick! so does it work with desktop as well?

it just refetches on desktop still since i didn't touch that code :o but i made sure there weren't any regressions from what's on prod!

PR #1975, opened 2022-09-22
[C-1152] Improve favorites tab performance

dylanjeffers:
### Description

Improves favorites tab performance with custom collection selectors and use of useProxySelector
Cuts down re-renders by 3x

### DIFF
--- a/packages/mobile/src/screens/favorites-screen/AlbumsTab.tsx
+++ b/packages/mobile/src/screens/favorites-screen/AlbumsTab.tsx
@@ -17,6 +17,7 @@ import { VirtualizedScrollView } from 'app/components/core'
 
 import { EmptyTab } from './EmptyTab'
 import { FilterInput } from './FilterInput'
+import { getAccountCollections } from './selectors'
 
 const { fetchSavedAlbums } = accountActions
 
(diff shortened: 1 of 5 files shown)


packages/mobile/src/screens/favorites-screen/AlbumsTab.tsx
```diff
@@ -60,7 +57,7 @@ export const AlbumsTab = () => {
           <CollectionList
             listKey='favorites-albums'
             scrollEnabled={false}
-            collection={userAlbums ?? []}
+            collection={(userAlbums as UserCollection[]) ?? []}
```
dylanjeffers, line 60: Having to cast this unnecessarily, since collection list assumes UserCollection, even though it only needs that when it's a "user" type card. we could probably improve this typing after reloaded

packages/mobile/src/screens/favorites-screen/PlaylistsTab.tsx
```diff
@@ -25,29 +32,21 @@ const messages = {
...
+  }, [dispatch])
 
-  const matchesFilter = (playlist: ExtendedCollection) => {
-    const matchValue = filterValue.toLowerCase()
-    return (
-      playlist.playlist_name.toLowerCase().indexOf(matchValue) > -1 ||
-      playlist.ownerName.toLowerCase().indexOf(matchValue) > -1
-    )
-  }
+  useEffectOnce(handleFetchSavedPlaylists)
```
dylanjeffers, line 41: I cant use `usefocusEffect` since this tab would lazy load, and we don't have a good loading indicator for this (fetching collections doesn't have a status associated with it) this would be a good improvement though eventually, since might as well reduce fetches as much as we can

sliptype (approved):
Awesome, looks good to me!

PR #1985, opened 2022-09-23
[C-1048] Reset scrubber on song repeat

dylanjeffers:
### Description

Fixes bug where scrubber doesn't reset when a song is on repeat. This is because the mediaKey was only changing when track_id changed. In reality though, we should actually just look at playCount.


packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
```diff
@@ -163,16 +164,16 @@ const NowPlayingDrawer = ({ translationAnim }: NowPlayingDrawerProps) => {
...
     getUser(state, track ? { id: track.owner_id } : {})
   )
-
-  const trackId = track?.track_id
   const [mediaKey, setMediaKey] = useState(0)
+
   useEffect(() => {
     setMediaKey((mediaKey) => mediaKey + 1)
-  }, [trackId])
+  }, [playCounter])
```
raymondjacobson, line 176: ;) noice

raymondjacobson (approved):

PR #1989, opened 2022-09-23
Fix TopSupporters, TrackScreen caching

dylanjeffers:
### Description

Fix TopSupporters runtime error when more than 6 supporters
Fix TrackScreen lineup cache issues (track screen lineups dont reset)

### DIFF
--- a/packages/mobile/src/screens/track-screen/TrackScreen.tsx
+++ b/packages/mobile/src/screens/track-screen/TrackScreen.tsx
@@ -1,11 +1,11 @@
-import { useEffect } from 'react'
+import { useCallback } from 'react'
 
 import {
-  lineupSelectors,
   trackPageLineupActions,
   trackPageActions,
   trackPageSelectors
 } from '@audius/common'
+import { useFocusEffect } from '@react-navigation/native'
 import { Text, View } from 'react-native'
 import { useDispatch, useSelector } from 'react-redux'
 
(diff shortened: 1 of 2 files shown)


packages/mobile/src/screens/track-screen/TrackScreen.tsx
```diff
@@ -68,16 +65,16 @@ export const TrackScreen = () => {
...
+    dispatch(tracksActions.reset())
     dispatch(fetchTrack(id, undefined, user?.handle, canBeUnlisted))
-    return () => {
-      dispatch(resetTrackPage())
-      dispatch(tracksActions.reset())
-    }
-  }, [canBeUnlisted, dispatch, id, trackUserExists, user?.handle])
+  }, [dispatch, canBeUnlisted, id, user?.handle])
+
+  useFocusEffect(handleFetchTrack)
```
raymondjacobson, line 77: great. call.

packages/mobile/src/screens/track-screen/TrackScreen.tsx
```diff
@@ -152,7 +149,6 @@ export const TrackScreen = () => {
         lineup={lineup}
         start={1}
         includeLineupStatus
-        selfLoad
```
raymondjacobson, line 155: does this impact scrolling pagination behavior? (i guess not necessary for track screen lineup anyway)

raymondjacobson (approved):

packages/mobile/src/screens/track-screen/TrackScreen.tsx
dylanjeffers: yeah exactly, we limit to 6 so shouldnt ever need to "loadMore", and the first 6 tracks get fetched in the `fetchTrack` saga chain. this was actually leading to a runtime error since it was fetching, but didnt have the right data, since it needs ownerHandle, etc which are calculated and used in saga's call

PR #1991, opened 2022-09-24
Update react-native-screens patchfile

dylanjeffers:
### Description

Issue where react-native-screens patchfile wasn't applying which breaks build. This is due to us upgrading it, but not "upgrading" the patchfile, since it needs to specify version and gitsha in the diff


raymondjacobson (approved):
https://www.youtube.com/watch?v=nLfsrNvclbg

PR #1993, opened 2022-09-24
[C-536] Cache makeStyles StyleSheets

dylanjeffers:
### Description

Drastically reduces calls (up to 10x less) to `Stylesheet.create` by using style cache. This majorly improves renders for small items, like track-tiles that would otherwise each create their own stylesheet.

### DIFF
--- a/packages/mobile/package.json
+++ b/packages/mobile/package.json
@@ -72,6 +72,7 @@
     "moment": "2.24.0",
     "node-libs-react-native": "1.2.1",
     "numeral": "2.0.6",
+    "object-hash": "^3.0.0",
     "path-dirname": "1.0.2",
     "proxy-memoize": "1.2.0",
     "query-string": "7.0.1",
--- a/packages/mobile/package.json
+++ b/packages/mobile/package.json
@@ -72,7 +72,7 @@
     "moment": "2.24.0",
     "node-libs-react-native": "1.2.1",
     "numeral": "2.0.6",
-    "object-hash": "^3.0.0",
+    "object-hash": "3.0.0",
     "path-dirname": "1.0.2",
     "proxy-memoize": "1.2.0",
     "query-string": "7.0.1",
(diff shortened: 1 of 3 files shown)


packages/mobile/package.json
```diff
@@ -72,6 +72,7 @@
     "moment": "2.24.0",
     "node-libs-react-native": "1.2.1",
     "numeral": "2.0.6",
+    "object-hash": "^3.0.0",
```
raymondjacobson, line 75: pin!

packages/mobile/src/styles/makeStyles.ts
```diff
@@ -61,7 +64,14 @@ export const makeStyles = <PropsT, T extends NamedStyles<T> = NamedStyles<any>>(
     const stylesheet = useMemo(() => {
       const theme = { palette, typography, spacing, type: themeVariant }
       const namedStyles = styles(theme, memoizedProps)
-      return StyleSheet.create(namedStyles)
+      const namedStylesHash = hash(namedStyles)
```
raymondjacobson, line 67: this is cool. 

is it worth us thinking about moving styles out of the render cycle? (not sure how we would do theme changes but maybe possible)

raymondjacobson (approved):

packages/mobile/package.json
dylanjeffers: omg the number of times...

packages/mobile/src/styles/makeStyles.ts
dylanjeffers: yeah i would love to do that, but how do we handle theming? there doesn't seem to be a good pattern online for using Stylesheet directly like this, so might be worth considering emotionjs or some other css-in-jss-native option. long term question though imo.

raymondjacobson: yes 100%

PR #1994, opened 2022-09-24
[C-1176] Improve ExploreScreen performance

dylanjeffers:
### Description

Some QoL updates for explore screen. Reduces re-renders by 5x. There is a lot more work here we can do though, by delaying fetching playlists/artist tab data until you actually go to that tab. Story added for this after reloaded.

### DIFF
--- a/packages/common/src/store/pages/explore/selectors.ts
+++ b/packages/common/src/store/pages/explore/selectors.ts
@@ -11,24 +11,16 @@ const getExplore = (state: CommonState) => state.pages.explore
 
 export const getExplorePlaylists = createSelector(
   (state: CommonState) => state.pages.explore.playlists,
-  getCollections,
-  getUsers,
-  (playlists, collections, users) => {
-    const explorePlaylists = playlists
-      .map((id) => collections[id])
-      .filter(Boolean)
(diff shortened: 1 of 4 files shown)


packages/common/src/store/pages/explore/selectors.ts
```diff
@@ -11,24 +11,16 @@ const getExplore = (state: CommonState) => state.pages.explore
...
-    const explorePlaylists = playlists
-      .map((id) => collections[id])
-      .filter(Boolean)
-      .map((collection) => ({
-        ...collection,
-        user: users[collection.playlist_owner_id] || {}
-      }))
-    return explorePlaylists
-  }
+  (state: CommonState) => state.collections.entries,
```
raymondjacobson, line 14: aren't we losing associated users here?

dylanjeffers: So how does that affect the tile? It seems like the user is not referenced on the tile itself?

raymondjacobson: oh you may be right. only thing i can think of is needing the handle for the click.

raymondjacobson (approved):

PR #1995, opened 2022-09-25
[C-1177] Move tipping requests to api client

raymondjacobson:
### Description

Makes a pretty substantial difference in the loading of the tip tile.
Part 2 of this will be to support the audiuslibs eager stuff in native, but this is the bigger low-hanging fruit

### Dragons

### How Has This Been Tested?

Locally vs. staging via web & mobile

### Feature Flags ###


packages/common/src/services/audius-api-client/AudiusAPIClient.ts
```diff
@@ -1444,6 +1474,71 @@ export class AudiusAPIClient {
     return adapted
   }
 
+  async getSupporting({ userId, limit = 25, offset = 0 }: GetSupportingArgs) {
+    console.log('get supporting')
```
dylanjeffers, line 1478: lawg

packages/web/src/common/store/tipping/sagas.ts
```diff
@@ -322,90 +319,86 @@ function* refreshSupportAsync({
...
+    [apiClient, apiClient.getSupporting],
+    supportingParams
+  )
+  const supportersForReceiverList = yield* call(
+    [apiClient, apiClient.getSupporters],
+    supportersParams
+  )
 
-    yield call(fetchUsers, userIds)
+  if (!supportingForSenderList || !supportersForReceiverList) return
```
dylanjeffers, line 354: this makes sense, but want to confirm that early exit is all good

packages/web/src/common/store/tipping/sagas.ts
```diff
@@ -434,10 +423,9 @@ function* fetchSupportingForUserAsync({
     account?.user_id === userId
       ? account.supporting_count
       : MAX_ARTIST_HOVER_TOP_SUPPORTING + 1
-  const supportingList = yield* call(fetchSupporting, {
-    encodedUserId,
-    limit,
-    audiusBackendInstance
+  const supportingList = yield* call([apiClient, apiClient.getSupporting], {
```
dylanjeffers, line 426: this makes sense, just so ya know ive mostly been doing `[apiClient, 'getSupporting']`, maybe a good norming thing? v minor tho

dylanjeffers (approved):
looks solid to me! lets get someone on money squad to review as well?

packages/web/src/common/store/tipping/sagas.ts
raymondjacobson: i like the member vs the string because IDE can click into it 😇

raymondjacobson:
@dylanjeffers can you take a quick pass again? i updated the .dev.tmpl to test this and then just realized we can remove those tmpl files now!

packages/web/src/common/store/tipping/sagas.ts
raymondjacobson: yep should be, but will wait for @sddioulde too. we try/caught things before and just didn't dispatch later actions

raymondjacobson:
Looking into type issues. Odd that CI passed before

audius-infra:
Preview this change https://demo.audius.co/rj-c-1177

packages/web/src/common/store/tipping/sagas.ts
sddioulde: hmm im not sure this early return is correct. it's saying if there is either no supporting for sender or no supporters for receiver then return. What if there is no supporting for sender but there are supporters for the receiver?

sddioulde (commented):
looks good overall but not sure about the early return

packages/web/src/common/store/tipping/sagas.ts
raymondjacobson: I really don't think this is an issue because apiclient will only return a nullable response if the request errors out. Otherwise response.data = [] and this will still be truthy.

I switched to an inline ternary just in case, but I think this is good.

PR #2000, opened 2022-09-26
[C-1167,C-1171] Fix missing tip data, fix tipping

dylanjeffers:
### Description

Fixes tipping header and tipping flow due to missing user data. relying on `useProfile` was incorrect because that removes all elements that are not specifically requested in the dependencies

### DIFF
--- a/packages/mobile/src/components/feed-tip-tile/SendTipButton.tsx
+++ b/packages/mobile/src/components/feed-tip-tile/SendTipButton.tsx
@@ -47,7 +47,8 @@ type SendTipButtonProps = {
   receiver: User
 }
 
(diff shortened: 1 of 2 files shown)


packages/mobile/src/components/feed-tip-tile/SendTipButton.tsx
```diff
@@ -47,7 +47,8 @@ type SendTipButtonProps = {
   receiver: User
 }
 
-export const SendTipButton = ({ receiver }: SendTipButtonProps) => {
+export const SendTipButton = (props: SendTipButtonProps) => {
+  const { receiver } = props
```
nicoback2, line 51: JC, why this change?

nicoback2 (approved):
amazing find

PR #2001, opened 2022-09-27
[C-1150, C-1165] Fix app-tab navigation

dylanjeffers:
### Description

- Fixes issues with navigation in context outside the bottom-bar tabs
- Removes redundant overflow menu options

### DIFF
--- a/packages/mobile/src/components/lineup-tile/CollectionTile.tsx
+++ b/packages/mobile/src/components/lineup-tile/CollectionTile.tsx
@@ -8,6 +8,7 @@ import type {
   CommonState
 } from '@audius/common'
 import {
+  removeNullable,
   useProxySelector,
   playerSelectors,
   FavoriteSource,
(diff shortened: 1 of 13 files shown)


packages/mobile/src/components/lineup-tile/CollectionTile.tsx
```diff
@@ -160,19 +161,13 @@ const CollectionTileComponent = ({
...
-        ? OverflowAction.UNFAVORITE
-        : OverflowAction.FAVORITE,
       is_album
         ? OverflowAction.VIEW_ALBUM_PAGE
         : OverflowAction.VIEW_PLAYLIST_PAGE,
       isOwner && !is_album ? OverflowAction.PUBLISH_PLAYLIST : null,
       isOwner && !is_album ? OverflowAction.DELETE_PLAYLIST : null,
       OverflowAction.VIEW_ARTIST_PAGE
-    ].filter(Boolean) as OverflowAction[]
+    ].filter(removeNullable)
```
raymondjacobson, line 170: nice

packages/mobile/src/screens/app-screen/AppTabNavigationProvider.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export const AppTabNavigationContext =
+  createContext<AppTabNavigationContextValue>({
+    navigation: null,
+    setNavigation: () => {}
+  })
+
+type AppTabNavigationProviderProps = { children: ReactNode }
+
+export const AppTabNavigationProvider = (
```
raymondjacobson, line 24: Kinda nitty, but I'm wondering if we should use this opportunity to call it something else besides Tabs which is overloaded with the top tabs swipe thing.

Maybe just AppNavigatorProvider? or BottomBarNavigatorProvider?

raymondjacobson (approved):
nice so many little cleanups!

packages/mobile/src/screens/app-screen/AppTabNavigationProvider.tsx
dylanjeffers: yeah im down for an update, we have AppTabs still, so that's why this is named the way it is. I'll add a tech-debt item to address all the references to AppTabs? and do samethnig like BottomTabNavigator/AppNavigator or something?

raymondjacobson: sounds good!

PR #2005, opened 2022-09-27
[C-1164] Fix collections screen loading issues

dylanjeffers:
### Description

Fixes collection screen track list loading issues. Essentially we weren't reseting collection state when entering a new collection screen.


sliptype (approved):
Awesome!

PR #2010, opened 2022-09-27
[C-1181] skip connection check if not in foreground

amendelsohn:
### Description

Skip updates to connectivity state when in background.
Manually trigger connectivity state update when app returns to foreground.

### Dragons

If we have background processes that rely on the connectivity state, we may need to rethink this. Offline mode comes to mind.

### How Has This Been Tested?

Tested on iPhone 11

### DIFF
--- a/packages/mobile/src/App.tsx
+++ b/packages/mobile/src/App.tsx
@@ -24,7 +24,7 @@ import { refreshConnectivity } from 'app/utils/connectivity'
 import { Drawers } from './Drawers'
 import ErrorBoundary from './ErrorBoundary'
 import { NotificationReminder } from './components/notification-reminder/NotificationReminder'
-import useAppState from './hooks/useAppState'
+import { useEnterForeground } from './hooks/useAppState'
 
 Sentry.init({
   dsn: Config.SENTRY_DSN
@@ -62,14 +62,9 @@ const App = () => {
     refreshConnectivity()
   })
 
-  useAppState(
-    // onEnterForeground
-    () => {
-      refreshConnectivity()
-    },
-    // onEnterBackground
-    () => {}
-  )
+  useEnterForeground(() => {
+    refreshConnectivity()
+  })
 
   return (
     <SafeAreaProvider>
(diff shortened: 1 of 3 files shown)


packages/mobile/src/utils/connectivity.ts
```diff
@@ -21,6 +22,7 @@ export const Connectivity: { netInfo: NetInfoState | null } = { netInfo: null }
 
 const updateConnectivity = (state: NetInfoState) => {
   Connectivity.netInfo = state
+  if (!(AppState.currentState === 'active')) return
```
dylanjeffers, line 25: !==

packages/mobile/src/App.tsx
```diff
@@ -58,6 +58,19 @@ const App = () => {
     setIsReadyToSetupBackend(!entropy)
   }, [])
 
+  useEffectOnce(() => {
+    refreshConnectivity()
+  })
+
+  useAppState(
```
dylanjeffers, line 65: Could we update this hook to not require both, if we only need to add callback for one of them?

packages/mobile/src/App.tsx
```diff
@@ -58,6 +58,19 @@ const App = () => {
     setIsReadyToSetupBackend(!entropy)
   }, [])
 
+  useEffectOnce(() => {
```
dylanjeffers, line 61: I think this is fine, but maybe we could add this into a saga? I think we do a similar thing with the mobile theme saga to avoid having one-off effects in the root?

packages/mobile/src/App.tsx
dylanjeffers: Also considering we have this here too, i wonder if this could also be moved to a saga. Really not a big deal though, maybe this is a good topic for our sync this week

dylanjeffers (approved):
Nice nice

packages/mobile/src/App.tsx
amendelsohn: Updated the hook. Let me tinker with the saga approach. I need to learn more about them anyway :)

packages/mobile/src/App.tsx
sliptype: Well we do have the entropy / backend setup effect here too. I kinda prefer keeping things out of sagas but either way works

sliptype (approved):
Looks good to me!

Kyle-Shanks (approved):

PR #2021, opened 2022-09-28
[C-1065][C-1066] Primitive download and playback for tracks offline

amendelsohn:
### Description

**Checkpoint for offline mode including:**
1. Can manually trigger download of all favorites
2. Stores mp3, album art, and track json in caches directory that the os allocates for Audius app
3. If available, playback will select the stored audio file instead of streaming it

_Note: This is not production-ready code, and much of this may be refactored as we encounter other concerns with download and playback. Looking for early feedback and concerns._

**Known issues:**
1. We aren't yet loading the track json back into state, so even if a track's information is available offline, we won't be able to load and play it back in a subsequent session
2. There are some debugging methods checked in for convenience
3. Ran into a couple type errors with existing store state, and didn't want to mess with them without digging deeper

### Dragons

Just need to ensure there's no impact to current behavior

### How Has This Been Tested?

iOS simulator

### Feature Flags ###

All of these changes are experimental and not ready for consumption.
There should be no UI changes to users without the `OFFLINE_MODE_ENABLED` feature flag.
The flag is not enabled for any prod users.


packages/mobile/src/services/offline-downloader/offline-storage.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export const isAudioAvailableOffline = async (track: Track) => {
+  const fullFilePath = path.join(...getLocalAudioPath(track))
+  return await exists(fullFilePath)
+}
+
+// Storage management
+
+// Debugging method to clear all downloaded content
```
nicoback2, line 52:
```suggestion
/** Debugging method to clear all downloaded content */
```
super nit but nicer for code editors

nicoback2: (same for func below)

packages/mobile/src/services/offline-downloader/offline-storage.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export const getArtFileNameFromUri = (uri: string) => {
+  return uri.split('/').slice(-1)[0]
+}
+
+// Audio
+
+export const getLocalAudioPath = (track: Track): [string, string] => {
+  // @ts-ignore route_id exists on track
+  const fileName = `${track.route_id.replaceAll('/', '_')}.mp3`
```
nicoback2, line 41: should we just change the track type?

packages/mobile/src/services/offline-downloader/offline-storage.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export const getLocalCoverArtPath = (track: Track): [string?, string?] => {
+  const uri = getCoverArtUri(track)
+  if (!uri) return []
+  const fileName = getArtFileNameFromUri(uri)
+  return [`${downloadsRoot}/tracks/${track.track_id}`, fileName]
+}
+
+export const getArtFileNameFromUri = (uri: string) => {
+  return uri.split('/').slice(-1)[0]
```
nicoback2, line 34: what if we just named them all the same thing like "Artwork"? (more predictable)

packages/mobile/src/services/offline-downloader/offline-storage.ts
nicoback2, line 41: similar to the artwork thing, could we just name these all the same thing?

packages/mobile/src/services/offline-downloader/offline-downloader.ts
```diff
@@ -0,0 +1,102 @@
...
+    }
+  }
+}
+
+export const getCoverArtUri = (track: Track) => {
+  // TODO: get other sizes
+  return track._cover_art_sizes?.['150x150']
+}
+
+const downloadIfNotExists = async (
```
nicoback2, line 82: Could you add a short description for the function above this? I was a bit confused by the name at first but couldn't think of a better name haha.

packages/mobile/src/services/offline-downloader/offline-downloader.ts
```diff
@@ -0,0 +1,102 @@
...
+  const track = getTrack(state, { id: trackId })
+  if (!track) return false
+
+  const coverArtUri = getCoverArtUri(track)
+  const [coverArtDirectory, coverArtFileName] = getLocalCoverArtPath(track)
+  if (coverArtUri && coverArtDirectory && coverArtFileName) {
+    await downloadIfNotExists(coverArtUri, coverArtDirectory, coverArtFileName)
+  }
+  await tryDownloadTrackFromEachCreatorNode(track)
+  await writeTrackJson(track, collection)
```
nicoback2, line 36: What do we do if either of these steps fail?

packages/mobile/src/services/offline-downloader/offline-downloader.ts
```diff
@@ -0,0 +1,102 @@
...
+      userId: track?.owner_id,
+      // @ts-ignore mismatch in an irrelevant part of state
+      currentUserId: getUserId(state)
+    })
+  )[0] as UserMetadata
+  const encodedTrackId = encodeHashId(track.track_id)
+  const creatorNodeEndpoints = user.creator_node_endpoint.split(',')
+  for (const creatorNodeEndpoint of creatorNodeEndpoints) {
+    const uri = `${creatorNodeEndpoint}/tracks/stream/${encodedTrackId}`
+    const [audioDirectory, audioFileName] = getLocalAudioPath(track)
```
nicoback2, line 70: nit: you can move this out of the for loop and just return null/failure before the loop if either of the values are empty

packages/mobile/src/services/offline-downloader/offline-downloader.ts
```diff
@@ -0,0 +1,102 @@
...
+    if (audioDirectory && audioFileName) {
+      await downloadIfNotExists(uri, audioDirectory, audioFileName)
```
nicoback2, line 72: nit: might be a slight time saver (or even just more clear to read) to just exit the loop as soon as the track is successfully downloaded

nicoback2 (approved):
so fire 🔥🔥🔥

packages/mobile/src/hooks/useOfflineTrack.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import {
+  getLocalAudioPath,
+  isAudioAvailableOffline
+} from 'app/services/offline-downloader'
+import { pathJoin } from 'app/utils/fileSystem'
+
+export const useOfflineTrackUri = (track: Track | null) => {
+  const [offlineSrc, setOfflineSrc] = useState<String | null>()
+  useEffect(() => {
```
sliptype, line 13: We can leverage `useAsync` from `react-use` to make this cleaner!

packages/mobile/src/screens/favorites-screen/FavoritesScreen.tsx
```diff
@@ -32,13 +35,20 @@ const favoritesScreens = [
...
   usePopToTopOnDrawerOpen()
+  const { isEnabled: isOfflineModeEnabled } = useFeatureFlag(
+    FeatureFlags.OFFLINE_MODE_ENABLED
+  )
 
   return (
     <Screen>
       <Header text='Favorites' />
-      <ScreenContent>
+      {isOfflineModeEnabled ? (
```
sliptype, line 45: Maybe a quick comment here that ScreenContent handles displaying the offline indicator. I was confused for a sec

packages/mobile/src/services/offline-downloader/offline-downloader.ts
```diff
@@ -0,0 +1,102 @@
...
+  getLocalAudioPath,
+  getLocalCoverArtPath,
+  getLocalTrackJsonPath
+} from './offline-storage'
+const { getUserId } = accountSelectors
+const { getTrack } = cacheTracksSelectors
+
+// Main entrypoint - perform all steps required to complete a download
+export const downloadTrack = async (trackId: number, collection: string) => {
+  const state = store.getState()
```
sliptype, line 26: I don't think we are doing this anywhere else, all this logic feels like it might actually belong in sagas. But I'm pretty anti-saga these days so would love to hear other people's thoughts

packages/mobile/src/services/offline-downloader/offline-downloader.ts
```diff
@@ -0,0 +1,102 @@
...
+    }
+  }
+
+  const pathToWrite = getLocalTrackJsonPath(track)
+  await RNFS.write(pathToWrite, JSON.stringify(trackToWrite))
+}
+
+const tryDownloadTrackFromEachCreatorNode = async (track: Track) => {
+  const state = store.getState()
+  const user = (
```
sliptype, line 59: Do we not already have this in the store?

packages/mobile/src/services/offline-downloader/offline-storage.ts
```diff
@@ -0,0 +1,80 @@
...
+  return path.join(getLocalTrackRoot(track), `${track.track_id}.json`)
+}
+
+// Cover Art
+
+export const getLocalCoverArtPath = (track: Track): [string?, string?] => {
+  const uri = getCoverArtUri(track)
+  if (!uri) return []
+  const fileName = getArtFileNameFromUri(uri)
+  return [`${downloadsRoot}/tracks/${track.track_id}`, fileName]
```
sliptype, line 30: Kinda weird that this function is returning a tuple but the function name is `getLocalCoverArtPath`

packages/mobile/src/utils/fileSystem.ts
```diff
@@ -0,0 +1,3 @@
+export const pathJoin = (...segments: (string | undefined)[]) => {
```
sliptype, line 1: You should be able to `import path from 'path'` and use `path.join()`

sliptype (approved):
Amazing work!!!

packages/mobile/src/services/offline-downloader/offline-downloader.ts
amendelsohn: It felt weird writing it too. Currently, we usually call this from a component that has the state, but I wanted to keep it reusable for the potential background process. However, we won't be able to use the state their either. Keeping it simple for now, but open to ideas.

packages/mobile/src/services/offline-downloader/offline-downloader.ts
amendelsohn: Really, we should check if all of the assets exist before writing `downloaded_time` to the track json. There are more steps that need to happen here, so I haven't gotten to that yet, but it's 100% something we need to worry about. I'll add a todo for the next round.

packages/mobile/src/services/offline-downloader/offline-downloader.ts
amendelsohn: Oh, we probably do. I was teaching myself how to use the API 😅

amendelsohn: Actually, wait. Not necessarily. If we're downloading all of our favorites we may not have loaded all of them yet. I think I have a mix of logic meant to work in the background and logic that doesn't. It's going to need some further separation and cleanup.

packages/mobile/src/services/offline-downloader/offline-storage.ts
amendelsohn: these will be named by the image size like `150x150.jpg`

packages/mobile/src/services/offline-downloader/offline-downloader.ts
amendelsohn: Discussed in standup. Seems like this is the way for now

PR #2022, opened 2022-09-28
[C-1194] Fix desktop profile state

dylanjeffers:
### Description

Fixes desktop profile state by ensuring the ProfilePageProvider dispatches actions with the current user's handle

### DIFF
--- a/packages/common/src/store/lineup/actions.ts
+++ b/packages/common/src/store/lineup/actions.ts
@@ -182,10 +182,11 @@ export class LineupActions {
     }
   }
 
-  updateLineupOrder(orderedIds: UID[]) {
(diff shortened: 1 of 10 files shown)


audius-infra:
Preview this change https://demo.audius.co/c-1194-profile-state

raymondjacobson (approved):

rickyrombo (approved):
I might be missing context here, but it looks good to me? What happened with the missing handles in actions?

dylanjeffers:
> I might be missing context here, but it looks good to me? What happened with the missing handles in actions?

No worries, yeah we have a new profile state where we have each entry specified by handle, rather then every profile page sharing the same state. The actions need to specify which user via handle to know which entry to update

PR #2028, opened 2022-09-29
[PAY-668] Set up premium content

sddioulde:
### Description

Depends on https://github.com/AudiusProject/audius-protocol/pull/3928 (now merged) and new libs version release (also done)

- Consume libs premium content changes
- Add premium content fields to track upload metadata, setting the premiumness to false by default. We have no UI for making tracks premium as of now.

### Dragons

Danger here is potentially borking the track upload flow. Testing the upload was fine but extra scrutiny on that piece is welcome.

### How Has This Been Tested?

Local client against remote dev box. Was able to upload and play tracks.

Testing track edits...
Update: track edit works!

No monitoring here really, only that tracks have new is_premium and premium_conditions fields. Requesting tracks from DNs returns those fields as well.

### Feature Flags ###

No feature flags necessary as it does not introduce new UI. Overall behavior should remain the same.


sddioulde:
edit track no longer erroring, but not able to confirm either. getting to the bottom of that...

audius-infra:
Preview this change https://demo.audius.co/sd-pc

packages/web/src/pages/upload-page/UploadPage.js
```diff
@@ -262,6 +262,7 @@ class Upload extends Component {
       this.state.stems
     )
     this.changePage(Pages.FINISH)
+    this.setState({ tracks })
```
rickyrombo, line 265: For my own context, what happened/why was this needed?

packages/web/src/common/store/player/sagas.ts
```diff
@@ -24,7 +26,13 @@ import {
   delay
 } from 'typed-redux-saga'
 
+import { waitForLibsInit } from 'services/audius-backend/eagerLoadUtils'
+
 import errorSagas from './errorSagas'
+
+// @ts-ignore
+const libs = (): AudiusLibs => window.audiusLibs
```
rickyrombo, line 34: We can't include libs in common sagas direcly afaik, maybe put the necessary functionality in AudiusBackend and use `getContext('audiusBackendInstance)`? cc: @dylanjeffers maybe

rickyrombo (changes requested):
Looks pretty good! I think we have to refrain from keeping libs in common sagas directly since common is used in mobile.

packages/web/src/common/store/player/sagas.ts
rickyrombo: wonder if we can't put a linter on common to check for `window` usages? Although since we have the `@ts-ignore` we'd probably miss this anyway

dylanjeffers: yeah the move is to grab them from context, happy to show you how, just dm

packages/web/src/pages/upload-page/UploadPage.js
sddioulde: i dont think we need this actually.
wanted to be consistent with the state before i merged this small client PR https://github.com/AudiusProject/audius-client/pull/2042/files

but pretty sure this does not really do anything once we switch to the FINISH page.
going to simply remove and make sure things still work

packages/web/src/common/store/player/sagas.ts
sddioulde: got it! thx. please re-review if latest change makes sense :)

rickyrombo (approved):

PR #2032, opened 2022-09-29
Only check isInternetReachable

nicoback2:
### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/utils/connectivity.ts
+++ b/packages/mobile/src/utils/connectivity.ts
@@ -9,12 +9,9 @@ import { dispatch } from './../store/store'
 export const checkConnectivity = (netInfo: NetInfoState | null) => {
   if (!netInfo) return true
 
-  const { isConnected, isInternetReachable, type } = netInfo
-  // Unknown type or null reachability means that netinfo hasn't been able to
-  // determine connection status. We may actually be connected despite
-  // `isConnected` being `false`
-  if (type === 'unknown' || isInternetReachable === null) return true
-  return isConnected && isInternetReachable
+  const { isInternetReachable } = netInfo
+
+  return isInternetReachable !== false
 }
 
 // Latest connectivity value


packages/mobile/src/utils/connectivity.ts
```diff
@@ -9,12 +9,9 @@ import { dispatch } from './../store/store'
 export const checkConnectivity = (netInfo: NetInfoState | null) => {
   if (!netInfo) return true
 
-  const { isConnected, isInternetReachable, type } = netInfo
-  // Unknown type or null reachability means that netinfo hasn't been able to
-  // determine connection status. We may actually be connected despite
-  // `isConnected` being `false`
-  if (type === 'unknown' || isInternetReachable === null) return true
-  return isConnected && isInternetReachable
```
nicoback2, line 17: Apparently `isConnected` can be null if unknown so maybe that was causing issues with this returning false. https://github.com/react-native-netinfo/react-native-netinfo
Also, I think we only want/need to check `isInternetReachable` since you can be connected to a network but not able to reach the internet.

amendelsohn: Yeah I was thinking of just triggering the internet ping wherever we're currently using the event listener.

Any pros/cons you can think of for using `isInternetReachable` as the connectivity state for the app? Are there places where we care about the difference between network off vs can't reach audius?

nicoback2: I don't think there are any places where we care about connected to network vs can't reach internet. I think we'd always just care about not being able to reach the internet. 
Also I'm pretty sure NetInfo sends a ping of its own when it can't determine if the internet is reachable through normal native means. Ctrl + F "reachabilityTest": https://github.com/react-native-netinfo/react-native-netinfo

Also also, looks like we can just disable `useNativeReachability` on NetInfo to make it ping a URL!

PR #2034, opened 2022-09-29
[C-1206] double check before declaring offline state

amendelsohn:
### Description

Be more optimistic on reachability checks.
Do a ping test whenever we would set `isReachable: false`. If the ping test passes, assume we're connected instead of trusting `NetInfoState`

### How Has This Been Tested?

iOS simulator, iOS device

Need to watch this as it goes out on Bounce and Prod TestFlight


packages/mobile/src/utils/reachability.ts
```diff
@@ -0,0 +1,81 @@
...
+      console.debug('Reachability call succeeded')
+      return true
+    }
+    console.debug('Reachability call failed')
+    return false
+  } catch {
+    console.debug('Reachability call failed')
+    return false
+  }
+}
```
amendelsohn, line 46: The ping test is the main addition

packages/mobile/src/utils/reachability.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  const newValue = checkNetInfoReachability(netInfoState)
+  if (!newValue) {
+    // Don't trust offline signal while app is not in focus
+    // We check on refocus
+    const appState = AppState.currentState
+    if (appState !== 'active') return
+
+    // Perform our own reachability test to be extra sure we're offline
+    const reachable = await pingTest()
```
amendelsohn, line 59: added ping test
moved app state test to only block disconnecting events

packages/mobile/src/utils/reachability.ts
```diff
@@ -0,0 +1,81 @@
+import { reachabilityActions } from '@audius/common'
+import type { NetInfoState } from '@react-native-community/netinfo'
+import NetInfo from '@react-native-community/netinfo'
+import { debounce } from 'lodash'
+import { AppState } from 'react-native'
+
+import { env } from 'app/services/env'
+import { dispatch } from 'app/store'
+
+const REACHABILITY_URL = env.REACHABILITY_URL
```
amendelsohn, line 10: In my testing this was available, and it looks like we're just grabbing it from the `.env` files where it's already populated:

```
// .env.dev; .env.stage
REACHABILITY_URL=https://staging.audius.co/204
```
```
// .env.prod
REACHABILITY_URL=https://audius.co/204
```

amendelsohn (commented):
Apologies for both changing and renaming the file in one PR

packages/mobile/src/utils/reachability.ts
raymondjacobson: def supportive of this ping test now in addition to the netinfo stuff. ideally we can move this to a route on discovery node so we don't depend on audius.co from here, but that can be left for another day

packages/mobile/src/utils/reachability.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export const forceRefreshConnectivity = async () => {
+  NetInfo.refresh()
+  const updatedNetInfoState = await NetInfo.fetch()
+  updateReachability(updatedNetInfoState)
+}
+
+/** Called on first app render */
+export const subscribeToNetworkStatusUpdates = () => {
+  NetInfo.addEventListener(
```
raymondjacobson, line 76: is this debounce actually hurting us? i wonder if it's debouncing the "back online" bit?

raymondjacobson (approved):

packages/mobile/src/utils/reachability.ts
amendelsohn: It would definitely also do that. I could maybe separate the debounce into just the part that handles going offline. I think we probably still want that to get delayed.

raymondjacobson: yeah i think that'd be a good call!

going from online => offline should be like pushing a rock up a hill and offline => online should be like letting it roll back down

sliptype (approved):
Nice! I agree with only applying the debounce when changing from online to offline

PR #2037, opened 2022-09-30
[C-1128] Add profile screen skeleton

dylanjeffers:
### Description

- Adds profile skeleton, which drastically improves UX for profiles loaded from search screen or deep-link, and other cases where the profile hasn't been fetched yet.
- Adds `noShimmer` option to `Skeleton` which is good for some skeletons that are small, like we have here for the bio element.
> Note: there is some lineup jank present when the profile data is first fetched present in this video. This will be addressed in a separate pr

https://user-images.githubusercontent.com/8230000/193168912-d12caf17-7979-4b25-b2ea-10f94ff9562a.mp4

### DIFF
--- a/packages/mobile/src/components/feed-tip-tile/FeedTipTileSkeleton.tsx
+++ b/packages/mobile/src/components/feed-tip-tile/FeedTipTileSkeleton.tsx
@@ -5,9 +5,8 @@ import { makeStyles } from 'app/styles'
 
 import { Tile } from '../core'
 
-const useStyles = makeStyles(({ spacing, palette }) => ({
+const useStyles = makeStyles(({ spacing }) => ({
   tile: {
-    display: 'flex',
     flexDirection: 'row',
     justifyContent: 'center',
     alignItems: 'center',
@@ -17,13 +16,11 @@ const useStyles = makeStyles(({ spacing, palette }) => ({
     paddingVertical: spacing(4)
   },
   header: {
-    display: 'flex',
     flexDirection: 'row',
     justifyContent: 'flex-start',
     alignItems: 'center'
   },
   metadata: {
-    display: 'flex',
     flexDirection: 'column',
     justifyContent: 'space-evenly'
   },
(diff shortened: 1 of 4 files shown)


packages/mobile/src/components/feed-tip-tile/FeedTipTileSkeleton.tsx
```diff
@@ -5,9 +5,8 @@ import { makeStyles } from 'app/styles'
 
 import { Tile } from '../core'
 
-const useStyles = makeStyles(({ spacing, palette }) => ({
+const useStyles = makeStyles(({ spacing }) => ({
```
raymondjacobson, line 8: Grr sorry

packages/mobile/src/components/skeleton/Skeleton.tsx
```diff
@@ -51,12 +52,13 @@ const Skeleton = ({ width, height, style }: SkeletonProps) => {
         })
       ).start()
     }
-  }, [shimmerPos, shimmerWidth])
+  }, [shimmerPos, shimmerWidth, noShimmer])
 
   return (
     <View style={[styles.view, { height, width }, style]}>
       <Animated.View
         onLayout={(e) => {
```
raymondjacobson, line 60: should we just not provide an onLayout at all?

packages/mobile/src/screens/profile-screen/ProfileScreenSkeleton.tsx
```diff
@@ -0,0 +1,177 @@
+import { memo } from 'react'
```
raymondjacobson, line 1: CRANKED

raymondjacobson (approved):

packages/mobile/src/components/skeleton/Skeleton.tsx
dylanjeffers: i think we should look into this. it seems like this is needed for cases where width is not specified. we can update our usages to ensure we always explicitly set it. seems nice to allow flex to help us sometimes tho.

PR #2038, opened 2022-09-30
[C-481] Fix profile header loading jank

dylanjeffers:
### Description

Fixes issue where profile lineups load underneath the profile header, and all the way at the top, then suddenly transition down. This is due to collapsible-tabs not knowing the height of the header on the first load. Adding a default value that is around the height of an average header.

### DIFF
--- a/packages/mobile/src/screens/profile-screen/ProfileTabNavigator.tsx
+++ b/packages/mobile/src/screens/profile-screen/ProfileTabNavigator.tsx
@@ -107,6 +107,7 @@ export const ProfileTabNavigator = ({
       <CollapsibleTabNavigator
         renderHeader={renderHeader}
         animatedValue={animatedValue}
+        headerHeight={1000}
       >
         {trackScreen}
         {albumsScreen}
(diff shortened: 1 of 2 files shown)


packages/mobile/src/screens/profile-screen/ProfileTabNavigator.tsx
```diff
@@ -107,6 +107,7 @@ export const ProfileTabNavigator = ({
       <CollapsibleTabNavigator
         renderHeader={renderHeader}
         animatedValue={animatedValue}
+        headerHeight={1000}
```
raymondjacobson, line 110: maybe comment / make a good named const for 1000. hard to know "why" ya know?

raymondjacobson (approved):

packages/mobile/src/screens/profile-screen/ProfileTabNavigator.tsx
dylanjeffers: oh weird, latest commit didn't get pushed which approaches this exactly!

PR #2039, opened 2022-09-30
Fix tracks/reposts lineups for artist with handles with capitials

dylanjeffers:
### Description

Fixes mobile tracks/reposts by ensuring handle is lowerCased, which we need since the profile state references lowercase handles. Ultimately i think we should move all the `.toLowerCase` code to the actions/reducers/selectors so the consumer doesn't have to think about this. This pr unblocks tho.


raymondjacobson (approved):
Agree w/ your conclusion.

Perhaps we can think of a way to just have `handle` and `handle_lc` set in store on the mutation after a fetch so before it hits the store

PR #2041, opened 2022-09-30
[C-1207] Fix stem downloads

dylanjeffers:
### Description

Fixes runtime error due to calling `new URL` in native. It was tricky since `URL` works when the Chrome debugger is active, which is why we missed this.


sliptype (approved):
Awesome! I think there may be some other usages of URL in mobile too

raymondjacobson (approved):
well done sir

nicoback2 (approved):
Noice

PR #2043, opened 2022-09-30
[PAY-614][PAY-541] Add recovery flow and modal for buying $AUDIO

rickyrombo:
### Description
- Checks for existing SOL or $AUDIO balance in the root wallet on page load, and finishes the buy $AUDIO flow as necessary
- Checks for existing SOL or $AUDIO balance in the root wallet when clicking the on-ramp buttons in case the user is trying to buy $AUDIO in the same session as a failed purchase
- Pops a new UI modal during a recovery
- To do the recovery, the buy $AUDIO flow now writes out the necessary metadata after each step to local storage. If that info doesn't exist, the metadata will default to unknown and empty string
- Refactors a bunch of the sagas to be more composable for the sake of the recovery flow

### Dragons

### How Has This Been Tested?

Sent SOL and $AUDIO to my root wallet and triggered the modal at both entry points. Tried with empty local storage and with populated local storage. Did not test end to end with coinbase, nor with a working identity service to actually save the metadata yet.

TODO:
- [ ] Add analytics to track how often this gets hit, which step it gets hit for, and how often it succeeds/fails
  - Will probably do in a follow up PR

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/mjp-buy-recovery

packages/web/src/store/application/ui/buy-audio/sagas.ts
```diff
@@ -371,87 +423,388 @@ Total: ${estimatedLamports.toNumber() / LAMPORTS_PER_SOL} SOL ($${
...
+    swapTransactionId,
+    cleanupTransactionId,
+    transferTransactionId,
+    estimatedUSD,
+    purchasedLamports,
+    purchasedAudioWei
+  } = localStorageState.transactionDetailsArgs
+
+  if (!transferTransactionId) {
+    throw new Error('Missing transactionDetailsArgs[transferTransactionId]')
```
piazzatron, line 446: Nit: if this happens maybe we should zero out local storage so the next session is fine

piazzatron (approved):
Let's merge this so we can test, and we can address any comments in subsequent PRs!

packages/common/src/utils/wallet.ts
```diff
@@ -117,6 +117,7 @@ export const convertJSBIToAmountObject = (amount: JSBI, decimals: number) => {
     : quotient.toString()
   return {
     amount: JSBI.toNumber(amount),
+    amountString: JSBI.toString(),
```
sddioulde, line 120: should this include the amount? looks like it does not

packages/web/src/store/application/ui/buy-audio/sagas.ts
```diff
@@ -371,87 +423,388 @@ Total: ${estimatedLamports.toNumber() / LAMPORTS_PER_SOL} SOL ($${
...
+  if (purchasedLamports !== new BN(estimatedSOL.amount)) {
+    console.warn(
+      `Warning: Purchase SOL amount differs from expected. Actual: ${
+        new BN(newBalance).sub(new BN(initialBalance)).toNumber() /
+        LAMPORTS_PER_SOL
+      } SOL. Expected: ${estimatedSOL.uiAmountString} SOL.`
+    )
+  }
+
+  const localStorage = yield* getContext('localStorage')
```
sddioulde, line 577: curious how come on line 500 we don't get the localStorage the same way as here. i see we do it this way too on line 668 for example

packages/web/src/store/application/ui/buy-audio/sagas.ts
```diff
@@ -682,31 +910,184 @@ function* startBuyAudioFlow({
...
+
+      const { audioSwappedSpl } = yield* swapStep({
+        exchangeAmount: exchangableBalance,
+        desiredAudioAmount: localStorageState.desiredAudioAmount,
+        rootAccount,
+        transactionHandler,
+        maxRetryCount,
+        retryDelayMs
+      })
+      yield* transferStep({
```
sddioulde, line 1006: yield* `call(transferStep, ...)`?

packages/web/src/store/application/ui/buy-audio/sagas.ts
```diff
@@ -682,31 +910,184 @@ function* startBuyAudioFlow({
...
+        } SOL, converting ${
+          exchangableBalance.toNumber() / LAMPORTS_PER_SOL
+        } SOL to AUDIO...`
+      )
+
+      yield* put(setVisibility({ modal: 'BuyAudioRecovery', visible: true }))
+      yield* put(setVisibility({ modal: 'BuyAudio', visible: false }))
+      didNeedRecovery = true
+
+      const { audioSwappedSpl } = yield* swapStep({
```
sddioulde, line 998: yield* `call(swapStep, ...)`?

packages/web/src/store/application/ui/buy-audio/sagas.ts
```diff
@@ -682,31 +910,184 @@ function* startBuyAudioFlow({
...
+      if (audioBalance.gt(new BN(0))) {
+        console.debug(
+          `Found existing $AUDIO balance of ${audioBalance}, transferring to user bank...`
+        )
+
+        yield* put(setVisibility({ modal: 'BuyAudioRecovery', visible: true }))
+        yield* put(setVisibility({ modal: 'BuyAudio', visible: false }))
+        didNeedRecovery = true
+
+        yield* transferStep({
```
sddioulde, line 1031: here too, use call effect?

sddioulde (commented):
epic! left few comments tho

packages/common/src/utils/wallet.ts
rickyrombo: yes! this is awkward...

packages/web/src/store/application/ui/buy-audio/sagas.ts
rickyrombo: I believe we do, just way at the top of the method (line 428)

sddioulde: ah ok

PR #2046, opened 2022-09-30
[C-1208] Track playback UID is not consistent when clicking into track

sliptype:
### Description

* When playing a track from the feed and navigating to the track screen, the play state did not match because the uids are different
* Match on the track id instead of the uid on track and collection screens. This means that no matter where the track was played from, it will display as playing on the track and collection screens

https://user-images.githubusercontent.com/19916043/193358988-b4cdcec1-de1f-4978-b958-c5ecd3f0523c.mov

### Dragons

This is currently broken in prod and most likely has been since the initial react UI work. Matching on id seems to be correct but maybe there are cases that I'm breaking with this change?

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/track-list/TrackList.tsx
+++ b/packages/mobile/src/components/track-list/TrackList.tsx
@@ -1,6 +1,6 @@
 import type { ReactElement } from 'react'
 
-import type { CommonState, ID, Nullable, UID } from '@audius/common'
+import type { ID, UID } from '@audius/common'
 import { playerSelectors } from '@audius/common'
 import type { FlatListProps } from 'react-native'
 import { FlatList, View } from 'react-native'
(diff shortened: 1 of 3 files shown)


raymondjacobson:
```
Match on the track id instead of the uid on track and collection screens. This means that no matter where the track was played from, it will display as playing on the track and collection screens
```

I think this is unfortunately not the right product behavior & why we came up with the concept of UIDs. If you play a track from your feed and then click into a playlist, that track would be playing. Or if this track was in your feed and there was also a playlist with the same track in your feed, they would be both playing.

dylanjeffers (approved):
this is great! so question regarding queue, when we play/pause a track that's in a different screen, what happens?

sliptype:
> this is great! so question regarding queue, when we play/pause a track that's in a different screen, what happens?

Not sure if I understand the question, if we play a track from a different screen it would start playing whatever lineup that track is part of

sliptype:
> ```
> Match on the track id instead of the uid on track and collection screens. This means that no matter where the track was played from, it will display as playing on the track and collection screens
> ```
> 
> I think this is unfortunately not the right product behavior & why we came up with the concept of UIDs. If you play a track from your feed and then click into a playlist, that track would be playing. Or if this track was in your feed and there was also a playlist with the same track in your feed, they would be both playing.

Gotcha, trying to understand how we achieve that on web since the uids are different for feed and track screen.

Interesting, it looks like on web playing a collection from the feed and then navigating to that collection also doesn't show it as playing:

https://user-images.githubusercontent.com/19916043/193360528-03e4862c-f464-4324-b689-4f126863d633.mov

That's probably a bug right?

PR #2048, opened 2022-10-01
[C-1179] Fix edit-profile

dylanjeffers:
### Description

Fixes edit-profile flow by adding a few improvements to profile state.
- Condenses three state variables, `updating`, `updateSuccess`, and `updateError` into an `updateStatus` variable that can be selected.
- Potentially improves the values of the various variables with the `updateProfile` `updateProfileSuccess` and updateProfileError` actions
- Moves `updateProfileSucceeded` to the onSuccess confirmer callback
- Updates profile screen to only fetch profile when we are not in "edit mode, ie waiting for edit confirmation, cause otherwise the values are overwritten

### DIFF
--- a/packages/common/src/store/pages/profile/reducer.ts
+++ b/packages/common/src/store/pages/profile/reducer.ts
@@ -178,7 +178,11 @@ const actionsMap = {
     return updateProfile(state, action, { mostUsedTags })
   },
   [UPDATE_PROFILE](state, action) {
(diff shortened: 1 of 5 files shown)


sliptype (approved):

audius-infra:
Preview this change https://demo.audius.co/dj-c-1179-fix-edit-profile

PR #2050, opened 2022-10-03
[C-1201,C-1202] Improve search results rendering and perf

dylanjeffers:
### Description

- Fixes rendering issues in search results page, where "no results" pops up for a moment, and then loading indicator appears.
- Improves rerender perf by using useProxySelector and simplified selectors.
- Improves reducer logic to be a bit more straightforward, and prevent uneccesary state changes

### DIFF
--- a/packages/mobile/src/screens/search-results-screen/tabs/AlbumsTab.tsx
+++ b/packages/mobile/src/screens/search-results-screen/tabs/AlbumsTab.tsx
@@ -1,21 +1,27 @@
-import { SearchKind, searchResultsPageSelectors } from '@audius/common'
-import { useSelector } from 'react-redux'
+import type { CommonState } from '@audius/common'
+import { useProxySelector, SearchKind } from '@audius/common'
 
 import { CollectionList } from 'app/components/collection-list/CollectionList'
 
 import { SearchResultsTab } from './SearchResultsTab'
 import { useFetchTabResultsEffect } from './useFetchTabResultsEffect'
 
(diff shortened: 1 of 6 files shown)


packages/mobile/src/screens/search-results-screen/tabs/AlbumsTab.tsx
```diff
@@ -1,21 +1,27 @@
...
 
 import { SearchResultsTab } from './SearchResultsTab'
 import { useFetchTabResultsEffect } from './useFetchTabResultsEffect'
 
-const { makeGetSearchAlbums } = searchResultsPageSelectors
-
-const getSearchAlbums = makeGetSearchAlbums()
+const selectSearchAlbums = (state: CommonState) =>
+  state.pages.searchResults.albumIds
+    ?.map((albumId) => {
```
nicoback2, line 11: for all these tabs might be worth to share this mapping/filtering logic w the `makeGetSearchX` functions in common (since they seem to be the same). and maybe even move this to a separate selectors file (either the one in common or a new one in mobile since this is mobile only)

nicoback2 (approved):

packages/mobile/src/screens/search-results-screen/tabs/AlbumsTab.tsx
dylanjeffers: yeah we should really replace the existing selectors in common with these, just wanted to go quickly here and make sure im not messing up desktop. added a tech-debt item to move these over to common tho!

PR #2052, opened 2022-10-03
[C-1213, C-1216] Fix album artwork corners

dylanjeffers:
### Description

Fixes TrackList album artwork and smart collection screens sharp corners

### DIFF
--- a/packages/mobile/src/components/track-list/TrackListItem.tsx
+++ b/packages/mobile/src/components/track-list/TrackListItem.tsx
@@ -89,7 +89,6 @@ export type TrackListItemProps = {
   hideArt?: boolean
   index: number
   isActive?: boolean
-  isLoading?: boolean
   isPlaying?: boolean
   isReorderable?: boolean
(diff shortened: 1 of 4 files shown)


packages/mobile/src/components/track-list/TrackListItem.tsx
```diff
@@ -89,7 +89,6 @@ export type TrackListItemProps = {
   hideArt?: boolean
```
nicoback2, line 89: What are the changes in this file for?

nicoback2 (approved):
1q but LGTM!

packages/mobile/src/components/track-list/TrackListItem.tsx
dylanjeffers: ah, yeah so we are not handling `isLoading` anymore, chatting it over with ray. it wasn't implemented, and since loading state is so short, were just not gonna deal with it

PR #2053, opened 2022-10-03
[C-1224] Prevent duplicate navigation pushes

dylanjeffers:
### Description

Prevents user from triggering multiple navigation requests by spamming the same link


raymondjacobson (approved):
yas

nicoback2 (approved):

PR #2054, opened 2022-10-04
Fix verified badge C-1178

nicoback2:
### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/screens/profile-screen/ProfileInfo.tsx
+++ b/packages/mobile/src/screens/profile-screen/ProfileInfo.tsx
@@ -86,8 +86,10 @@ export const ProfileInfo = (props: ProfileInfoProps) => {
     'name',
     'handle',
     'does_current_user_follow',
-    'does_follow_current_user'
+    'does_follow_current_user',
+    'is_verified'
   ])
+  console.log('beep profile', profile)
 
   const { name, handle, does_current_user_follow, does_follow_current_user } =
     profile
--- a/packages/mobile/src/screens/profile-screen/ProfileInfo.tsx
+++ b/packages/mobile/src/screens/profile-screen/ProfileInfo.tsx
@@ -89,7 +89,6 @@ export const ProfileInfo = (props: ProfileInfoProps) => {
     'does_follow_current_user',
     'is_verified'
   ])
-  console.log('beep profile', profile)
 
   const { name, handle, does_current_user_follow, does_follow_current_user } =
     profile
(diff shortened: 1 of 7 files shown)


packages/mobile/src/screens/profile-screen/ProfileInfo.tsx
```diff
@@ -86,7 +86,8 @@ export const ProfileInfo = (props: ProfileInfoProps) => {
     'name',
     'handle',
     'does_current_user_follow',
-    'does_follow_current_user'
+    'does_follow_current_user',
+    'is_verified'
```
raymondjacobson, line 90: oh dang, this is kind of scary because nowhere in this file do we reference is_verified explicitly. wondering if we can be more deliberate than to pass `profile` down all the way through the component tree?

raymondjacobson (approved):

packages/mobile/src/screens/profile-screen/ProfileInfo.tsx
nicoback2: yeah for sure. i just typed the `useSelectProfile` hook to be correct because before it coerced the return type to be the full User object. this would have prevented this bug

nicoback2: agree its more reacty and better to pass the individual values of `profile` that we need in the presentational components, but a bunch of components (including two more in this same file) follow the same pattern of passing in a whole `profile` object so I just left that for now

dylanjeffers: yeah this is super scary, and my bad honestly, what should we do?

packages/mobile/src/screens/profile-screen/selectors.ts
```diff
@@ -16,7 +16,7 @@ const { getAccountUser, getUserId } = accountSelectors
 /*
  * Selects profile user and ensures rerenders occur only for changes specified in deps
  */
-export const useSelectProfileRoot = (deps: Array<keyof User>) => {
+export const useSelectProfileRoot = <K extends keyof User>(deps: K[]) => {
```
dylanjeffers, line 19: omg you figured thils out!??

dylanjeffers (approved):
i am awestruck with the deps picking, great great job

packages/mobile/src/screens/profile-screen/ProfileInfo.tsx
raymondjacobson: the manual typing is good. probably stop passing objects through is best or selecting from redux if you need a complex object and not depending on parent to pass it

PR #2055, opened 2022-10-04
[C-1027] Improve pull to refresh interaction

dylanjeffers:
### Description

This is a fairly complicated pr, but suffice to say this ensures the user has more control over their interactions when pull-t-refresh is activated, following the twitter pull-to-refresh logic.

Main things that are improved:
- User can scroll through their content while pull-to-refresh is running. Before the user was forced back up to the top of the scroll-view when pull-to-refresh finished.
- The pull-to-refresh icon is removed from view when user scrolls in the section view.
- Pull-to-refresh will not continue to trigger if the user has scrolled past pull distance, in line with twitter

### Dragons

These improvements only apply to not-collapsible-scroll-view pull-to-refreshes, ie not profile ones. This will be addressed in a separate pr

### DIFF
--- a/packages/mobile/src/components/core/PullToRefresh.tsx
+++ b/packages/mobile/src/components/core/PullToRefresh.tsx
@@ -19,17 +19,17 @@ import { colorize } from 'app/utils/colorizeLottie'
 import { useThemeColors } from 'app/utils/theme'
 
 const PULL_DISTANCE = 75
+const DEBOUNCE_TIME_MS = 1000
 
-const useStyles = (topOffset: number) =>
-  makeStyles(() => ({
-    root: {
-      width: '100%',
-      position: 'absolute',
-      height: 20,
-      top: topOffset,
-      zIndex: 10
-    }
-  }))
+const useStyles = makeStyles((_, topOffset: number) => ({
+  root: {
+    width: '100%',
+    position: 'absolute',
+    height: 20,
+    top: topOffset,
+    zIndex: 10
+  }
+}))
 
 const interpolateTranslateY = (scrollAnim: Animated.Value) =>
   scrollAnim.interpolate({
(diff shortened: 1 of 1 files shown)


packages/mobile/src/components/core/PullToRefresh.tsx
```diff
@@ -1,5 +1,7 @@
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 
+import { setImmediate } from 'timers'
```
raymondjacobson, line 3: I don't think we need an import here

packages/mobile/src/components/core/PullToRefresh.tsx
```diff
@@ -66,46 +68,67 @@ export const useOverflowHandlers = ({
...
+      setIsMomentumScroll(true)
+      if (isRefreshing && currentYOffset.current <= 0) {
+        scrollTo(-50)
+      }
+    },
+    [setIsMomentumScroll, scrollTo, isRefreshing]
+  )
 
-  const handleScroll = attachToScroll(scrollAnim, { listener: onScroll })
+  const handleScrolll = attachToScroll(scrollAnim, { listener: handleScroll })
```
raymondjacobson, line 124: attachedScrollHandler? dunno

packages/mobile/src/components/core/PullToRefresh.tsx
```diff
@@ -218,16 +245,23 @@ export const PullToRefresh = ({
...
     return () => {
       if (listenerRef.current) {
         scrollAnim?.removeListener(listenerRef.current)
       }
     }
-  }, [scrollAnim, onRefresh, isRefreshDisabled])
+  }, [scrollAnim, onRefresh, isRefreshDisabled, didHitTop])
 
-  return scrollAnim ? (
+  return scrollAnim && shouldShowRefresh ? (
```
raymondjacobson, line 264: rather than a boolean to gate it, does it make sense to tweak opacity with animated so it happens outside of the JS loop? (and then to hide it)

raymondjacobson (approved):

packages/mobile/src/components/core/PullToRefresh.tsx
dylanjeffers: oh yeah hahah, that was temporary

packages/mobile/src/components/core/PullToRefresh.tsx
dylanjeffers: oh yeah that's smart, yeah can try that out

dylanjeffers: updated this, great call

PR #2056, opened 2022-10-04
Prevent rendering root until account fetched

dylanjeffers:
Improves TTI for first render by preventing rendering of sign-in screen until we know user account has been attempted to be fetched


sliptype (approved):
Makes sense!

PR #2059, opened 2022-10-04
[C-1226] Fix refresh stuck in loop

dylanjeffers:
### Description

Fixes issue on profile screen where refresh get's stuck in a loop. This was due to not getting the right `refreshStatus` and some tricky issues with when we should trigger refresh in `PullToRefresh`


raymondjacobson (approved):
I think this seems reasonable. Will wait till it's on device to play around a bit :)

dylanjeffers:
> I think this seems reasonable. Will wait till it's on device to play around a bit :)

yeah it took a while and from what i can tell this is the best i can get it :D

PR #2063, opened 2022-10-04
[C-1233] hide edit profile when offline

amendelsohn:
### Description

Hide the edit profile button because changing the profile requires connectivity


sliptype (approved):
Love it! HideIfOffline could also be OnlineOnly, but I think it's fine either way

Kyle-Shanks (approved):

PR #2064, opened 2022-10-04
[C-1200] Improve lineup and bottom-tab perf

dylanjeffers:
### Description

- Improves all lineup perf by ensuring we don't trigger a full rerender on it when we are out of view. This significantly improves navigating to profile from feed.
- Improves bottom-tab-navigation by splitting up the reading of bottom-bar-navigation object, and setting of it. Having them together meant that each bottom bar would trigger a full rerender on the one that was being left, slowing that initial navigation.

### DIFF
--- a/packages/mobile/src/components/lineup/Lineup.tsx
+++ b/packages/mobile/src/components/lineup/Lineup.tsx
@@ -186,7 +186,6 @@ export const Lineup = ({
 
   const handleInView = useCallback(() => {
     dispatch(actions.setInView(true))
-    return () => dispatch(actions.setInView(false))
   }, [dispatch, actions])
 
   useFocusEffect(handleInView)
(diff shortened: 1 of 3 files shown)


packages/mobile/src/components/lineup/Lineup.tsx
```diff
@@ -186,7 +186,6 @@ export const Lineup = ({
 
   const handleInView = useCallback(() => {
     dispatch(actions.setInView(true))
-    return () => dispatch(actions.setInView(false))
```
nicoback2, line 189: noice

packages/mobile/src/screens/app-screen/AppTabNavigationProvider.tsx
```diff
@@ -27,14 +34,21 @@ export const AppTabNavigationProvider = (
   const { children } = props
   const [navigation, setNavigation] = useState<Nullable<AppTabNavigation>>(null)
 
-  const context = useMemo(
+  const navigationContext = useMemo(
     () => ({ navigation, setNavigation }),
```
nicoback2, line 38: should we remove `setNavigation` from this context?

nicoback2 (approved):
lGTM

packages/mobile/src/screens/app-screen/AppTabNavigationProvider.tsx
dylanjeffers: oh yeah 100% nice catch. it shouldn't technically matter, but prob good practice

PR #2069, opened 2022-10-04
[C-1254] Fix debounced navigation

dylanjeffers:
### Description

Refactors debounced navigation solution with a custom push function.
Inspiration here: https://github.com/react-navigation/react-navigation/issues/3839#issuecomment-377159605
dispatch docs: https://reactnavigation.org/docs/navigation-prop/#dispatch

### Dragons
Instead of using the nearest navigator, we are defaulting to the AppStack Navigator, which should result in consistent behavior. Without this, when we were in a collapsible tab bar, we didn't have access to the previous route, so it wouldn't work.

### DIFF
--- a/packages/mobile/src/screens/app-screen/AppTabNavigationProvider.tsx
+++ b/packages/mobile/src/screens/app-screen/AppTabNavigationProvider.tsx
@@ -1,20 +1,19 @@
 import type { ReactNode } from 'react'
 import { useMemo, createContext, useState } from 'react'
 
-import type { Nullable } from '@audius/common'
 import type { NativeStackNavigationProp } from '@react-navigation/native-stack'
 
 import type { AppTabScreenParamList } from './AppTabScreen'
 
-type AppTabNavigation = NativeStackNavigationProp<AppTabScreenParamList>
+export type AppTabNavigation = NativeStackNavigationProp<AppTabScreenParamList>
 
 type AppTabNavigationContextValue = {
-  navigation: Nullable<AppTabNavigation>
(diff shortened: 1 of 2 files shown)


packages/mobile/src/screens/app-screen/AppTabNavigationProvider.tsx
```diff
@@ -32,7 +31,7 @@ export const AppTabNavigationProvider = (
   props: AppTabNavigationProviderProps
 ) => {
   const { children } = props
-  const [navigation, setNavigation] = useState<Nullable<AppTabNavigation>>(null)
+  const [navigation, setNavigation] = useState({} as AppTabNavigation)
```
sliptype, line 34: Why was this needed?

sliptype (approved):
Very interesting, great find

packages/mobile/src/screens/app-screen/AppTabNavigationProvider.tsx
dylanjeffers: yeah it's pretty bad, but since we are using this as the main navigator now, there is a moment where it's null as it's setting up, which was causing runtime issues. Was thinking this is a bit easier to do here, than cascade the null down to the useNavigate, which was messing with it's returned type, causing useNavigation() to return Nullable<Navigator>, so casted early. lmk though, open to suggestions

sliptype: Oof, yeah this is probably fine for know

PR #2074, opened 2022-10-05
[C-1269] Fix now-playing drawer and scrubber

dylanjeffers:
### Description

A lot to unpack here.

At it's simplest, the app-background -> app-foreground scrubber position recalculation was happening too often, which was incorrectly setting it for tracks that are short. Explicitly checking previous app-state before calculating prevents that recalc.

However as I was digging around I noticed a pretty major performance regression, which is that we were literally un-rendering, and then quickly re-rendering the entire the now-playing drawer at the end of every song.... this was due to logic in now-playing drawer to require user and track before rendering. This was further complicating the scrubber calc logic cause the component would always run its effects cause it was a remount. I was able to get it to always stick around and handle the null user/track case, which is great! the now-playing drawer is visibly smoother in all ways when a track finishes.

I also updated the styles for each of the components that were using `createStyles` since those would call a `Stylesheet.create` every rerender, which aint the move now that useStyles caches so well.

### DIFF
--- a/packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
+++ b/packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
@@ -271,53 +271,46 @@ const NowPlayingDrawer = (props: NowPlayingDrawerProps) => {
           { paddingTop: staticTopInset.current, paddingBottom: insets.bottom }
         ]}
       >
-        {isPlayBarShowing ? (
-          <>
-            <View style={styles.playBarContainer}>
-              <PlayBar
-                track={track}
-                user={user}
-                onPress={onDrawerOpen}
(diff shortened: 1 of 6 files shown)


sliptype (approved):
Oh hell yes, great work!

packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
```diff
@@ -271,7 +271,7 @@ const NowPlayingDrawer = (props: NowPlayingDrawerProps) => {
           { paddingTop: staticTopInset.current, paddingBottom: insets.bottom }
         ]}
       >
-        {track && user && (
+        {isPlayBarShowing ? (
```
raymondjacobson, line 274: great call.

PR #2075, opened 2022-10-05
[C-1166] Avoid only primary image

raymondjacobson:
### Description

The logic that handles fetching images from secondaries & legacy image formats is unfortunately split into two places:

Magic image component in native: https://github.com/AudiusProject/audius-client/blob/ee0b668b1165d7f980cf12b8fe59b8dfe5e77ece/packages/mobile/src/components/image/UserImage.tsx#L56

Libs: https://github.com/AudiusProject/audius-protocol/blob/6fdd8a88cb3d72da43743def17255c0fcc40edc7/libs/src/api/File.ts#L94

I think probably we want to move everything to magic image component in native, but this in the near-term makes all "Dynamic Images" work since the magic image thing in native isn't built to handle multiple sizes yet.

Curious for thoughts on this PR...!

At least this doesn't incur double fetch and we are doing some preloading!

### Dragons

### How Has This Been Tested?

Simulator, prod 

Simulator Screen Shot - iPhone 14 Pro - 2022-10-04 at 21 27 09

### Feature Flags ###

### DIFF
--- a/packages/common/src/services/audius-backend/AudiusBackend.ts
+++ b/packages/common/src/services/audius-backend/AudiusBackend.ts
@@ -206,7 +206,7 @@ type WaitForLibsInit = () => Promise<unknown>
 
 type AudiusBackendParams = {
   claimDistributionContractAddress: Maybe<string>
-  disableImagePreload?: boolean
+  imagePreloader?: (url: string) => Promise<boolean>
   env: Env
   ethOwnerWallet: Maybe<string>
   ethProviderUrls: Maybe<string[]>
@@ -260,7 +260,7 @@ type AudiusBackendParams = {
 
 export const audiusBackend = ({
   claimDistributionContractAddress,
-  disableImagePreload,
+  imagePreloader,
   env,
   ethOwnerWallet,
   ethProviderUrls,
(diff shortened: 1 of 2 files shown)


packages/common/src/services/audius-backend/AudiusBackend.ts
```diff
@@ -464,17 +464,28 @@ export const audiusBackend = ({
...
-    if (disableImagePreload) {
-      return firstImageUrl
-    }
-
     if (primary) {
-      // Attempt to fetch/load the image using the first creator node gateway
-      const preloadedImageUrl = await preloadImage(firstImageUrl)
+      if (imagePreloader) {
+        try {
+          const preloaded = await imagePreloader(firstImageUrl)
```
dylanjeffers, line 470: so this preloads the image which then get's used by the Image component when it's paired with a url that matches what's been preloaded?

dylanjeffers: This also prevents a double fetch since we dont go through the cidCache + preloadImage section, and then the Image itself?

dylanjeffers (approved):
sweet i think im getting it, if my comments are on track?

packages/common/src/services/audius-backend/AudiusBackend.ts
raymondjacobson: ^ yep that's exactly right.

this is all overly complex & there are demons everywhere

dylanjeffers: well i started looking at this pr really not understanding the state of our images, so this was v helpful

PR #2076, opened 2022-10-05
[C-1205,C-1273] Improve splash screen animation, root screen management

dylanjeffers:
### Description

After lots of testing, I believe the source of the white screen is due to the final frames of the splash being fully white, and lag associated with removing the Lottie animation from the view when animation is finished. When there is lag (due to so many things competing for main thread), the animation is paused in it's final frames, hence the white screen issue. To avoid this, I added an opacity animation that is delayed and quick on the back half of the animation, resulting in 0 opacity for the final frames of the lottie animation as it's being removed. It looks much better I think!

While experimenting with reducing the main thread churn as sign-in/feed is loading in (which could potentially improve speed at which the splash screen is removed from dom), I realized we could drastically improve animations from sign-in -> feed, and account-settings -> sign-out -> sign-in, by making a root stack that contains both of them, instead of conditional rendering outside nav flow. This flow is buttery smooth now, which is makes the app feel a lot more profesh, very stoked about this.

I also went ahead and cleaned up the RootScreen.tsx file and moved things to separate files as it was getting pretty overwhelming!

### DIFF
--- a/packages/mobile/src/hooks/useNavigation.ts
+++ b/packages/mobile/src/hooks/useNavigation.ts
@@ -64,7 +64,7 @@ export const useNavigation = <
         return StackActions.push(screen as string, params)
       }
 
-      nativeNavigation?.dispatch(customPushAction)
+      nativeNavigation.dispatch(customPushAction)
     },
     [nativeNavigation]
   )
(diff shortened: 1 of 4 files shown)


raymondjacobson (approved):
this makes so much sense. of course we had a separate stack for sign on when it was implemented in native alongside all the webview stuff

sliptype (approved):
Amazing stuff! Yup makes a lot of sense

PR #2078, opened 2022-10-05
[C-1246] Fix smart collections

dylanjeffers:
### Description

Basically the `resetCollection` action was incorrectly typed, requiring CID and userCID, but smart collections don't have CID. Updating type and updating useFocusEffect in collections-screen to call reset even if CID/userCID not there.

- Also updated the smart collection code to conform a bit more to our standards
- Removed the console.warn about missing user/collection for collection screen, since this is possible with normal behavior, we should add a skeleton eventually tho


nicoback2 (commented):
amazing!

nicoback2 (approved):

sliptype (approved):

PR #2082, opened 2022-10-05
[C-977] Add Screen change analytics

sliptype:
### Description

* Add analytics on screen change on mobile
* Primary routes are recorded via the NavigationContainer `onStateChange`
* Secondary routes are recorded via effects in the corresponding components
   This is necessary because in some cases the data needed to build the url is not available in the navigation state. Track screen for example records a url like this: `/KSHMRmusic/anywheres-home` but we only have track id available in the navigation state

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/navigation-container/NavigationContainer.tsx
+++ b/packages/mobile/src/components/navigation-container/NavigationContainer.tsx
@@ -1,8 +1,10 @@
 import type { ReactNode } from 'react'
+import { useRef } from 'react'
 
 import { accountSelectors } from '@audius/common'
 import type { LinkingOptions } from '@react-navigation/native'
 import {
+  useNavigationContainerRef,
   getStateFromPath,
   NavigationContainer as RNNavigationContainer
 } from '@react-navigation/native'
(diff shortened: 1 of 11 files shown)


packages/mobile/src/screens/profile-screen/ProfileScreen.tsx
```diff
@@ -104,6 +106,15 @@ export const ProfileScreen = () => {
     }
   }, [status])
 
+  // Record screen view
+  useEffect(() => {
```
dylanjeffers, line 110: just wondering why some are effect and some are focusEffect?

dylanjeffers (approved):
sweet great changes, lmk if you want to discuss putting it in screen component

packages/mobile/src/screens/profile-screen/ProfileScreen.tsx
sliptype: I was using useEffect when the url is based on data inside the component and useFocusEffect when the url is a static value. Pushing this into the screen component means I can use useEffect across the board with a dependency on the url

packages/mobile/src/components/navigation-container/NavigationContainer.tsx
```diff
@@ -157,7 +164,27 @@ const NavigationContainer = (props: NavigationContainerProps) => {
...
   return (
-    <RNNavigationContainer linking={linking} theme={navigationThemes[theme]}>
+    <RNNavigationContainer
+      linking={linking}
+      theme={navigationThemes[theme]}
+      ref={navigationRef}
+      onReady={() => {
+        routeNameRef.current = getPrimaryRoute(navigationRef.getRootState())
+      }}
+      onStateChange={async () => {
```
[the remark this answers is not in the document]
sliptype: I double checked and these aren't causing any extra renders. Also this is how the docs recommend to do it: https://reactnavigation.org/docs/screen-tracking/

dylanjeffers (approved):
awesome changes

PR #2083, opened 2022-10-05
[C-1184] Fetch lineups after coming back online

amendelsohn:
### Description
Problem: 
Lineups are not loading when returning to online after the app started offline.

Solution: 
This fix switches `setupBackend` to a best-effort approach with retrying each time we return to online.

### Dragons

Changing the init logic for `setupBackend` could lead us to bad race conditions or failed requests. We may need to add `waitForBackendSetup` to more of our fetch sagas to ensure they only run after the `apiClient` is ready. I'd welcome guidance here from someone who knows the setup logic better.

### How Has This Been Tested?

Simulator and device
1. Launch app in airplane mode
2. Observe offline indicator
3. Turn off airplane mode
4. Optionally navigate in app while offline
5. Content on current page should load as normal
6. Navigate around after, content should continue to load normally

Keep an eye on error counts in test and prod. See if it's causing an increase in random crashes/logged errors


packages/mobile/src/components/reachability/ToggleReachability.tsx
```diff
@@ -0,0 +1,38 @@
...
+import IconNoWifi from 'app/assets/images/iconNoWifi.svg'
+import { Switch } from 'app/components/core'
+const { getIsReachable } = reachabilitySelectors
+export const ToggleReachability = () => {
+  const dispatch = useDispatch()
+  const reachability = useSelector(getIsReachable)
+
+  useEffectOnce(() => {
+    // dispatch(reachabilityActions.setUnreachable())
+  })
```
amendelsohn: Unnecessary

amendelsohn:
Whoops, included some extraneous changes I didn't mean to. Update on the way.

packages/common/src/store/reachability/reducer.ts
```diff
@@ -6,7 +6,7 @@ import { ReachabilityState } from './types'
 type ReachabilityActions = ActionType<typeof actions>
 
 const initialState = {
-  networkReachable: true
+  networkReachable: false
```
amendelsohn, line 9: This is a shift in the initial assumption of connectivity. It helps the sagas block properly when they should.

TODO: test this in web and desktop to make sure it doesn't break initial load

packages/mobile/src/screens/root-screen/RootScreen.tsx
```diff
@@ -61,6 +64,10 @@ export const RootScreen = ({ isReadyToSetupBackend }: RootScreenProps) => {
     }
   }, [accountStatus])
 
+  useBecomeReachable(() => {
+    if (isReadyToSetupBackend && !isBackendSetup) dispatch(setupBackend())
+  })
+
```
amendelsohn, line 70: We may need something similar in web/desktop

packages/web/src/common/store/lineup/sagas.js
```diff
@@ -116,16 +118,24 @@ function* fetchLineupMetadatasAsync(
...
   action
 ) {
-  const initLineup = yield select(lineupSelector)
+  const initLineup = yield* select(lineupSelector)
+  const establishedReachability = yield* call(awaitReachability)
+  // If we couldn't connect, show the error page
+  // and just sit here waiting for reachability.
+  if (!establishedReachability) {
+    yield* put(lineupActions.fetchLineupMetadatasFailed())
+    yield* take(reachabilityActions.SET_REACHABLE)
```
rickyrombo, line 127: Clever! Does this work if we toggle between network states a few times?

packages/web/src/common/store/pages/profile/lineups/feed/sagas.js
```diff
@@ -12,10 +12,9 @@ import {
   waitForAccount,
   makeUid
 } from '@audius/common'
-import { select, call, takeEvery, put } from 'redux-saga/effects'
-
 import { getConfirmCalls } from 'common/store/confirmer/selectors'
 import { LineupSagas } from 'common/store/lineup/sagas'
+import { select, call, takeEvery, put } from 'typed-redux-saga'
```
rickyrombo, line 17: nit: Maybe stick with the `redux-saga/effects` unless we're moving this file to typescript? Alternatively, move this file to typescript? :)

packages/web/src/common/store/backend/sagas.ts
```diff
@@ -44,7 +44,7 @@ export function* waitForBackendSetup() {
   }
 }
 
-function* awaitReachability() {
+export function* awaitReachability() {
```
rickyrombo, line 47: Is this still used in this file? Curious if it should still belong here?

rickyrombo (approved):

packages/mobile/src/screens/root-screen/RootScreen.tsx
```diff
@@ -60,6 +63,10 @@ export const RootScreen = ({ isReadyToSetupBackend }: RootScreenProps) => {
     }
   }, [accountStatus])
 
+  useBecomeReachable(() => {
+    if (isReadyToSetupBackend && !isBackendSetup) dispatch(setupBackend())
```
sliptype, line 67: I don't think it would matter but we could also add the `!isBackendSetup` condition to the useEffect that does the initial backend setup above. Maybe we could break into a helper function because both places are doing the same thing?

packages/web/src/common/store/lineup/sagas.js
```diff
@@ -137,14 +148,19 @@ function* fetchLineupMetadatasAsync(
 
       // Let page animations on mobile have time to breathe
       // TODO: Get rid of this once we figure out how to make loading better
-      const isNativeMobile = yield getContext('isNativeMobile')
+      const isNativeMobile = yield* getContext('isNativeMobile')
       if (!isNativeMobile && isMobileWeb()) {
         yield delay(100)
```
sliptype, line 153: I wonder if we can get rid of this entirely now

packages/web/src/common/store/lineup/sagas.js
```diff
@@ -137,14 +148,19 @@ function* fetchLineupMetadatasAsync(
...
       }
 
+      // TODO: await reachability (race vs timeout?)
       const lineupMetadatasResponse = yield call(lineupMetadatasCall, action)
+      console.log(`metadatasResponse: ${lineupMetadatasResponse}`)
```
sliptype, line 158: Log

packages/web/src/common/store/lineup/sagas.js
```diff
@@ -137,14 +148,19 @@ function* fetchLineupMetadatasAsync(
 
       // Let page animations on mobile have time to breathe
       // TODO: Get rid of this once we figure out how to make loading better
-      const isNativeMobile = yield getContext('isNativeMobile')
+      const isNativeMobile = yield* getContext('isNativeMobile')
       if (!isNativeMobile && isMobileWeb()) {
         yield delay(100)
       }
 
+      // TODO: await reachability (race vs timeout?)
```
sliptype, line 156: Is this todo done by lines 123-130?

packages/web/src/common/store/backend/sagas.ts
```diff
@@ -44,28 +42,7 @@ export function* waitForBackendSetup() {
...
-  return !!action
-}
-
 export function* setupBackend() {
-  const establishedReachability = yield* call(awaitReachability)
-  // If we couldn't connect, show the error page
-  // and just sit here waiting for reachability.
-  if (!establishedReachability) {
-    console.error('No internet connectivity')
-    yield* put(accountActions.fetchAccountNoInternet())
```
sliptype, line 64: So we aren't dispatching this action anymore right? Not sure if it's necessary. But might be good to track down

packages/web/src/common/store/reachability/sagas.ts
```diff
@@ -0,0 +1,20 @@
...
+  reachabilityActions,
+  reachabilitySelectors,
+  getContext
+} from '@audius/common'
+import { delay, take, select, race } from 'typed-redux-saga'
+
+const { getIsReachable } = reachabilitySelectors
+const REACHABILITY_TIMEOUT_MS = 8 * 1000
+
+export function* awaitReachability() {
```
sliptype, line 11: I know you didn't change it but this function name feels off to me. It doesn't actually await being in a reachable state, it just gets the current reachability

Maybe we could rename it to something like `getReachability`. We could even have an `awaitReachability` that does
```ts
yield take(reachabilityActions.SET_REACHABLE)
```
for us

packages/mobile/src/hooks/useReachability.ts
```diff
@@ -0,0 +1,43 @@
...
+export const useReachability = (
+  onBecomeReachable: OnBecomeReachable | null,
+  onBecomeUnreachable: OnBecomeUnreachable | null
+) => {
+  const isInternetReachable = useSelector(getIsReachable)
+  const [lastReachabilityState, setLastReachabilityState] =
+    useState(isInternetReachable)
+
+  useEffect(() => {
+    if (isInternetReachable !== lastReachabilityState) {
```
sliptype, line 19: Nice!

sliptype (approved):

audius-infra:
Preview this change https://demo.audius.co/amendel-initialize-after-offline

audius-infra:
Preview this change https://demo.audius.co/amendel-initialize-after-offline

packages/common/src/store/reachability/reducer.ts
amendelsohn: seems to be working on desktop and web

PR #2092, opened 2022-10-06
[C-1279] Improve drawer perf, improve bottom-bar navigation

dylanjeffers:
### Description

Greatly improves drawer performance:
 - memoizes the panResponder, which is good practice. Reduces rerenders by 10x
 - refactors to makeStyles, which will prevent tons of duplicate Stylesheet.create calls, (basically on ever render!)

Solidly improves bottom-bar performance:
- Ensures that now-playing drawer (an expensive component) does not rerender during any bottom-bar tab navigation.

### Dragons

I removed the `isOpen` dependency on the panResponder @raymondjacobson to further reduce recompute. Everything seems to work, but a callout to you if there was a special reason for it.


sliptype (approved):
Omg this is awesome, nice finds

PR #2099, opened 2022-10-06
Only fetch favorites/profiles once

dylanjeffers:
### Description

Improves bottom bar and profile perf by only fetching data once. Users can still see updated data for profile with pull-to-refresh.

### DIFF
--- a/packages/mobile/src/screens/favorites-screen/TracksTab.tsx
+++ b/packages/mobile/src/screens/favorites-screen/TracksTab.tsx
@@ -15,8 +15,8 @@ import {
   savedPageSelectors,
   tracksSocialActions
 } from '@audius/common'
-import { useFocusEffect } from '@react-navigation/native'
 import { useDispatch, useSelector } from 'react-redux'
+import { useEffectOnce } from 'react-use'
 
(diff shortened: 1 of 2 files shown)


raymondjacobson (approved):

sliptype (changes requested):
Doing this on the favorites screen feels like it will cause problems because there is no pull to refresh there.

Like when a user favorites a track then it doesn't show up in their favorites list right?

dylanjeffers:
> Doing this on the favorites screen feels like it will cause problems because there is no pull to refresh there.
> 
> Like when a user favorites a track then it doesn't show up in their favorites list right?

So there is optimistic updates, so on device it is fine, its just if you favorite on web it wont show up. what do we think?

sliptype:
> > Doing this on the favorites screen feels like it will cause problems because there is no pull to refresh there.
> > Like when a user favorites a track then it doesn't show up in their favorites list right?
> 
> So there is optimistic updates, so on device it is fine, its just if you favorite on web it wont show up until you kill app. maybe we do a fetch for when we come back from background? what do we think?

I forgot about optimistic updates, this is fine then for now

sliptype (approved):

PR #2107, opened 2022-10-07
[C-1040] Remove extraneous mobile logs

dylanjeffers:
### Dragons

Not sure what to do about the optimizely line. it appears that variable doesn't exist or have a default?


audius-infra:
Preview this change https://demo.audius.co/dj-remove-extraneous-logs

raymondjacobson:
Not sure I understand the intent of this PR? I think most of these logs are actually good to know about / we should address the root cause rather than just removing the logs
I also definitely want to keep the remote config var b/c it can help the network progress in instances where solana is down

dylanjeffers:
> Not sure I understand the intent of this PR? I think most of these logs are actually good to know about / we should address the root cause rather than just removing the logs I also definitely want to keep the remote config var b/c it can help the network progress in instances where solana is down

So logs themselves are a bit annoying vs errors and warnings? do we want to log all the various branches for fingerprint for example? one that we should def include is issues related to dates, and rendering "null" for some pages. happy to bring back the optimizely thing though

piazzatron (approved):
Looks reasonable to me w the exception of the unhealthySlotDiffPlays - I'm supportive of not adding logs for standard happypath things like starting fingerprint + analytics.

PR #2109, opened 2022-10-07
BuyAudio: Fix for if wallet has no transactions

rickyrombo:
### Description

If the wallet initially has no transactions, this code would set the initial transaction to the first transaction after the polling starts, which is already the purchase transaction in stripe's case, so the polling would timeout.

### Dragons

### How Has This Been Tested?

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/mjp-no-txs

piazzatron (approved):

PR #2120, opened 2022-10-11
Fix favorites page filtering on old tables

Kyle-Shanks:
### Description
Quick update to fix favorites for favorites old table. it was using the new fetch logic, which is slower and not needed

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A


audius-infra:
Preview this change https://demo.audius.co/kj-Fix-favorites-filtering

sliptype (approved):
Nice! Makes sense

PR #2123, opened 2022-10-11
[C-1316] Fix getState undefined crash when opening reopening app on android

sliptype:
### Description

`navigation.getState` is not defined for a brief period when reopening the app on android. Opted to use `useNavigationState` instead for safety

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/lineup-tile/TrackTile.tsx
+++ b/packages/mobile/src/components/lineup-tile/TrackTile.tsx
@@ -19,6 +19,7 @@ import {
   shareModalUIActions,
   RepostType
 } from '@audius/common'
+import { useNavigationState } from '@react-navigation/native'
 import { useDispatch, useSelector } from 'react-redux'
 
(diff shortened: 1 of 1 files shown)


nicoback2 (approved):

dylanjeffers (commented):
awesome! i am a little worried about potentially not having the right navigator here, for example i think the track tiles in CollapsibleSectionList use navigation from tab-nav, but we actually want the bottom-bar stack nav? what do we think?

sliptype:
> awesome! i am a little worried about potentially not having the right navigator here, for example i think the track tiles in CollapsibleSectionList use navigation from tab-nav, but we actually want the bottom-bar stack nav? what do we think?

We actually want the navigation from the tab-nav, this is checking that we are on the `Tracks` tab (which happens to only be on Profile). This condition evaluates to false everywhere else, like feed

This should be ok for now, but yeah I think we could do something better here in general

PR #2125, opened 2022-10-11
[C-1309] Fix playlists/collectibles not loading

dylanjeffers:
### Description

Fixes issue where users with captials in their handle mess up with some of our selectors. By updating our profile selectors to always `toLowerCase`, this fixes any other issues.

I may remove some cases where we are manually `toLowerCase`ing things in the client, but this ultimately should fix things

### Dragons
Looking at clayton's account, his collectibles only have one entry. is this expected?


sliptype (approved):
RIP

PR #2126, opened 2022-10-11
[C-1312, C-1313, C-1314] Remove old table components and rename new table components

Kyle-Shanks:
### Description
Big update to remove all references of the old table components and use the new table components
Also renaming the new tables to remove the `test` part of the name

### Dragons

Did a lot of looking and error checking, but there is a slight chance that I missed something. app loaded fine locally though

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A


audius-infra:
Preview this change https://demo.audius.co/kj-Remove-old-tables-and-rename-new-tables

Kyle-Shanks:
> Wooooo, so satisfying to get rid of the old tables! Can we also remove the feature flag itself?

Yes, theres a ticket for that, just want to make sure this code goes out before removing it 👍

PR #2128, opened 2022-10-12
Dual writes when setting artist pick

michellebrier:
### Description
Currently artist pick lives in the identity service. With this change, we'll also write `artist_pick_track_id` to the users table in discovery when a user pins a track in the client. This is the first part in the migration to completely to reading and writing the artist pick from discovery.

### How Has This Been Tested?
Web:
- Run client with changes locally. Test against staging data, which has API updates to show the artist_pick_track_id on a user in discovery.
- Set artist pick in UI. Verify artist pick was set via API.
- Remove artist pick. Verify change with API.
- Set new artist pick in UI. Verify change with API.
- Delete track. Verify change with API

Mobile:
- Make sure tracks list correctly reflects each change above ^

### Feature Flags ###
`updateCreator` is gated by the `USER_ENTITY_MANAGER_ENABLED` feature flag.

### DIFF
--- a/packages/common/src/services/audius-backend/AudiusBackend.ts
+++ b/packages/common/src/services/audius-backend/AudiusBackend.ts
@@ -2158,11 +2158,23 @@ export const audiusBackend = ({
 
   /**
    * Sets the artist pick for a user
+   * @param {User} userMetadata
+   * @param {number} userId
    * @param {number?} trackId if null, unsets the artist pick
    */
-  async function setArtistPick(trackId: Nullable<ID> = null) {
+  async function setArtistPick(
+    userMetadata: User,
+    userId: ID,
+    trackId: Nullable<ID> = null
+  ) {
     await waitForLibsInit()
     try {
+      // Dual write to the artist_pick_track_id field in the
+      // users table in the discovery DB. Part of the migration
+      // of the artist pick feature from the identity service
+      // to the entity manager in discovery.
+      updateCreator(userMetadata, userId)
+
       const { data, signature } = await signData()
       return await fetch(`${identityServiceUrl}/artist_pick`, {
         method: 'POST',
--- a/packages/web/src/common/store/cache/users/sagas.js
+++ b/packages/web/src/common/store/cache/users/sagas.js
@@ -407,7 +407,8 @@ export function* fetchUserSocials({ handle }) {
           tiktok_handle: socials.tikTokHandle || null,
           website: socials.website || null,
           donation: socials.donation || null,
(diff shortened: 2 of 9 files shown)


packages/common/src/services/audius-backend/AudiusBackend.ts
```diff
@@ -2135,11 +2135,23 @@ export const audiusBackend = ({
...
+    userId: ID,
+    trackId: Nullable<ID> = null
+  ) {
     await waitForLibsInit()
     try {
+      // Dual write to the artist_pick_track_id field in the
+      // users table in the discovery DB. Part of the migration
+      // of the artist pick feature from the identity service
+      // to the entity manager in discovery.
+      updateCreator(userMetadata, userId)
```
michellebrier, line 2153: `updateCreator` in libs is gated by the user entity manager flag. The user metadata update logic is the same as the old flow but there is all the new encapsulating entity manager logic in `entity_manager.py` that will be run so it seems risky to not gate this with the user entity manager flag @jowlee

jowlee: It shouldn't be gated on entity manager flag. There is a params `useEntityManager` that's passed to it which toggles the flow to the update to the user's metadata from the current User contract to being doing with the entity manager contract. 
On the discovery indexing side, it be calling the same method to update the user - but I do agree that it's stacking changes that could lead to a bug. 
We could wait to roll out until entity manager it out

jowlee (approved):

jowlee:
You probs want to update the @audius/sdk dep version

packages/web/src/common/store/cache/users/sagas.js
```diff
@@ -407,7 +407,8 @@ export function* fetchUserSocials({ handle }) {
           tiktok_handle: socials.tikTokHandle || null,
           website: socials.website || null,
           donation: socials.donation || null,
-          _artist_pick: socials.pinnedTrackId || null
+          _artist_pick: socials.pinnedTrackId || null,
+          artist_pick_track_id: socials.pinnedTrackId | null
```
michellebrier, line 411: Set `artist_pick_track_id` in the user metadata with the value from the identity service when populating the user in the cache. This is necessary for the case during the migration where, after backfilling the users table in discovery, there is a track id in the users table but `artist_pick_track_id` in the cache is null. Upon saving an update to user metadata, the backfilled artist pick track id in the users table would be incorrectly overwritten.

audius-infra:
Preview this change https://demo.audius.co/mbrier/plat-411/dual_writes_pinned_track_id

PR #2134, opened 2022-10-12
[C-1295] Update new tables to allow drag to add to playlist

Kyle-Shanks:
### Description
Quick updates to allow tracks within new tables to be able to be dragged into playlists in the side nav

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/web/src/components/dragndrop/Draggable.js
+++ b/packages/web/src/components/dragndrop/Draggable.js
@@ -17,6 +17,7 @@ const Draggable = (props) => {
     kind,
     link,
     id,
+    index,
     isOwner,
     drag,
     drop,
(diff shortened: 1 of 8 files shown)


dylanjeffers (commented):
Knowing nothing, is it stable to use index for drag-drop vs something like a key/id?

audius-infra:
Preview this change https://demo.audius.co/kj-Update-new-tables-to-allow-drag-track-to-playlist

Kyle-Shanks:
> Knowing nothing, is it stable to use index for drag-drop vs something like a key/id?

Good thought. The index here is only being used in the new code that I added for the new tables and is only used for reorder logic, which has been stable so far for all of the testing. the track id will be used for the id prop and now matches the pattern that the other drag and drop components use around the app so this should work

dylanjeffers (approved):
makes sense!

PR #2136, opened 2022-10-12
Refactor bottom-tab-bar

dylanjeffers:
### Description

Refactors bottom-tab-bar by removing overloaded BottomTabBarButton, that was aware of every lottie animation file. Now we have separate components for each button. This also makes it a bit easier to develop the new nav-bar overhaul.

I was also hoping this would help with performance, but turns out the issue we are seeing where there are multiple active buttom-buttons is due to lottie animation.reset not working as expected. An upgrade could potentially help there, which ill look into soon.


sliptype (approved):
Clean!

PR #2137, opened 2022-10-12
[C-1327] Fix deep linking for tracks with diacritics

sliptype:
### Description

* Slug was not being decoded
* Handle does not need to be decoded because only alphanumeric + underscores is valid
* CollectionScreen is unaffected bc it pulls the id from the collection name

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/screens/track-screen/TrackScreen.tsx
+++ b/packages/mobile/src/screens/track-screen/TrackScreen.tsx
@@ -75,7 +75,7 @@ export const TrackScreen = () => {
     dispatch(
       fetchTrack(
         id,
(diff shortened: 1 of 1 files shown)


dylanjeffers (approved):
Simple enough. Should we do the same for collectibles? I'm are there other pages I wonder?

sliptype:
> Simple enough. Should we do the same for collectibles? I'm are there other pages I wonder?

We don't have deep linking to collectibles right now. Do you mean collections? I checked collections and they are unaffected bc they parse a numeric id from the collection "name" currently. Profiles are also unaffected bc handles can't have diacritics

PR #2147, opened 2022-10-14
[C-1385] Stream mp3 instead of segments

amendelsohn:
### Description

Switch playback on both Web and Mobile to use mp3 stream endpoint instead of hls segmented streaming.

### Dragons

Playback is a core feature. Could break it.

### How Has This Been Tested?

Sim and web local.

Tried adding bogus endpoints on mobile and it correctly skips the bad ones and streams from valid endpoints.

Existing playback monitoring.

### Feature Flags ###

N/A


audius-infra:
Preview this change https://demo.audius.co/amendel-stream-mp3

packages/mobile/src/components/audio/Audio.tsx
```diff
@@ -58,6 +58,14 @@ const styles = StyleSheet.create({
   }
 })
 
+const getStreamUris = (track: Track, trackOwner: UserMetadata) => {
+  const encodedTrackId = encodeHashId(track.track_id)
+  const creatorNodeEndpoints = trackOwner.creator_node_endpoint.split(',')
+  return creatorNodeEndpoints.map(
```
dylanjeffers, line 64: so is it the case that hls would figure out which endpoints were the best, wheras now we are just kinda guessing? assuming this isn't a big deal, but thought id ask so i understand any potential tradeoffs

dylanjeffers (approved):
so much simpler!

packages/mobile/src/components/audio/Audio.tsx
```diff
@@ -333,34 +359,27 @@ export const Audio = () => {
...
-  const gateways = trackOwner
-    ? audiusBackendInstance.getCreatorNodeIPFSGateways(
-        trackOwner.creator_node_endpoint
-      )
-    : []
+  useEffect(() => {
+    if (track && trackOwner) {
+      setStreamUris(getStreamUris(track, trackOwner))
+    }
+  }, [track, trackOwner])
```
sliptype, line 366: Could be cleaner as a memo imo:

```suggestion
  const streamUris = useMemo(() => {
    if (track && trackOwner) {
      return getStreamUris(track, trackOwner)
    }
    return null
  }, [track, trackOwner])
```

packages/web/src/services/audio-player/AudioPlayer.ts
```diff
@@ -207,102 +151,18 @@ export class AudioPlayer {
...
-            ? `${info.title} by ${info.artist}`
-            : 'Audius'
-      }
-    }
-
-    this.duration = segments.reduce(
-      (duration, segment) => duration + parseFloat(segment.duration),
-      0
-    )
+    // TODO: Test to make sure that this doesn't break anything
```
sliptype, line 154: Lol, love that we still have this todo

sliptype (approved):
Omg so much deleted code, this is great! Trying to think of places where this could break... Would be good to test the lock screen MusicControls on mobile

packages/mobile/src/components/audio/Audio.tsx
amendelsohn: I need this to only call the state setter once per track update, since we're using the state as a queue to track which ones we haven't tried yet

packages/mobile/src/components/audio/Audio.tsx
amendelsohn: So I don't know the HLS stuff very well, but I think we were giving it the choice of endpoints to grab each segment. Before, we passed all the gateways as options which the player can use as it pleases. Presumably, it tries each one.
```
const gateways = audiusBackendInstance.getCreatorNodeIPFSGateways(
  trackOwner.creator_node_endpoint
)

const m3u8 = hlsUtils.generateM3U8Variants({
  segments: track.track_segments,
  gateways
})
```
The new logic will simply try each endpoint and fall back to the next if it errors out.
See `onError` in the [video player usage example](https://github.com/react-native-video/react-native-video/blob/master/API.md#usage)

packages/mobile/src/components/audio/Audio.tsx
sliptype: Ah gotcha, makes sense

amendelsohn: Though @piazzatron just brought up that we could use the Discovery endpoint's redirect instead. That might be better than doing the fallback in client?

packages/mobile/src/components/audio/Audio.tsx
dylanjeffers: Thanks for context

packages/mobile/src/components/audio/Audio.tsx
sliptype: Oh yeah that could be better, fewer requests and also avoids error states in the `Video` component

audius-infra:
Preview this change https://demo.audius.co/amendel-stream-mp3

raymondjacobson:
> Omg so much deleted code, this is great! Trying to think of places where this could break... Would be good to test the lock screen MusicControls on mobile

We never implemented scrubbing on lock screen controls, just FYI FWIW to any readers

amendelsohn:
> Omg so much deleted code, this is great! Trying to think of places where this could break... Would be good to test the lock screen MusicControls on mobile

Media controls working on iOS

PR #2148, opened 2022-10-14
[C-1341] Add notifications bottom tab screen

dylanjeffers:
### Description

Implements notifications bottom-tab-bar
Implements notifications screen
Updates account-drawer touch targets

### DIFF
--- a/packages/mobile/src/screens/notifications-screen/NotificationsScreen.tsx
+++ b/packages/mobile/src/screens/notifications-screen/NotificationsScreen.tsx
@@ -6,6 +6,7 @@ import { useDispatch } from 'react-redux'
 
 import { Screen } from 'app/components/core'
 import { Header } from 'app/components/header'
+import { usePopToTopOnDrawerOpen } from 'app/hooks/usePopToTopOnDrawerOpen'
 
 import { ScreenContent } from '../ScreenContent'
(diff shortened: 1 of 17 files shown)


packages/mobile/src/screens/notifications-screen/NotificationsScreen.tsx
```diff
@@ -17,6 +18,7 @@ const messages = {
 }
 
 export const NotificationsScreen = () => {
+  usePopToTopOnDrawerOpen()
```
sliptype, line 21: Do you need this? Because this only makes sense for screens to will be navigated to from the notification drawer

sliptype (approved):

packages/mobile/src/screens/notifications-screen/NotificationsScreen.tsx
dylanjeffers: so it has the navigation setting in there as well, so that's main reason why. but once we remove the old way, we can remove it and just have the stack nav setting (assuming we still need that, even)

sliptype: Yeah makes sense, easiest just to leave it!

PR #2152, opened 2022-10-17
[C-1345] redesign primary/secondary screens

dylanjeffers:
### Description

Redesigns primary and secondary screens.

Primary screens now have a smaller screen header with an icon
Secondary screens have a redesigned headerTitle with uppercase and icon

https://user-images.githubusercontent.com/8230000/196255437-80b7562e-58e8-4959-843d-e4abacd7a693.mov


packages/mobile/src/components/core/Screen/ScreenHeader.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { View } from 'react-native'
+import type { SvgProps } from 'react-native-svg'
+
+import { GradientIcon, GradientText } from 'app/components/core'
+import { useFeatureFlag } from 'app/hooks/useRemoteConfig'
+import type { StylesProp } from 'app/styles'
+import { makeStyles } from 'app/styles'
+
+type ScreenHeaderProps = {
+  children?: ReactChild
```
sliptype, line 14: Is `ReactChild` the preferred way of doing this? Could use ReactNode to allow multiple children right?

sliptype (approved):
Good stuff!

PR #2154, opened 2022-10-17
[C-1283][C-1386] Offline Mode - Favorites download and playback

amendelsohn:
image

### Description

Offline track storage and playback for favorites tracks tab.

Includes:
1. UI toggle to trigger favorites
2. Store tracks on disk
3. Show which tracks are download-ing/-ed
4. Load favorites UI properly when offline
5. Playback tracks while offline

### Dragons

1. Am I shooting myself in the foot by reusing the saves lineup?
2. Anything I missed that should be behind the feature flag?

### How Has This Been Tested?

Sim and device
Note that sim does not always get the reachability signal, so device test is more reliable

In the future, we will add metrics to track usage, but that's a later change.

### Feature Flags ###

`OFFLINE_MODE_ENABLED`


audius-infra:
Preview this change https://demo.audius.co/amendel-load-favorites-ui-from-disk

packages/mobile/src/hooks/useLoadOfflineTracks.ts
```diff
@@ -0,0 +1,89 @@
...
+import { useFeatureFlag } from 'app/hooks/useRemoteConfig'
+import { loadTracks } from 'app/store/offline-downloads/slice'
+
+import {
+  getTrackJson,
+  listTracks,
+  verifyTrack
+} from '../services/offline-downloader/offline-storage'
+
+export const useLoadOfflineTracks = async (collection: string) => {
```
nicoback2, line 22: Sorry we def talked about this before, but remind me why we're putting this in a hook instead of a saga?

packages/mobile/src/hooks/useLoadOfflineTracks.ts
```diff
@@ -0,0 +1,89 @@
...
+    if (!isOfflineModeEnabled) return
+
+    const trackIds = await listTracks()
+    const tracks: UserTrackMetadata[] = []
+    const savesLineupTracks: (UserTrackMetadata & { uid: string })[] = []
+    const cacheTracks: { uid: string; id: number; metadata: Track }[] = []
+    const cacheUsers: { uid: string; id: number; metadata: UserMetadata }[] = []
+
+    for (const trackId of trackIds) {
+      const verified = await verifyTrack(trackId)
```
nicoback2, line 38: ditto q re: error handling

packages/mobile/src/hooks/useLoadOfflineTracks.ts
```diff
@@ -0,0 +1,89 @@
...
+          ) {
+            savesLineupTracks.push(lineupTrack)
+          }
+        })
+        .catch(() => console.warn('Failed to load track from disk', trackId))
+    }
+
+    dispatch(cacheActions.add(Kind.TRACKS, cacheTracks, false, true))
+    dispatch(cacheActions.add(Kind.USERS, cacheUsers, false, true))
+    dispatch(loadTracks(savesLineupTracks as unknown as Track[]))
```
nicoback2, line 71: what's the type issue here?

packages/mobile/src/screens/favorites-screen/FavoritesScreen.tsx
```diff
@@ -41,13 +53,20 @@ export const FavoritesScreen = () => {
...
+
   return (
     <Screen>
       <ScreenHeader
         text={messages.header}
         icon={IconFavorite}
         styles={{ icon: { marginLeft: 3 } }}
-      />
+      >
+        <DownloadToggle
```
nicoback2, line 65: Why not have `DownloadToggle` get the tracks from the store (instead of passing them in)?

packages/mobile/src/screens/favorites-screen/FavoritesScreen.tsx
```diff
@@ -41,13 +53,20 @@ export const FavoritesScreen = () => {
...
+          collection={DOWNLOAD_REASON_FAVORITES}
+          tracks={savedTracks.entries}
```
nicoback2, line 67: are all the saved tracks loaded at once on this screen (as opposed to loading in more when you scroll)? also should we disable this if they're not done fetching yet (or the fetch errored out)?

packages/mobile/src/screens/favorites-screen/FavoritesScreen.tsx
```diff
@@ -41,13 +53,20 @@ export const FavoritesScreen = () => {
...
   return (
     <Screen>
       <ScreenHeader
         text={messages.header}
         icon={IconFavorite}
         styles={{ icon: { marginLeft: 3 } }}
-      />
+      >
+        <DownloadToggle
+          collection={DOWNLOAD_REASON_FAVORITES}
```
nicoback2, line 66: Eventually this'll also download all your saved albums and playlists too right?

nicoback2:
overall looks good to me - would be great to get more eyes on this since it's a big boi

packages/mobile/src/components/offline-downloads/DownloadToggle.tsx
```diff
@@ -0,0 +1,69 @@
...
+        })
+      } else {
+        // TODO: remove only downloads associated with this collection
+        purgeAllDownloads()
+      }
+    },
+    [collection, isOfflineModeEnabled, tracks]
+  )
+
+  if (!isOfflineModeEnabled) return null
```
sliptype, line 62: It might be better to conditionally render this component based on the feature flag at the parent level. Then non of the hooks in here will get run

packages/mobile/src/hooks/useLoadOfflineTracks.ts
```diff
@@ -0,0 +1,89 @@
...
+  const { isEnabled: isOfflineModeEnabled } = useFeatureFlag(
+    FeatureFlags.OFFLINE_MODE_ENABLED
+  )
+  const dispatch = useDispatch()
+
+  useAsync(async () => {
+    if (!isOfflineModeEnabled) return
+
+    const trackIds = await listTracks()
+    const tracks: UserTrackMetadata[] = []
```
sliptype, line 32: Is `tracks` being used? Other than pushing things to it

packages/mobile/src/screens/favorites-screen/TracksTab.tsx
```diff
@@ -126,11 +129,15 @@ export const TracksTab = () => {
...
-    [dispatch, isPlaying, playingUid]
+    [playingUid, isPlaying, dispatch, isReachable, isOfflineModeEnabled]
   )
 
   const isLoading = savedTracksStatus !== Status.SUCCESS
-  const hasNoFavorites = savedTracks.entries.length === 0
+  let tracks = savedTracks.entries
+  if (!isReachable && isOfflineModeEnabled) {
+    tracks = Object.values(offlineTracks)
+  }
```
sliptype, line 139: nit but I think this would be better as an expression
```suggestion
const tracks =
  !isReachable && isOfflineModeEnabled
    ? Object.values(offlineTracks)
    : savedTracks.entries
```

sliptype (approved):
Amazing stuff! Looks mostly good to me

audius-infra:
Preview this change https://demo.audius.co/amendel-load-favorites-ui-from-disk

packages/mobile/src/hooks/useLoadOfflineTracks.ts
amendelsohn: We didn't really have a good reason one way or the other. Personally, I think it's easier to reason about the hooks since they get called very directly from the components that rely on them. However, we use sagas so much everywhere that I'm not sure it's worth introducing a new pattern now.

packages/mobile/src/screens/favorites-screen/FavoritesScreen.tsx
amendelsohn: The same toggle component will need to get tracks from the current collection or from favorites. I figured it'd be easier to pass it in from the parent than to bake in the knowledge of where to select it from state based on `collection_id | 'favorites'`

packages/mobile/src/screens/favorites-screen/FavoritesScreen.tsx
amendelsohn: No, in fact I think I need to rewrite the downloader to work properly for saves. I'm going to tackle that in a future PR though as I don't currently have the queueing mechanism for downloads. My understanding is that the `favorites` endpoint returns every favorited item (not paginated) so we can queue them, fetch each one from discovery as needed, and save it.

PR #2155, opened 2022-10-17
[C-1347] Improve navigation performance

sliptype:
### Description

Reduces rerenders when navigating by updating our custom `useNavigation` hook to:
* Use react-navigation `useNavigation` instead of pulling the navigation from the AppTabScreen context
  > Components that need the context navigation now pass it to `useNavigation` as the customNavigation, so their pushes are also de-duped
* De-dupe navigation actions by saving the last action in a ref and comparing when a new action comes in. The saved action is reset after 500ms so that it can be performed again after the initial transition is complete. This worked better than trying to reset on `beforeRemove` because that attached a listener for every instance of the `useNavigation` hook
* Improve types

### Dragons

**This touches navigation so should be considered dangerous!**

### How Has This Been Tested?

Tested navigation:
* Navigation from the now playing drawer pushes to the correct stack
* Navigation from the notifications drawer pushes to the correct stack
* Navigation from Trending Playlists and other overflow menus push to the correct stack
* Duplicate navigation is prevented on all screens

Profiled performance
* Consumers of `useNavigation` do not rerender when a navigation occurs

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
+++ b/packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
@@ -30,6 +30,7 @@ import Drawer, {
 } from 'app/components/drawer'
 import { Scrubber } from 'app/components/scrubber'
 import { useDrawer } from 'app/hooks/useDrawer'
+import { useNavigation } from 'app/hooks/useNavigation'
 import { AppDrawerContext } from 'app/screens/app-drawer-screen'
 import { AppTabNavigationContext } from 'app/screens/app-screen'
 import { getAndroidNavigationBarHeight } from 'app/store/mobileUi/selectors'
@@ -98,7 +99,10 @@ export const NowPlayingDrawer = memo(function NowPlayingDrawer(
   props: NowPlayingDrawerProps
 ) {
   const { translationAnim } = props
-  const { navigation } = useContext(AppTabNavigationContext)
+  const { navigation: contextNavigation } = useContext(AppTabNavigationContext)
+  const navigation = useNavigation({
+    customNavigation: contextNavigation
+  })
   const dispatch = useDispatch()
   const insets = useSafeAreaInsets()
   const androidNavigationBarHeight = useSelector(getAndroidNavigationBarHeight)
--- a/packages/mobile/src/hooks/useNavigation.ts
+++ b/packages/mobile/src/hooks/useNavigation.ts
@@ -29,6 +29,17 @@ type UseNavigationOptions<NavigationProp extends RNNavigationProp<any>> = {
   customNavigation?: NavigationProp
 }
 
+/**
+ * Custom wrapper around react-navigation `useNavigation`
+ *
+ * Features:
+ * - Prevent duplicate navigation pushes
+ * - Apply contextual params to all routes
+ *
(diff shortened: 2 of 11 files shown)


packages/mobile/src/screens/app-drawer-screen/useAppDrawerNavigation.ts
```diff
@@ -20,27 +17,35 @@ export const useNotificationNavigation = () => {
...
-      }
+  const navigation = useNavigation()
 
-  const navigation = useNavigation<
-    AppTabScreenParamList & ProfileTabScreenParamList
-  >(navigationOptions)
+  const notificationNavigation = useMemo(
+    () => ({
+      ...navigation,
+      ...drawerHelpers
```
sliptype, line 25: @dylanjeffers this is touching your stuff. It seems that `drawerHelpers` really isn't a navigator and we were only able to treat it as one because of the merging going on in `useNavigation`. I pulled the merge out here so that the types of `useNavigation` could be more accurate

dylanjeffers: sounds good to me, thanks for headsup

packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
```diff
@@ -98,7 +99,10 @@ export const NowPlayingDrawer = memo(function NowPlayingDrawer(
   props: NowPlayingDrawerProps
 ) {
   const { translationAnim } = props
-  const { navigation } = useContext(AppTabNavigationContext)
+  const { navigation: contextNavigation } = useContext(AppTabNavigationContext)
+  const navigation = useNavigation({
```
dylanjeffers, line 103: nice nice

packages/mobile/src/hooks/useNavigation.ts
```diff
@@ -1,95 +1,73 @@
...
-        ) {
-          // react-navigation considers this a no-op
-          return CommonActions.navigate(lastRoute)
-        }
-        return StackActions.push(screen as string, params)
+    (...config: PerformNavigationConfig<ParamList>) => {
+      if (!isEqual(lastNavAction.current, config)) {
+        ;(navigation as NativeStackNavigationProp<any>).push(...config)
+        lastNavAction.current = config
+        setTimeout(() => {
```
dylanjeffers, line 53: okay this is interesting, what's the idea here? imo maybe it's okay to not allow a user to push to the same page they are currently on no matter what?

dylanjeffers (approved):
makes a ton of sense to me!

packages/mobile/src/hooks/useNavigation.ts
sliptype: So it's to account for popping off the stack and then trying to push the previous thing again. Only the push action is stored here so the next push would get cancelled

sliptype: That one only gets called when we call it, not when swiping back. I went down the path of using an event handler but it then gets called it every instance of useNavigation

dylanjeffers: tricky, well sg!

PR #2164, opened 2022-10-24
[C-1370] Fix app initialization sequence and assumptions for offline and online startup

amendelsohn:
### Description

Fix client init race conditions when starting offline vs online.

Changes:
1. Assume we're offline in initial state, but check right away.
2. Optimistically fetch the account before checking reachability. Fetch it again when returning online the first time.
3. Account exists no longer necessitates client initialized, so check for both where appropriate in sagas.

### Dragons

Changing the app initialization is dangerous and has potential to lock users out of sessions if done improperly. This needs careful QA.

### How Has This Been Tested?

Works on both simulator and physical device starting in offline and online modes and switching between the two.

TODO:
- [x] Test on web and desktop

We should watch our error logs as this rolls out.

### Feature Flags ###

N/A


packages/common/src/services/remote-config/remote-config.ts
```diff
@@ -74,7 +74,7 @@ export const remoteConfig = <
   let client: Client | undefined
 
   const emitter = new EventEmitter()
-  emitter.setMaxListeners(20)
+  emitter.setMaxListeners(100)
```
dylanjeffers, line 77: This for the error you were noticing earlier?

packages/web/src/common/store/tipping/sagas.ts
```diff
@@ -200,6 +202,7 @@ function* sendTipAsync() {
   const { waitForRemoteConfig } = yield* getContext('remoteConfigInstance')
   const isNativeMobile = yield* getContext('isNativeMobile')
   yield call(waitForRemoteConfig)
+  yield* call(waitForBackendSetup)
```
dylanjeffers, line 205: So do all these essentially wait until we've come online and then initialized backend to run?

dylanjeffers (approved):
This makes sense, are there cases we are missing with the waitForBackend call? Is there a way to abstract this away, like an "authenticated" route pattern, where these all require backend kinda thing?

packages/web/src/common/store/tipping/sagas.ts
amendelsohn: Yeah, that's right. The waitForAccount gate was opening since I added the optimistic offline account load. In the past, we could assume account meant backend was present too, but that's not true anymore.

packages/common/src/services/remote-config/remote-config.ts
amendelsohn: Yessir. Thanks to @sliptype for the tip

raymondjacobson: woah interesting, why do they even have a max?

amendelsohn:
> This makes sense, are there cases we are missing with the waitForBackend call? Is there a way to abstract this away, like an "authenticated" route pattern, where these all require backend kinda thing?

Yes, however, it isn't necessarily grouped easily by something like route, at least to my knowledge. We want some of our sagas to work offline but not others.

This change also doesn't solve the issue where the online -> offline transition can leave the client initialized but the requests will fail. Still, it ensures we will never get an exception for calling the client before it exists, which is not an error we handle as well as failed requests.

We will need to block or divert some requests while in offline mode anyway (e.g. tracking plays and other metrics) so perhaps we can find a clean pattern for that. It could be worth codifying the state change more explicitly.

audius-infra:
Preview this change https://demo.audius.co/amendel-open-offline

packages/common/src/services/remote-config/remote-config.ts
amendelsohn: I _think_ it's to warn you if you're creating tons of them in a loop? We are creating one per component that calls `useFeatureFlag` so it's kinda in a loop, but we also know that we're not making 1000s of them given other limitations on the size of UI lists.

packages/web/src/common/store/backend/sagas.ts
```diff
@@ -56,6 +56,9 @@ function* awaitReachability() {
 }
 
 export function* setupBackend() {
+  // Optimistically fetch account, then do it again later when we're sure we're connected
+  // This ensures we always get the cached account when starting offline if available
+  yield* put(accountActions.fetchAccount())
```
nicoback2, line 61: It might be better to make a separate action/saga to get the cached account if it exists. bc `fetchAccount` makes a backend call if the account is not cached (and the saga is `takeEvery`)

packages/web/src/common/store/pages/feed/lineup/sagas.ts
```diff
@@ -43,6 +44,7 @@ function* getTracks({
   offset: number
   limit: number
 }): Generator<any, FeedItem[], any> {
+  yield call(waitForBackendSetup)
```
nicoback2, line 47:
```suggestion
  yield* call(waitForBackendSetup)
```

packages/web/src/common/store/pages/feed/lineup/sagas.ts
```diff
@@ -15,8 +15,9 @@ import {
   CommonState,
   waitForAccount
 } from '@audius/common'
-import { select, all } from 'redux-saga/effects'
+import { select, all, call } from 'redux-saga/effects'
```
nicoback2, line 18:
```suggestion
import { select, all } from 'redux-saga/effects'
import { call } from 'typed-redux-saga'
```

nicoback2 (approved):
Left a few comments but LGTM overall! I do think we should centralize this `waitForX` logic at some point - seems very error prone to have to always remember to wait for account and/or wait for backend when writing sagas like these.

packages/web/src/common/store/user-list/utils.ts
```diff
@@ -72,6 +74,7 @@ export function createUserListProvider<T, U = void>({
     currentPage: number
     pageSize: number
   }) {
+    yield* call(waitForBackendSetup)
```
sliptype, line 77: I think it would be a little better to always have the `waitForBackendSetup` and `waitForAccount` lines together if they are both needed

packages/web/src/common/store/social/tracks/sagas.ts
```diff
@@ -587,6 +589,7 @@ export function* watchRecordListen() {
       if (isNativeMobile) return
       console.debug('Listen recorded for track', action.trackId)
 
+      yield* call(waitForBackendSetup)
```
sliptype, line 592: So couldn't we get rid of the `waitForAccount`s since the account status is guaranteed to exist if the backend setup is complete?

sliptype (approved):

packages/web/src/common/store/playlist-library/sagas.ts
```diff
@@ -119,6 +119,7 @@ function* watchUpdatePlaylistLibraryWithTempPlaylist() {
       const playlistLibrary =
         removePlaylistLibraryDuplicates(rawPlaylistLibrary)
 
+      yield call(waitForBackendSetup)
```
Kyle-Shanks, line 122:
```suggestion
      yield* call(waitForBackendSetup)
```

packages/web/src/common/store/social/tracks/sagas.ts
amendelsohn: Yes, though I have some concerns over having the same assumption come back to bite us in the future. I'm thinking of changing it to one call that does both like `waitForBackendAndAccount` just to be explicit

audius-infra:
Preview this change https://demo.audius.co/amendel-open-offline

audius-infra:
Preview this change https://demo.audius.co/amendel-open-offline

audius-infra:
Preview this change https://demo.audius.co/amendel-open-offline

PR #2165, opened 2022-10-25
[C-1368, C-1354, C-1364] Address mobile-nav qa

dylanjeffers:
### Description

Addresses mobile-nav qa

- Updates drawerStyle to "front", which makes transitions to nav-bar items smoother, and prevents double profile image.
- Prevents account-picture-header clipping due to negative margins
- Improves account-picture-header rerender perf by preventing opacity animation


raymondjacobson (approved):
Remove all negative margin? I thought we wanted to do maybe -1 spacing or somethin

"front" is interesting 👀

dylanjeffers:
> Remove all negative margin? I thought we wanted to do maybe -1 spacing or somethin
> 
> "front" is interesting 👀

yeah i thought our last convo was about removing it all for the photo so there is no cutoff. but if we want to do spacing(-1) instead, im here for it as a compromise

PR #2169, opened 2022-10-25
Remove getTrackListens endpoint from audiusBackend

sabrina-kiam:
### Description
This diff removes the getTrackListens endpoint in audiusBackend.ts. This endpoint, which goes through identity service, is being deprecated and replaced by getUserListenCountsMonthly which goes through discovery provider.

### Dragons
n/a

### How Has This Been Tested?
ran my client against staging to confirm artist dashboard still loads with other endpoint and site doesn't crash

n/a


sliptype (approved):

sabrina-kiam:
oops yes!

PR #2171, opened 2022-10-25
[C-1261] Android sign on overflow

sliptype:
### Description

* Improve layout of sign on screen so there is no overflow on android devices
* Update `windowSoftInputMode` so that the android keyboard overlays on top of the app, similar to ios

Before
before

After
after

> Side note: does the space between the email input and the sign up button seem too large to anyone else?

### Dragons

THIS TOUCHES SIGN ON

Also changing the keyboard mode could be dangerous, but I thoroughly tested cases where it could be problematic (outlined below)

### How Has This Been Tested?

Tested on pixel 3 which was exhibiting the problem, as well as OnePlus 7T and ios devices

Tested scroll behavior on android with new keyboard mode on
* Search
* Edit Profile
* Create / edit playlist

### Feature Flags ###

### DIFF
--- a/packages/mobile/android/app/src/main/AndroidManifest.xml
+++ b/packages/mobile/android/app/src/main/AndroidManifest.xml
@@ -35,7 +35,7 @@
         android:label="@string/app_name"
         android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode"
         android:screenOrientation="portrait"
-        android:windowSoftInputMode="adjustResize">
+        android:windowSoftInputMode="adjustPan">
         <intent-filter>
             <action android:name="android.intent.action.MAIN" />
             <category android:name="android.intent.category.LAUNCHER" />
--- a/packages/mobile/src/screens/signon/SignOn.tsx
+++ b/packages/mobile/src/screens/signon/SignOn.tsx
@@ -48,7 +48,6 @@ import { useThemeColors } from 'app/utils/theme'
 import type { SignOnStackParamList } from './types'
 
 const { getAccountUser } = accountSelectors
-const isAndroid = Platform.OS === 'android'
 const image = backgImage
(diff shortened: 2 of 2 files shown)


packages/mobile/android/app/src/main/AndroidManifest.xml
```diff
@@ -35,7 +35,7 @@
         android:label="@string/app_name"
         android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode"
         android:screenOrientation="portrait"
-        android:windowSoftInputMode="adjustResize">
```
dylanjeffers, line 38: coolio, so this makes it similar to ios? we double checked search, filters, and edit-profile?

packages/mobile/src/screens/signon/SignOn.tsx
```diff
@@ -732,46 +730,48 @@ const SignOn = ({ navigation }: SignOnProps) => {
...
-                  setShowInvalidEmailError(!isValidEmailString(email))
-                  // wait a bit for email validation to come back
-                  setTimeout(() => setAttemptedEmail(true), 1000)
-                }
-              }}
-            />
-            {passwordInputField()}
-            {errorView()}
-            <MainButton isWorking={isWorking} isSignin={isSignin} />
+            <SafeAreaView edges={['top']} style={styles.drawerContent}>
```
dylanjeffers, line 733: nice work figuring out what works here

dylanjeffers (approved):
love these improvements, was bugging me for a while

packages/mobile/android/app/src/main/AndroidManifest.xml
sliptype: Yes exactly. I check those, what do you mean by filters?

nicoback2 (approved):

PR #2175, opened 2022-10-25
[C-1349] Improve StatusBar colors

sliptype:
### Description

* Properly set StatusBar color based on theme
* Make StatusBar translucent during splash screen on android

Android splash screen
image
Android after splash screen
image

### Dragons

Could cause weird state where the StatusBar is not visible due to being white on white. But I believe I fixed all the cases where this could happen

### How Has This Been Tested?

Tested on both android and ios. Using system theme and custom theme in settings

### Feature Flags ###


nicoback2 (approved):
nice!

PR #2180, opened 2022-10-26
[C-1369] Fix account screen crash on android

sliptype:
### Description

This crash seems to only happen when certain images (maybe image size/format) are used as a cover photo. I thought this was fixed by upgrading react-native-screens but I was still seeing the crash on my account. Other people are having the same issue: https://github.com/Kureev/react-native-blur/issues/461. We aren't using the blur on android anyway so I went back to disabling it, which fixes the crash

* Also update react-native-screen patch filename so that it is applied to the new version of the packages

### Dragons

### How Has This Been Tested?

### Feature Flags ###


dylanjeffers (approved):
Dope

nicoback2 (approved):

PR #2192, opened 2022-10-27
Pass retry through for client handle validation

jowlee:
### Description
Iteration on #2183
Because `404` response codes trigger a discovery node reelection, the check for user not existing will trigger this logic and results in longer method response times. 
This change is to pass along a flag for `retry` so that checking for an existing handle will not trigger reslection and be instant!
Not sure if there's a mobile component to this that I'm missing @sliptype 

Fixes PLAT-432

### Dragons

### How Has This Been Tested?
Ran locally against stage and validated that no re-selection were happening for non-existent handles. 

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/jowlee-user-handle

sliptype (approved):
Awesome, this is all in `common` so this takes care of mobile too

PR #2193, opened 2022-10-28
[C-1376] Kickoff mobile upload

dylanjeffers:
### Description

- Adds mobile-upload feature flag
- Adds link in account-drawer
- Adds upload screen
- Adds document-picker
- Adds complete-track screen skeleton


packages/mobile/src/screens/upload-screen/ParamList.ts
```diff
@@ -0,0 +1,6 @@
+import type { DocumentPickerResponse } from 'react-native-document-picker'
+
+export type UploadParamList = {
+  Upload: undefined
```
Kyle-Shanks, line 4: should this always be undefined or an optional type?

Kyle-Shanks (approved):

packages/mobile/src/screens/upload-screen/ParamList.ts
dylanjeffers: yeah we have undefined when there isn't any params, not sure how else to do it...

PR #2194, opened 2022-10-28
Add notification navigation hook and update profile screen to accept id param

Kyle-Shanks:
### Description
Add notification navigation hook to be used by notification components and push notifications

This can be updated to have separate handlers for each notification type, but wanted to get the basics down

### Dragons

The structure of notifications coming back from identity is not the same as client side, so need to do a lot of verification to make sure that this will work for push notifications

### How Has This Been Tested?

Manually tested with a lot of notif types, but need to do testing on device to work out the kinks with push notifications

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/mobile/src/hooks/useNotificationNavigation.ts
+++ b/packages/mobile/src/hooks/useNotificationNavigation.ts
@@ -1,0 +1,191 @@
+import { useCallback, useMemo } from 'react'
+
+import type {
+  AddTrackToPlaylistNotification,
+  AnnouncementNotification,
+  ChallengeRewardNotification,
+  FavoriteNotification,
+  FollowNotification,
+  MilestoneNotification,
+  ReactionNotification,
+  RemixCosignNotification,
+  RemixCreateNotification,
+  RepostNotification,
+  SupporterDethronedNotification,
+  SupporterRankUpNotification,
+  SupportingRankUpNotification,
+  TierChangeNotification,
+  TipReceiveNotification,
+  TipSendNotification,
+  TrendingTrackNotification,
+  UserSubscriptionNotification
+} from '@audius/common'
+import {
+  notificationsUserListActions,
+  tippingActions,
+  Achievement,
+  Entity,
+  NotificationType
+} from '@audius/common'
+import type { AppState } from 'audius-client/src/store/types'
+import { useDispatch, useStore } from 'react-redux'
+
+import { useNavigation } from './useNavigation'
+
+const { setNotificationId } = notificationsUserListActions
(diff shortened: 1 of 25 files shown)


packages/mobile/src/notifications.ts
```diff
@@ -54,51 +43,37 @@ const getPlatformConfiguration = () => {
...
     console.info(`Received notification ${JSON.stringify(notification)}`)
     if (notification.userInteraction || Platform.OS === 'android') {
       track(
         make({
           eventName: EventNames.NOTIFICATIONS_OPEN_PUSH_NOTIFICATION,
           ...(notification.message
             ? {
-                title: notification.message.title,
-                body: notification.message.body
+                title: notification.message.title ?? notification.title,
```
dylanjeffers, line 69: oh interesting, is this the new format of notifs?

packages/mobile/src/notifications.ts
```diff
@@ -54,51 +43,37 @@ const getPlatformConfiguration = () => {
 class PushNotifications {
   lastId: number
   token: Token | null
-  drawerHelpers: DrawerNavigationHelpers | null
-  bottomTabNavigation: Nullable<BottomTabNavigation>
+  navigation: any | null
```
dylanjeffers, line 46: ideally we have a type here, something like `NativeNavigationStackProp<AppTabParamList>` or smthg

dylanjeffers (approved):
dopee

Kyle-Shanks:
Added commit to update all of the notif component to use the new hook. Re-requesting review bc of that

dylanjeffers (commented):
makes sense, but not the hugest fan of this pattern since all the notifs now know about eachother, what do you think?

sliptype (approved):
Wow, this is awesome! Nice job

packages/mobile/src/hooks/useNotificationNavigation.ts
```diff
@@ -0,0 +1,191 @@
...
+
+/**
+ * Navigator for notifications
+ *
+ * Uses the useNavigation hook under the hood
+ */
+export const useNotificationNavigation = () => {
+  const navigation = useNavigation()
+  const dispatch = useDispatch()
+  const store = useStore<AppState>()
```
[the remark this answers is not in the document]
Kyle-Shanks: doesn't look like it does 👍

packages/mobile/src/notifications.ts
Kyle-Shanks: yea the title and body isn't always in message so just covering that case. Can clean this up after more testing probably

PR #2199, opened 2022-10-29
Use confirmer wrapper when updating user

michellebrier:
### Description
Call `updateProfileAsync`, which wraps the call to `audiusBackendInstance.updateCreator` in confirmer logic, rather than calling `updateCreator` directly.

### How Has This Been Tested?
Open confirmer. For each of the following: verify confirmer action, verify change persistence via API
- Set artist pick
- Unset artist pick
- Delete track

N/A

### Feature Flags ###
N/A


sliptype (approved):
lgtm!

PR #2201, opened 2022-10-31
Fix upload

jowlee:
### Description
Not sure how this was ever working before, but calling close on the progress channel puts an undefined action, so we would typically use the channelCanceller to close the channel because it would take the last undefined action and then call cancel
ref: https://github.com/AudiusProject/audius-client/blob/8cc3a29a7582323dbd19c1fa9a6ce0f45daf3c1b/packages/common/src/utils/sagaHelpers.ts#L47-L53
But since the current logic called close, the action channel dispatcher is called with an action of undefined which is put and then throws an error on the webpage
ref: https://github.com/AudiusProject/audius-client/blob/8cc3a29a7582323dbd19c1fa9a6ce0f45daf3c1b/packages/common/src/utils/sagaHelpers.ts#L40-L45

### Dragons

### How Has This Been Tested?
Ran locally

### Feature Flags ###

### DIFF
--- a/packages/common/src/utils/sagaHelpers.ts
+++ b/packages/common/src/utils/sagaHelpers.ts
@@ -39,8 +39,10 @@ export function* batchYield(
  */
 export function* actionChannelDispatcher(channel: EventChannel<any>) {
   while (true) {
-    const action: Action<any> = yield take(channel)
-    yield put(action)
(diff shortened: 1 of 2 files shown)


nicoback2 (approved):

audius-infra:
Preview this change https://demo.audius.co/jowlee-fix-upload

PR #2202, opened 2022-10-31
Use feature flag for mp3 stream toggle

amendelsohn:
### Description
This reverts https://github.com/AudiusProject/audius-client/pull/2147 and uses a feature flag to switch playback from hls to mp3.

Once this has been verified in prod, we can go back and rip out the unused hls code as originally done in https://github.com/AudiusProject/audius-client/pull/2147

See the non-revert commit [9313d83](https://github.com/AudiusProject/audius-client/pull/2202/commits/9313d83f0de8352f102584ca26327c430125c00d) for effective changes.

### Dragons

Affects all of track streaming.

### How Has This Been Tested?

Local web and sim
Tested with feature flag on and off. The correct endpoint was used in both cases

Metrics for track play/listen will drop if malfunctioning

### Feature Flags ###

`STREAM_MP3`
- `true`: use mp3
- `false`: use hls

### DIFF
--- a/packages/common/src/services/audio-player.ts
+++ b/packages/common/src/services/audio-player.ts
@@ -1,3 +1,4 @@
+import { TrackSegment } from '../models'
 import { Nullable } from '../utils'
 
 export type AudioInfo = {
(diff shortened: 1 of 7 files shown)


sliptype (approved):
Good call!

audius-infra:
Preview this change https://demo.audius.co/amendel-stream-mp3-feature

raymondjacobson (approved):

PR #2205, opened 2022-11-01
Update secondary screen background color

dylanjeffers:
### Description

Then new mobile-nav-overhaul includes a change to the secondary screen background color, which was missed.


raymondjacobson (approved):
code lgtm! wondering if the feature flag was worth it all in the end

dylanjeffers:
> code lgtm! wondering if the feature flag was worth it all in the end 

It def was to actually get the code merged without blocking main and normal release cycle imo

PR #2212, opened 2022-11-02
Fix bad linking from a element

jowlee:
### Description
with new a tags for better seo, we need to `preventDefault` when they are clicked on and it was missed in a couple spots.
I added wrapper on click which prevents default in the root component which should now cover the base case too just in case

### How Has This Been Tested?
Ran locally

### Feature Flags ###


sliptype (approved):
Nice, thanks!

PR #2215, opened 2022-11-03
[C-1402] Migrate upload store to common

dylanjeffers:
### Description

- Migrates upload actions/reducer/selectors/types to common
- Refactors actions/reducers to typescript
- Updates references


sliptype (approved):
Nice! thanks for refactoring to TS

packages/web/src/components/table/Table.tsx
```diff
@@ -201,7 +201,7 @@ export const Table = ({
     }
   }, [columns, defaultSorter, isVirtualized, onSort, sortBy, useLocalSort])
 
-  useEffect(handleSortChange, [sortBy])
```
Kyle-Shanks, line 204: this may cause issues with how many times the handleSortChange function is called. There was a warning here, but the intent was so only that the handleSortChange function called then sortBy was changed, not the function itself. Would love to discuss how to better fix this

dylanjeffers: Ohhh nice yeah good call, I'll just add an ignore comment then

Kyle-Shanks (approved):
awesome stuff 🔥

PR #2218, opened 2022-11-03
[PAY-720] [PAY-721] Stripe OnRamp hotfixes

rickyrombo:
### Description

See Linear items/commits.

Noticed these issues while testing in Portugal. I think these could both be pretty critical

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/web/src/services/Location.ts
+++ b/packages/web/src/services/Location.ts
@@ -21,15 +21,23 @@ export type Location = {
   utc_offset: string
 }
 
+type CachedLocation = {
+  location: Location
+  expires: number
+}
+const CACHED_LOCATION_TTL_MS = 5000
+let cachedLocation: CachedLocation | null = null
 export const getLocation = async (): Promise<Location | null> => {
   try {
-    const cached_location = localStorage.getItem('ipapi_location')
-    if (cached_location) {
-      return JSON.parse(cached_location)
+    if (cachedLocation && cachedLocation.expires < Date.now()) {
+      return cachedLocation.location
     }
     const res = await fetch('https://ipapi.co/json/')
-    const json = res.json()
-    localStorage.setItem('ipapi_location', JSON.stringify(json))
+    const json = await res.json()
+    cachedLocation = {
+      location: json,
+      expires: Date.now() + CACHED_LOCATION_TTL_MS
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://demo.audius.co/mjp-cancel-stripe

packages/web/src/services/Location.ts
```diff
@@ -23,8 +23,14 @@ export type Location = {
 
 export const getLocation = async (): Promise<Location | null> => {
   try {
+    const cached_location = localStorage.getItem('ipapi_location')
```
raymondjacobson, line 26: Would an in memory thing here be better? I potentially worry about persisting this across sessions and forgetting about it or something.

NBD though, this seems fine as is.

raymondjacobson (approved):

packages/web/src/services/Location.ts
piazzatron: Yeah I agree @raymondjacobson, we only need this per session so let's just stick it in memory

piazzatron (approved):

packages/web/src/services/Location.ts
rickyrombo: yea had the same thought after publishing this but was passing out earlier, updated now!

packages/web/src/services/Location.ts
```diff
@@ -21,10 +21,24 @@ export type Location = {
   utc_offset: string
 }
 
+type CachedLocation = {
+  location: Location
+  expires: number
+}
+const CACHED_LOCATION_TTL_MS = 5000
+let cachedLocation: CachedLocation | null = null
```
raymondjacobson, line 29: idk if this is a me thing or not, but i like

const cachedLocation = {
  loc: null
}

cachedLocation.null = X


just because mutable things scare me to some degree b/c you might not know where it's coming from.

just personal pref tho, pls ignore me unless you decide you really agree

raymondjacobson (approved):

audius-infra:
Preview this change https://demo.audius.co/mjp-cancel-stripe

dylanjeffers:
we are hitting rate limits looks like :/

PR #2232, opened 2022-11-07
Update notif navigation hook to handle the push notification payloads from identity

Kyle-Shanks:
### Description
Updating the notif navigation hook to handle the push notif payloads from identity. There are some weird differences and quirks with the formatting compared to the types on the client side so there are ts-ignores in places for that reason.

### Dragons

This has not been thoroughly tested bc push notifications do not work consistently on the locally built apps and I could not get logs out of the TestFlight stage or prod apps. Based on the structure of the identity payloads in the code and that I have seen, these should mostly work

### How Has This Been Tested?

Some testing, but need to perform more testing when this is on Staging so see if the push notifications call this code properly

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/common/src/store/notifications/types.ts
+++ b/packages/common/src/store/notifications/types.ts
@@ -99,6 +99,24 @@ export type FollowNotification = BaseNotification & {
   userIds: ID[]
 }
 
+export type FollowPushNotification = {
+  type: NotificationType.Follow
+  actions: [
+    {
+      blocknumber: number
+      actionEntityId: ID
+      actionEntityType: Entity.User
+    }
+  ]
+  initiator: ID
+  blocknumber: ID
+  timestamp: string
+  metadata: {
+    followee_user_id: ID
+    follower_user_id: ID
+  }
+}
+
 export type RepostNotification = BaseNotification & {
   type: NotificationType.Repost
   entityId: ID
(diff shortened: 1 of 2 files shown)


dylanjeffers (changes requested):
looking great, wondering if there is a clean way to handle initiator without having to resort to @ts-ignores!

packages/common/src/store/notifications/types.ts
```diff
@@ -99,20 +99,78 @@ export type FollowNotification = BaseNotification & {
   userIds: ID[]
 }
 
+export type FollowPushNotification = {
```
dylanjeffers, line 102: can we define a "PushNotification" type that covers most of the stuff like, initiator, blocknumber, timestamp etc, and also have an Action type, and then we fill in the specifics for each one?

dylanjeffers: maybe it's fine as is to be as explicit as possible? what you think?

packages/mobile/src/hooks/useNotificationNavigation.ts
```diff
@@ -75,9 +92,16 @@ export const useNotificationNavigation = () => {
...
+        notification:
+          | AddTrackToPlaylistNotification
+          | AddTrackToPlaylistPushNotification
       ) => {
-        navigation.navigate('Collection', { id: notification.playlistId })
+        navigation.navigate('Collection', {
+          id:
+            'playlistId' in notification
+              ? notification.playlistId
+              : notification.metadata.playlistId
```
dylanjeffers, line 103: love this!

dylanjeffers (approved):
clutch!! nice work

sliptype (approved):

packages/common/src/store/notifications/types.ts
Kyle-Shanks: I wanted to, but there isn't enough consistency in the structures and I can't see them all so will leave as is for now. Good thought though, may come back to this

PR #2234, opened 2022-11-08
[C-1435] Add native TagInput

dylanjeffers:
### Description

Adds native tag-input
- Adds startAdornment to TextInput to implement TagInput
- Implements native track TagField
- Updates various component spacings to further align with spec

### Dragons
- Unable to do inline-flex in react-native, but still looks okay imo
- Single-frame jumps which will be addressed in final-final qa

https://user-images.githubusercontent.com/8230000/200453054-3a766d9a-d050-4102-8160-1b85e173b1e5.MP4


packages/mobile/src/components/core/TagInput.tsx
```diff
@@ -0,0 +1,134 @@
...
+  )
+
+  const handleKeyPress = useCallback(
+    (e: NativeSyntheticEvent<TextInputKeyPressEventData>) => {
+      const { key: keyValue } = e.nativeEvent
+
+      if (!inputValue && keyValue === 'Backspace') {
+        onChangeText?.(tags.slice(0, -1).join(','))
+      } else if (inputValue && (keyValue === ' ' || keyValue === ',')) {
+        onChangeText?.([...tags, trimToAlphaNumeric(inputValue)].join(','))
```
sliptype, line 114: Do we want to dedupe tags?

sliptype (approved):
Looks solid!

is iconSave the best name for the icon?

packages/mobile/src/components/core/TagInput.tsx
dylanjeffers: Yes great call, we also prob want to limit char count too

dylanjeffers:
> Looks solid!
> 
> 
> 
> is iconSave the best name for the icon?

Thanks! So that is the name of that icon in stems and web so for went with it for consistency

PR #2240, opened 2022-11-09
[C-1443] Fix error page analytics

sliptype:
### Description

There has been a large drop off in error page analytics since february https://analytics.amplitude.com/audius/chart/hghn2r9/edit/hzfdg28?source=copy+url

`redirectRoute` is defaulted to `ERROR_PAGE`. But error page analytics were not being recorded when `redirectRoute` exists

* Update error saga logic so analytics are recorded when redirectRoute is `ERROR_PAGE`

### Dragons

Perhaps it would be better to have an effect in the error page component itself? Because currently these analytics are triggered on every error, so potentially multiple times even though the user is only seeing the error page once

### How Has This Been Tested?

Tested by throwing errors in react hierarchy and dispatching error actions in sagas. Also realized that actual errors getting thrown in sagas do not redirect to the error page (just get a blank screen)

### Feature Flags ###

### DIFF
--- a/packages/web/src/store/errors/sagas.ts
+++ b/packages/web/src/store/errors/sagas.ts
@@ -21,17 +21,16 @@ function* handleError(action: errorActions.HandleErrorAction) {
   }
 
   if (action.shouldRedirect) {
-    if (action.redirectRoute) {
-      yield put(pushRoute(action.redirectRoute))
-    } else {
+    const redirectRoute = action.redirectRoute ?? ERROR_PAGE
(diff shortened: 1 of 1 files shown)


audius-infra:
Preview this change https://demo.audius.co/sk-c-1443-fix-error-page-analytics

dylanjeffers (approved):
ah nice catch, yeah i wonder if it makes more sense to be in error page. maybe not since someone could technically spam that page?

PR #2245, opened 2022-11-10
[PAY-695] Get signatures for nft-gated premium tracks

sddioulde:
### Description

The client sees which nfts it believes the user owns, and makes a request to DN to get premium content signatures for premium tracks gates by those nfts collections.
https://github.com/AudiusProject/audius-protocol/pull/4241 is the protocol counterpart PR (already merged)

The client store state keeps a map of tracks and their premium content signatures, if any. This will be used to stream premium tracks, and to also indicate which premium tracks the logged in user has access to, which will be relevant as we develop UI for premium content.

### Dragons

none

### How Has This Been Tested?

- Used a prod eth web3 provider url, made a request to the endpoint with hardcoded user wallet and nft contract addresses, got a response of track ids to their premium signatures.
- Tested the e2e flow with the client uploading a nft-gated premium track, and having a user that owns an nft from the collection make the request to get the signature. Used prod DN 4 with a specific commit of mine.

n/a

### Feature Flags ###

Going to add a premium_content_enabled feature flag which will hide this, although it is not necessary yet, but we could use that flag for other upcoming premium content changes.


audius-infra:
Preview this change https://demo.audius.co/sd-pc-nft

packages/web/src/common/store/premiumContent/sagas.ts
```diff
@@ -0,0 +1,266 @@
...
+        return
+
+      // Set the token ids for ERC1155 nfts as the balanceOf contract method
+      // which will be used to determine ownership requires the user's
+      // wallet address and token ids
+
+      // const { nft_collection: nftCollection } = premiumConditions
+      // todo: fix the string nft_collection to be an object
+      // temporarily parse it into object here for now
+      const nftCollection = JSON.parse(
```
sddioulde, line 92: this is a weird thing that i'll remove later. have to do this right now because the nft_collection is somehow coming back as a string. will fix

packages/web/src/common/store/premiumContent/sagas.ts
```diff
@@ -0,0 +1,266 @@
...
+        if (
+          'standard' in nftCollection &&
+          nftCollection.standard === 'ERC1155'
+        ) {
+          trackMap[trackId] = ethContractMap[nftCollection.address]
+        } else {
+          trackMap[trackId] = []
+        }
+      } else if (nftCollection.chain === Chain.Sol) {
+        // todo
```
sddioulde, line 115: this will be done later, once we implement getting signature for solana nft gated tracks

audius-infra:
Preview this change https://demo.audius.co/sd-pc-nft

packages/common/src/services/audius-api-client/AudiusAPIClient.ts
```diff
@@ -1544,6 +1557,43 @@ export class AudiusAPIClient {
...
+
+    // To avoid making a POST request and thereby introducing a new pattern in the DN,
+    // we build a param string that represents the info we need to verify nft collection ownership.
+    // The trackMap is a map of track ids -> token ids.
+    // If the nft collection is not ERC1155, then there are no token ids.
+    // We append the track ids and token ids as query params, making sure they're the same length
+    // so that DN knows which token ids belong to which track ids.
+    // Example:
+    // trackMap: { 1: [1, 2], 2: [], 3: [1]}
+    // query params: '?track_ids=1&token_ids=1-2&track_ids=2&token_ids=&track_ids=3&token_ids=1'
```
rickyrombo, line 1577: (ik this is outside the scope of this pr, but i think we might want clean this up and make it order agnostic by using JSON in the query string or our own parseable thing like `?track_token_id_pairs=1:1,2&track_token_id_pairs=2:` or something? worth thinking about)

packages/web/src/common/store/profile/sagas.js
```diff
@@ -158,6 +167,11 @@ export function* fetchSolanaCollectibles(user) {
     console.log('profile has no Solana NFTs')
   }
 
+  const currentUserId = yield select(getUserId)
+  if (currentUserId === user.user_id) {
+    yield put(solNFTsFetched())
```
rickyrombo, line 172: If we `put` this action before the cache update, won't the saga trigger and think the collectibles aren't there yet?

rickyrombo (approved):

packages/web/src/common/store/profile/sagas.js
sddioulde: good catch!!!

audius-infra:
Preview this change https://demo.audius.co/sd-pc-nft

audius-infra:
Preview this change https://demo.audius.co/sd-pc-nft

PR #2247, opened 2022-11-10
Update push notification types, update notif navigation hook, and remove userInteraction check in onNotification method

Kyle-Shanks:
### Description
Lots of small updates to get push notification navigation working. The biggest change was the removal of the userInteraction check in the onNotification function. Also added more types and a notif type enum for push notifications

### Dragons

The removal of the userInteraction check could cause some unexpected behavior so I'll keep an eye out for that

### How Has This Been Tested?

Lots of manual testing, lots more to do

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/common/src/store/notifications/types.ts
+++ b/packages/common/src/store/notifications/types.ts
@@ -54,6 +54,38 @@ export enum NotificationType {
   SupporterDethroned = 'SupporterDethroned'
 }
 
+export enum PushNotificationType {
+  Follow = 'Follow',
+  FavoriteTrack = 'FavoriteTrack',
+  FavoritePlaylist = 'FavoritePlaylist',
+  FavoriteAlbum = 'FavoriteAlbum',
+  RepostTrack = 'RepostTrack',
+  RepostPlaylist = 'RepostPlaylist',
+  RepostAlbum = 'RepostAlbum',
+  MilestoneListen = 'MilestoneListen',
+  MilestoneRepost = 'MilestoneRepost',
+  MilestoneFavorite = 'MilestoneFavorite',
+  MilestoneFollow = 'MilestoneFollow',
+  CreateTrack = 'CreateTrack',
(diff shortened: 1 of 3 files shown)


packages/common/src/store/notifications/types.ts
```diff
@@ -196,6 +234,70 @@ export type MilestoneNotification = BaseNotification &
       }
   )
 
+export type MilestoneFollowPushNotification = {
+  // TODO: Not the full structure. Need to verify the fields that come back from identity
```
sliptype, line 238: Do we still need all these todos?

packages/mobile/src/notifications.ts
```diff
@@ -62,21 +62,21 @@ class PushNotifications {
...
-                body: notification.message.body ?? notification.message
-              }
-            : {})
-        })
-      )
-
-      this.navigation?.navigate(notification.data.data)
-    }
+    // NOTE: Removed the check for userInteraction
+    // It was not being set to true when the user pressed the push notification
```
sliptype, line 66: Does the app navigate when it's open and a new push notif comes in now? Seems like the behavior is different on ios and android

sliptype (approved):
Nice!

packages/mobile/src/notifications.ts
Kyle-Shanks: Yea, I'll have to check this.

Kyle-Shanks: It was, compromise seems to be checking if the app is in the foreground or background when the notif comes in

packages/common/src/store/notifications/types.ts
Kyle-Shanks: Yea they're helpful for people walking around this code to know what is complete vs half baked

packages/mobile/src/notifications.ts
sliptype: Sounds like a plan!

PR #2249, opened 2022-11-11
[C-1430] Stop propagation on repost button click in tracks tables

Kyle-Shanks:
### Description
Stopping propagation so that clicking the repost button will not play/pause the track

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A


dylanjeffers (approved):
Not sure if it matters but I usually do stopprop before actions

Kyle-Shanks:
> Not sure if it matters but I usually do stopprop before actions

I thought that as soon as I committed the code. Why do we normally do that? This feels like a milk before the cereal thing

audius-infra:
Preview this change https://demo.audius.co/kj-stop-propagation-on-table-repost-butotn-click

PR #2253, opened 2022-11-11
[C-1419][C-1073] Offline download and playback for collections

amendelsohn:
### Description
https://www.loom.com/share/ae90e11718dc4a6e9e0346c0120cd9fe
image

#### Offline downloads for collections

1. Download toggle for collections
2. We track which collection(s) are responsible for a track's download
3. Removing download for a collection or favorites will not remove tracks that are still downloaded for another reason
4. Collection download status is persisted across app sessions, so toggles will have the correct state

### Dragons

I was careful to keep styling consistent when flag is off, but need a bit more QA there.

### How Has This Been Tested?

iPhone simulator; turn off host wi-fi to simulate offline status

### Feature Flags ###

`OFFLINE_MODE_ENABLED` - hoping to turn this on for internal users after this change ships


packages/mobile/src/services/offline-downloader/offline-downloader.ts
```diff
@@ -51,10 +70,31 @@ export const downloadTrack = async (trackId: number, collection: string) => {
...
+          downloaded_from_collection:
+            trackJson.offline?.downloaded_from_collection?.concat(
+              collection
+            ) ?? [collection]
+        }
+      }
+      await writeTrackJson(trackIdString, trackToWrite)
+      store.dispatch(loadTrack(track))
+      store.dispatch(completeDownload(trackIdString))
+      return
```
dylanjeffers, line 91: do we want a catch here similar to what we do with collection download?

packages/mobile/src/services/offline-downloader/offline-storage.ts
```diff
@@ -70,21 +71,42 @@ export const getTrackJson = async (
...
+): Promise<boolean> => {
+  const audioFile = await exists(getLocalAudioPath(trackId))
   // TODO: check for all required art
-  // const artFile = exists(path.join(getLocalTrackDir(trackId), '150x150.jpg'))
+  // const artFile = await exists(
+  //   path.join(getLocalTrackDir(trackId), '150x150.jpg')
+  // )
   const artFile = true
-  const jsonFile = exists(getLocalTrackJsonPath(trackId))
+  const jsonFile = await exists(getLocalTrackJsonPath(trackId))
```
dylanjeffers, line 95: no await here since we can put in it allSettled?

packages/mobile/src/services/offline-downloader/offline-storage.ts
```diff
@@ -70,21 +71,42 @@ export const getTrackJson = async (
...
 
   const results = await Promise.allSettled([audioFile, artFile, jsonFile])
-  const [audioExists, artExists, jsonExists] = results
+  const booleanResults = results.map(
```
dylanjeffers, line 98: maybe instead of a map -> every, we just return every and in that do the stuff we did in map?

packages/mobile/src/services/offline-downloader/offline-storage.ts
```diff
@@ -101,6 +123,75 @@ export const purgeAllDownloads = async () => {
...
+
+export const getOfflineCollections = async () => {
+  try {
+    const offlineCollectionsAsync = await AsyncStorage.getItem(
+      '@offline_collections'
+    )
+
+    if (!offlineCollectionsAsync) return []
+
+    return Object.entries(JSON.parse(offlineCollectionsAsync))
```
dylanjeffers, line 180: cooll

dylanjeffers (approved):
wow so much! this has helped me understand a lot better. as a general architecture thing, i wonder if using some kinda redux middleware for saving tracks to local-storage might be useful? then can view offlne saving as if it was just tracks being saved to state? dunno if that makes sense, also not something that needs to happen

amendelsohn:
> wow so much! this has helped me understand a lot better. as a general architecture thing, i wonder if using some kinda redux middleware for saving tracks to local-storage might be useful? then can view offlne saving as if it was just tracks being saved to state? dunno if that makes sense, also not something that needs to happen

Yeah so I've been just trying to get all the logic down in the simplest way possible, and viewing those architectural concerns as a problem to be tackled when I get to the download queueing work. I think a saga could work, or having an action that just dispatches a track to the queue to be handled by the consumer.

packages/mobile/src/components/track-list/TrackListItem.tsx
```diff
@@ -20,7 +20,7 @@ import UserBadges from 'app/components/user-badges'
 import { font, makeStyles } from 'app/styles'
 import { useThemeColors } from 'app/utils/theme'
 
-import { TrackDownloadIndicator } from '../offline-downloads/TrackDownloadIndicator'
+import { DownloadStatusIndicator } from '../offline-downloads/DownloadStatusIndicator'
```
sliptype, line 23: I would prefer an absolute import here

packages/mobile/src/services/offline-downloader/offline-downloader.ts
```diff
@@ -51,10 +72,31 @@ export const downloadTrack = async (trackId: number, collection: string) => {
...
+      const trackToWrite: UserTrackMetadata = {
+        ...trackJson,
+        offline: {
+          download_completed_time:
+            trackJson.offline?.download_completed_time ?? Date.now(),
+          last_verified_time:
+            trackJson.offline?.last_verified_time ?? Date.now(),
+          downloaded_from_collection:
+            trackJson.offline?.downloaded_from_collection?.concat(
+              collection
```
sliptype, line 86: So will the collection be "favorites" if a track is just in the user's favorites?

packages/mobile/src/services/offline-downloader/offline-downloader.ts
```diff
@@ -29,16 +29,35 @@ import {
...
-export const downloadTrack = async (trackId: number, collection: string) => {
+/** Main entrypoint - perform all steps required to complete a download for each track */
+export const downloadCollection = async (
+  collection: string,
+  trackIds: number[]
+) => {
+  store.dispatch(startDownload(collection))
+  markCollectionDownloaded(collection, true)
+
+  Promise.allSettled(
```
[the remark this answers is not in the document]
amendelsohn: No reason. For some reason I just was in a promise mindset when writing it lol. I think the await is more readable so I'll swap it out

packages/mobile/src/hooks/useLoadOfflineTracks.ts
```diff
@@ -10,24 +9,30 @@ import moment from 'moment'
...
+  const isOfflineModeEnabled = useIsOfflineModeEnabled()
+
   const dispatch = useDispatch()
 
   useAsync(async () => {
     if (!isOfflineModeEnabled) return
 
+    const offlineCollections = await getOfflineCollections()
+    offlineCollections?.forEach((collection) => {
+      dispatch(completeDownload(collection))
```
sliptype, line 33: Why do we dispatch `completeDownload` here?

packages/mobile/src/services/offline-downloader/offline-downloader.ts
amendelsohn: I think the outer try on line 74 should cover it, right?

sliptype (approved):
Looks good! Agree with dylan's comments, and have a few qs

packages/mobile/src/services/offline-downloader/offline-storage.ts
amendelsohn: that would be cleaner, but we won't know inside the map which promise result is which, so we can't log the appropriate message. I could bind the name to each promise, but I think that will result in less readable code

packages/mobile/src/components/offline-downloads/DownloadToggle.tsx
```diff
@@ -4,62 +4,142 @@ import type { Track } from '@audius/common'
...
+    transform: [{ scaleX: 0.8 }, { scaleY: 0.8 }]
+  },
+  toggleContainer: {},
+  flex1: {},
+  iconTitle: {},
+  labelText: {},
+  purple: {}
+}))
+
+const useLabeledStyles = makeStyles(({ palette }) => ({
```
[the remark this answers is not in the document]
amendelsohn: I played around with this and had issues with type errors (`makeStyles` didn't like having different shapes in the conditional styles). Happy to huddle and chat about this today and see if we can find a cleaner solution. Could also make a separate component, but might result in too much duplication.

amendelsohn (commented):
Thanks for the feedback!

packages/mobile/src/hooks/useLoadOfflineTracks.ts
amendelsohn: We're getting the stored list of collections marked for download. Calling `completeDownload` marks them as downloaded in state so we keep the toggle on across multiple sessions. We do this on completion of the track downloads for the collection, but next session we would forget.

PR #2255, opened 2022-11-14
Update push notificaiton navigation to work when the app is closed

Kyle-Shanks:
### Description
Updates to allow the app to check for and open the pressed push notification when the app is opened

### Dragons

There may be issues with updating how the ref is created in the NavigationContainer component

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/mobile/src/components/navigation-container/NavigationContainer.tsx
+++ b/packages/mobile/src/components/navigation-container/NavigationContainer.tsx
@@ -3,9 +3,9 @@ import { useRef } from 'react'
 
 import { accountSelectors } from '@audius/common'
 import {
-  useNavigationContainerRef,
   getStateFromPath,
-  NavigationContainer as RNNavigationContainer
+  NavigationContainer as RNNavigationContainer,
+  createNavigationContainerRef
 } from '@react-navigation/native'
 import { useSelector } from 'react-redux'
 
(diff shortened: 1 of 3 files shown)


packages/mobile/src/components/navigation-container/NavigationContainer.tsx
```diff
@@ -20,6 +20,9 @@ const { getAccountUser } = accountSelectors
 type NavigationContainerProps = {
   children: ReactNode
 }
+
+export const navigationRef = createNavigationContainerRef()
```
dylanjeffers, line 24: sweet so the `useRef` one was creating a new one each time?

packages/mobile/src/notifications.ts
```diff
@@ -79,6 +83,16 @@ class PushNotifications {
     }
   }
 
+  // Method used to open the push notification that the user pressed while the app was closed
+  popInitial = () => {
```
dylanjeffers, line 87: so this one pops and also opens, so maybe something like `openInitialNotification` or onInitialNotification or smthg?

packages/mobile/src/screens/app-screen/AppTabNavigationProvider.tsx
```diff
@@ -55,6 +58,16 @@ export const AppTabNavigationProvider = (
...
   }, [notifNavigation])
 
+  useInterval(
+    () => {
+      if (navigationRef.isReady()) {
+        PushNotifications.popInitial()
+        setIsNavigationReady(true)
+      }
+    },
+    isNavigationReady ? null : 1000
```
dylanjeffers, line 68: could maybe define the magic number, but nbd

dylanjeffers (approved):
this is a clutch feature

PR #2258, opened 2022-11-14
Read subscriptions from discovery behind feature flag

michellebrier:
### Description
Read subscriptions from discovery if the `READ_SUBSCRIBERS_FROM_DISCOVERY_ENABLED` flag is on.

### Dragons
N/A

### How Has This Been Tested?
- Turn flag on with breakpoint
- Go to a user you are subscribed to. Make sure discovery endpoint is hit, not the identity endpoint. Make sure the subscribe bell is toggled
- Go to a user you are not subscribed to. Make sure discovery endpoint is hit, not the identity endpoint. Make sure the subscribe bell is not toggled
- Repeat with flag off

### Feature Flags ###
`READ_SUBSCRIBERS_FROM_DISCOVERY_ENABLED`

### DIFF
--- a/packages/common/src/services/audius-backend/AudiusBackend.ts
+++ b/packages/common/src/services/audius-backend/AudiusBackend.ts
@@ -2779,26 +2779,25 @@ export const audiusBackend = ({
 
   /**
    * Returns whether the current user is subscribed to userId.
-   * @returns Boolean
    */
   async function getUserSubscribed(userId: ID) {
     await waitForLibsInit()
     const account = audiusLibs.Account.getCurrentUser()
     if (!account) return
 
-    const readSubscribersFromDiscoveryEnabled =
+    const isreadSubscribersFromDiscoveryEnabled =
       (await getFeatureEnabled(
         FeatureFlags.READ_SUBSCRIBERS_FROM_DISCOVERY_ENABLED
       )) ?? false
 
-    if (readSubscribersFromDiscoveryEnabled) {
+    if (isreadSubscribersFromDiscoveryEnabled) {
       // Read subscribers from discovery
       try {
-        const subscribers = await audiusLibs.User.getUserSubscribers(
+        const subscribers: APIUser[] = await audiusLibs.User.getUserSubscribers(
           encodeHashId(userId)
         )
-        const subscriberIds = subscribers.map((s: APIUser) =>
-          decodeHashId(s.id)
+        const subscriberIds = subscribers.map((subscriber) =>
+          decodeHashId(subscriber.id)
         )
         return subscriberIds.includes(account.user_id)
       } catch (e) {
--- a/packages/common/src/services/audius-backend/AudiusBackend.ts
+++ b/packages/common/src/services/audius-backend/AudiusBackend.ts
@@ -62,8 +62,10 @@ import {
   uuid,
   Maybe,
   encodeHashId,
+  decodeHashId,
   Timer
 } from '../../utils'
+import { APIUser } from '../audius-api-client/types'
 
 import { MonitoringCallbacks } from './types'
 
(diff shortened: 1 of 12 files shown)


packages/common/src/services/audius-backend/AudiusBackend.ts
```diff
@@ -2775,29 +2777,56 @@ export const audiusBackend = ({
...
-    try {
-      const { data, signature } = await signData()
-      return await fetch(
-        `${identityServiceUrl}/notifications/subscription?userId=${userId}`,
-        {
-          headers: {
-            [AuthHeaders.Message]: data,
-            [AuthHeaders.Signature]: signature
+
+    const readSubscribersFromDiscoveryEnabled =
```
dylanjeffers, line 2789: nit, but we try to prefix booleans with `is`

packages/common/src/services/audius-backend/AudiusBackend.ts
```diff
@@ -2775,29 +2777,56 @@ export const audiusBackend = ({
...
+        FeatureFlags.READ_SUBSCRIBERS_FROM_DISCOVERY_ENABLED
+      )) ?? false
+
+    if (readSubscribersFromDiscoveryEnabled) {
+      // Read subscribers from discovery
+      try {
+        const subscribers = await audiusLibs.User.getUserSubscribers(
+          encodeHashId(userId)
+        )
+        const subscriberIds = subscribers.map((s: APIUser) =>
```
dylanjeffers, line 2800: personal nit, i like still doing `subscriber` for the map argument, what you think?

packages/common/src/services/audius-backend/AudiusBackend.ts
```diff
@@ -2775,29 +2777,56 @@ export const audiusBackend = ({
...
+
+    const readSubscribersFromDiscoveryEnabled =
+      (await getFeatureEnabled(
+        FeatureFlags.READ_SUBSCRIBERS_FROM_DISCOVERY_ENABLED
+      )) ?? false
+
+    if (readSubscribersFromDiscoveryEnabled) {
+      // Read subscribers from discovery
+      try {
+        const subscribers = await audiusLibs.User.getUserSubscribers(
```
dylanjeffers, line 2797: is there any way `getUserSubscribers` could return `ApiUser[]` so we don't have to type subscriber in the `.map`?

packages/common/src/services/audius-backend/AudiusBackend.ts
```diff
@@ -2775,29 +2777,56 @@ export const audiusBackend = ({
     }
   }
 
+  /**
+   * Returns whether the current user is subscribed to userId.
+   * @returns Boolean
```
dylanjeffers, line 2782: with typescript, the @returns js-doc is not needed, we can either rely on the implicit return type, or you can explicitly type it with 
```
async function getUserSubscribed(userId: ID): Promise<boolean> {
```

dylanjeffers (approved):
nicee

packages/common/src/services/audius-backend/AudiusBackend.ts
michellebrier: added in 37d3ffce99d3ea6ae2137e569042e69810a15b64

packages/common/src/services/audius-backend/AudiusBackend.ts
michellebrier: sure, added in 37d3ffce99d3ea6ae2137e569042e69810a15b64

packages/common/src/services/audius-backend/AudiusBackend.ts
michellebrier: added in 37d3ffce99d3ea6ae2137e569042e69810a15b64

audius-infra:
Preview this change https://demo.audius.co/mbrier/plat-456/read-subscriptions-from-discprov

PR #2259, opened 2022-11-15
[C-1482] Fix visualizer on mp3 stream

raymondjacobson:
### Description

Also fixes some behavior around fadeIn/fadeOut that has been bugging me for a while.
I'm not entirely sure how this fixes it, but my guess is due to the `302` and us not handling the CORS properly.

### Dragons

### How Has This Been Tested?

Locally vs. staging.

I suggest that we cherry pick this onto the release branch after it goes in and re-enable mp3 stream when this is out.

Please verify my changes!

### Feature Flags ###

### DIFF
--- a/packages/web/src/pages/visualizer/VisualizerProvider.tsx
+++ b/packages/web/src/pages/visualizer/VisualizerProvider.tsx
@@ -98,7 +98,15 @@ const Visualizer = ({
 
   // Rebind audio
   useEffect(() => {
-    if (audioPlayer.audioCtx && playing) Visualizer1?.bind(audioPlayer)
+    if (playing) {
+      if (audioPlayer.audioCtx) {
+        Visualizer1?.bind(audioPlayer)
+      } else {
+        audioPlayer.audio.addEventListener('canplay', () => {
+          Visualizer1?.bind(audioPlayer)
+        })
+      }
+    }
   }, [isVisible, playing, currentQueueItem])
 
   useEffect(() => {
--- a/packages/web/src/services/audio-player/AudioPlayer.ts
+++ b/packages/web/src/services/audio-player/AudioPlayer.ts
@@ -16,8 +16,8 @@ const FADE_IN_EVENT = new Event('fade-in')
 const FADE_OUT_EVENT = new Event('fade-out')
 const VOLUME_CHANGE_BASE = 10
 const BUFFERING_DELAY_MILLISECONDS = 1000
-const FADE_IN_TIME = 320
-const FADE_OUT_TIME = 400
+const FADE_IN_TIME_MILLISECONDS = 320
+const FADE_OUT_TIME_MILLISECONDS = 400
 
 // In the case of audio errors, try to resume playback
 // by nudging the playhead this many seconds ahead.
@@ -375,7 +375,7 @@ export class AudioPlayer {
 
   pause = () => {
     if (this.gainNode) {
-      this.gainNode.gain.setValueAtTime(1, 0)
+      this.gainNode.gain.setValueAtTime(1, this.audioCtx?.currentTime ?? 0)
     }
     this.audio.addEventListener('fade-out', this._pauseInternal)
     this._fadeOut()
@@ -433,10 +433,10 @@ export class AudioPlayer {
     if (this.gainNode) {
       setTimeout(() => {
         this.audio.dispatchEvent(FADE_IN_EVENT)
-      }, FADE_IN_TIME)
+      }, FADE_IN_TIME_MILLISECONDS)
       this.gainNode.gain.exponentialRampToValueAtTime(
         1,
-        this.audioCtx!.currentTime + FADE_IN_TIME / 1000.0
+        this.audioCtx!.currentTime + FADE_IN_TIME_MILLISECONDS / 1000.0
       )
     } else {
       this.audio.dispatchEvent(FADE_IN_EVENT)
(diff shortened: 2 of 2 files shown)


audius-infra:
Preview this change https://demo.audius.co/rj-c-1482

packages/web/src/services/audio-player/AudioPlayer.ts
```diff
@@ -141,20 +143,20 @@ export class AudioPlayer {
...
-      if (!this.audioCtx && !shouldSkipAudioContext) {
+      if (!this.audioCtx) {
         // Set up WebAudio API handles
         const AudioContext = window.AudioContext || window.webkitAudioContext
         try {
           this.audioCtx = new AudioContext()
           this.gainNode = this.audioCtx.createGain()
-          this.gainNode.connect(this.audioCtx.destination)
           this.source = this.audioCtx.createMediaElementSource(this.audio)
           this.source.connect(this.gainNode)
```
dylanjeffers, line 155: So moving this after makes a difference?

packages/web/src/services/audio-player/AudioPlayer.ts
```diff
@@ -15,7 +15,9 @@ declare global {
 const FADE_IN_EVENT = new Event('fade-in')
 const FADE_OUT_EVENT = new Event('fade-out')
 const VOLUME_CHANGE_BASE = 10
-const BUFFERING_DELAY_MILLISECONDS = 500
+const BUFFERING_DELAY_MILLISECONDS = 1000
+const FADE_IN_TIME = 320
```
dylanjeffers, line 19: Nice

packages/web/src/pages/visualizer/VisualizerProvider.tsx
```diff
@@ -98,7 +98,15 @@ const Visualizer = ({
 
   // Rebind audio
   useEffect(() => {
-    if (audioPlayer.audioCtx && playing) Visualizer1?.bind(audioPlayer)
+    if (playing) {
+      if (audioPlayer.audioCtx) {
+        Visualizer1?.bind(audioPlayer)
+      } else {
+        audioPlayer.audio.addEventListener('canplay', () => {
```
dylanjeffers, line 105: So is this for mp3 part of it basically?

dylanjeffers (approved):
Noice

packages/web/src/services/audio-player/AudioPlayer.ts
```diff
@@ -370,6 +374,9 @@ export class AudioPlayer {
   }
 
   pause = () => {
+    if (this.gainNode) {
+      this.gainNode.gain.setValueAtTime(1, 0)
+    }
```
amendelsohn, line 379: Disclaimer: I still don't fully understand the purpose of this call, so feel free to ignore this suggestion...

All the [examples](https://developer.mozilla.org/en-US/docs/Web/API/BaseAudioContext/createGain#examples) I've seen are using currentTime to set the gain.

To me this makes more sense, since in the other place we call `setValueAtTime`, we're inside a condition that guarantees currentTime is 0:
`this.gainNode.gain.setValueAtTime(1, this.audioCtx?.currentTime ?? 0)`

I tried both, and it was hard to tell the difference, but I think the fade was less severe on the currentTime version. I'm still also confused about the currentTime not moving when we skip issue. Is that still happening?

packages/web/src/services/audio-player/AudioPlayer.ts
```diff
@@ -219,9 +221,11 @@ export class AudioPlayer {
       this.gainNode = null
       this.source = null
       this.audioCtx = null
-      this._initContext(/* shouldSkipAudioContext */ true)
-      this.audio.setAttribute('preload', 'none')
-      this.audio.setAttribute('src', forceStreamSrc)
+
+      this._initContext()
+      this.audio.preload = 'none'
+      this.audio.crossOrigin = 'anonymous'
```
amendelsohn, line 227: Does this fix the tracks that weren't playing?

amendelsohn (approved):
Wow! Can't believe you got it working so quickly! 👏

packages/web/src/services/audio-player/AudioPlayer.ts
raymondjacobson: yea

packages/web/src/services/audio-player/AudioPlayer.ts
raymondjacobson: It doesn't actually make a diff, I just thought it was a little more sane to follow what we saw online.

raymondjacobson: hmm really wow

packages/web/src/services/audio-player/AudioPlayer.ts
raymondjacobson: I think your change is correct

packages/web/src/pages/visualizer/VisualizerProvider.tsx
raymondjacobson: This makes sure the stuff is all connected before the visualizer tries to bind otherwise it may attach an analyser node to nothing. This is solved by a setTimeout too if we want. LMK if you have thoughts on a nicer way

amendelsohn: No, I think this makes sense. Was just curious if there could be a race condition since both sides are being done in `'canplay'` listeners.

packages/web/src/services/audio-player/AudioPlayer.ts
raymondjacobson: I'm still confused about it too. It does still happen, but I don't think it's actively breaking anything. Worth a follow up investigation at some point, probably, but I think ok to leave for now

audius-infra:
Preview this change https://demo.audius.co/rj-c-1482

PR #2275, opened 2022-11-16
[C-1495] Bump target api to 31

sliptype:
### Description

* Bump android target api to 31
* Add activity:exported to fix build errors
* This should fix our CI builds

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/android/app/src/main/AndroidManifest.xml
+++ b/packages/mobile/android/app/src/main/AndroidManifest.xml
@@ -35,7 +35,9 @@
         android:label="@string/app_name"
         android:configChanges="keyboard|keyboardHidden|orientation|screenSize|uiMode"
         android:screenOrientation="portrait"
-        android:windowSoftInputMode="adjustPan">
+        android:windowSoftInputMode="adjustPan"
+        android:exported="true"
+        >
         <intent-filter>
             <action android:name="android.intent.action.MAIN" />
             <category android:name="android.intent.category.LAUNCHER" />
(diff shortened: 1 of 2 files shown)


packages/mobile/android/app/src/main/AndroidManifest.xml
```diff
@@ -79,7 +83,7 @@
       <!-- < Only if you're using GCM or localNotificationSchedule() > -->
       <service
           android:name="com.dieam.reactnativepushnotification.modules.RNPushNotificationListenerServiceGcm"
-          android:exported="false" >
+          android:exported="true" >
           <intent-filter>
```
nicoback2, line 87: how did you know that these are all supposed to be `true`? I had a hard time parsing this explanation lol
```android:exported Whether or not the broadcast receiver can receive messages from sources outside its application — "true" if it can, and "false" if not. If "false", the only messages the broadcast receiver can receive are those sent by components of the same application or applications with the same user ID. The default value depends on whether the broadcast receiver contains intent filters. The absence of any filters means that it can be invoked only by Intent objects that specify its exact class name. This implies that the receiver is intended only for application-internal use (since others would not normally know the class name). So in this case, the default value is "false". On the other hand, the presence of at least one filter implies that the broadcast receiver is intended to receive intents broadcast by the system or other applications, so the default value is "true".
```

nicoback2 (approved):
looks good, just had a q!

packages/mobile/android/app/src/main/AndroidManifest.xml
sliptype: I understood it as "if something outside of the app is meant to trigger this activity (indicated by the presence of `intent-filter`s) the value should be true"

nicoback2: ahh tyty

PR #2280, opened 2022-11-17
[C-1515] Allow repeat pushes via useNavigation inside nested navigator

sliptype:
### Description

Pushing via `useNavigation` inside a nested navigator (such as a tab navigator inside a stack navigator) was deduping any subsequent pushes to that same screen because the `transitionEnd` handler was being attached on the inner navigator instead of the stack navigator

* Add `getNearestStackNavigator` utility
* Use utility to push and attach the `transitionEnd` handler to the nearest stack instead of the nearest navigator

### Dragons

Slight perf implications. I considered computing the nearest stack navigator outside of the `performCustomPush` function but that would mean every instance of useNavigation is doing that work. I think computing it inside `performCustomPush` is better and in most cases it will only need to traverse one level up the navigation hierarchy anyway

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/utils/navigation.ts
+++ b/packages/mobile/src/utils/navigation.ts
@@ -1,4 +1,4 @@
-import type { NavigationState } from '@react-navigation/native'
+import type { NavigationProp, NavigationState } from '@react-navigation/native'
 
 /**
  * Navigation state selector that selects the current route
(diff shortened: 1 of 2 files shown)


packages/mobile/src/utils/navigation.ts
```diff
@@ -21,3 +21,19 @@ export const getPrimaryRoute = (state: NavigationState) => {
...
+ * Given a navigator, get the nearest stack navigator in the hierarchy
+ */
+export const getNearestStackNavigator = (navigator: NavigationProp<any>) => {
+  if (navigator.getState?.()?.type === 'stack') {
+    return navigator
+  }
+  const parent = navigator.getParent()
+
+  if (!parent) {
+    return undefined
```
amendelsohn, line 35: should we return `navigator` in this case?

amendelsohn (approved):
Ah so that's what was happening. Nice!

packages/mobile/src/utils/navigation.ts
sliptype: I think it's safer to have the consumer explicitly check that there is a result, let me fix up the types so that consumers are forced to check

PR #2281, opened 2022-11-17
[C-1239] Fix bottom-tab-bar highlighting

dylanjeffers:
### Description

Fixes issue where multiple bottom-tab-bars can be highlighted at once. The issue is due to lottie not properly managing animation reset when "progress" prop also changes. 

To simplify things, we now have two lottie animations per button. Animation 1 is as it was before, which is to animate in when button becomes active. Animation two is a static/idle animation, which is visible when the button is inactive. Together, this ensures the inactive->active animation is smooth (only animation1 is visible), and active->inactive is immediate and smooth (only animation 2 is visible)


sliptype (approved):
Good thinking! Might be worth a comment in the code describing the reasoning

raymondjacobson (approved):

PR #2282, opened 2022-11-17
[PAY-731] Update streaming logic for premium content

sddioulde:
### Description

Update streaming logic for premium content.
Also move premium content store stuff to common module.

https://github.com/AudiusProject/audius-protocol/pull/4319 related

### Dragons

none

### How Has This Been Tested?

locally v remote services

n/a

### Feature Flags ###

yup. premium_content_enabled, which is False for prod. this feature flag is not made use of in this PR, but in a previous one. it will continue to be used for other premium content features that should be hidden


packages/common/src/store/premiumContent/reducer.ts
```diff
@@ -1,20 +1,20 @@
-import { ID, PremiumContentSignature } from '@audius/common'
+import { ID, PremiumContentSignature } from 'models'
```
dharit-tan, line 1: just double-checking, you're moving these over from common because we don't plan on using them on mobile?

dharit-tan (approved):

packages/common/src/store/premiumContent/reducer.ts
sddioulde: other way around. moving them over TO common because we DO plan on using them in mobile.

PR #2283, opened 2022-11-18
[C-1399] Fix chromecast

sliptype:
### Description

Chromecast was impressively broken, I'm not sure how it ever worked tbh

* Parts of the object being passed to `loadMedia` were undefined
  * `track.uri` - this might be because of the streaming change? Fixed this by using the same logic as in `Audio` to create the streaming endpoint
  * `track.largeArtwork` - straight up doesn't exist in the codebase, where did this go from? where did it go? something something cotton-eyed joe
  * `track.artist` also didn't exist on the track object. Fixed this by following the pattern in `Audio` and using the `getUser` selector to get the owner
* Scrubber position was being updated by both the GoogleCast client and the Audio player, resulting in flickering time values
* Time also wasn't resetting when the next queued track started playing. Fixed this by resetting the global currentTime when a new track starts in `GoogleCast.tsx`

May have a follow up pr to get rid of `global.progress` in favor of a context provider

### Dragons

Changes are relatively confined to casting

### How Has This Been Tested?

Tested casting on ios and android

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/audio/Audio.tsx
+++ b/packages/mobile/src/components/audio/Audio.tsx
@@ -272,7 +272,7 @@ export const Audio = () => {
         elapsedTime.current = seek
 
         // If we are casting, don't update the internal
-        // seek clock
+        // seek clock. This is already handled by the effect in GoogleCast.tsx
         if (!isCasting) {
           global.progress.currentTime = seek
         }
@@ -350,14 +350,23 @@ export const Audio = () => {
           setListenLoggedForTrack(false)
         )
       }
-      global.progress = progress
+
+      if (!isCasting) {
+        // If we aren't casting, update the progress
+        global.progress = progress
+      } else {
+        // If we are casting, only update the seekable duration
+        // The currentTime is set via the effect in GoogleCast.tsx
+        global.progress.seekableDuration = progress.seekableDuration
+      }
     },
     [
       track,
       currentUserId,
       listenLoggedForTrack,
       isOfflineModeEnabled,
-      isReachable
+      isReachable,
+      isCasting
     ]
   )
   const { value: offlineTrackUri, loading } = useOfflineTrackUri(
--- a/packages/mobile/src/components/audio/Audio.tsx
+++ b/packages/mobile/src/components/audio/Audio.tsx
@@ -355,7 +355,7 @@ export const Audio = () => {
         // If we aren't casting, update the progress
         global.progress = progress
       } else {
-        // If we are casting, only update the seekable duration
+        // If we are casting, only update the seekableDuration
         // The currentTime is set via the effect in GoogleCast.tsx
         global.progress.seekableDuration = progress.seekableDuration
       }
--- a/packages/mobile/src/components/audio/GoogleCast.tsx
+++ b/packages/mobile/src/components/audio/GoogleCast.tsx
@@ -1,6 +1,12 @@
-import { useCallback, useEffect } from 'react'
+import { useCallback, useEffect, useMemo } from 'react'
 
-import { castActions, playerSelectors } from '@audius/common'
+import {
+  castActions,
+  encodeHashId,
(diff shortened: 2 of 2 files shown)


packages/mobile/src/components/audio/GoogleCast.tsx
```diff
@@ -70,12 +104,20 @@ export const useChromecast = () => {
...
+    }
+  }, [prevTrack, track])
+
   // Load media when the cast connects
   useEffect(() => {
     if (castState === CastState.CONNECTED) {
       loadCast(track, global.progress.currentTime ?? 0)
     }
-  }, [loadCast, track, castState])
+  }, [loadCast, track, prevTrack, castState])
```
raymondjacobson, line 120: does this work if you replay the same song? maybe it should be using play counter

raymondjacobson (approved):

packages/mobile/src/components/audio/GoogleCast.tsx
sliptype: Ah, no it does not. Good idea, I think counter is the move

packages/mobile/src/components/audio/Audio.tsx
```diff
@@ -350,14 +350,23 @@ export const Audio = () => {
...
       }
-      global.progress = progress
+
+      if (!isCasting) {
+        // If we aren't casting, update the progress
+        global.progress = progress
+      } else {
+        // If we are casting, only update the seekableDuration
+        // The currentTime is set via the effect in GoogleCast.tsx
+        global.progress.seekableDuration = progress.seekableDuration
```
dylanjeffers, line 360: Do we need to also set this for when we are not casting?

packages/mobile/src/components/audio/GoogleCast.tsx
```diff
@@ -1,53 +1,89 @@
...
   const loadCast = useCallback(
-    (track, startTime) => {
-      if (client && track) {
+    async (track, startTime) => {
+      if (client && track && owner && streamingUri) {
+        const gateways = audiusBackendInstance.getCreatorNodeIPFSGateways(
+          owner.creator_node_endpoint
+        )
+
+        const imageUrl = await audiusBackendInstance.getImageUrl(
```
dylanjeffers, line 62: Can't use the useTrackArtwork hook for this?

dylanjeffers: I guess cause we need it in this callback cause it's one time shot?

packages/mobile/src/components/audio/GoogleCast.tsx
```diff
@@ -1,53 +1,89 @@
...
               type: 'musicTrack',
               images: [
                 {
-                  url: track.largeArtwork
+                  url: imageUrl
                 }
               ],
               title: track.title,
-              artist: track.artist
+              artist: owner?.name
```
dylanjeffers, line 79: Owner should be there yeah?

dylanjeffers (approved):
Sweet, and great call on counter @raymondjacobson

packages/mobile/src/components/audio/Audio.tsx
sliptype: Yes, in that case we update the entire progress object. Which includes seekableDuration

packages/mobile/src/components/audio/GoogleCast.tsx
sliptype: I did that at first but that hook downloads the image and then returns the url. Which makes sense when caching but we only need the url in this case

packages/mobile/src/components/audio/GoogleCast.tsx
sliptype: Oh yup don't need to optional chain here

PR #2286, opened 2022-11-19
[C-1326] Fix open in app drawer to open the app properly

Kyle-Shanks:
### Description
Fixed the Open in app drawer to use deep link to open the mobile app properly

### Dragons

May be some slight bugs using a different link, but appears to be working

### How Has This Been Tested?

Manually tested

N/a

### Feature Flags ###

N/A


packages/web/src/components/app-redirect-popover/components/AppRedirectPopover.tsx
```diff
@@ -119,7 +119,8 @@ const AppRedirectPopover = ({
   const onClick = () => {
     onBeforeClickApp()
     const pathname = getPathname()
-    const newHref = `https://redirect.audius.co${APP_REDIRECT}${pathname}`
+    const newHref = `audius://${pathname}`
```
raymondjacobson, line 122: nice. this redirect.audius.co thing is super broken.
does this work if the app is not installed?

packages/mobile/src/components/navigation-container/NavigationContainer.tsx
```diff
@@ -124,6 +124,9 @@ const NavigationContainer = (props: NavigationContainerProps) => {
       }
     },
     getStateFromPath: (path, options) => {
+      // Add leading slash if it is missing
+      if (path[0] !== '/') path = `/${path}`
```
dylanjeffers, line 128: Catch of the century

dylanjeffers (approved):
Lgtm, also just checking that the redirect part is needed for this pr or if it's for the wallet connect piece?

packages/web/src/components/app-redirect-popover/components/AppRedirectPopover.tsx
sliptype: No this doesn't work when the app isn't installed, I assume that's why we had the redirect thing?

dylanjeffers (approved):
Nice

PR #2287, opened 2022-11-19
[C-1069] Use persistent queue for offline downloads

amendelsohn:
### Description

Handle track downloads in a persistent queue
1. Works across sessions
2. Limits parallel downloads (better for large collections/favorites without blocking everything)
3. Change to how collection download status is tracked (status -> boolean)

### Dragons

### How Has This Been Tested?

- [x] works on iOS Sim

TODO:
- [x] Test on device
- [x] Test if downloads continue with app in background
- [ ] Test performance with thousands of tracks

EDIT: Not all of the above works, but I'm going to follow-up with another PR later to get proper reviews

### Feature Flags ###

All behind `OFFLINE_MODE_ENABLED`


packages/mobile/ios/Podfile.lock
```diff
@@ -15,9 +15,9 @@ PODS:
     - React-Core (= 0.66.5)
     - React-jsi (= 0.66.5)
     - ReactCommon/turbomodule/core (= 0.66.5)
-  - ffmpeg-kit-ios-https (5.1.LTS)
-  - ffmpeg-kit-react-native/https-lts (5.1.0):
-    - ffmpeg-kit-ios-https (= 5.1.LTS)
+  - ffmpeg-kit-ios-https (4.5.1.LTS)
```
dylanjeffers, line 18: So I think we don't want this downgrade? To fix remove your pods, checkout pod lock from main, and then run pod install again

packages/mobile/package-lock.json
```diff
@@ -17516,7 +17516,7 @@
     "node-dir": {
       "version": "0.1.17",
       "resolved": "https://registry.npmjs.org/node-dir/-/node-dir-0.1.17.tgz",
-      "integrity": "sha512-tmPX422rYgofd4epzrNoOXiE8XFZYOcCq1vD7MAXCDO+O+zndlA2ztdKKMa+EeuBG5tHETpr4ml4RGgpqDCCAg==",
+      "integrity": "sha1-X1Zl2TNRM1yqvvjxxVRRbPXx5OU=",
```
dylanjeffers, line 17519: Also looks like we got the sha change for lots of stuff. For this, remove node modules, checkout lock file from main, and install from root

packages/mobile/package.json
```diff
@@ -101,6 +101,7 @@
     "react-native-google-cast": "4.2.4",
     "react-native-haptic-feedback": "1.11.0",
     "react-native-image-picker": "4.7.3",
+    "react-native-job-queue": "^0.5.0",
```
dylanjeffers, line 104: Pin plz

packages/mobile/src/App.tsx
```diff
@@ -63,6 +66,7 @@ const App = () => {
 
   useEffectOnce(() => {
     subscribeToNetworkStatusUpdates()
+    isOfflineModeEnabled && startDownloadWorker()
```
dylanjeffers, line 69: I think we want to do if else for calling functions

packages/mobile/src/components/track-list/TrackListItem.tsx
```diff
@@ -234,7 +234,7 @@ export const TrackListItem = ({
             <UserBadges user={track.user} badgeSize={12} hideName />
           </Text>
         </View>
-        <DownloadStatusIndicator itemId={track_id} />
+        <DownloadStatusIndicator trackId={track_id} />
```
dylanjeffers, line 237: Nice change

packages/mobile/src/services/offline-downloader/offline-download-queue.ts
```diff
@@ -0,0 +1,57 @@
...
+  if (!track || !user) {
+    // TODO: try getting it from the API
+    store.dispatch(errorDownload(trackIdString))
+    return false
+  }
+
+  store.dispatch(startDownload(trackIdString))
+
+  queue.addJob(
+    TRACK_DOWNLOAD_WORKER,
```
dylanjeffers, line 37: This is sooo clutch

packages/mobile/src/services/offline-downloader/offline-downloader.ts
```diff
@@ -47,36 +42,30 @@ export const downloadCollection = async (
...
-  }
-
-  store.dispatch(startDownload(trackIdString))
+export const downloadTrack = async ({
+  track,
+  user,
+  collection
+}: TrackDownloadWorkerPayload) => {
+  const trackIdStr = track.track_id.toString()
+  console.log(
```
dylanjeffers, line 56: Just wondering if we want log?

dylanjeffers (approved):
This code is soooo clean

packages/mobile/ios/Podfile.lock
dylanjeffers: Might need to also do the thing for package lock first as well. Lmk if you pairing can help on this

amendelsohn: You said you looked into this and mine was good. Is this correct or not?

package-lock.json
```diff
@@ -2589,7 +2589,7 @@
     "concat-map": {
       "version": "0.0.1",
       "resolved": "https://registry.npmjs.org/concat-map/-/concat-map-0.0.1.tgz",
-      "integrity": "sha512-/Srv4dswyQNBfohGpz9o6Yb3Gz3SrUDqBH5rTuhGR7ahtlbYKnVxw2bCFMRljaA7EXHaXZ8wsHdodFvbkhKmqg==",
+      "integrity": "sha1-2Klr13/Wjfd5OnMDajug1UBdR3s=",
```
amendelsohn, line 2592: @dylanjeffers this also feels wrong. It should stay 512, right? I thought that was a npm versioning thing

PR #2289, opened 2022-11-20
[C-1411] Improve loading splash

raymondjacobson:
### Description

Use https://github.com/zoontek/react-native-bootsplash for loading splash screen
Change a few loading state things to make white screen go away

Some painful slow parts while libs init happens / early access mode shows / feed loads. I think there's more work that could be done here. But I think this is better as is than prod currently.

https://user-images.githubusercontent.com/2731362/203182423-84928795-c4ec-4fc4-ba87-4c1bff5e6a89.MP4

### Dragons

### How Has This Been Tested?

- Open app from offline mode (shows offline a little slower, but that's fine b/c we're optimizing for happy path)
- Sign out / sign in

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/rj-c-1411

audius-infra:
Preview this change https://demo.audius.co/rj-c-1411

packages/common/src/store/reachability/reducer.ts
```diff
@@ -6,7 +6,7 @@ import { ReachabilityState } from './types'
 type ReachabilityActions = ActionType<typeof actions>
 
 const initialState = {
-  networkReachable: false
+  networkReachable: true
```
sliptype, line 9: cc @amendelsohn You think this is safe?

packages/mobile/src/components/bottom-tab-bar/BottomTabBar.tsx
```diff
@@ -104,10 +104,11 @@ export const BottomTabBar = (props: BottomTabBarProps) => {
     <Animated.View
       style={[styles.root, interpolatePostion(translationAnim, insets.bottom)]}
     >
-      <SafeAreaView
-        style={styles.bottomBar}
-        edges={['bottom']}
-        pointerEvents='auto'
+      <View
```
sliptype, line 107: Hmmm, we might want to test this + NowPlaying bar on android with navigation bar / navigation gestures enabled

packages/mobile/src/screens/root-screen/RootScreen.tsx
```diff
@@ -53,28 +57,26 @@ export const RootScreen = ({ isReadyToSetupBackend }: RootScreenProps) => {
...
+        }
+      />
+      {foundUser ? (
         <Stack.Navigator
           screenOptions={{ gestureEnabled: false, headerShown: false }}
         >
           {updateRequired ? (
             <Stack.Screen name='UpdateStack' component={UpdateRequiredScreen} />
-          ) : !hasAccount ? (
+          ) : !foundUser ? (
```
sliptype, line 73: Can it ever get to this branch? `foundUser` is the condition for the outer ternary

packages/mobile/src/screens/splash-screen/SplashScreen.tsx
```diff
@@ -1,120 +1,96 @@
...
-  const [animationFinished, setAnimationFinished] = useState(false)
-  const accountStatus = useSelector(getAccountStatus)
-  const theme = useSelector(getTheme)
-  const systemAppearance = useSelector(getSystemAppearance)
-  const backgroundOpacityAnim = useRef(new Animated.Value(1))
-
-  useEffectOnce(() => {
-    if (Platform.OS === 'ios') {
-      // Hide the StatusBar on ios
-      updateStatusBarTheme(theme, systemAppearance)
```
sliptype, line 43: We don't have to do anything with the status bar colors anymore? Even on android 👀

If so that's pretty awesome

Can you test starting the app with different default OS themes enabled? I think the splashscreen was responsible for the initial setting of the status bar color

packages/web/src/common/store/account/sagas.js
```diff
@@ -240,12 +240,12 @@ export function* fetchLocalAccountAsync() {
   const cachedAccountUser = yield call([localStorage, 'getAudiusAccountUser'])
   const currentUserExists = yield call([localStorage, 'getCurrentUserExists'])
   if (cachedAccount && cachedAccountUser && !cachedAccountUser.is_deactivated) {
+    yield put(fetchAccountSucceeded(cachedAccount))
```
sliptype, line 243: Could this be dangerous? Something listening to account succeeded might depend on it being in the cache

sliptype (approved):
Awesome stuff! a few questions

packages/common/src/store/reachability/reducer.ts
raymondjacobson: I tested:
- going into offline mode with app open and clicking around
- going into app from already offline

Both seem to work well. Lmk if I should test more!

packages/mobile/src/components/bottom-tab-bar/BottomTabBar.tsx
raymondjacobson: Will test on android more thoroughly

packages/mobile/src/screens/splash-screen/SplashScreen.tsx
raymondjacobson: It seems to work fine! Unless I'm missing something ⁉️

packages/web/src/common/store/account/sagas.js
raymondjacobson: I think I can drop this line now actually. Testing

packages/mobile/src/screens/splash-screen/SplashScreen.tsx
sliptype: Cool, it's probably fine!

dylanjeffers (approved):
looks awesome, and great callouts @sliptype. if you dont mind updating lock @raymondjacobson that would be great

packages/mobile/ios/AudiusReactNative/BootSplash.storyboard
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
raymondjacobson: yeah. we kinda were before, but this is now generated from the tool that does the react native side too so things really match up (no stutter if you're familiar with that old UI switching from splash screen => lottie)

packages/common/src/store/reachability/reducer.ts
amendelsohn: Hmm so iirc the reason for this was to keep us from making and failing a bunch of requests that we assume would succeed.
The tricky case is:
1. start the app offline
2. wait for the initial feed lineup fetch to fail
3. go back online
4. make sure the lineup request gets sent again on reconnect

packages/mobile/src/screens/splash-screen/SplashScreen.tsx
raymondjacobson: You were right. Updated.

packages/mobile/src/components/bottom-tab-bar/BottomTabBar.tsx
raymondjacobson: needed some casing

packages/mobile/src/screens/root-screen/RootScreen.tsx
raymondjacobson: stupid! undid this change. it wasn't doing much

packages/common/src/store/reachability/reducer.ts
raymondjacobson: 1-4 seems to work fine for me with the changes here! 🤔 

https://user-images.githubusercontent.com/2731362/203259305-668066e2-ca1a-4d86-86df-b804e7e095b8.MP4

I'm going to merge this so it gets build to staging apps, but will follow on

amendelsohn: I remember it mattered whether I actually waited for the previous request to fire and fail while offline. Basically, I think when we do it quickly, either the saga or the request can still be queued up and just proceed when connected. It looks like you turned the network back on quickly after opening. We should make sure it still works after waiting ~20s or until the `feedFetchMetadatas` call fails (if we're still making it at all on startup).

amendelsohn: Pretty sure this happened because if we default to `reachable = true`, we initialize the backend and apiClient on startup before we get a chance to switch reachable to false, and also may briefly render the Feed screen content before we switch to the offline indicator. This may not be a problem anymore, but it's worth testing.

PR #2293, opened 2022-11-22
[C-1535] Add mobile-upload analytics, fix remix field

dylanjeffers:
### Description

Adds upload analytics to mimic existing desktop analytics.
Adds page-views for upload flow, to get another way to collect metrics about user's flow through upload


sliptype (approved):
Looks great!

PR #2295, opened 2022-11-22
Improve mobile image fetching

sliptype:
### Description

Improves image fetching performance on mobile by reducing:
* Request volume
* Animations
* Redux actions

#### Request Volume
Before https://github.com/AudiusProject/audius-client/pull/2075 the state of the world for fetching images was:
  1. Race requests to all nodes in replica set
  2. Cache the result
  3. Use the resulting url as the source for the Image
  
But in react-native the caching didn't do anything and the Image component had to fetch the image again anyway. https://github.com/AudiusProject/audius-client/pull/2075 fixed this in cases where the image is available on the primary (most cases) but it still falls back to the original process otherwise.

This PR implements image fetching with a combination of the  `useContentNodeImage` hook and the react-native `Image` component. `Image` is responsible for picking the correct size and `useContentNodeImage` is responsible for iterating through the replica set if the image is not found.

> One downside to this is that we are no longer displaying a smaller, cached version of an image while the larger image is loading. This can still be added via cache introspection, but I wanted to limit the scope of this pr

#### Animations
The app was idling at ~40% cpu utilization because all of the DynamicImage skeletons were not unmounted. Xcode Instruments was very helpful in debugging this

Removing the skeletons when they aren't needed results in the app idling around ~5% cpu usage

Screen Shot 2022-11-22 at 4 01 29 PM
Screen Shot 2022-11-22 at 4 18 36 PM

#### Redux actions
Because we are relying on the `Image` component to fetch the correct size, we do not need to dispatch redux actions to trigger image fetching. This prevents some more rerenders

#### DynamicImage changes
Made some improvements to DynamicImage to allow it to work with the new image process. There is now a nested `ImageLoader` component that is responsible for showing a skeleton while an image loads. `DynamicImage` is now only responsible for transitioning between 2 images with different `source`s

### Some other things I tried

Fast Image https://github.com/DylanVann/react-native-fast-image
This unfortunately doesn't support automatic size selection. But might be interesting to consider in the future

react-native-shimmer https://github.com/oblador/react-native-shimmer
This was an attempt to reduce the skeleton animations perf impact further by running the animations in native land, but the library is old and did work on `View`s

### Dragons

ALL MOBILE IMAGES AFFECTED!!!

### How Has This Been Tested?
- [x] NowPlaying drawer
- [x]  coverphoto defaults
- [x]  legacy images
- [x]  search
- [x]  smart collections
- [x]  edit profile
- [x]  suggested follows
etc

Should probably test on Android too

### Feature Flags ###


packages/mobile/src/hooks/useContentNodeImage.ts
```diff
@@ -0,0 +1,148 @@
...
+    const legacyImageSources = createImageSourcesForEndpoints({
+      endpoints,
+      sizes,
+      createUri: (endpoint) => () => `${endpoint}${cid}`
+    })
+
+    return [...newImageSources, ...legacyImageSources]
+  }, [cid, endpoints, sizes])
+
+  const handleError = useCallback(() => {
```
dylanjeffers, line 118: So is it up to consumer to call this? Guessing it's image component that tries and fails?

dylanjeffers (approved):
Incredible work

packages/mobile/src/hooks/useContentNodeImage.ts
sliptype: Exactly! But for usability I created the `TrackImage`, `UserImage`, etc components that automatically call this

PR #2310, opened 2022-11-28
[C-1548] Integrate native WalletConnect

raymondjacobson:
### Description

Vendors in wallet connect

TODO:
1. Move token dashboard sagas to common and wire up
Basically this code, but inside that saga
```
const connector = useWalletConnect()
connector.signMessage(...), etc.
```

2. Wire up phantom stuff (probably differently than is currently done b/c we never want to touch the privkey) into the same modal experience

3. Wire up solana phone as another option (but quite similar to phantom)

### Dragons

### How Has This Been Tested?

Many tests necessary

### Feature Flags ###

### DIFF
--- a/packages/mobile/package.json
+++ b/packages/mobile/package.json
@@ -178,7 +178,6 @@
     "jest": "24.9.0",
     "jetifier": "1.6.8",
     "metro-react-native-babel-preset": "0.67.0",
-    "npm-force-resolutions": "0.0.10",
     "patch-package": "6.4.7",
     "react-devtools-core": "4.14.0",
     "react-test-renderer": "16.9.0",
--- a/packages/mobile/package.json
+++ b/packages/mobile/package.json
@@ -33,6 +33,7 @@
     "typecheck:trace": "tsc --generateTrace tsc_trace",
     "enable-hermes:ios": "node enableHermesiOS.js",
     "disable-hermes:ios": "node disableHermesiOS.js",
+    "preinstall": "npm-force-resolutions",
     "postinstall": "patch-package"
   },
   "dependencies": {
@@ -65,6 +66,7 @@
     "@sayem314/react-native-keep-awake": "^1.1.0",
     "@sentry/react-native": "3.2.8",
     "@svgr/core": "5.5.0",
+    "@walletconnect/react-native-dapp": "1.8.0",
     "array.prototype.flat": "1.2.5",
     "audius-client": "1.4.3",
     "babel-plugin-transform-remove-console": "6.9.4",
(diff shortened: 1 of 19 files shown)


raymondjacobson:
Not ready to merge, but tagging you for vis @dylanjeffers , we can sync tm on this!

packages/mobile/package.json
```diff
@@ -172,6 +178,7 @@
     "jest": "24.9.0",
     "jetifier": "1.6.8",
     "metro-react-native-babel-preset": "0.67.0",
+    "npm-force-resolutions": "0.0.10",
```
dylanjeffers, line 181: cool!

packages/mobile/src/App.tsx
```diff
@@ -79,21 +81,30 @@ const App = () => {
...
-                <Airplay />
-                <RootScreen isReadyToSetupBackend={isReadyToSetupBackend} />
-                <Drawers />
-                <Modals />
-                <Audio />
-                <OAuth />
-                <NotificationReminder />
-                <OfflineDownloader />
-              </NavigationContainer>
+              <WalletConnectProvider
```
dylanjeffers, line 84: so do we need this wrapped in the root vs just around the wallet-connect screen?

packages/mobile/src/components/drawer/Drawer.tsx
```diff
@@ -217,16 +219,17 @@ export const springToValue = ({
...
+      friction = 10 + 2 * (1 - Math.min(drawerHeight / FULL_DRAWER_HEIGHT, 1))
       break
   }
   Animated.spring(animation, {
     toValue: value,
     tension,
     friction,
     useNativeDriver: true,
-    velocity
+    velocity,
```
dylanjeffers, line 231: maybe we add a "animationConfig" prop for various things, but i think this is all good :)

packages/mobile/src/screens/wallet-connect/components/LinkedWallets.tsx
```diff
@@ -0,0 +1,132 @@
...
+  )
+}
+
+export const LinkedWallets = () => {
+  const styles = useStyles()
+
+  const { connectedEthWallets, connectedSolWallets } =
+    useSelector(getAssociatedWallets)
+
+  const wallets = [
```
dylanjeffers, line 90: will memoize this in future, maybe even we do this all in the selector

dylanjeffers (approved):
awesome lift here, main callout is if we need wallet-connect-provider at root of app, or whether we can isolate to wallet-connect-modal. in the future, that could help us from having to load extra js on initial load

packages/mobile/src/App.tsx
raymondjacobson: we don't! we can wrap further down!

dylanjeffers: word word

packages/mobile/src/screens/wallet-connect/components/LinkedWallets.tsx
raymondjacobson: yeah sounds good. i originally was just iterating over and rendering, but then thought we should use a FlatList here for scrolling and it expects a single array of data

PR #2317, opened 2022-11-29
Fetch and store ablum art for oflfine downloads

amendelsohn:
### Description
- Populate and fetch `_cover_art_sizes`
- Verify cover art in `verifyTrack`
- Fetch track json from API on job execution so we don't need it in state

### Feature Flags ###

`OFFLINE_MODE_ENABLED`


packages/mobile/src/hooks/useLocalTrackImage.ts
```diff
@@ -0,0 +1,53 @@
...
+                uri: getLocalCoverArtPath(trackId, size.toString()),
+                width: parseInt(size.split('x')[0]),
+                height: parseInt(size.split('x')[1])
+              })
+            )
+        : [],
+    [trackId]
+  )
+
+  const handleError = useCallback(() => {
```
sliptype, line 34: I don't think we need this error handler or the imageSourceIndex in this case because the result of this hook is an array of sizes which the Image component can accept and automatically select the correct size

sliptype (approved):
This looks great! I have a branch with some more changes to the image performance and you actually included some of the same type changes too 🙂

packages/mobile/src/components/image/TrackImage.tsx
```diff
@@ -27,14 +28,24 @@ export const useTrackImage = (
...
-  track: Parameters<typeof useTrackImage>[0]
+  track: Track
   user?: Parameters<typeof useTrackImage>[1]
 } & DynamicImageProps
 
 export const TrackImage = (props: TrackImageProps) => {
   const { track, user, ...imageProps } = props
 
+  const { source: localSource, handleError: localHandleError } =
+    useLocalTrackImage(track?.track_id?.toString())
```
dylanjeffers, line 39: track is required yeah?

packages/mobile/src/components/image/TrackImage.tsx
```diff
@@ -27,14 +28,24 @@ export const useTrackImage = (
...
 type TrackImageProps = {
-  track: Parameters<typeof useTrackImage>[0]
+  track: Track
   user?: Parameters<typeof useTrackImage>[1]
 } & DynamicImageProps
 
 export const TrackImage = (props: TrackImageProps) => {
   const { track, user, ...imageProps } = props
 
+  const { source: localSource, handleError: localHandleError } =
```
dylanjeffers, line 38: this is cool. so im wondering where is the best place to introduce this local hook... would it make sense to build into the existing useTrackImage? that way we can kinda isolate the sources and it's usage is the same across the app, like "i dont care where i get it from, i just expect to get something when i call getImage?"

packages/mobile/src/services/offline-downloader/offline-downloader.ts
```diff
@@ -79,19 +129,20 @@ export const downloadTrack = async ({
...
     await downloadCoverArt(track)
     await tryDownloadTrackFromEachCreatorNode(track)
-    await writeUserTrackJson(track, user, collection)
+    await writeUserTrackJson(track, collection)
     const verified = await verifyTrack(trackIdStr, true)
     if (verified) {
       store.dispatch(loadTrack(track))
       store.dispatch(completeDownload(trackIdStr))
     } else {
       store.dispatch(errorDownload(trackIdStr))
```
dylanjeffers, line 138: could use `failJob` here?

dylanjeffers (approved):
sweett, lmk what you think re merging useImage vs useLocalImage

packages/mobile/src/components/image/TrackImage.tsx
amendelsohn: Yeah I was debating which way to do it. I think that would make sense. I guess there's also a question of does album art change regularly? Should we always pull from discovery if we're online, or is it okay to default to the local cache and maybe miss an update until the offline verifier comes back around to it.

dylanjeffers: Yeah my guess is like eager from local, make a fetch if online, and break otherwise?

packages/mobile/src/hooks/useLocalTrackImage.ts
amendelsohn: Ok, I tried doing this and realized I'm unable to get it to work with both the remote and local fetching. Even with a hard-coded local source it's sometimes showing the blank image... I might need to pair on this a bit with someone who knows how `DynamicImage` and `ImageLoader` work

PR #2318, opened 2022-11-29
Fix logged out offline startup

raymondjacobson:
### Description

My understanding of how/why this happened is because the nav stack used to be not rendered until account came back as SUCCESS or ERROR. Since we always render it to have a faster startup time, we need to be careful around whether the sign on stack shows.

### Dragons

### How Has This Been Tested?

I think this feels right -- I wasn't able to repro, but I also think the build version on the app store is different even than "Release mode" build from xcode.

https://user-images.githubusercontent.com/2731362/204440960-73fb98e0-dd77-4e35-a928-6c516ffa0100.MP4

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/screens/root-screen/RootScreen.tsx
+++ b/packages/mobile/src/screens/root-screen/RootScreen.tsx
@@ -57,19 +57,17 @@ export const RootScreen = ({ isReadyToSetupBackend }: RootScreenProps) => {
     () => dispatch(enterBackground())
   )
 
+  const canDismissSplashScreen =
+    accountStatus === Status.SUCCESS || accountStatus === Status.ERROR
   return (
     <>
-      <SplashScreen
-        canDismiss={
-          accountStatus === Status.SUCCESS || accountStatus === Status.ERROR
-        }
-      />
+      <SplashScreen canDismiss={canDismissSplashScreen} />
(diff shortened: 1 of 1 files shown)


raymondjacobson:
This round of testing also highlighted a second issue with the profile pic not loading in when opening offline (always in debug mode, sometimes in release mode). I think this could be related to @sliptype 's images changes.
Here's a solution that I think would work:

https://github.com/AudiusProject/audius-client/pull/2319

packages/mobile/src/screens/root-screen/RootScreen.tsx
```diff
@@ -57,19 +57,17 @@ export const RootScreen = ({ isReadyToSetupBackend }: RootScreenProps) => {
     () => dispatch(enterBackground())
   )
 
+  const canDismissSplashScreen =
```
amendelsohn, line 60: Maybe name this like `accountFetchResolved` or add a comment. It took me a bit to figure out that the status was of the fetch request not the account's status (deactivated, etc.)

amendelsohn (approved):
Awesome, thank you! I do think this should fix it. As long as the local account fetch also sets the status, which it must since it's working in your demo clip.

sliptype (approved):

PR #2319, opened 2022-11-29
Always fetch content node images from cache

raymondjacobson:
### Description

See comment and related PR  https://github.com/AudiusProject/audius-client/pull/2318

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/hooks/useContentNodeImage.ts
+++ b/packages/mobile/src/hooks/useContentNodeImage.ts
@@ -28,7 +28,11 @@ const createImageSourceWithSizes = ({
     return {
       width,
       height: width,
-      uri: createUri(size)
+      uri: createUri(size),
+      // A CID is a unique identifier of a piece of content,
+      // so we can always rely on a cached value
+      // https://reactnative.dev/docs/images#cache-control-ios-only
+      cache: 'force-cache'
     }
   })
 }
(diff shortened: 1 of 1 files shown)


packages/mobile/src/hooks/useContentNodeImage.ts
```diff
@@ -28,7 +28,11 @@ const createImageSourceWithSizes = ({
     return {
       width,
       height: width,
-      uri: createUri(size)
+      uri: createUri(size),
+      // A CID is a unique identifier of a piece of content,
+      // so we can always rely on a cached value
+      // https://reactnative.dev/docs/images#cache-control-ios-only
```
amendelsohn, line 34: ios only? 👀
is the bug also on android?

amendelsohn (approved):

packages/mobile/src/hooks/useContentNodeImage.ts
raymondjacobson: I think we need to build something better here long term (sounds like @sliptype has some progress on it too!).

I believe that react native image does pretty intelligently cache, but since this flag exists, I figure worth turning on for iOS?
https://rangle.io/blog/image-caching-in-react-native/

raymondjacobson: Confirmed that the image does load offline on android, so the caching must be present in some fashion. Can't guarantee this always, though I guess

PR #2324, opened 2022-11-29
[C-1572] Add react-native-threads, dominantColors thread

dylanjeffers:
### Description

Adds react-native-threads
Adds dominantColors thread implementation


packages/mobile/threads/dominantColors.thread.js
```diff
@@ -0,0 +1,141 @@
...
+
+// listen for messages
+self.onmessage = (imageUrl) => {
+  dominantRgb(imageUrl)
+  // console.log('received message from thread:', message)
+  // self.postMessage(JSON.stringify(defaultDominantColors))
+}
+
+// send a message, strings only
+self.postMessage('hello')
```
nicoback2, line 141: is this supposed to be here?

packages/mobile/threads/dominantColors.thread.js
```diff
@@ -0,0 +1,141 @@
...
+const defaultDominantColors = [
+  { r: 82, g: 224, b: 222 },
+  { r: 110, g: 82, b: 224 },
+  { r: 224, g: 82, b: 167 }
+]
+
+// listen for messages
+self.onmessage = (imageUrl) => {
+  dominantRgb(imageUrl)
+  // console.log('received message from thread:', message)
```
nicoback2, line 136: remove comments?

packages/mobile/threads/dominantColors.thread.js
```diff
@@ -0,0 +1,141 @@
+import { self } from 'react-native-threads'
```
nicoback2, line 1: Wonder if we should throw the code that can be shared between this and the web worker version (packages/web/src/workers/dominantColor.worker.js) into common? 

Could be a TODO for later if it's a lot of work since idt this logic will change very often. But at the very least would be good to add a comment about the existence of the web version/file so that people know to change both if needed.

nicoback2 (approved):
thank you!!!!!! this is amazing

packages/mobile/threads/dominantColors.thread.js
dylanjeffers: no, nice catchhh

packages/mobile/threads/dominantColors.thread.js
dylanjeffers: yeah i definitely hear you, the main issue however is that you cant import code from node_modules in threads it looks like... that's why i had to vendor in jimp.min.js instead of installing and importing from node_modules. what you think?

PR #2348, opened 2022-12-05
[C-1615] Fix android fullscreen modals

dylanjeffers:
### Description

Fixes issue where modals are not full screen on android, due to their positioning in navigation hierarchy. moving to the top solves the issue, at the compromise of encapsulation.

### DIFF
--- a/packages/mobile/android/app/src/main/res/values-night/colors.xml
+++ b/packages/mobile/android/app/src/main/res/values-night/colors.xml
@@ -1,0 +1,3 @@
+<resources>
+    <color name="colorPrimary">#C74BD3</color>
+</resources>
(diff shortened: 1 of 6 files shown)


packages/mobile/android/app/src/main/res/values-night/colors.xml
```diff
@@ -0,0 +1,3 @@
+<resources>
```
sliptype, line 1: Why is this file needed?

sliptype (approved):

packages/mobile/android/app/src/main/res/values-night/colors.xml
dylanjeffers: good catch, its meant for another pr, which is general android mobile-upload fixes. this is for the date-picker interestingly enough

PR #2349, opened 2022-12-05
[Share to IG] Fix crash share to IG C-1612

nicoback2:
### Description
Share to IG feature was crashing the app due to a memory issue: https://sentry.io/organizations/audius/issues/3491156239/?query=is%3Aunresolved&referrer=issue-stream
This was only happening on release builds which is why I didn't see it when developing locally.
Fixed by moving the dominant color calculation outside of a react-native-thread and into just a normal async function. Seemed to have no impact on the loading time or the UI interactions/music playback etc.

From react-native-thread repo:
```
The main tradeoff of using this library is memory usage, as creating new JS processes can have significant overhead. Be sure to benchmark your app's memory usage and other resources before using this library! Alternative solutions include using runAfterInteractions or the [Interaction Manager](https://facebook.github.io/react-native/docs/interactionmanager.html), and I recommend you investigate those thoroughly before using this library.
```
Maybe worth trying `runAfterInteractions` if we do end up seeing issues with this change.
I left the react-native-thread stuff in so we can get this feature into QA quicker and in case we want to figure out how to make it work.

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/share-drawer/useShareToStory.tsx
+++ b/packages/mobile/src/components/share-drawer/useShareToStory.tsx
@@ -4,13 +4,13 @@ import EventEmitter from 'events'
 import path from 'path'
 
 import type { Color, Nullable, ShareModalContent } from '@audius/common'
-import { modalsActions, encodeHashId, ErrorLevel, uuid } from '@audius/common'
+import { encodeHashId, ErrorLevel, modalsActions, uuid } from '@audius/common'
 import {
   activateKeepAwake,
   deactivateKeepAwake
 } from '@sayem314/react-native-keep-awake'
 import type { FFmpegSession } from 'ffmpeg-kit-react-native'
-import { FFmpegKitConfig, FFmpegKit, ReturnCode } from 'ffmpeg-kit-react-native'
+import { FFmpegKit, FFmpegKitConfig, ReturnCode } from 'ffmpeg-kit-react-native'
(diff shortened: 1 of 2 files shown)


raymondjacobson:
Defer to @dylanjeffers 's review on the actual code, but LGTM -- i'm curious if we can figure out a way to reproduce this more easily locally?

https://stackoverflow.com/a/4717157

nicoback2:
> Defer to @dylanjeffers 's review on the actual code, but LGTM -- i'm curious if we can figure out a way to reproduce this more easily locally?
> 
> https://stackoverflow.com/a/4717157

Oh interesting yeah this might be useful. I was able to repro it locally by just running the app with the release config `"react-native run-ios --device \"Poop\" --configuration Release"` (Poop is my phone lol)

sliptype (approved):

dylanjeffers (approved):
this all def makes sense to me! would be interested to check out that memory warning thing ray pointed out, but i agree that ideally we don't need threads.

dylanjeffers:
@nicoback small nit, re pr titles, are we doing [C-blah] first or [Story] with C-blah at the end?

nicoback2:
> @nicoback q re pr titles, are we doing [C-blah] first or [Story] with C-blah at the end?

oh idk im super inconsistent lol, wdyt?

PR #2351, opened 2022-12-06
[C-1595] Fix sign out

raymondjacobson:
### Description
Fix sign out / sign in behavior. I felt like a small sensible refactor was due here -- I made SignOnScreen and HomeScreen both rendered at the same time so we can control via navigation what screen you're on as a user -- I think this is better? But open to feedback.

Some other issues:
- sign out sagas connected twice :oops:
- wait for account wasn't taking into consideration IDLE state
- sign out wasn't doing a page navigation but implicily relying on account state to change, which makes it hard to track
- sign in sometimes clears state before animations finish so you can see some error messages at the tail end of signing in sometimes in dev mode

### Dragons

### How Has This Been Tested?

- sign in
- sign in & sign out & sign in
- sign in & sign out & sign in different account
- sign in & sign out & quit app & sign in
- sign in & quit app & open app

### Feature Flags ###

### DIFF
--- a/packages/common/src/services/audius-backend/AudiusBackend.ts
+++ b/packages/common/src/services/audius-backend/AudiusBackend.ts
@@ -863,18 +863,11 @@ export const audiusBackend = ({
     )
   }
 
-  async function getAccount(fromSource = false) {
+  async function getAccount() {
     await waitForLibsInit()
     try {
-      let account
-      if (fromSource) {
-        const wallet = audiusLibs.Account.getCurrentUser().wallet
-        account = await audiusLibs.discoveryProvider.getUserAccount(wallet)
-        audiusLibs.userStateManager.setCurrentUser(account)
-      } else {
-        account = audiusLibs.Account.getCurrentUser()
-        if (!account) return null
-      }
+      const account = audiusLibs.Account.getCurrentUser()
+      if (!account) return null
 
       // If reading the artist pick from discovery, set _artist_pick on
       // the user to the value from discovery (set in artist_pick_track_id
--- a/packages/mobile/src/screens/root-screen/RootScreen.tsx
+++ b/packages/mobile/src/screens/root-screen/RootScreen.tsx
@@ -59,7 +59,6 @@ export const RootScreen = ({ isReadyToSetupBackend }: RootScreenProps) => {
     () => dispatch(enterBackground())
   )
 
-  console.reportErrorsAsExceptions = false
   useEffect(() => {
     if (
       !isLoaded &&
--- a/packages/mobile/src/screens/root-screen/RootScreen.tsx
+++ b/packages/mobile/src/screens/root-screen/RootScreen.tsx
@@ -1,6 +1,7 @@
 import { useEffect, useState } from 'react'
 
 import { accountSelectors, Status } from '@audius/common'
+import { getHasAccount } from '@audius/common/dist/store/account/selectors'
 import type { NavigatorScreenParams } from '@react-navigation/native'
 import { createNativeStackNavigator } from '@react-navigation/native-stack'
 import { setupBackend } from 'audius-client/src/common/store/backend/actions'
@@ -42,6 +43,7 @@ type RootScreenProps = {
 export const RootScreen = ({ isReadyToSetupBackend }: RootScreenProps) => {
   const dispatch = useDispatch()
   const accountStatus = useSelector(getAccountStatus)
+  const hasAccount = useSelector(getHasAccount)
   const { updateRequired } = useUpdateRequired()
   const [isLoaded, setIsLoaded] = useState(false)
 
@@ -81,7 +83,7 @@ export const RootScreen = ({ isReadyToSetupBackend }: RootScreenProps) => {
         >
           {updateRequired ? (
             <Stack.Screen name='UpdateStack' component={UpdateRequiredScreen} />
-          ) : accountStatus === Status.SUCCESS ? (
+          ) : hasAccount ? (
             <Stack.Screen name='HomeStack' component={AppDrawerScreen} />
           ) : (
             <Stack.Screen name='SignOnStack' component={SignOnScreen} />
--- a/packages/mobile/src/screens/root-screen/RootScreen.tsx
+++ b/packages/mobile/src/screens/root-screen/RootScreen.tsx
@@ -1,7 +1,6 @@
 import { useEffect, useState } from 'react'
 
 import { accountSelectors, Status } from '@audius/common'
-import { getHasAccount } from '@audius/common/dist/store/account/selectors'
 import type { NavigatorScreenParams } from '@react-navigation/native'
 import { createNativeStackNavigator } from '@react-navigation/native-stack'
 import { setupBackend } from 'audius-client/src/common/store/backend/actions'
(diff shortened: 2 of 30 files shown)


sliptype (approved):
This all looks great but I'm a little worried about rendering both the signonstack and homestack at the same time. Are the lineups for the feed trying to load while on the signon screen?

raymondjacobson:
> This all looks great but I'm a little worried about rendering both the signonstack and homestack at the same time. Are the lineups for the feed trying to load while on the signon screen?

They do wait for you to have an account it seems but fair callout. I can undo that part of the change, but I am worried that the implicit navigation is a bit error prone.

raymondjacobson:
> > This all looks great but I'm a little worried about rendering both the signonstack and homestack at the same time. Are the lineups for the feed trying to load while on the signon screen?
> 
> They do wait for you to have an account it seems but fair callout. I can undo that part of the change, but I am worried that the implicit navigation is a bit error prone. 

So I think if this is a real concern we should actually switch this to render no HomeStack or SignOnStack until account is in success or error state. wdyt?

packages/mobile/src/store/sign-out/sagas.ts
```diff
@@ -6,24 +6,24 @@ import { audiusBackendInstance } from 'app/services/audius-backend-instance'
...
 
+    yield* call(disablePushNotifications)
     yield* call([localStorage, 'clearAudiusAccount'])
     yield* call([localStorage, 'clearAudiusAccountUser'])
     yield* call([audiusBackendInstance, 'signOut'])
     yield* call([localStorage, 'removeItem'], 'theme')
 
-    yield* put(setSearchHistory([]))
+    yield* put(resetAccount())
+    yield* put(clearHistory())
```
dylanjeffers, line 26: great add, thank you

raymondjacobson:
New changes, please re-review!
- After discussion, switched back to the "render either or" set up for the RootScreen (this is an improvement of what it was before whether sometimes both would be rendered)
- Cleaned up a lot of code around regaining reachability
- Added waitForRead to a few more places
- Fixed and tested C-1289, which clears the WebApp entropy on sign out

packages/common/src/services/audius-backend/AudiusBackend.ts
```diff
@@ -863,18 +863,11 @@ export const audiusBackend = ({
     )
   }
 
-  async function getAccount(fromSource = false) {
```
raymondjacobson, line 866: This was unused

audius-infra:
Preview this change https://demo.audius.co/rj-c-1595

packages/web/src/common/store/account/sagas.js
```diff
@@ -182,34 +181,30 @@ function* onSignedIn({ payload: { account, isSignUp = false } }) {
...
   const remoteConfigInstance = yield getContext('remoteConfigInstance')
-  const localStorage = yield getContext('localStorage')
   const isNativeMobile = yield getContext('isNativeMobile')
   const isElectron = yield getContext('isElectron')
   const fingerprintClient = yield getContext('fingerprintClient')
 
   yield put(accountActions.fetchAccountRequested())
 
-  if (!fromSource) {
-    yield fetchLocalAccountAsync()
```
sliptype, line 196: So we don't need to call this anymore?

packages/mobile/src/screens/root-screen/RootScreen.tsx
```diff
@@ -1,11 +1,11 @@
-import { useEffect } from 'react'
+import { useEffect, useState } from 'react'
 
 import { accountSelectors, Status } from '@audius/common'
+import { getHasAccount } from '@audius/common/dist/store/account/selectors'
```
sliptype, line 4: I think this import is wrong, we should be destructuring from accountSelectors or something like that

packages/web/src/common/store/account/sagas.js
```diff
@@ -182,34 +181,30 @@ function* onSignedIn({ payload: { account, isSignUp = false } }) {
...
-  if (!account || account.is_deactivated) {
+  if (account.is_deactivated) {
     yield put(
       fetchAccountFailed({
-        reason: account ? 'ACCOUNT_DEACTIVATED' : 'ACCOUNT_NOT_FOUND'
+        reason: 'ACCOUNT_DEACTIVATED'
       })
     )
-    // Clear local storage users if present
-    yield call([localStorage, 'clearAudiusAccount'])
```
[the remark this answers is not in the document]
raymondjacobson: I don't think so, this causes a ton of issues though

packages/web/src/common/store/account/sagas.js
raymondjacobson: Yeah discussed with @amendelsohn  a bit -- I can't find the utility of this b/c we fetchLocal account twice on startup currently

packages/mobile/src/screens/root-screen/RootScreen.tsx
raymondjacobson: My bad thx

packages/web/src/common/store/account/sagas.js
amendelsohn: It also didn't look like `fromSource` was ever true, but possible we missed something

sliptype: I couldn't find any instances of it being true either. But was worried since we are doing the local fetch in the false case. But if we are already doing fetchLocal as part of the `getAccount` logic we should be good

packages/web/src/common/store/account/sagas.js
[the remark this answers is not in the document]
raymondjacobson: I tested pretty heavily after this change, but it makes account restore fail every other session so you're forced down the path of fetching account from remote

packages/web/src/common/store/account/sagas.js
raymondjacobson: It was never called with true lol!

audius-infra:
Preview this change https://demo.audius.co/rj-c-1595

packages/mobile/src/screens/signon/SignOn.tsx
```diff
@@ -379,8 +379,12 @@ const SignOn = ({ navigation }: SignOnProps) => {
   useEffect(() => {
     if (signedIn && accountUser) {
       setIsWorking(false)
-      setEmail('')
-      setPassword('')
+
+      // Debounce resetting state
+      setTimeout(() => {
```
dylanjeffers, line 384: Do we need to cancel this in a cleanup task?

dylanjeffers (approved):
Awesome work parsing through all this complexity

packages/mobile/src/screens/signon/SignOn.tsx
raymondjacobson: i think we always want to do this, so probs not?

dylanjeffers: well it's just a matter of ensuring we dont call 'setState' on an unmounted component, minor react thing, but good practice

raymondjacobson: oh gotcha, will put up PR, yea

PR #2352, opened 2022-12-06
[C-1624] Fix eager load

raymondjacobson:
### Description

Fix eager load so that feed/trending/whatever loads quickly on your first app open
This change is big, but not that scary. It does 3 things:
- Rename `waitForAccountAndBackend` to be `waitForWrite` -- you need this to write data
- Add a new `waitForAccountAndReachability` (no backend) and calls it `waitForRead` -- you need this to read data
- Changes web3.utils.sha3 to use a (tiny 3kb gzip) js-sha3 library to not block on libs init for trending/track ranks

https://user-images.githubusercontent.com/2731362/205843733-7dbc0ac4-f8d5-4c83-b070-b37afc8f3243.MP4

### Dragons

### How Has This Been Tested?

Release build on phone, tested offline mode usual behaviors as well.

### Feature Flags ###

### DIFF
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -112,6 +112,7 @@
     "idb-keyval": "3.2.0",
     "inherits": "2.0.4",
     "jimp": "0.6.8",
+    "js-sha3": "0.8.0",
     "js-yaml": "3.13.1",
     "jsmediatags": "3.8.1",
     "lerp": "1.0.3",
--- a/packages/web/src/common/store/account/sagas.js
+++ b/packages/web/src/common/store/account/sagas.js
@@ -21,7 +21,7 @@ import { identify } from 'common/store/analytics/actions'
 import { retrieveCollections } from 'common/store/cache/collections/utils'
 import { addPlaylistsNotInLibrary } from 'common/store/playlist-library/sagas'
 import { updateProfileAsync } from 'common/store/profile/sagas'
-import { waitForBackendAndAccount } from 'utils/sagaHelpers'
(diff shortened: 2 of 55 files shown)


raymondjacobson:
The visual stutter in the images fading in in the lineup drives me crazy.... maybe something can be done there next

audius-infra:
Preview this change https://demo.audius.co/rj-c-1624

dylanjeffers:
> The visual stutter in the images fading in in the lineup drives me crazy.... maybe something can be done there next

funny you mention this, i was looking into that! i think its due to the crazy number of saga calls that get triggered as a follow on to a lineup coming back. and on top of that, tons of animations for the images to come in, which is probably overkill. maybe we only do the anims for the first x number of lineup items, maybe we make them faster, or just remove them

sliptype:
> > The visual stutter in the images fading in in the lineup drives me crazy.... maybe something can be done there next
> 
> funny you mention this, i was looking into that! i think its due to the crazy number of saga calls that get triggered as a follow on to a lineup coming back. and on top of that, tons of animations for the images to come in, which is probably overkill. maybe we only do the anims for the first x number of lineup items, maybe we make them faster, or just remove them. also we should look into all the extra calls/action that happens post lineup coming back and see if we can simplify there

Lol I was also thinking about this! So on lineups we animate both the tile content and the image itself, which could be what is causing the stuttering. Since we wait until the image is loaded to fade in the tile content, we could probably disable the image animation altogether in lineups

PR #2357, opened 2022-12-07
[C-1632] Queue tracks in favorited collections

amendelsohn:
### Description

Favorites toggle now queues download for every track in each favorited collection.
Formerly only queued tracks favorited individually.

### Dragons

This change moves the fetch for favorited albums and playlists to the top level of favorites. This could slow things down, but didn't appear to have a significant impact.

### How Has This Been Tested?

- [x] Per-collection tracking still works with combining the favorites toggle and collection toggles
- [x] Queue doesn't explode with hundreds of queued tracks

### Feature Flags ###

`OFFLINE_MODE_ENABLED`


packages/mobile/src/screens/favorites-screen/FavoritesScreen.tsx
```diff
@@ -48,10 +56,38 @@ const favoritesScreens = [
 
 export const FavoritesScreen = () => {
   usePopToTopOnDrawerOpen()
+  const dispatch = useDispatch()
   const isOfflineModeEnabled = useIsOfflineModeEnabled()
 
   const savedTracks = useProxySelector(getTracks, [])
 
+  const handleFetchSavedPlaylists = useCallback(() => {
```
sliptype, line 64: Are these `useCallback`s valuable? Seems like we could just define the function and pass it to `useEffectOnce`

packages/mobile/src/screens/favorites-screen/FavoritesScreen.tsx
```diff
@@ -48,10 +56,38 @@ const favoritesScreens = [
...
+
+  const userCollections = useSelector((state: CommonState) =>
+    getAccountCollections(state, '')
+  )
+
+  const allSavesTrackIds = useMemo(() => {
+    const allIds = (savedTracks?.entries ?? [])
+      .map((track: Track) => track.track_id)
+      .concat(
+        userCollections.flatMap((collection) =>
```
sliptype, line 84: Flatmap, let's gooo

sliptype (approved):
Awesome!

PR #2362, opened 2022-12-07
Queue optimizations; batch action for tracks

amendelsohn:
### Description

Some fixes and optimizations for the download queue.

- Batch action for setting `LOADING` state on many tracks (lower redux overhead)
- Prune failed jobs from queue on startup
- Set `ERROR` state more reliably on failed jobs


dylanjeffers (approved):
Makes senseee

PR #2363, opened 2022-12-08
[C-1563] Fetch and download all paginated track favorites

amendelsohn:
### Description

Fetch the small version of all the user's favorites so we don't have to paginate the lineup to know what to download.

Todo:
- [ ] Sort the API results by date favorited so the on-screen tracks go first (could hack it by using the saves lineup tracks)

### Dragons

Performance with large numbers of tracks

### How Has This Been Tested?

Simulator

Todo:
- [ ] Test on device

### Feature Flags ###

`OFFLINE_MODE_ENABLED`

### DIFF
--- a/packages/mobile/src/hooks/useFetchAllFavoritedTrackIds.ts
+++ b/packages/mobile/src/hooks/useFetchAllFavoritedTrackIds.ts
@@ -5,11 +5,15 @@ import { useAsync } from 'react-use'
 
 import { apiClient } from 'app/services/audius-api-client'
 
+import { useIsOfflineModeEnabled } from './useIsOfflineModeEnabled'
+
 const { getUserId } = accountSelectors
 
 export const useFetchAllFavoritedTrackIds = () => {
   const currentUserId = useSelector(getUserId)
+  const isOfflineModeEnabled = useIsOfflineModeEnabled()
   return useAsync(async () => {
+    if (!isOfflineModeEnabled) return []
     let trackIds: number[] = []
     let loadMore = true
     let offset = 0
--- a/packages/mobile/src/hooks/useFetchAllFavoritedTrackIds.ts
+++ b/packages/mobile/src/hooks/useFetchAllFavoritedTrackIds.ts
@@ -1,10 +1,5 @@
 import type { APIFavorite } from '@audius/common'
-import {
-  FavoriteType,
-  decodeHashId,
-  encodeHashId,
-  accountSelectors
-} from '@audius/common'
+import { decodeHashId, encodeHashId, accountSelectors } from '@audius/common'
 import { useSelector } from 'react-redux'
 import { useAsync } from 'react-use'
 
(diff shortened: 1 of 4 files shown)


packages/mobile/src/hooks/useFetchAllFavoritedTrackIds.ts
```diff
@@ -0,0 +1,46 @@
...
+export const useFetchAllFavoritedTrackIds = () => {
+  const currentUserId = useSelector(getUserId)
+  const isOfflineModeEnabled = useIsOfflineModeEnabled()
+  return useAsync(async () => {
+    if (!isOfflineModeEnabled) return []
+    let trackIds: number[] = []
+    let loadMore = true
+    let offset = 0
+    while (loadMore) {
+      const url = apiClient.makeUrl(
```
amendelsohn, line 21: I tried doing this using the common `AudiusAPIClient` but was having trouble. It always used the `/full/` endpoint even when I passed in the `PathType.VersionPath`. Probably missed something or built it wrong; not sure.

sliptype: Strange, I feel like it should work. Did you look at one of the other non-full endpoints as a reference? https://github.com/AudiusProject/audius-client/blob/main/packages/common/src/services/audius-api-client/AudiusAPIClient.ts#L1277

packages/mobile/src/services/offline-downloader/offline-downloader.ts
```diff
@@ -47,7 +47,7 @@ export const downloadCollection = async (
 ) => {
   store.dispatch(addCollection(collection))
   persistCollectionDownloadStatus(collection, true)
-  store.dispatch(batchStartDownload(trackIds.map(toString)))
+  store.dispatch(batchStartDownload(trackIds.map((id) => id.toString())))
```
sliptype, line 50: This is a fun one

packages/mobile/src/services/offline-downloader/offline-storage.ts
```diff
@@ -67,12 +67,14 @@ export const listTracks = async (): Promise<string[]> => {
...
   trackId: string
 ): Promise<UserTrackMetadata> => {
-  const trackJson = await readFile(getLocalTrackJsonPath(trackId))
   try {
+    const trackJson = await readFile(getLocalTrackJsonPath(trackId))
     return JSON.parse(trackJson)
   } catch (e) {
-    console.error(e)
-    return e
+    if (e instanceof SyntaxError) {
```
sliptype, line 74: Is this when the file doesn't exist?

sliptype (approved):

packages/mobile/src/hooks/useFetchAllFavoritedTrackIds.ts
amendelsohn: Yeah pretty sure I had it just like that. I'm gonna merge this and circle back to it. This works for now and there's some more critical feature work still left on my plate.

PR #2374, opened 2022-12-09
[C-1649] Reset libs on app start

sliptype:
### Description

There is some preserved state between separate runs of the app (especially on android). Sometimes top level code like configuring the store or declaring the `audiusLibs` variable is not run. This causes an issue where consumers of libs think it is already initted but it is in a broken state

Working case:
image

Broken case:
image

Manually clearing libs in an effect in App fixes the problem because this effect is run before the store configuration or any sagas. It also runs every time the app is opened

Will make another PR to address similar issue with store

### Dragons

Could be dangerous if this effect were to run after the libs initialization. But that seems to be impossible

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/App.tsx
+++ b/packages/mobile/src/App.tsx
@@ -33,6 +33,7 @@ import { NotificationReminder } from './components/notification-reminder/Notific
 import { OfflineDownloader } from './components/offline-downloads/OfflineDownloader'
 import { useEnterForeground } from './hooks/useAppState'
 import { WalletConnectProvider } from './screens/wallet-connect'
+import { setLibs } from './services/libs'
 
 Sentry.init({
   dsn: Config.SENTRY_DSN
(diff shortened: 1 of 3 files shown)


Kyle-Shanks (approved):

packages/mobile/src/App.tsx
```diff
@@ -58,6 +59,11 @@ const Modals = () => {
 }
 
 const App = () => {
+  // Reset libs so that we get a clean app start
+  useEffectOnce(() => {
```
dylanjeffers, line 63: wondering if we want to do `useLayoutEffect` here to make sure it happens before the first render? i think with [] deps on it, it should only once yeah?

dylanjeffers (changes requested):
just want to double check with my comment before going forward, thanks

packages/mobile/src/App.tsx
sliptype: I can change it yeah, don't think it makes much of a difference though because I was seeing App render multiple times before the sagas even started running

dylanjeffers: oh chill sg, no worries then

dylanjeffers (approved):

PR #2377, opened 2022-12-10
Fix offline playback

amendelsohn:
### Description
The `waitForWrite` change broke offline playback. Added a special case for offline and deferred the `waitForWrite` check if we're offline and offline mode is on.


raymondjacobson (approved):
Looks good to me! I think also potentially a `waitForRead` + some sort of `waitForWrite` on the recording of a listen or something

audius-infra:
Preview this change https://demo.audius.co/amendel-fix-offline-playback

PR #2387, opened 2022-12-12
[PAY-740] Use Identity endpoint for location

piazzatron:
### Description
- Use new Identity endpoint for getting location (that PR to go out first)
- Cache isEU in local storage for 1 wk

### Dragons

None

### How Has This Been Tested?

Against stage

### Feature Flags ###

### DIFF
--- a/packages/common/src/services/local-storage/LocalStorage.ts
+++ b/packages/common/src/services/local-storage/LocalStorage.ts
@@ -1,6 +1,7 @@
 import { CURRENT_USER_EXISTS_LOCAL_STORAGE_KEY } from '@audius/sdk/dist/core'
 
 import { User } from 'models/User'
+import { Nullable } from 'utils'
 
 // TODO: the following should come from @audius/libs/dist/core when
 // discoveryProvider/constants is migrated to typescript.
@@ -34,7 +35,7 @@ export class LocalStorage {
     return await this.localStorage.getItem(key)
   }
 
-  async getJSONValue(key: string) {
+  async getJSONValue<T>(key: string): Promise<T | null> {
     const val = await this.getValue(key)
     if (val) {
       try {
@@ -45,6 +46,23 @@ export class LocalStorage {
       }
     }
     return null
+  }
+
+  async getExpiringJSONValue<T>(key: string): Promise<T | null> {
+    const res: Nullable<{ value: T; expiry: number }> = await this.getJSONValue(
+      key
+    )
+    if (!res) {
+      return null
+    }
+
+    const isExpired = res.expiry < Date.now()
+    if (isExpired) {
+      this.localStorage.removeItem(key)
+      return null
+    }
+
+    return res.value
   }
 
   setItem = async (key: string, value: string) => {
@@ -60,6 +78,14 @@ export class LocalStorage {
     await this.setValue(key, string)
   }
 
+  setExpiringJSONValue(key: string, value: any, ttlSeconds: number) {
+    const expiring = {
+      value,
+      expiry: Date.now() + ttlSeconds * 1000
+    }
+    this.localStorage.setItem(key, JSON.stringify(expiring))
+  }
+
   async removeItem(key: string) {
     await this.localStorage.removeItem(key)
   }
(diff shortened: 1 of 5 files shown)


packages/web/src/utils/gdpr.ts
```diff
@@ -1,14 +1,47 @@
 import { getLocation } from 'services/Location'
 
 const DISMISSED_COOKIE_BANNER_KEY = 'dismissCookieBanner'
+const IS_EU_KEY = 'isEU'
+const IS_EU_CACHE_TTL_MS = 7 * 24 * 3600 * 1000
```
rickyrombo, line 5: 7 days feels long, no?

rickyrombo (approved):

packages/web/src/utils/gdpr.ts
piazzatron: I think it's fine, how often are ppl flying in and out of EU

audius-infra:
Preview this change https://demo.audius.co/piazz-location-endpoint

packages/common/src/services/local-storage/LocalStorage.ts
```diff
@@ -34,7 +35,7 @@ export class LocalStorage {
     return await this.localStorage.getItem(key)
   }
 
-  async getJSONValue(key: string) {
+  async getJSONValue<T>(key: string): Promise<T | null> {
```
rickyrombo, line 38: I like this, but also I've been thinking more about how this isn't "correct" from a static typing perspective. The true value is "unknown" as anything could be in there. (Similar to API requests). Part of me feels like using `as` here to coerce types makes it explicit that it's an unsafe thing, similar to a cast. The other part of me loves the handy type parameter. No need to change, just thought I'd share my musings.

packages/common/src/services/local-storage/LocalStorage.ts
```diff
@@ -47,6 +48,23 @@ export class LocalStorage {
     return null
   }
 
+  async getExpiringJSONValue<T>(key: string): Promise<T | null> {
+    const res: Nullable<{ value: T; expiry: number }> = await this.getJSONValue(
```
rickyrombo, line 52: nit: I personally prefer the type parameter in the method call vs the variable type:
`const res = await this.getJSONValue<{value: T; expiry: number} | null>(key)`

packages/web/src/store/application/ui/buy-audio/sagas.ts
```diff
@@ -1216,8 +1206,8 @@ function* watchRecovery() {
  */
 function* recoverOnPageLoad() {
   const localStorage = yield* getContext('localStorage')
-  const savedLocalStorageState: BuyAudioLocalStorageState | null = yield* call(
-    [localStorage, localStorage.getJSONValue],
+  const savedLocalStorageState = yield* call(
+    (val) => localStorage.getJSONValue<BuyAudioLocalStorageState>(val),
```
rickyrombo, line 1210: why this rewrite?

packages/web/src/store/application/ui/buy-audio/sagas.ts
```diff
@@ -137,6 +137,17 @@ const defaultBuyAudioLocalStorageState: BuyAudioLocalStorageState = {
   provider: OnRampProvider.UNKNOWN
 }
 
+function* getLocalStorageStateWithFallback() {
```
rickyrombo, line 140: <3 thanks for this

rickyrombo (approved):

packages/web/src/store/application/ui/buy-audio/sagas.ts
piazzatron: There's some issue with babel where you can't have `,getJSONValue<BuyAudioLocalStorage>]` without it freaking out, even tho it's totally valid TS syntax. Had to actually call the function instead of using the nice context form of `call`

packages/common/src/services/local-storage/LocalStorage.ts
piazzatron: fair! I'm more of a fan of the former, it usually feels odd to me to pass a type itself as a param

packages/common/src/services/local-storage/LocalStorage.ts
piazzatron: Very good point! The issue I have with `as` is that it's a bit of a broken window thing - imo ideally we'd have a codebase free of `as`, since it's usually a sign you're doing something wrong/taking a shortcut & having `as` as happy path feels a bit dangerous. But point taken that implicitly you're performing a dangerous coercion here, and maybe it's better to make it explicit. If it was Rust, we'd get around this with`dangerouslyGetJSONValue`. Runtime types wen

rickyrombo: Haha yeah, I mean on one hand that's the idea - forcing the use of `as` makes it scarier and hopefully makes the coder double check, but on the other hand totally agree that could lead to desensitizing to `as` anywhere

packages/web/src/store/application/ui/buy-audio/sagas.ts
rickyrombo: :( sad

PR #2389, opened 2022-12-12
[C-1665] Fix deleteWallet bug

dylanjeffers:
### Description

Fixes interesting edge-case where accountUser's associated_wallets were not updating correctly when deleting a wallet. This resulted in bad user experience where deleting an eth-wallet, then deleting a sol-wallet resulted in the eth-wallet being re-added to associated-wallets.

The crux of the issue had to do with two things:
1. Our deleteWallet success call did not tell redux to update accountUser's associatedWallets
2. Our update-cache merger did not respect deletions from associated-wallet object, meaning that the removed value would remain.

### DIFF
--- a/packages/common/src/store/cache/reducer.ts
+++ b/packages/common/src/store/cache/reducer.ts
@@ -87,6 +87,14 @@ export const mergeCustomizer = (
     return objValue
   }
 
+  if (key === 'associated_wallets') {
+    return srcValue
+  }
+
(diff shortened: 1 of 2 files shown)


sliptype (approved):
Nice find!

audius-infra:
Preview this change https://demo.audius.co/dj-c-1665-fix-delete-wallet-bug

PR #2393, opened 2022-12-13
Bump @audius/sdk version to 1.0.26

dharit-tan:
### Description
Bump `@audius/sdk` version

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -28,7 +28,7 @@
     "url": "https://github.com/AudiusProject/audius-client/issues"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://demo.audius.co/rt-sdk

sliptype (approved):
There is an npm script in the root that will automatically do this for you btw!

dharit-tan:
> There is an npm script in the root that will automatically do this for you btw!

Yeah I saw! I tried it but it also updated a different package and changed the shas, so I decided to just bump manually.

PR #2376, opened 2022-12-10
[C-1654] Only show individual tracks on offline tracks tab

amendelsohn:
### Description

Tracks that were downloaded as part of a favorited collection via the favorites toggle were showing up in the tracks tab lineup. The tracks tab should only show individual tracks.

### Feature Flags ###

`OFFLINE_MODE_ENABLED`

PR #2394, opened 2022-12-13
[C-1663] Update DownloadStatusIndicator to handle collectionIds

Kyle-Shanks:
### Description
* Update DownloadStatusIndicator to handle collections
*  Add download indicator to lineup tiles and collection cards
* Moved the call to fetch download track status to OfflineDownloader.tsx from the saved page so that download status could show up on the feed and trending screens, etc.

Didn't update the styles of the Card component bc that seemed like a large change across all cards that should be in its own PR

### Dragons

This relies on the collection being in state when the indicator tries to fetch the track for the collection and determine if it is in the process of being downloaded. Chatted with @amendelsohn and this is planned to be added as part of this project

### How Has This Been Tested?

Manually tested

IMG_1342
IMG_1343
IMG_1344

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/mobile/src/components/card/Card.tsx
+++ b/packages/mobile/src/components/card/Card.tsx
@@ -6,7 +6,9 @@ import { Text, View } from 'react-native'
 
 import { Tile } from 'app/components/core'
 import UserBadges from 'app/components/user-badges/UserBadges'
-import { makeStyles } from 'app/styles'
+import { flexRowCentered, makeStyles } from 'app/styles'
+
+import { DownloadStatusIndicator } from '../offline-downloads'
 
 export type CardType = 'user' | 'collection'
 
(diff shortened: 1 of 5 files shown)


packages/mobile/src/components/offline-downloads/OfflineDownloader.tsx
```diff
@@ -1,11 +1,14 @@
...
 
 import { useIsOfflineModeEnabled } from 'app/hooks/useIsOfflineModeEnabled'
+import { useLoadOfflineTracks } from 'app/hooks/useLoadOfflineTracks'
+import { DOWNLOAD_REASON_FAVORITES } from 'app/services/offline-downloader'
 import { startDownloadWorker } from 'app/services/offline-downloader/offline-download-queue'
 
 export const OfflineDownloader = () => {
   const isOfflineModeEnabled = useIsOfflineModeEnabled()
   const [initialized, setInitialized] = useState(false)
+  useLoadOfflineTracks(DOWNLOAD_REASON_FAVORITES)
```
Kyle-Shanks, line 11: Question here: useLoadOfflineTracks requires a collection, but not sure if this should be necessary. Can I update the hook to not require this prop/should a different collection be passed here?

amendelsohn: Great point. https://github.com/AudiusProject/audius-client/pull/2376 is merged and has the fix, so you should be able to merge that in and remove the argument now.

Kyle-Shanks: fixed, thanks 👍

amendelsohn (approved):
Thank you!!

PR #2398, opened 2022-12-13
[PAY-747] Handle solana-nft-gated premium content

sddioulde:
### Description

- Add logic to pass in track ids for solana-nft-gated tracks to get their premium content signatures.
- Add function `isCollectionNFT` to `SolanaClient` to check whether a mint address represents an NFT that denotes a collection. This will later be used by the premium content UI to ensure solana nfts which gate track access are valid.
- Update Collectible and sol nft collection types.
- Simplify logic to get sol nft chain metadata.

### Dragons

none

### How Has This Been Tested?

Tested by hardcoding solana collectibles from another profile and premium tracks gated on some solana nft collection and seeing that the client would build and make the correct request to DN to fetch the nft-gated track signatures.

n/a

### Feature Flags ###

existing `premium_content_enabled` feature flag

### DIFF
--- a/packages/common/src/models/Collectible.ts
+++ b/packages/common/src/models/Collectible.ts
@@ -1,6 +1,7 @@
 import { Nullable } from '../utils/typeUtils'
 
 import { Chain } from './Chain'
+import { Metadata } from "@metaplex-foundation/mpl-token-metadata"
 
 export type CollectiblesMetadata = {
   [key: string]: object
@@ -40,5 +41,5 @@ export type Collectible = {
   assetContractAddress: Nullable<string>
 
   // solana nfts
-  chainMetadata: Nullable<string>
+  chainMetadata: Nullable<Metadata>
 }
--- a/packages/common/src/models/Collectible.ts
+++ b/packages/common/src/models/Collectible.ts
@@ -42,5 +42,5 @@ export type Collectible = {
   assetContractAddress: Nullable<string>
 
   // solana nfts
-  chainMetadata: Nullable<Metadata>
+  solanaChainMetadata: Nullable<Metadata>
 }
--- a/packages/common/src/models/Collectible.ts
+++ b/packages/common/src/models/Collectible.ts
@@ -32,8 +32,13 @@ export type Collectible = {
   dateLastTransferred: Nullable<string>
   externalLink: Nullable<string>
   permaLink: Nullable<string>
-  assetContractAddress: Nullable<string>
   chain: Chain
   wallet: string
   duration?: number
+
+  // ethereum nfts
+  assetContractAddress: Nullable<string>
+
+  // solana nfts
+  chainMetadata: Nullable<string>
 }
--- a/packages/common/src/models/Collectible.ts
+++ b/packages/common/src/models/Collectible.ts
@@ -1,7 +1,8 @@
+import { Metadata } from '@metaplex-foundation/mpl-token-metadata'
+
 import { Nullable } from '../utils/typeUtils'
 
 import { Chain } from './Chain'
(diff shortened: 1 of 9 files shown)


audius-infra:
Preview this change https://demo.audius.co/sd-pc-sol-nft

audius-infra:
Preview this change https://demo.audius.co/sd-pc-sol-nft

audius-infra:
Preview this change https://demo.audius.co/sd-pc-sol-nft

packages/common/src/models/Collectible.ts
```diff
@@ -32,8 +34,13 @@ export type Collectible = {
...
-  assetContractAddress: Nullable<string>
   chain: Chain
   wallet: string
   duration?: number
+
+  // ethereum nfts
+  assetContractAddress: Nullable<string>
+
+  // solana nfts
+  chainMetadata: Nullable<Metadata>
```
piazzatron, line 45: could we name these something more self documenting

packages/common/src/services/solana-client/SolanaClient.ts
```diff
@@ -145,123 +157,31 @@ export class SolanaClient {
     }
   }
 
-  /**
```
piazzatron, line 148: so much deleted code, my heart sings!

packages/common/src/services/solana-client/solCollectibleHelpers.ts
```diff
@@ -272,8 +275,9 @@ const metaplexNFTToCollectible = async (
     description: nft.description,
     externalLink: nft.external_url,
     isOwned: true,
-    chain: Chain.Sol
-  } as Collectible
+    chain: Chain.Sol,
+    chainMetadata
+  } as unknown as Collectible
```
piazzatron, line 280: what's w the cast to `Unknown`?

packages/common/src/services/solana-client/solCollectibleHelpers.ts
```diff
@@ -316,8 +321,9 @@ const starAtlasNFTToCollectible = async (
     name: nft.name,
     description: nft.description,
     isOwned: true,
-    chain: Chain.Sol
-  } as Collectible
+    chain: Chain.Sol,
+    chainMetadata
+  } as unknown as Collectible
```
piazzatron, line 326: Same here

piazzatron (approved):
Another pair of eyes would be good but LGTM!

packages/common/src/models/Collectible.ts
sddioulde: hmmm, kind of hard to come up with a name here. was going to just call it `metadata`, as you can even see that its actual type from the library, but people may confuse it to the nft metadata obtained from the nft uri.

it basically represents this picture below, which is data from the metadata account, i added the `chain` prefix to show it's metadata represented on-chain.

Screen Shot 2022-12-19 at 9 44 11 AM

any better names come to mind?

audius-infra:
Preview this change https://demo.audius.co/sd-pc-sol-nft

packages/common/src/models/Collectible.ts
piazzatron: makes sense, how about `solanaChainMetadata` just so it's clear which chain this is from & can lose the comment?

piazzatron: If we really wanted this to be correct, it could be a discriminated union on `Chain` where the solana type only exists if it's a solana chain and eth only if it's an eth chain :)

audius-infra:
Preview this change https://demo.audius.co/sd-pc-sol-nft

PR #2400, opened 2022-12-14
[C-1671] Finalize wallet-connect flow

dylanjeffers:
### Description

Finalizes wallet connect flow ui, logic, edge-cases

Android still needs a once-over


audius-infra:
Preview this change https://demo.audius.co/dj-c-1671-finalize-ios-wallet-connect

packages/mobile/src/screens/wallet-connect/components/LinkedWallet.tsx
```diff
@@ -0,0 +1,167 @@
...
+    borderRadius: 14,
+    padding: 2,
+    marginRight: spacing(2)
+  },
+  address: {
+    flexDirection: 'row',
+    alignItems: 'center'
+  },
+  copyIcon: {
+    marginBottom: 2,
```
Kyle-Shanks, line 70: spacing

packages/mobile/src/screens/wallet-connect/types.ts
```diff
@@ -14,7 +19,11 @@ export type SignedMessageParams = {
 }
 
 export type WalletConnectParamList = {
-  Wallets: undefined | ConnectNewWalletParams | SignedMessageParams
+  Wallets:
```
Kyle-Shanks, line 22: is there a reason for undefined vs `?:` here?

dylanjeffers: yeah good q, i tihnk the undefined can be removed actually, i think the idea here is that params can be undefined, but they are always present? though in reality that means the same thing! will look into

Kyle-Shanks (approved):
Looks good. awesome stuff

PR #2403, opened 2022-12-14
Fix other share drawers

nicoback2:
### Description
Share drawers other than share track were erroring out due to:
` TypeError: undefined is not an object (evaluating 'trackImage.source[2].uri')
`

Turns out the `trackImage` not null check before didn't work because `useTrackImage` doesn't actually return null if its input is null...

### Dragons

### How Has This Been Tested?

### Feature Flags ###


dylanjeffers (approved):
so fast! nice

PR #2420, opened 2022-12-18
[C-1087] Check for updates to offline items on startup

amendelsohn:
### Description

Sync process to:
1. Sync list of favorited tracks and download/remove tracks to match the updated favorites
2. Sync tracks on downloaded collections, and add/remove to match the updated track list
3. Check each track that hasn't been checked in the last week, removing if it's deleted, or updating metadata/art if changed

Future work:
1. Re-download the mp3 if it changes for a track
2. Check for newly favorited/unfavorited collections and fetch/remove their metadata/tracks. We may want to be careful with this as actions taken on other platforms can silently remove tracks from device.

### Dragons

Worst failure modes:
1. Failing to remove a track that's been taken down
2. Deleting users tracks that they wanted to keep

### How Has This Been Tested?

Simulator

### Feature Flags ###

`OFFLINE_MODE_ENABLED`


packages/mobile/src/hooks/useFetchAllFavoritedTrackIds.ts
```diff
@@ -9,38 +9,42 @@ import { useIsOfflineModeEnabled } from './useIsOfflineModeEnabled'
...
+    )
+    const { data: result } = await fetch(url).then((response) =>
+      response.json()
+    )
+
+    loadMore = result.length > 0
+    offset += result.length
+    trackIds = trackIds.concat(
+      result
+        .filter((trackSave) => trackSave.favorite_type === 'SaveType.track')
```
dylanjeffers, line 34: does this handle the tracks that have been deleted? or guessing that doesn't come back from the api?

packages/mobile/src/hooks/useLoadOfflineTracks.ts
```diff
@@ -99,6 +103,7 @@ export const useLoadOfflineTracks = () => {
     dispatch(cacheActions.add(Kind.TRACKS, cacheTracks, false, true))
     dispatch(cacheActions.add(Kind.USERS, cacheUsers, false, true))
     dispatch(loadTracks(lineupTracks))
+    dispatch(doneLoadingFromDisk())
```
dylanjeffers, line 106: this is an interesting new case

packages/mobile/src/services/offline-downloader/offline-sync.ts
```diff
@@ -0,0 +1,253 @@
...
+  const state = store.getState()
+  const cacheTracks = getTracks(state, {})
+  const currentUserId = getUserId(state as unknown as CommonState)
+  if (!currentUserId) return
+
+  const staleCachedTracks = Object.entries(cacheTracks)
+    .filter(
+      ([id, track]) =>
+        track.offline &&
+        moment()
```
dylanjeffers, line 212: cool! not that we are making this push at all rn, but im hoping we get off moment at some point. do you have any alternative date libraries you've used in past? ive personally used dayjs cause it's so small and pretty effective

packages/mobile/src/services/offline-downloader/offline-sync.ts
```diff
@@ -0,0 +1,253 @@
...
+
+  staleCachedTracks.forEach(async (staleTrack) => {
+    const updatedTrack: UserTrackMetadata = await apiClient.getTrack({
+      id: staleTrack.track_id,
+      currentUserId
+    })
+
+    // If track should not be available
+    if (
+      !updatedTrack.is_available ||
```
dylanjeffers, line 226: nice this is a great block, i wonder if we use this in other spots around the app? might be a nice helper just cause i bet it's hard as a new dev to know what qualifies as "not playable'

packages/mobile/src/utils/populateCoverArtSizes.ts
```diff
@@ -0,0 +1,41 @@
...
+  CoverArtSizes,
+  Track,
+  UserMetadata
+} from '@audius/common'
+import { DefaultSizes, SquareSizes } from '@audius/common'
+
+import { audiusBackendInstance } from 'app/services/audius-backend-instance'
+
+export const populateCoverArtSizes = async <
+  T extends Pick<Track | Collection, 'cover_art_sizes' | 'cover_art'> & {
```
dylanjeffers, line 12: niceee, i might make that a helper type just to keep the inline def from getting too wild, nit tho.

dylanjeffers (approved):
wow a lot of complex and well written code, nice work!

packages/mobile/src/services/offline-downloader/offline-sync.ts
amendelsohn: oh, nice. No, I'm just familiar with moment and js Date

packages/mobile/src/services/offline-downloader/offline-sync.ts
amendelsohn: Totally. I'm not even 100% that I got them all or that all of these should be blocked. I can move it to a utility

packages/mobile/src/utils/populateCoverArtSizes.ts
amendelsohn: Yeah also I wasn't sure if I should just write the type explicitly instead of using pick. Either works fine, but not sure if one is more maintainable.

PR #2422, opened 2022-12-19
[PAY-757] Prevent use of premium tracks as remix parents

sddioulde:
### Description

Sets premium track as invalid when trying to upload track as a remix of a premium track.

related to https://github.com/AudiusProject/audius-protocol/pull/4499

Also, when building premium content UI, we need to ensure that the "This is a remix switch" is non-existing when uploading a premium track, so premium track will not be a remix child.

### Dragons

part of track upload flow, but should be fine

### How Has This Been Tested?

Sets the track as invalid so user cannot proceed to use premium track as remix parent.

n/a

### Feature Flags ###

### DIFF
--- a/packages/web/src/common/store/remix-settings/sagas.ts
+++ b/packages/web/src/common/store/remix-settings/sagas.ts
@@ -68,6 +68,11 @@ function* watchFetchTrack() {
           })) as unknown as Track
         }
         if (track) {
+          // Premium tracks cannot be used as remix track parents
+          if (track.is_premium) {
+            yield* put(fetchTrackFailed())
+            return
+          }
           yield* put(fetchTrackSucceeded({ trackId: track.track_id }))
           return
         }


audius-infra:
Preview this change https://demo.audius.co/sd-pc-remix

packages/web/src/common/store/remix-settings/sagas.ts
```diff
@@ -68,6 +68,11 @@ function* watchFetchTrack() {
           })) as unknown as Track
         }
         if (track) {
+          // Premium tracks cannot be used as remix track parents
+          if (track.is_premium) {
+            yield* put(fetchTrackFailed())
```
rickyrombo, line 73: I wonder if this is the best place to "fail"? I'd be confused as a user as to what's happening - wondering if maybe it makes sense to have this check a bit closer to the UI, like [here](https://github.com/AudiusProject/audius-client/blob/85c75169cc3f67ea425844e9870c23cc8e36e5a7/packages/web/src/components/remix-settings-modal/ConnectedRemixSettingsModal.tsx#L61)? Technically the fetch succeeds, but the user shouldn't be allowed to set that track as the remix?

Wondering if it makes sense to show the successful track fetch but with a notice/"disabled" look about it being premium? (disclaimer: I'm not in the loop with design here)

rickyrombo (approved):

packages/web/src/common/store/remix-settings/sagas.ts
sddioulde: yeah i've gone back and forth myself about this, thought about putting it right where you mentioned actually, then ended going with this.
we'll have our first design sync on this tomorrow and i'll bring it up. thanks for raising

PR #2425, opened 2022-12-20
[C-1675] Fix useCallback

dylanjeffers:
### Description

Fixes incorrect useCallback type in @types/react 16.8. This is resolved in v18, but we need to do a bit of work first before upgrading react. 

The alternative is to just take the type from v18, and override useCallback, which is defined in useCallback.d.ts

### DIFF
--- a/packages/web/src/types/react/useCallback.d.ts
+++ b/packages/web/src/types/react/useCallback.d.ts
@@ -1,0 +1,13 @@
+declare namespace React {
+  /**
+   * `useCallback` will return a memoized version of the callback that only changes if one of the `inputs`
+   * has changed.
+   *
+   * @version 16.8.0
+   * @see https://reactjs.org/docs/hooks-reference.html#usecallback
+   */
+  // A specific function type would not trigger implicit any.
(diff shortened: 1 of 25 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-c-1675-use-callback

packages/web/src/types/react/useCallback.d.ts
```diff
@@ -0,0 +1,13 @@
...
+   * `useCallback` will return a memoized version of the callback that only changes if one of the `inputs`
+   * has changed.
+   *
+   * @version 16.8.0
+   * @see https://reactjs.org/docs/hooks-reference.html#usecallback
+   */
+  // A specific function type would not trigger implicit any.
+  // See https://github.com/DefinitelyTyped/DefinitelyTyped/issues/52873#issuecomment-845806435 for a comparison between `Function` and more specific types.
+  // tslint:disable-next-line ban-types
+  function useCallback<T extends Function>(callback: T, deps: DependencyList): T
```
rickyrombo, line 12: 🦸‍♂️

rickyrombo (approved):

PR #2426, opened 2022-12-20
[C-1625] Fix mobile profile follow button

dylanjeffers:
### Description

Fixes case where users were following themselves in mobile app, since the follow button incorrectly rendered in cases where user went to their own profile from search/track-screen.

Note: doing toLowerCase check for params just to make sure we are checking case where deeplink has lowercased handle

### DIFF
--- a/packages/mobile/src/screens/profile-screen/ProfileInfo.tsx
+++ b/packages/mobile/src/screens/profile-screen/ProfileInfo.tsx
@@ -1,5 +1,6 @@
-import { FollowSource } from '@audius/common'
+import { accountSelectors, FollowSource } from '@audius/common'
 import { View, Text } from 'react-native'
+import { useSelector } from 'react-redux'
 
 import { FollowButton, FollowsYouChip } from 'app/components/user'
 import UserBadges from 'app/components/user-badges'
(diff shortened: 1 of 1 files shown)


packages/mobile/src/screens/profile-screen/ProfileInfo.tsx
```diff
@@ -93,19 +96,27 @@ export const ProfileInfo = (props: ProfileInfoProps) => {
...
-        style={styles.followButton}
-        profile={profile}
-        onPress={onFollow}
-        followSource={FollowSource.PROFILE_PAGE}
-      />
-    </>
-  )
+  const isOwner =
+    params.handle === 'accountUser' ||
+    params.handle?.toLowerCase() === accountHandle?.toLowerCase() ||
```
raymondjacobson, line 101: possible if these are undefined we would check undefined === undefined? probs ok even so

raymondjacobson (approved):

packages/mobile/src/screens/profile-screen/ProfileInfo.tsx
dylanjeffers: yeah good callout, we have a check that handle needs to be present for isOwner to actually be there later on, so should be good

PR #2428, opened 2022-12-20
Fix offline playback with new track player

amendelsohn:
### Description

Change to use `react-native-trackplayer` broke offline playback due to asynchronous local image fetching.

TODO:
- [ ] check if this breaks the lock screen controls

### DIFF
--- a/packages/mobile/src/components/audio/Audio.tsx
+++ b/packages/mobile/src/components/audio/Audio.tsx
@@ -376,7 +376,7 @@ export const Audio = () => {
   })
 
   let source
-  if (offlineTrackUri) {
+  if (offlineTrackUri && trackImageSource) {
     source = {
       type: TrackType.Default,
       uri: offlineTrackUri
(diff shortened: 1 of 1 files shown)


packages/mobile/src/components/audio/Audio.tsx
```diff
@@ -420,9 +420,9 @@ export const Audio = () => {
     const newUri = source.uri
     if (currentUriRef.current !== newUri) {
       currentUriRef.current = newUri
-      const imageUrl = trackImageSource?.source[2].uri ?? DEFAULT_IMAGE_URL
+      const imageUrl = trackImageSource?.source?.[2]?.uri ?? DEFAULT_IMAGE_URL
       const nextImageUrl =
-        nextTrackImageSource?.source[2].uri ?? DEFAULT_IMAGE_URL
+        nextTrackImageSource?.source?.[2]?.uri ?? DEFAULT_IMAGE_URL
```
raymondjacobson, line 425: do we get artwork on the lock screen when offline?

raymondjacobson (approved):

packages/mobile/src/components/audio/Audio.tsx
amendelsohn: I still need to test that. We did, but haven't tried on device since the player change. Sim isn't showing me lock screen controls

PR #2429, opened 2022-12-20
[C-1543] Fix left-nav zero state vanity metrics

dylanjeffers:
### Description

Fixes left-nav zero state vanity metrics following design
- Uses playlist count when user has no tracks
- Adds vertical dividers
- Refactors vanity metrics section


Kyle-Shanks (approved):
Neat

PR #2432, opened 2022-12-20
[PAY-772] Update table styles to remove the space between the header text and sorters

Kyle-Shanks:
### Description
Small styling update. Should remove the space between the text and sorters so there is no gap in the clickable area

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/web/src/components/table/Table.module.css
+++ b/packages/web/src/components/table/Table.module.css
@@ -59,8 +59,8 @@
   line-height: 41px;
   user-select: none;
 }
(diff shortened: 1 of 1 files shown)


audius-infra:
Preview this change https://demo.audius.co/kj-Update-table-styles-to-remove-space-between-title-and-sorters

dharit-tan (approved):
Nice wow thought it would be way more complicated.

PR #2433, opened 2022-12-20
[C-880] Prevent track count change when filtering

dylanjeffers:
### Description

When filtering tracks, we don't want to change the metric for number of tracks


Kyle-Shanks (approved):

sliptype (approved):

audius-infra:
Preview this change https://demo.audius.co/dj-c-880-filtering-tracks

PR #2437, opened 2022-12-20
[C-1692] Prevent infinite failed requests while offline

sliptype:
### Description

* Check reachability in AudiusApiClient so retries stop once we enter an unreachable state. This means there are still some requests that fail before we get into the unreachable state, but they stop fairly quickly. Would be ideal to have the unreachable state happen faster, but bc we do a ping check it takes a bit
* Check reachability in `getNotifications`. This function gets run on an interval, so we can prevent failed requests by returning early

The reachability handlers not being in sagas made this kind of interesting. I'm definitely still in favor of moving away from sagas but the border between saga <> non-saga can be annoying

### Dragons

I noticed that coming back online doesn't seem to be working. Once my wifi is reconnected and I press reload in the app, nothing happens. Not sure if you are aware of this @amendelsohn but I could dig in if you want me to

### How Has This Been Tested?

iOS simulator

### Feature Flags ###

### DIFF
--- a/packages/common/src/services/audius-api-client/AudiusAPIClient.ts
+++ b/packages/common/src/services/audius-api-client/AudiusAPIClient.ts
@@ -491,6 +491,7 @@ export class AudiusAPIClient {
   remoteConfigInstance: RemoteConfigInstance
   localStorage: LocalStorage
   env: Env
+  isReachable?: boolean = true
 
   constructor({
     audiusBackendInstance,
@@ -506,6 +507,10 @@ export class AudiusAPIClient {
     this.remoteConfigInstance = remoteConfigInstance
     this.localStorage = localStorage
     this.env = env
+  }
+
+  setIsReachable(isReachable: boolean) {
+    this.isReachable = isReachable
   }
 
   async getTrending({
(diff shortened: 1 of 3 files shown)


packages/common/src/services/audius-api-client/AudiusAPIClient.ts
```diff
@@ -508,6 +509,10 @@ export class AudiusAPIClient {
     this.env = env
   }
 
+  setIsReachable(isReachable: boolean) {
+    this.isReachable = isReachable
```
amendelsohn, line 513: Makes me slightly nervous to keep another separate state. We're risking bugs where it gets out of sync, but I think it's worth it.

amendelsohn (approved):
Nice, thank you! This will help a lot. I was primarily worried about these draining battery life, so even if we can cut them off after the ping test, I think that's good enough.

amendelsohn:
> I noticed that coming back online doesn't seem to be working. Once my wifi is reconnected and I press reload in the app, nothing happens. Not sure if you are aware of this @amendelsohn but I could dig in if you want me to

Hmm there have been changes to the backend reachability and saga blocking since I last checked that. Is it happening on device as well as the simulator? I do know that the NetworkInfo update events sometimes just get dropped on the sim.

sliptype:
> > I noticed that coming back online doesn't seem to be working. Once my wifi is reconnected and I press reload in the app, nothing happens. Not sure if you are aware of this @amendelsohn but I could dig in if you want me to
> 
> Hmm there have been changes to the backend reachability and saga blocking since I last checked that. Is it happening on device as well as the simulator? I do know that the NetworkInfo update events sometimes just get dropped on the sim.

Let me try on device and I'll get back to you

packages/common/src/services/audius-api-client/AudiusAPIClient.ts
sliptype: Yeah I wish I could reach into the redux state here but there isn't a clean way to do it with sagas that I could think of. I could export `store` from our top level `store.ts` file like we are doing with `dispatch` but then it creates this giant circular dependency that I would like to avoid

audius-infra:
Preview this change https://demo.audius.co/sk-c-1692-prevent-failed-reqs-offline

PR #2438, opened 2022-12-21
[C-1078] Add offline mode confirmation drawers

dylanjeffers:
### Description

Adds three types of offline confirmation drawers to ensure user means to delete downloaded tracks
- RemoveDownloadedFavoritesDrawer opens when user switches off favorites track downloading
- RemoveDownloadedCollectionDrawer opens when user switches off playlist/album download
- UnfavoriteDownloadedCollectionDrawer opens when user unfavorites a downloaded collection

Additional changes
- Creates ConfirmationDrawer base component, and implements above three drawers + refactors other various confirmation drawers
- Stictly types drawer data to ensure we are typesafe across opening/closing drawers

https://user-images.githubusercontent.com/8230000/208819129-a4609258-8af6-483c-9eea-dc93559fc7ad.MP4

### Dragons
There is a bug currently with AnimatedButton that prevents us from not animating the FavoriteButton from active to inactive when the confirmation dialog appears... Will def dig into this eventually though.

There is also a bug with switches where it kinda glitches closed and then open when pressing and opening. This appears to be a bug with the react-native switch, but i think we can hack a fix if we capture touches before passing it along to the switch. going to look into this eventually too when we are doing final ui pass.


packages/mobile/src/components/drawers/UnfavoriteDownloadedCollection.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const collectionIdStr = collectionId.toString()
+
+  const tracksForDownload = useSelector((state: CommonState) => {
+    const collection = state.collections.entries[collectionId]
+    if (!collection) return []
+    const { tracks } = collection.metadata
+    if (!tracks) return []
+
+    const downloadReason = {
+      is_from_favorites: false,
```
amendelsohn, line 35: Now this is interesting... I think we actually want to remove it whether it's from favorites or not. We could have originally downloaded the collection as part of the favorites umbrella, or on the collection page. In either case, un-favoriting the collection should cause the download to be removed.

I think the solution here is actually to add two items to `tracksForDownload`: one with `is_from_favorites: true` and one that's `false`. That should still work with `removeCollectionDownload` as it does a best-effort and only removes the data once the track's remaining `downloadReasons` list is empty

packages/mobile/src/components/offline-downloads/DownloadToggle.tsx
```diff
@@ -93,7 +93,9 @@ export const DownloadToggle = ({
   labelText,
   isFavoritesDownload
 }: DownloadToggleProps) => {
-  const styles = useStyles({ labelText })
+  const styleProps = useMemo(() => ({ labelText }), [labelText])
```
amendelsohn, line 96: Oh, nice optimization, thank you!

packages/mobile/src/components/offline-downloads/DownloadToggle.tsx
```diff
@@ -110,21 +112,45 @@ export const DownloadToggle = ({
...
       } else {
-        collectionIdStr &&
-          removeCollectionDownload(collectionIdStr, tracksForDownload)
+        if (!isFavoritesDownload && collectionIdStr) {
+          // we are trying to remove download from a collection page
+          dispatch(
+            setVisibility({
+              drawer: 'RemoveDownloadedCollection',
+              visible: true,
+              data: { collectionId: collectionIdStr, tracksForDownload }
```
amendelsohn, line 129: Cool! I was expecting the handler to be passed in, but this works

packages/mobile/src/components/drawers/RemoveDownloadedCollectionDrawer.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  const handleConfirm = useCallback(() => {
+    removeCollectionDownload(collectionId, tracksForDownload)
+  }, [collectionId, tracksForDownload])
+
+  return (
+    <ConfirmationDrawer
+      drawerName={drawerName}
+      messages={messages}
+      onConfirm={handleConfirm}
```
amendelsohn, line 29: if the user clicks off the modal, that'll trigger `onCancel` and do nothing, is that correct?

amendelsohn (approved):
Wow! Super nice consolidation into ConfirmationDrawer.

Thank you so much for doing this. Would have taken me at least twice as long!

packages/mobile/src/Drawers.tsx
```diff
@@ -95,7 +98,10 @@ const nativeDrawersMap: { [DrawerName in Drawer]?: ComponentType } = {
   DownloadTrackProgress: DownloadTrackProgressDrawer,
   ForgotPassword: ForgotPasswordDrawer,
   DeleteConfirmation: DeleteConfirmationDrawer,
-  ShareToStoryProgress: ShareToStoryProgressDrawer
+  ShareToStoryProgress: ShareToStoryProgressDrawer,
+  RemoveDownloadedCollection: RemoveDownloadedCollectionDrawer,
+  RemoveDownloadedFavorites: RemoveDownloadedFavoritesDrawer,
+  UnfavoriteDownloadedCollection: UnfavoriteDownloadedCollectionDrawer
```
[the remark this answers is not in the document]
dylanjeffers: hahahah, eye me wares and purchase those you take fancy

packages/mobile/src/components/drawers/ConfirmationDrawer.tsx
```diff
@@ -0,0 +1,111 @@
...
+          style={styles.headerIcon}
+          height={spacing(5)}
+          width={spacing(5)}
+          fill={neutral}
+        />
+        <Text fontSize='xl' weight='heavy' textTransform='uppercase'>
+          {messages.header}
+        </Text>
+      </View>
+      <Text fontSize='large' style={styles.description}>
```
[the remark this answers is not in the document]
dylanjeffers: i use the props for all the typography related things, and then style for like margins and other things not exposed as props

packages/mobile/src/components/offline-downloads/DownloadToggle.tsx
dylanjeffers: yeah im trying to avoid passing non-serializable things into redux store when possible

packages/mobile/src/components/drawers/RemoveDownloadedCollectionDrawer.tsx
dylanjeffers: yeah exactly, the NativeDrawer simply calls onClose

packages/mobile/src/components/drawers/RemoveDownloadedFavorites.tsx
```diff
@@ -0,0 +1,32 @@
...
+const messages = {
+  header: 'Are You Sure?',
+  description:
+    'Are you sure you want to remove all of your downloaded favorites?',
+  confirm: 'Remove All Downloads'
+}
+
+const drawerName = 'RemoveDownloadedFavorites'
+
+export const RemoveDownloadedFavoritesDrawer = () => {
```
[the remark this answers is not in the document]
dylanjeffers: yeah i can see potentially merging them, but with the text differences etc, thought might as well separate for now

PR #2440, opened 2022-12-21
[C-1673] Add collection to favorites on download

amendelsohn:
### Description

Automatically favorite any collection that is downloaded individually.


dylanjeffers (approved):
cool to have a new favorite source

PR #2442, opened 2022-12-21
[C-1674] Fix queue controls for offline playback

amendelsohn:
### Description

Queue controls (e.g. the Next button) were not working in offline playback.
The kind was missing from the lineup tracks, so we were failing to translate them to items in the playback queue state.


dylanjeffers (approved):
nice catch

PR #2443, opened 2022-12-21
[C-1478] Clear offline downloads when switching accounts

dylanjeffers:
### Description

Clears downloads when signing out. Adds a saga for this, are we trying to avoid using sagas though?

### DIFF
--- a/packages/mobile/src/hooks/useLoadOfflineTracks.ts
+++ b/packages/mobile/src/hooks/useLoadOfflineTracks.ts
@@ -49,7 +49,6 @@ export const useLoadOfflineTracks = () => {
       uid: string
       metadata: CollectionMetadata
     }[] = []
-    console.log('okay adding collection for some reason...', offlineCollections)
     for (const collectionId of offlineCollections) {
       try {
         if (collectionId === DOWNLOAD_REASON_FAVORITES) continue
--- a/packages/mobile/src/hooks/useLoadOfflineTracks.ts
+++ b/packages/mobile/src/hooks/useLoadOfflineTracks.ts
@@ -49,6 +49,7 @@ export const useLoadOfflineTracks = () => {
       uid: string
       metadata: CollectionMetadata
     }[] = []
+    console.log('okay adding collection for some reason...', offlineCollections)
     for (const collectionId of offlineCollections) {
       try {
         if (collectionId === DOWNLOAD_REASON_FAVORITES) continue
--- a/packages/mobile/src/services/offline-downloader/offline-storage.ts
+++ b/packages/mobile/src/services/offline-downloader/offline-storage.ts
@@ -221,7 +221,6 @@ export const verifyTrack = async (
 }
 
 export const purgeAllDownloads = async (withLogs?: boolean) => {
-  const trackIds = await listTracks()
   if (withLogs) {
     console.log(`Before purge:`)
   }
@@ -232,7 +231,6 @@ export const purgeAllDownloads = async (withLogs?: boolean) => {
     console.log(`After purge:`)
   }
   await readDirRec(downloadsRoot)
-  store.dispatch(unloadTracks(trackIds))
 }
 
 export const purgeDownloadedTrack = async (trackId: string) => {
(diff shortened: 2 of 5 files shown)


packages/mobile/src/hooks/useLoadOfflineTracks.ts
```diff
@@ -48,6 +48,7 @@ export const useLoadOfflineTracks = () => {
       uid: string
       metadata: CollectionMetadata
     }[] = []
+    console.log('okay adding collection for some reason...', offlineCollections)
```
amendelsohn, line 51: debug log?

packages/mobile/src/services/offline-downloader/offline-storage.ts
```diff
@@ -219,18 +220,19 @@ export const verifyTrack = async (
...
   await RNFS.mkdir(downloadsRoot)
-  console.log(`After purge:`)
+  if (withLogs) {
+    console.log(`After purge:`)
+  }
   await readDirRec(downloadsRoot)
-  trackIds.forEach((trackId) => {
-    store.dispatch(unloadTrack(trackId))
-  })
+  store.dispatch(unloadTracks(trackIds))
```
amendelsohn, line 235: We should also get the collectionIds and call `offlineDownloadsActions.removeCollection` on each of them.

I wrote this before you could download collections.

packages/mobile/src/store/offline-downloads/slice.ts
```diff
@@ -82,8 +82,20 @@ const slice = createSlice({
...
+      })
+    },
     doneLoadingFromDisk: (state) => {
       state.isDoneLoadingFromDisk = true
+    },
+    clearOfflineDownloads: (state) => {
+      state.collections = initialState.collections
+      state.tracks = initialState.tracks
+      state.downloadStatus = initialState.downloadStatus
+      state.isDoneLoadingFromDisk = initialState.isDoneLoadingFromDisk
```
amendelsohn, line 98: does `state = {...initialState}` work?
I suppose this is more explicit in case we add more unrelated state to this slice

packages/mobile/src/store/sagas.ts
```diff
@@ -144,6 +145,7 @@ export default function* rootSaga() {
     ...mobileUiSagas(),
     ...uploadSagas(),
     ...remixSettingsSagas(),
+    ...offlineDownloadSagas(),
```
amendelsohn, line 148: haha it took me a while but I just did this too!

amendelsohn (approved):
Looks great!

Julian had mentioned putting some sort of warning on the UI before signing out. I think we already have that, so I wonder if we should put it as an addendum to the signOut confirmation drawer if any downloads are on? Might need a design pass.

dylanjeffers:
> Looks great!
> 
> 
> 
> Julian had mentioned putting some sort of warning on the UI before signing out. I think we already have that, so I wonder if we should put it as an addendum to the signOut confirmation drawer if any downloads are on? Might need a design pass.

Yeah we have the sign out confirmation drawer, could just add more text to it if we have downloads?

packages/mobile/src/services/offline-downloader/offline-storage.ts
dylanjeffers: so given that, ill just remove this unloadTracks dispatch, cause the removeDownloads reducer sets the offline-download state to initialState

packages/mobile/src/store/offline-downloads/slice.ts
dylanjeffers: it really weirdly doesnt... i was really surprised

amendelsohn: huh. I mean this is fine too

dylanjeffers: lets circle back on this in final qa cause i really didnt understand why it didnt work

packages/mobile/src/services/offline-downloader/offline-storage.ts
amendelsohn: yeah that works
unlink of the root should delete everything

Come to think of it, can you confirm if you haven't that this doesn't break the directory? (i.e. that downloads still work after this gets called) I think the mkdir I had on 230 should handle that, but want to make sure.

PR #2445, opened 2022-12-21
Don't stop loading if one collection fails

amendelsohn:
### Description

`useLoadOfflineTracks` was choking if any collection failed to load from disk. Likely due to a corrupted file that was half written before last time. Wrapping it in a try catch lets us move on and still load the rest.

### DIFF
--- a/packages/mobile/src/hooks/useLoadOfflineTracks.ts
+++ b/packages/mobile/src/hooks/useLoadOfflineTracks.ts
@@ -28,6 +28,7 @@ import {
   getOfflineCollections,
   getTrackJson,
   listTracks,
+  purgeDownloadedCollection,
   verifyTrack
 } from '../services/offline-downloader/offline-storage'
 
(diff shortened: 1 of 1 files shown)


Kyle-Shanks (approved):

packages/mobile/src/hooks/useLoadOfflineTracks.ts
```diff
@@ -49,20 +49,24 @@ export const useLoadOfflineTracks = () => {
...
         })
+        if (collection.user) {
+          cacheUsers.push({
+            id: collection.user.user_id,
+            uid: makeUid(Kind.USERS, collection.user.user_id),
+            metadata: collection.user
+          })
+        }
+      } catch (e) {
+        console.warn('Failed to load offline collection', collectionId)
```
amendelsohn, line 69: I wonder if we should just remove the folder in this case...

PR #2446, opened 2022-12-21
[C-1698] Fix first offline playback stalled

amendelsohn:
### Description

First playback was failing to play any audio. Subsequent plays were working.

Changes:
1. Wait for (should be snappy) filesystem check for the offline source before selecting one
2. Memoize the source objects to reduce rerenders

### Dragons

Changing playback is scary. This is not feature flagged, so we should make sure it performs well enough.

### How Has This Been Tested?

Simulator

Do we have existing playback speed metrics?

### Feature Flags ###

Notably, this is **NOT** behind the `OFFLINE_MODE_ENABLED` flag


packages/mobile/src/components/audio/Audio.tsx
```diff
@@ -375,90 +373,117 @@ export const Audio = () => {
...
-      if (playing) await TrackPlayer.play()
-
-      const isPodcast = track?.genre === Genre.PODCASTS
-      if (isPodcast !== isPodcastRef.current) {
-        isPodcastRef.current = isPodcast
-        await updatePlayerOptions(isPodcast)
+    if (
+      loadingOfflineTrack ||
+      loadingNextOfflineTrack ||
+      currentUriRef.current === newUri
```
Kyle-Shanks, line 442: Would this block non-offline tracks from playing if the offline track is still loading? Not sure if thats how it works, but wanted to check

amendelsohn: Yes. I don't think that should take much time, but we can do some perf tests. This was the only way I was able to get it to work, but I'm open to suggestions. Certainly don't want to impose any significant delay on all plays. Maybe we can check the offline downloads state and short circuit if it's not in there

amendelsohn: To clarify, it doesn't need to read the whole file, just verify that the file at the specified location exists

Kyle-Shanks: sounds good. i'm gonna checkout the branch and test playback a bit just to be safe, but this seems ok to me

amendelsohn: Thank you! I actually think the short circuit may be worth it anyway. I'll look into adding it tonight

Kyle-Shanks (approved):
looks good 👍

PR #2447, opened 2022-12-22
[C-1646][C-1647] Handle more offline lineup states

raymondjacobson:
### Description

Resolves two issues (mostly):

[C-1646]: Add favorite created_at data to the `offlineDownloads` slice and forces the lineups to sort by either favorite or playlist added timestamp accordingly

[C-1647]: Make sure lineups (collection page & favorites page) allow you to deal with switching back & forth between reachability states. This involved a few things:
* Fix a couple bugs around promises where things were not being awaited properly (missing data in the `offlineDownloads` slice)
* Restructure the `useOfflineCollectionLineup` hook to accept a bit more lineup functionality so that the component itself can forget about conditionally rendering certain things. I think this is ok, but maybe not ideal. Hard to get better without significant code restructuring, but the mental modal here is that components should worry about what data should be loaded up front rather than worry about what it needs to render.
* A useEffect in the hook that will re-trigger fetching the original content when reachability is regained. This importantly will not fetch the data from the network because it presumably is already in the cache. (maybe not even a good thing anymore, but fyi)

### Dragons

- This impacts lineups a little bit even with the feature flag turned off.
- This has a somewhat weird UX caveat where switching between online & offline lineups seem to yield different `uid`s which makes it seems like you're looking at a totally new list of tracks versus the one that you might have been playing. Audio still does work and you're able to play anything else, but it looks like you're listening to some other queue of tracks. This PR was getting a little crazy, so wanted to draw the line here, but I think a bit more love is needed.

### How Has This Been Tested?

Simulator extensively with a Button to switch reachability state, on device to cursorily test, but I do think substantial more testing is required.

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/hooks/useFetchAllFavoritedTrackIds.ts
+++ b/packages/mobile/src/hooks/useFetchAllFavoritedTrackIds.ts
@@ -1,4 +1,4 @@
-import type { APIFavorite } from '@audius/common'
+import type { APIFavorite, ID } from '@audius/common'
 import { decodeHashId, encodeHashId, accountSelectors } from '@audius/common'
 import { useSelector } from 'react-redux'
 import { useAsync } from 'react-use'
@@ -10,7 +10,8 @@ import { useIsOfflineModeEnabled } from './useIsOfflineModeEnabled'
 const { getUserId } = accountSelectors
 
 export const fetchAllFavoritedTrackIds = async (currentUserId: number) => {
-  let trackIds: number[] = []
+  let tracksAndTimestamps: Array<{ trackId: ID; favoriteCreatedAt: string }> =
+    []
   let loadMore = true
   let offset = 0
   // TODO: store results in state to avoid duplicate fetching
@@ -29,15 +30,16 @@ export const fetchAllFavoritedTrackIds = async (currentUserId: number) => {
 
     loadMore = result.length > 0
     offset += result.length
-    trackIds = trackIds.concat(
+    tracksAndTimestamps = tracksAndTimestamps.concat(
       result
         .filter((trackSave) => trackSave.favorite_type === 'SaveType.track')
-        .map((trackSave: APIFavorite) =>
-          decodeHashId(trackSave.favorite_item_id)
-        )
+        .map((trackSave: APIFavorite) => ({
+          trackId: decodeHashId(trackSave.favorite_item_id),
+          favoriteCreatedAt: trackSave.created_at
+        }))
     )
   }
-  return trackIds
+  return tracksAndTimestamps
 }
 
 export const useFetchAllFavoritedTrackIds = () => {
--- a/packages/mobile/src/hooks/useLoadOfflineTracks.ts
+++ b/packages/mobile/src/hooks/useLoadOfflineTracks.ts
@@ -1,3 +1,5 @@
+import { useEffect, useState } from 'react'
+
 import type {
   CollectionMetadata,
   Track,
@@ -5,13 +7,13 @@ import type {
   UserTrackMetadata
 } from '@audius/common'
 import {
-  collectionPageLineupActions,
-  savedPageTracksLineupActions,
   Kind,
   makeUid,
   cacheActions,
   reachabilitySelectors
 } from '@audius/common'
+import { getCollection } from '@audius/common/dist/store/cache/collections/selectors'
+import type { LineupActions } from '@audius/common/dist/store/lineup/actions'
 import { useDispatch, useSelector } from 'react-redux'
 import { useAsync } from 'react-use'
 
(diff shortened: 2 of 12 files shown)


packages/mobile/src/hooks/useLoadOfflineTracks.ts
```diff
@@ -76,36 +78,39 @@ export const useLoadOfflineTracks = () => {
...
-            })
-            if (track.user) {
-              cacheUsers.push({
-                id: track.user.user_id,
-                uid: makeUid(Kind.USERS, track.user.user_id),
-                metadata: track.user
-              })
-            }
-            lineupTracks.push(lineupTrack)
+    await Promise.all(
```
raymondjacobson, line 81: This was dispatching the below actions before this work was done

packages/mobile/src/screens/favorites-screen/FavoritesScreen.tsx
```diff
@@ -80,6 +81,7 @@ export const FavoritesScreen = () => {
             is_from_favorites: true,
             collection_id: collection.playlist_id.toString()
           }
+          // TODO: include a favorite_created_at timestamp for sorting offline collections
```
raymondjacobson, line 84: this is the todo remaining from 1647

amendelsohn: so we're not currently sorting the favorited collections?

packages/mobile/src/screens/favorites-screen/TracksTab.tsx
```diff
@@ -129,17 +132,7 @@ export const TracksTab = () => {
...
-    () =>
-      !isReachable && isOfflineModeEnabled
-        ? Object.values(offlineTracks).filter((track) =>
-            track.offline?.reasons_for_download.some(
-              (reason) => reason.collection_id === DOWNLOAD_REASON_FAVORITES
-            )
-          )
-        : savedTracks.entries,
-    [isOfflineModeEnabled, isReachable, offlineTracks, savedTracks.entries]
-  )
```
amendelsohn, line 142: this is a great change, thank you!

packages/mobile/src/services/offline-downloader/offline-sync.ts
```diff
@@ -99,21 +99,22 @@ const syncFavorites = async () => {
...
+    ({ trackId }) => !oldTrackIds.has(trackId)
   )
   const removedTrackIds = [...oldTrackIds].filter(
     (trackId) => !newTrackIds.has(trackId)
   )
 
   const tracksForDownload: TrackForDownload[] = addedTrackIds.map(
     (addedTrack) => ({
-      trackId: addedTrack,
+      trackId: addedTrack.trackId,
```
amendelsohn, line 112: huh. Does `fetchAllFavoritedTrackIds` not return `number[]`? Why do we need `.trackId` here?

amendelsohn (approved):
This looks great! Thank you! :ray:🚀

packages/mobile/src/screens/favorites-screen/FavoritesScreen.tsx
raymondjacobson: that's right. i think less jarring of a ux issue versus playlist contents being wrongly ordered but this is still a todo

packages/mobile/src/services/offline-downloader/offline-sync.ts
raymondjacobson: should rename some vars. this now as id and created at

packages/mobile/src/hooks/useFetchAllFavoritedTrackIds.ts
```diff
@@ -10,7 +10,8 @@ import { useIsOfflineModeEnabled } from './useIsOfflineModeEnabled'
 const { getUserId } = accountSelectors
 
 export const fetchAllFavoritedTrackIds = async (currentUserId: number) => {
-  let trackIds: number[] = []
+  let tracksAndTimestamps: Array<{ trackId: ID; favoriteCreatedAt: string }> =
```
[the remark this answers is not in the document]
raymondjacobson: oops

packages/mobile/src/hooks/useLoadOfflineTracks.ts
```diff
@@ -114,40 +119,105 @@ export const useLoadOfflineTracks = () => {
...
+          const bTime = b.offline?.favorite_created_at
+          const aTime = a.offline?.favorite_created_at
+          if (aTime && bTime) {
+            if (aTime > bTime) {
+              return -1
+            }
+            return 1
+          }
+          return 0
+        })
```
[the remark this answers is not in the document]
raymondjacobson: nice!

packages/mobile/src/hooks/useLoadOfflineTracks.ts
```diff
@@ -114,40 +119,105 @@ export const useLoadOfflineTracks = () => {
...
+    isReachable,
+    collectionId,
+    setHasGoneOffline,
+    collection?.playlist_contents.track_ids
+  ])
+
+  // If we go back online, trigger a refetch of the original content
+  useEffect(() => {
+    if (isReachable && hasGoneOffline) {
+      fetchOnlineContent()
```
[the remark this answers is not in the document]
raymondjacobson: Technically this can be only set once because it's like a "have we done the initial work" check, but I think longer term a nicer model would be to have a single loader hook manage all of this -- everything from first fetch to offline fetches and later refetches.

```
const useLoadLineup = (lineup: 'favorites' | ...) => {
  const doesSupportOffline = lineup === 'favorites' || ...
  if (doesSupportOffline) {
     // reachability stuff
  }

  useFocusEffect() { ... dispatch () }
}

const lineup = useLoadLineup('favorites')

return <Lineup lineup={lineup} />
```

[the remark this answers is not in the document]
raymondjacobson: (But i made this change anyway)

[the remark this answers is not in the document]
amendelsohn: Well I was wondering if you go offline -> online -> offline, do we not need to call the `lineupMetadatasSucceeded` again?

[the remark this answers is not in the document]
raymondjacobson: the hasGoneOffline is just to make sure that we don't call `fetchOnlineContent` on the first component render, afterwards it should be called every time isReachable becomes true again

PR #2452, opened 2022-12-22
Secret override for offline mode

amendelsohn:
### Description

Add a secret dev switch to enable offline mode for testing. Required because we can't reach optimizely when we're offline

### Dragons

Is this an obscure enough trigger? We could put it somewhere less discoverable.

### How Has This Been Tested?

On iOS device

### DIFF
--- a/packages/web/src/common/store/player/sagas.ts
+++ b/packages/web/src/common/store/player/sagas.ts
@@ -75,7 +75,7 @@ export function* watchPlay() {
   const isOfflineModeEnabled = yield* call(
     getFeatureEnabled,
     FeatureFlags.OFFLINE_MODE_ENABLED
-  ) ?? false
+  ) ?? true
 
   yield* takeLatest(play.type, function* (action: ReturnType<typeof play>) {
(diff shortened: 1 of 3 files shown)


packages/web/src/common/store/player/sagas.ts
```diff
@@ -75,7 +75,7 @@ export function* watchPlay() {
   const isOfflineModeEnabled = yield* call(
     getFeatureEnabled,
     FeatureFlags.OFFLINE_MODE_ENABLED
-  ) ?? false
+  ) ?? true
```
sliptype, line 78: Do we still want this?

sliptype (approved):
Sneaky, nice!

packages/web/src/common/store/player/sagas.ts
amendelsohn: Yes, unfortunately we don't have context in common to do this properly. However, it shouldn't be possible to get here unless you've already managed to download and trigger a play for an offline track so... I think that's fine?

audius-infra:
Preview this change https://demo.audius.co/amendel-mobile-feature-flag-override

PR #2453, opened 2022-12-22
Fix favorited tracks download

amendelsohn:
### Description

Fix some bugs with fetching and downloading favorited tracks.

### DIFF
--- a/packages/mobile/src/components/offline-downloads/DownloadToggle.tsx
+++ b/packages/mobile/src/components/offline-downloads/DownloadToggle.tsx
@@ -186,7 +186,7 @@ export const DownloadToggle = ({
         <Switch
           value={isCollectionMarkedForDownload}
           onValueChange={handleToggleDownload}
-          disabled={!collectionId || disabled}
+          disabled={disabled}
         />
       </View>
     </View>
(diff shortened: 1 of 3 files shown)


packages/mobile/src/components/offline-downloads/DownloadToggle.tsx
```diff
@@ -186,7 +186,7 @@ export const DownloadToggle = ({
         <Switch
           value={isCollectionMarkedForDownload}
           onValueChange={handleToggleDownload}
-          disabled={!collectionId || disabled}
```
Kyle-Shanks, line 189: Was there a reason we were checking for collectionId previously?

amendelsohn: Yeah, but I added the isFavoritesDownload flag since then, and the check up above should cover this. the check we would use here is `(!collectionId && !isFavoritesDownload) || disabled`

amendelsohn: Yeah, but I added the isFavoritesDownload flag since then, and the check up above should cover this. the check we would use here is `(!collectionId && !isFavoritesDownload) || disabled`

Kyle-Shanks: cool sounds good

Kyle-Shanks (approved):
Nice 😎

PR #2456, opened 2022-12-23
[C-1083, C-1080] Add offline play count recording

dylanjeffers:
### Description

Adds offline play count recording using react-native-queue
Additional changes: 
- Divides out worker types into separate files
- Adds helper for reseting workers
- starts/stops the queue based on reachability status

### DIFF
--- a/packages/mobile/src/components/audio/Audio.tsx
+++ b/packages/mobile/src/components/audio/Audio.tsx
@@ -275,10 +275,10 @@ export const Audio = () => {
 
     if (position > RECORD_LISTEN_SECONDS && !listenLoggedForTrack) {
       setListenLoggedForTrack(true)
-      if (isOfflineModeEnabled && !isReachable) {
-        queue.addJob<PlayCountWorkerPayload>(PLAY_COUNTER_WORKER, { trackId })
-      } else if (isReachable) {
+      if (isReachable) {
         dispatch(recordListen(trackId))
+      } else if (isOfflineModeEnabled && !isReachable) {
+        queue.addJob<PlayCountWorkerPayload>(PLAY_COUNTER_WORKER, { trackId })
       }
     }
 
--- a/packages/mobile/src/components/audio/Audio.tsx
+++ b/packages/mobile/src/components/audio/Audio.tsx
@@ -273,11 +273,7 @@ export const Audio = () => {
     const duration = await TrackPlayer.getDuration()
     const position = await TrackPlayer.getPosition()
 
-    if (
-      position > RECORD_LISTEN_SECONDS &&
-      !listenLoggedForTrack &&
-      isReachable
-    ) {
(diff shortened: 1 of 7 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-c-1083-offline-play-count

packages/mobile/src/components/audio/Audio.tsx
```diff
@@ -268,13 +273,13 @@ export const Audio = () => {
...
-      isReachable
-    ) {
+    if (position > RECORD_LISTEN_SECONDS && !listenLoggedForTrack) {
       setListenLoggedForTrack(true)
-      dispatch(recordListen(trackId))
+      if (isOfflineModeEnabled && !isReachable) {
+        queue.addJob<PlayCountWorkerPayload>(PLAY_COUNTER_WORKER, { trackId })
+      } else if (isReachable) {
+        dispatch(recordListen(trackId))
+      }
```
amendelsohn, line 282: super nitpick, but I might put the reachable case first since it's the most common one

packages/mobile/src/components/offline-downloads/DownloadToggle.tsx
```diff
@@ -186,7 +186,7 @@ export const DownloadToggle = ({
         <Switch
           value={isCollectionMarkedForDownload}
           onValueChange={handleToggleDownload}
-          disabled={disabled}
+          disabled={(!collectionId && !isFavoritesDownload) || disabled}
```
amendelsohn, line 189: Did removing this break something? I took it out cause we had the check above where we return null in this case on 158.

packages/mobile/src/services/offline-downloader/workers/playCounterWorker.ts
```diff
@@ -0,0 +1,17 @@
...
+const { recordListen } = tracksSocialActions
+
+export const PLAY_COUNTER_WORKER = 'play_counter_worker'
+
+export type PlayCountWorkerPayload = { trackId: number }
+
+const countPlay = async (payload: PlayCountWorkerPayload) => {
+  const { trackId } = payload
+  store.dispatch(recordListen(trackId))
+}
```
amendelsohn, line 15: So simple!!

amendelsohn (approved):
Super clean!

packages/mobile/src/components/offline-downloads/DownloadToggle.tsx
dylanjeffers: Yeah lmk what you think, wasn't sure if a bug cause the favorites switch was unresponsive otherwise

amendelsohn: Interesting. I did check in a change to main to fix this, so maybe try again without? It used to just be `!collectionId || disabled` which was unresponsive.

PR #2459, opened 2022-12-23
[C-1721] Fix case where cached track is incomplete

dylanjeffers:
### Description

Fixes interesting case where a track object exists, but doesn't have an id, or any real metadata. this happens due to a design choice to optimistically set a track's remix values before fetch has completed, resulting in a track object of `{_remixes: [..], _remixCount: 0}` for a few ticks before the full object comes in. This caused a bug on mobile side where we think we have a full track object and try to do `track?.track_id.toString()` The fix chosen in this pr is to continue using the "searchTrack" object before the fetchedTrack comes back with everything, using the existence of an id as that indicator.

@raymondjacobson i know you have context on this code, how do you feel about it? i think we might want to include the track_id in this cache change?

### DIFF
--- a/packages/mobile/ios/Podfile.lock
+++ b/packages/mobile/ios/Podfile.lock
@@ -803,7 +803,7 @@ SPEC CHECKSUMS:
   FirebaseInstallations: 60edbf7e11d91ae4c751d26c200dfd037099abe0
   fmt: ff9d55029c625d3757ed641535fd4a75fedc7ce9
   GCDWebServer: 2c156a56c8226e2d5c0c3f208a3621ccffbe3ce4
-  glog: 5337263514dd6f09803962437687240c5dc39aa4
+  glog: 85ecdd10ee8d8ec362ef519a6a45ff9aa27b2e85
   google-cast-sdk-no-bluetooth: 5313156110d6e71fa336f6d742f1ee641e91fdca
   GoogleAppMeasurement: ae033c3aad67e68294369373056b4d74cc8ae0d6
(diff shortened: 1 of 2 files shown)


packages/mobile/ios/Podfile.lock
```diff
@@ -888,4 +888,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 86c7e909196d7920299e68babc12fcaf4882d3e8
 
-COCOAPODS: 1.11.2
+COCOAPODS: 1.11.3
```
sliptype, line 891: Why cocoapods changes

sliptype (approved):
Seems good to me! setting the track_id in the cache update like you mentioned would accomplish the same thing right? I think either solution is fine

packages/mobile/ios/Podfile.lock
Kyle-Shanks: this might have been me originally actually. whenever i pod install with lerna, the version is set to 1.11.2. I think its supposed to be at 1.11.3

dylanjeffers: @Kyle-Shanks yeah this is the result of me clearing out mobile node_modules and running npm i, and just seeing what happens, might help you if it keeps reverting you

dylanjeffers:
> Seems good to me! setting the track_id in the cache update like you mentioned would accomplish the same thing right? I think either solution is fine

The only issue with adding track_id is that it might break other things, since our types expect track to be `Track | SearchTrack`, so yeah its a tricky one...

PR #2460, opened 2022-12-23
[C-1724] Add track screen skeleton

dylanjeffers:
### Description

Adds track screen skeleton, to improve perceived load when clicking a deep link track. This extra helps cause coming into the audius app, things feel extra broken having a blank screen as first experience.

TODO: collection pages

### DIFF
--- a/packages/mobile/src/screens/track-screen/TrackScreenSkeleton.tsx
+++ b/packages/mobile/src/screens/track-screen/TrackScreenSkeleton.tsx
@@ -131,7 +131,6 @@ export const TrackScreenSkeleton = () => {
               (elementWidth: number, i) => (
                 <StaticSkeleton
                   key={i}
-                  noShimmer
                   style={[styles.descriptionText, { width: elementWidth }]}
                 />
               )
--- a/packages/mobile/src/screens/track-screen/TrackScreenSkeleton.tsx
+++ b/packages/mobile/src/screens/track-screen/TrackScreenSkeleton.tsx
@@ -2,7 +2,7 @@ import { times, random } from 'lodash'
 import { View } from 'react-native'
 
 import { Screen, Tile, Text, Divider } from 'app/components/core'
-import Skeleton from 'app/components/skeleton/Skeleton'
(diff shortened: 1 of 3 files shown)


Kyle-Shanks (approved):
clean stuff, nice

sliptype (approved):
LETSSS GOOO

I've been meaning to do this, thank you!

I will say that we maybe don't need all this complexity in the skeleton, when I was profiling the perf of animations it turned out that having many animations on screen like this is a big perf hit. So maybe we can simplify a bit?

dylanjeffers:
> LETSSS GOOO
> 
> I've been meaning to do this, thank you!
> 
> I will say that we maybe don't need all this complexity in the skeleton, when I was profiling the perf of animations it turned out that having many animations on screen like this is a big perf hit. So maybe we can simplify a bit?

totally agreed, how about i remove the shimmer effect for every one except maybe the image?

dylanjeffers:
> > LETSSS GOOO
> > I've been meaning to do this, thank you!
> > I will say that we maybe don't need all this complexity in the skeleton, when I was profiling the perf of animations it turned out that having many animations on screen like this is a big perf hit. So maybe we can simplify a bit?
> 
> totally agreed, how about i remove the shimmer effect for every one except maybe the image?

The other option im thinking, is that we can somehow use a "SkeletonProvider" that manages the single animated value instance, that all these use to do a coordinated anim?

packages/mobile/src/screens/track-screen/TrackScreenSkeleton.tsx
```diff
@@ -0,0 +1,150 @@
...
+            <StaticSkeleton style={styles.metric} />
+            <StaticSkeleton style={styles.metric} />
+            <StaticSkeleton style={styles.metric} />
+          </View>
+          <View style={styles.description}>
+            {times(random(5, 15), () => random(20, 100)).map(
+              (elementWidth: number, i) => (
+                <StaticSkeleton
+                  key={i}
+                  noShimmer
```
Kyle-Shanks, line 134: can lose this `noShimmer` now right?

PR #2461, opened 2022-12-23
C-1725 Fix sign-up with Instagram

nicoback2:
### Description
For some reason we are only getting IG id and handle back from the auth response (so no profile pic, no full name, etc). Not sure if this is a change to the IG API that we have to account for in Identity, or maybe a regression in the Identity code or something else. But this fixes the silent failure on the UI side for now. The UX experience is still not great though because basically no info gets filled out from this.

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/common/src/services/oauth/formatSocialProfile.ts
+++ b/packages/common/src/services/oauth/formatSocialProfile.ts
@@ -109,10 +109,9 @@ export const formatInstagramProfile = async (
     instagramProfile.full_name.length > MAX_DISPLAY_NAME_LENGTH
   ) {
     requiresUserReview = true
-    instagramProfile.full_name = instagramProfile.full_name!.slice(
-      0,
-      MAX_DISPLAY_NAME_LENGTH
-    )
+    if (instagramProfile.full_name) {
+      instagramProfile.full_name.slice(0, MAX_DISPLAY_NAME_LENGTH)
+    }
   }
 
   return {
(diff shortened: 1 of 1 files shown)


packages/common/src/services/oauth/formatSocialProfile.ts
```diff
@@ -109,10 +109,9 @@ export const formatInstagramProfile = async (
     instagramProfile.full_name.length > MAX_DISPLAY_NAME_LENGTH
   ) {
     requiresUserReview = true
-    instagramProfile.full_name = instagramProfile.full_name!.slice(
-      0,
-      MAX_DISPLAY_NAME_LENGTH
-    )
+    if (instagramProfile.full_name) {
+      instagramProfile.full_name.slice(0, MAX_DISPLAY_NAME_LENGTH)
```
dylanjeffers, line 113: Slice doesn't change orignal array right?

dylanjeffers (changes requested):
Nice catch, small not for setting sliced value

packages/common/src/services/oauth/formatSocialProfile.ts
nicoback2: oh yeah lol

nicoback2: i forgot to put the assignment back when i moved `instagramProfile.full_name` it to the if statement 💀 thx for catching

raymondjacobson (approved):
lgtm. we _should_ have this going forward with insta fixed

nicoback2:
Ugh IG removed the other fields https://stackoverflow.com/questions/50086945/how-to-get-instagram-profile-picture-via-api https://developers.facebook.com/docs/instagram-basic-display-api/reference/user#fields

dylanjeffers (approved):
nicee

PR #2462, opened 2022-12-23
[c-1719] QA update for Rate CTA Drawer

Kyle-Shanks:
### Description
Small styling updates from QA for the Rate CTA drawer

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A


packages/mobile/src/components/rate-cta-drawer/RateCtaDrawer.tsx
```diff
@@ -52,6 +52,9 @@ const useStyles = makeStyles(({ palette, typography, spacing }) => ({
     width: 18,
     marginRight: spacing(2)
   },
+  buttonText: {
+    fontSize: typography.fontSize.large
```
dylanjeffers, line 56: Is there a way to configure the button using props and defaults? Just hoping to prevent one off styling, like we could use the size prop on button maybe? That makes the whole button bigger as well though. Happy to have a convo about this

dylanjeffers (approved):
Approved but also let's see if we can fudge it to work with defaults or make a note of incorrect default values to improve design system

packages/mobile/src/components/rate-cta-drawer/RateCtaDrawer.tsx
Kyle-Shanks: yea i agree. there were a few things about the drawer that needed one-off styling to conform with the designs. I feel like a lot of this will get tackled when we make the big design system push

PR #2463, opened 2022-12-23
Offline mode tweaks

amendelsohn:
### Description

Tweaks to get offline mode working:
1. Increase timeout on queue
2. Set concurrency to 1 on queue
3. Fix feature flag check in saga that blocks offline playback

### DIFF
--- a/packages/mobile/src/services/offline-downloader/offline-download-queue.ts
+++ b/packages/mobile/src/services/offline-downloader/offline-download-queue.ts
@@ -44,7 +44,7 @@ const removeExistingWorkers = () => {
 export const startDownloadWorker = async () => {
   queue.stop()
   queue.configure({
-    concurrency: 10,
+    concurrency: 1,
     updateInterval: 10
   })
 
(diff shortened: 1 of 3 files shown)


packages/mobile/src/services/offline-downloader/offline-download-queue.ts
```diff
@@ -44,7 +44,7 @@ const removeExistingWorkers = () => {
 export const startDownloadWorker = async () => {
   queue.stop()
   queue.configure({
-    concurrency: 10,
+    concurrency: 1,
```
raymondjacobson, line 47: how come only 1?

raymondjacobson (approved):

packages/mobile/src/services/offline-downloader/offline-download-queue.ts
amendelsohn: It was failing when running in parallel. I think we can track it down and fix it, but wanted to get something up and working for now.

amendelsohn: It could also be that it just overloaded my internet here. Which is a valid case anyway, but the change is likely not necessary for most users unless we also broke something in the recent changes. Obviously it used to work at 10.

raymondjacobson: sounds good!

PR #2464, opened 2022-12-24
Fix background reset issues for track player on android 12+

Kyle-Shanks:
### Description
Fix the issue where calling reset on the Trackplayer when the app is in the background breaks the app on android 12+

### Dragons

Touches audio playback so we should handle with care.
This also infinitely adds to the queue for the track player which could cause memory issues after long sessions of listening. This should be fixed in the future when we update our TrackPlayer implementation

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A


dylanjeffers (approved):
very sweet, epic

sliptype (approved):

PR #2465, opened 2022-12-24
Improve tipping perf, only fetch recent tip

piazzatron:
### Description

Net negative PR!!!

- Feed Tips were slow because we would fetch many users from DN and have to call populate_metadata on each one. This change has us only fetch a single, most recent tip. A lot of really detailed logic by @sddioulde wrote goes away in this world - seems like a sane tradeoff since if you're using Audius with any frequency anyways, you're caught up to the head and only see the most recent tip.

- **Also** fixes a bug where we had the wrong type for followee_supporters and would throw these away in the saga, causing us to never show more than the initial tip sender in the feed tile. Now we get multiple names:

Screen Shot 2022-12-23 at 6 36 15 PM

- Also caches the sender + receiver users so we don't have to refetch them.

### Dragons

I have not tested this on mobile yet

### How Has This Been Tested?

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/piazz-better-tip-fetching

rickyrombo (approved):

sddioulde (approved):

sddioulde:
nice!

piazzatron:
looks like i broke something with the dismiss button. investigating

packages/common/src/store/tipping/slice.ts
```diff
@@ -26,10 +24,8 @@ const initialState: TippingState = {
     error: null,
     source: 'profile'
   },
-  recentTips: [],
-  storage: null,
   tipToDisplay: null,
-  showTip: true
+  showTip: false
```
piazzatron, line 28: Set this as false initially so that in the case where we're dismissed, we don't show a loading tile

PR #2466, opened 2022-12-24
Fix api-client types by fixing @audius/common absolute imports

dylanjeffers:
### Description

Fixes issue where ApiClient consumers were not getting the correct input/return types, meaning things silently became type "any", yikes. When going to definition however, the types are correct, ie @audius/common is internally consistent, but its build is not. nice catch @piazzatron 

Issue ultimately due to absolute imports in @audius/common. This for whatever reason silently breaks the build with no warning, propagating "any" types to consumers.

- Fixed absolute imports
- Fixed emergent type bugs in @audius/web-client now that AudiusApi actually typed

### DIFF
--- a/packages/web/src/common/store/premiumContent/sagas.ts
+++ b/packages/web/src/common/store/premiumContent/sagas.ts
@@ -170,7 +170,7 @@ function* updateNewPremiumContentSignatures({
  * Halts if not all nfts have been fetched yet. Similarly, does not proceed if no tracks are in the cache yet.
  * Skips tracks whose signatures have already been previously obtained.
  */
-function* updateNFTGatedTrackAccess(
+async function* updateNFTGatedTrackAccess(
   action:
     | ReturnType<typeof ethNFTsFetched>
     | ReturnType<typeof solNFTsFetched>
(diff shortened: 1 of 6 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-fix-api-client-types

packages/web/src/common/store/search-bar/sagas.ts
```diff
@@ -9,7 +9,7 @@ import { getSearch } from './selectors'
 
 const getUserId = accountSelectors.getUserId
 
-export function* getSearchResults(searchText: string) {
+export async function* getSearchResults(searchText: string) {
```
rickyrombo, line 12: I don't think this should be async?

packages/web/src/common/store/premiumContent/sagas.ts
```diff
@@ -170,7 +170,7 @@ function* updateNewPremiumContentSignatures({
  * Halts if not all nfts have been fetched yet. Similarly, does not proceed if no tracks are in the cache yet.
  * Skips tracks whose signatures have already been previously obtained.
  */
-function* updateNFTGatedTrackAccess(
+async function* updateNFTGatedTrackAccess(
```
rickyrombo, line 173: same here?

rickyrombo (approved):

packages/web/src/common/store/premiumContent/sagas.ts
dylanjeffers: Great catches, I was doing an await to check types cause it's better than call, and forgot to change back

PR #2469, opened 2022-12-26
[PAY-776] Add premium content to track upload flow on desktop - Part 1

sddioulde:
### Description

Part 1 of the web track upload flow for premium content as I do not want this PR to become large.

Add (feature-flagged) TrackAvailabilityModal which is what determines whether track uploaded is premium.
Also update the modal button trigger and track metadata state based on track availability selection.
Also publish new stems.

To do (in subsequent PR):
- turn selection rows in modal into radio buttons
- add collectible dropdown
- update track metadata accordingly
- test

Related to https://github.com/AudiusProject/audius-protocol/pull/4556 for the updated track metadata schema.

Note: PR looks big but that's mosty because of the package lock.

### Dragons

None

### How Has This Been Tested?

Run dapp locally and interact with the track upload flow.

N/A

### Feature Flags ###

Existing premium_content_enabled flag, plus additional nft_gate_enabled and special_access_enabled flags.


packages/web/package.json
```diff
@@ -51,8 +51,8 @@
   },
   "dependencies": {
     "@audius/common": "1.5.4",
-    "@audius/sdk": "1.0.27",
-    "@audius/stems": "1.5.2",
+    "@audius/sdk": "1.0.29",
+    "@audius/stems": "1.5.2-beta.3",
```
sddioulde, line 55: going to publish and use real version

packages/web/src/components/track-availability-modal/TrackAvailabilityModal.module.css
```diff
@@ -0,0 +1,132 @@
...
+  margin-bottom: 12px;
+}
+
+.doneButtonContainer {
+  display: flex;
+  justify-content: center;
+}
+
+.doneButton {
+  font-size: var(--font-l) !important;
```
sddioulde, line 119: let me know if you have ideas on how to not use the `important` here

audius-infra:
Preview this change https://demo.audius.co/sd-pc-track-upload

packages/common/src/store/collectibles/reducer.ts
```diff
@@ -0,0 +1,33 @@
...
+
+type CollectiblesState = {
+  collectibles: { [id: ID]: Collectible[] }
+}
+
+const initialState = {
+  collectibles: {}
+}
+
+const collectiblesReducer = createReducer<
```
piazzatron, line 18: Will do a more thorough review soon, but why are we using createReducer instead of redux-toolkit `create-slice` here?

packages/web/src/components/data-entry/FormTile.js
```diff
@@ -306,30 +376,45 @@ const BasicForm = (props) => {
...
 
-  const [isUnlistedModalOpen, setIsUnlistedModalOpen] = useState(false)
+  const [isAvailabilityModalOpen, setIsAvailabilityModalOpen] = useState(false)
   const [hideRemixes, setHideRemixes] = useState(
     !(props.defaultFields?.field_visibility?.remixes ?? true)
   )
 
   // Need to update two fields in the metadata.
-  const didUpdateUnlistedState = (newState) => {
+  const didUpdateAvailabilityState = (newState) => {
```
rickyrombo, line 417: nit: I'm a little worried about "availability" being the term of choice here, since I think it may be overloaded from trusted notifier/content availability

packages/web/src/components/data-entry/FormTile.module.css
```diff
@@ -184,6 +184,14 @@
   font-weight: var(--font-demi-bold);
 }
 
+.trackAvailabilityButton {
+  height: 35px !important;
```
rickyrombo, line 188: any chance we can lose the `!important` by using more specific selectors or something?

packages/web/src/components/track-availability-modal/TrackAvailabilityModal.module.css
rickyrombo: typically what i would do is something more specific ie `.doneButtonContainer .doneButton` etc, or I make the button text wrapped in a span and do the styles on that (dunno if our button component allows that)

rickyrombo (approved):

packages/web/src/components/data-entry/FormTile.js
```diff
@@ -59,6 +66,69 @@ const Divider = (props) => {
   )
 }
 
+const TrackAvailabilityButton = (props) => {
```
piazzatron, line 69: Can we move this to a separate TS file so we're not adding any more JS?

sddioulde: this is meant to be removed after launch, only there for the feature flag so seems to me that we don't need to separate it. lmk if you feel strongly

piazzatron: that's fine if just feature flagged. but let's add a comment tracking linear issue then here

PR #2470, opened 2022-12-27
[C-1728] Add collection screen skeleton

dylanjeffers:
### Description

- Adds collection-screen skeleton
- Refactors track-screen skeleton to entity-skeleton


sliptype (approved):
Looks great!

PR #2476, opened 2022-12-27
[C-1660] Clear feed on sign-out

dylanjeffers:
### Description

Clears feed on sign out so new signed-in user get's their feed.
- Refactors AsyncStorage keys into one file to make it easier to understand storage
- Uses this file to aid in clearing storage keys on sign-out

### DIFF
--- a/packages/mobile/src/constants/storage-keys.ts
+++ b/packages/mobile/src/constants/storage-keys.ts
@@ -1,0 +1,7 @@
+export const DEVICE_TOKEN = '@device-token'
+export const IS_MOBILE_USER = 'is-mobile-user'
+export const RATE_CTA_STORAGE_KEY = '@rate-cta-response'
+export const IP_STORAGE_KEY = 'user-ip-timetstamp'
+export const SESSION_COUNT_KEY = '@session-count'
+export const ENTROPY_KEY = 'hedgehog-entropy-key'
+export const THEME_STORAGE_KEY = 'theme'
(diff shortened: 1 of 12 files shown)


packages/mobile/src/store/account/sagas.ts
```diff
@@ -2,11 +2,11 @@ import { accountActions, getContext } from '@audius/common'
...
 import { updateProfileAsync } from 'common/store/profile/sagas'
 import { takeEvery, call } from 'typed-redux-saga'
+
+import { IS_MOBILE_USER } from 'app/constants/storage-keys'
 const { signedIn } = accountActions
 
-export const RESET_REQUIRED_KEY = 'password-reset-required'
-export const ENTROPY_KEY = 'hedgehog-entropy-key'
-export const IS_MOBILE_USER_KEY = 'is-mobile-user'
+// const RESET_REQUIRED_KEY = 'password-reset-required'
```
nicoback2, line 9: delete?

packages/mobile/src/constants/storage-keys.ts
```diff
@@ -0,0 +1,7 @@
+export const DEVICE_TOKEN = '@device-token'
```
sliptype, line 1: Love the consistency here 😂

packages/mobile/src/store/account/sagas.ts
sliptype, line 9: Is this not being used at all?

sliptype (approved):
This is way cleaner!

packages/mobile/src/constants/storage-keys.ts
dylanjeffers: It's sad too cause we can't really change them, lol

packages/mobile/src/store/account/sagas.ts
dylanjeffers: So there's a commented out section below it that used it, so I kept this around so when we do work on that, we have the key avail?

nicoback2 (approved):

PR #2477, opened 2022-12-27
[C-1581] Mobile Complete w/ TikTok

sliptype:
### Description

* Adds complete w/ tiktok on mobile
Simulator Screen Shot - iPhone 14 Pro - 2022-12-27 at 16 12 29
* This is only the web based oauth flow (like twitter and ig), will look into native flow next
* Adds custom color prop to `Button`, it will darken 10% programmatically on press

### Dragons

* Touches Twitter and IG flow as well so should test thoroughly

### How Has This Been Tested?

### Feature Flags ###

The TikTok button and flow is behind `Complete_Profile_With_TikTok` feature flag. The other UI changes are not behind a flag


packages/mobile/ios/Podfile.lock
```diff
@@ -814,7 +814,7 @@ SPEC CHECKSUMS:
   Permission-Notifications: 2662b6885535a98892b8e48da04b74fc898d70f8
   PromisesObjC: 68159ce6952d93e17b2dfe273b8c40907db5ba58
   Protobuf: 235750e4696ff59fb07d949a9dbbc92b3c0700fe
-  RCT-Folly: 803a9cfd78114b2ec0f140cfa6fa2a6bafb2d685
+  RCT-Folly: a21c126816d8025b547704b777a2ba552f3d9fa9
```
dylanjeffers, line 817: Been seeing these change a lot... I wonder if you were to install fresh what would happen?

packages/mobile/src/components/core/Button.tsx
```diff
@@ -267,21 +271,32 @@ export const Button = (props: ButtonProps) => {
...
-    common: primaryDark1,
-    commonAlt: neutralLight10,
-    destructive: accentRedDark1
-  }
+  const pressColor = useMemo(() => {
+    // If a custom color is specified
+    // darken the color by 10% for the press state
+    if (customColor) {
+      const c = new Color(customColor)
+      return c.darken(0.1).hex()
```
dylanjeffers, line 279: I've used polish is in the past and have really liked it, curious what you think

packages/mobile/src/components/core/Button.tsx
```diff
@@ -267,21 +271,32 @@ export const Button = (props: ButtonProps) => {
...
+    }
+
+    // If no custom color is specified
+    // derive the press state color for the variant
+    return {
```
dylanjeffers, line 284: This is a great addition

packages/mobile/src/store/oauth/actions.ts
```diff
@@ -154,7 +181,7 @@ export const setTwitterError = (error: any): SetTwitterErrorAction => ({
 
 export const setInstagramInfo = (
   uuid: string,
-  profile: TwitterAccountPayload,
+  profile: InstagramAccountPayload,
```
dylanjeffers, line 184: Good change 😀

packages/mobile/src/store/oauth/reducer.ts
```diff
@@ -27,6 +31,14 @@ type InstagramInfo = {
   instagramId?: any
 }
 
+type TikTokInfo = {
+  uuid: string
+  profile: TikTokProfile
+  profileImage: any
```
dylanjeffers, line 37: Any thoughts here?

dylanjeffers (approved):
Great changes

packages/mobile/ios/Podfile.lock
sliptype: Let me try removing pods

packages/mobile/src/components/core/Button.tsx
sliptype: polished looks really nice but we already have a dependency on `color` here https://github.com/AudiusProject/audius-client/blob/main/packages/mobile/src/utils/colorizeLottie.ts#L11 so opted to use that

Maybe we could update `colorizeLottie` to use polish?

dylanjeffers: Good to know we already use color, yeah possibly! It would be useful for the glassmorphic now playing, maybe we can update then

packages/mobile/ios/Podfile.lock
sliptype: Yeah I still have the same change, I'll leave it in for now

packages/mobile/src/store/oauth/reducer.ts
sliptype: Yeah I updated it now to a more specific type, was just `any` because that was the case for twitter and ig. I may go through those flows to improve this stuff too after I get the tiktok stuff in a good spot

PR #2483, opened 2022-12-29
[C-1738] Fix mobile crash

sliptype:
### Description
The addition of `@metaplex-foundation/mpl-token-metadata` in https://github.com/AudiusProject/audius-client/pull/2398 caused a crash on mobile when loading a profile with solana collectibles (It was not possible to repro locally on ios and only able to repro locally on android without the debugger attached)

This was the error being thrown: https://sentry.io/organizations/audius/issues/3832938092/?query=bigint&referrer=issue-stream&statsPeriod=14d

This PR:
* Polyfills BigInt with `big-integer` 
* Patches @noble packages for compatibility `big-integer`. `big-integer` does not support `0x` prefixes, instead uses a radix argument

Solana docs specify a min version of react-native of 0.70 which would fix this problem also. We should prioritize upgrading so that we don't need this hacky fix for long https://github.com/solana-labs/solana-web3.js/blob/34b107635cd21d9e61c705ded316064df06950a0/README.md#compatibility

### Dragons

The only thing that could potentially break here is the solana nft premium content stuff, but this fix should keep that working too!

### How Has This Been Tested?

Android locally, will test on ios once on stage

### Feature Flags ###

### DIFF
--- a/packages/common/package-lock.json
+++ b/packages/common/package-lock.json
@@ -2073,6 +2073,12 @@
         "@walletconnect/window-getters": "^1.0.0"
       }
     },
+    "@yarnpkg/lockfile": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/@yarnpkg/lockfile/-/lockfile-1.1.0.tgz",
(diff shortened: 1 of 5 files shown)


nicoback2 (approved):
Wow amazing!

PR #2485, opened 2022-12-29
Replace react-native-dark-mode with react-native-dynamic

dylanjeffers:
### Description

- Replaces legacy react-native-dark-mode with react-native-dynamic, which is needed to upgrade react-native
- Removes  difficult to understand systemAppearance sagas
- Adds ThemeProvider which sets up base theme and listens to system appearance changes
- Moves App.tsx into app dir, and migrates a few other loose app-wide files into it too


sliptype (approved):
Nice!

PR #2486, opened 2022-12-29
Remove react-native-static-server

dylanjeffers:
### Description

- Removes react-native-static-server which is needed for react-native upgrade

### Dragons
- This removes our failsafe for users on old versions not getting signed out. Personally, I feel this is an okay compromise considering this is blocking us from upgrading react-native, which is currently a blocker preventing us from working with latest crypto libraries. Lets all decide together on what's best though


raymondjacobson:
We should first look at stats to see if any users are driving sessions from versions prior to async storage stuff

dylanjeffers:
> We should first look at stats to see if any users are driving sessions from versions prior to async storage stuff

yes lets, my guess also is that we have required users to upgrade to a reloaded version at least once yeah? so that would mean anyone not on new system hasnt used app in months?

sliptype:
Looks like less that 1% of users on ios and android on a weekly basis are on old versions

dylanjeffers:
> Looks like less than 1% of users on ios and android on a weekly basis are on old versions

if we roll this out, should we put up a required update for folks behind a specific version?

sliptype:
> > Looks like less than 1% of users on ios and android on a weekly basis are on old versions
> 
> if we roll this out, should we put up a required update for folks behind a specific version?

No because the only degradation will be that they have to sign in when they update

sliptype (approved):

PR #2487, opened 2022-12-29
Force polyfill of BigInt on ios as well

sliptype:
### Description

Forcing the polyfill of bigint on ios so that the package patches will work

### Feature Flags ###


dylanjeffers (approved):
lgtm

PR #2489, opened 2022-12-30
[C-1733] Refactor suggested follows

dylanjeffers:
### Description

Refactors suggested follows using core components and style system.
- Improves rerender perf by 15x
- Refactors "secondary" and "primaryAlt" button styles according to design system naming
- Adds xs size option for buttons
- Adds "activeUnderline" option for text-button to replicate existing requirement for "select for me" text-button
- Fixes bug where "ListEmptyComponent" could still trigger pull-to-refresh behavior


packages/mobile/src/assets/images/iconWand.svg
```diff
@@ -1,6 +1,6 @@
...
... C10.6907926,44.196726 9.11522725,44.2759229 8.09356721,43.3536063 C8.06237813,43.3255361 8.03196877,43.2966862 8.00233914,43.2670566 L4.7695906,40.0366472 C3.79543426,39.0622828 3.79543426,37.4827466 4.7695906,36.5083821 C4.79922022,36.4787525 4.82962958,36.4499026 4.86003894,36.4218324 L14.0189733,28.1587419 L14.0194932,25.2865497 C14.0194932,24.6406026 14.510409,24.1093151 15.1395004,24.0454274 L15.2670565,24.0389864 L18.5859733,24.0387419 L40.3578948,4.3984405 C40.8503066,3.95480021 41.6036659,3.97363419 42.0732944,4.44132549 Z M37.7231969,26.5341131 L39.8019494,30.6923977 L43.9610137,32.7719298 L39.8019494,34.851462 L37.7231969,39.0097466 L35.6444445,34.851462 L31.4853801,32.7719298 L35.6444445,30.6923977 L37.7231969,26.5341131 Z M10.2768031,4.07797266 L12.3555555,8.23625727 L16.5146199,10.3157894 L12.3555555,12.3953216 L10.2768031,16.5536062 L8.19805064,12.3953216 L4.0389863,10.3157894 L8.19805064,8.237037 L10.2768031,4.07797266 Z M21.5048733,4.07797266 L22.7524366,6.57309937 L25.2475634,7.82066273 L22.7524366,9.06822608 L21.5048733,11.5633528 L20.2573099,9.06822608 L17.7621832,7.82066273 L20.2573099,6.57309937 L21.5048733,4.07797266 Z" id="Combined-Shape" fill="#000"></path>
```
raymondjacobson, line 4: what changed here?

packages/mobile/src/components/core/Button.tsx
```diff
@@ -213,15 +228,22 @@ export type ButtonProps = Omit<RNButtonProps, 'title'> &
     IconProps?: SvgProps
     fullWidth?: boolean
     noText?: boolean
-    size?: 'small' | 'medium' | 'large'
+    size?: 'xs' | 'small' | 'medium' | 'large'
```
raymondjacobson, line 231: aww lil guy

packages/mobile/src/components/core/SectionList.tsx
```diff
@@ -168,9 +163,15 @@ export const SectionList = forwardRef<RNSectionList, SectionListProps>(
       PlayBarChin
     )
 
+    const areSectionsEmpty = sections.every(
```
raymondjacobson, line 166: why move into component?

packages/mobile/src/hooks/usePressScaleAnimation.ts
```diff
@@ -6,30 +6,32 @@ export const usePressScaleAnimation = (
   scaleTo = 0.97,
   useNativeDriver = true
 ) => {
-  const scale = useRef(new Animated.Value(1)).current
+  const scaleAnim = useRef(new Animated.Value(1))
```
raymondjacobson, line 9: we do this pattern a lot. is that a source of extra renders?

raymondjacobson (approved):

packages/mobile/src/assets/images/iconWand.svg
dylanjeffers: fill=#000 wasn't specified so it was not actually using style-system fill

packages/mobile/src/components/core/Button.tsx
dylanjeffers: dawwww

packages/mobile/src/components/core/SectionList.tsx
dylanjeffers: to prevent calling twice, since im reffing in two spots, but yeah didn't need to. when i tried to make a const with a single call, i was getting `const areSectionsEmptyyy = areSectionsEmpty(sections)` so to avoid a clever var name, just used it directly, whatever you think tho

packages/mobile/src/hooks/usePressScaleAnimation.ts
dylanjeffers: yeah i actually dont know for certain, i think probably not, but i tried it here and meant to check again on it and forgot. can remove to be consistent though

packages/mobile/src/components/core/SectionList.tsx
raymondjacobson: carry on

packages/mobile/src/hooks/usePressScaleAnimation.ts
dylanjeffers: example in RN docs does .current off the rip, so prob fine: 
image

the tradeoff is having to put in callback deps which is annoying, but not a huge deal

audius-infra:
Preview this change https://demo.audius.co/dj-c-1733-refactor-suggested-follows

PR #2490, opened 2022-12-30
[C-1737] Upgrade react-native to 0.67.5

dylanjeffers:
### Description

- Upgrades react-native one minor version from 0.66.5 to 0.67.5
- Aligns deps with expected versions


packages/mobile/package.json
```diff
@@ -111,9 +111,9 @@
     "react-native-in-app-review": "4.2.1",
     "react-native-job-queue": "0.5.0",
     "react-native-keyboard-aware-scroll-view": "^0.9.5",
-    "react-native-linear-gradient": "github:react-native-linear-gradient/react-native-linear-gradient.git#9aece37cdf7d33ed52a1747b36e9e84f3c7a49ef",
+    "react-native-linear-gradient": "2.6.2",
```
sliptype, line 114: I think we had to use a specific commit for `react-native-linear-gradient` because there was an issue with the text mask gradient (like in the screen headers) where the gradient would show as a rectangle before the text mask was applied. Let's just double check that the headers look right on android!

sliptype (approved):
Amazing! Excited to get rid of the web sync stuff 🙌 that's the last vestige of the web app

packages/mobile/package.json
dylanjeffers: gotcha, yeah i checked the release, and it seems to have marcus's commit!

PR #2493, opened 2022-12-30
[C-1732] Add mobile end of feed notice

dylanjeffers:
### Description

Adds end of feed notice similar to prod
Moves suggested follows and end of feed to sub components for modularity

### Dragons

This was all eye-balled based on desktop look and feel. Prob worth a Design pass.

### DIFF
--- a/packages/mobile/src/components/lineup/Lineup.tsx
+++ b/packages/mobile/src/components/lineup/Lineup.tsx
@@ -524,10 +524,7 @@ export const Lineup = ({
         onScroll={handleScroll}
         ListHeaderComponent={header}
         ListFooterComponent={
-          <>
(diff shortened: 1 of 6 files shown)


sliptype (approved):

Kyle-Shanks (approved):
Looks good! Double checking to make sure that there is logic to make sure that the footer only shows when you are at the actual end of your feed and not just the end of the current lineup while the feed is loading more in?

dylanjeffers:
> Looks good! Double checking to make sure that there is logic to make sure that the footer only shows when you are at the actual end of your feed and not just the end of the current lineup while the feed is loading more in?

So smart, no there isn't logic, nice one

PR #2496, opened 2023-01-03
Enable hermes and flipper by default for ios

dylanjeffers:
### Description

Enables hermes and flipper by default for ios
Also fixes react-native patch, nice find @sliptype 

### Dragons

Requires migration to flipper. Using flipper and it works well, so i see this as a good patch for us


sliptype (approved):
Awesome! Could you do a quick post in eng-client or document on notion of to install/use flipper?

audius-infra:
Preview this change https://demo.audius.co/dj-c-enable-hermes-flipper

PR #2498, opened 2023-01-03
[C-1741] Upgrade react-native to 0.69.7

dylanjeffers:
### Description

Upgrades react-native another minor version from 0.68.5 to 0.69.7
Upgrades react from 17 to 18
Various other upgrades needed to keep build happy

### Dragons

Upgrading react to 18 (and redux) just in mobile works since we have the metro.config aliases that force usage, even for code from web/common. We should migrate our entire monorepo after, though.


sliptype (approved):
HUGE

PR #2501, opened 2023-01-03
[QA-269] Update counter effect logic to only reset track position when the track index is the same

Kyle-Shanks:
### Description
Update logic for the counter effect to only reset the playback position when the track index has not changed. This will reset it when we restart the track, but not when playing a new track.

There was some wonkiness with the effect being called too many times due to the function updating when the queue index updated so there is some hacky counterRef logic to help with that.

### Dragons

N/A

### How Has This Been Tested?

Manually tested on iOS

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/mobile/src/components/audio/Audio.tsx
+++ b/packages/mobile/src/components/audio/Audio.tsx
@@ -326,11 +326,26 @@ export const Audio = () => {
     }
   }, [seek, setSeekPosition])
 
+  // Keep track of the track index the last time counter was updated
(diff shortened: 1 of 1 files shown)


dylanjeffers (approved):
Sweet so this prevents the choppy transition between tracks?

Kyle-Shanks:
> Sweet so this prevents the choppy transition between tracks?

Yea exactly, the track ends and is not restarted when we don't want it to

PR #2502, opened 2023-01-03
[C-17496 Upgrade react-native to 0.70.6 (latest)

dylanjeffers:
### Description

Upgrades react-native from 0.69 to 0.70, the final upgrade for now.
Adds all the "new-architecture" react-native code. I tried enabling it, but there are a bunch of dependencies that don't fully support it yet


sliptype (approved):
Unfortunate that solana phone doesn't work with 0.70, seeing as one of the main reasons we chose to upgrade was to support @solana/web3.js 😂

Let's make sure to test this thoroughly on ios and android once it's available!

Actually quite interesting that with the New Architecture react-native also moved away from message passing: https://reactnative.dev/docs/the-new-architecture/why

dylanjeffers:
> Unfortunate that solana phone doesn't work with 0.70, seeing as one of the main reasons we chose to upgrade was to support @solana/web3.js 😂
> 
> Let's make sure to test this thoroughly on ios and android once it's available!
> 
> Actually quite interesting that with the New Architecture react-native also moved away from message passing: https://reactnative.dev/docs/the-new-architecture/why

Yeah indeed, im going to keep trying to get the solana stuff to work, and if not just hide it behind a build flag for solana phone only. ahh interesting about the new architecture.

dylanjeffers:
> Unfortunate that solana phone doesn't work with 0.70, seeing as one of the main reasons we chose to upgrade was to support @solana/web3.js 😂
> 
> Let's make sure to test this thoroughly on ios and android once it's available!
> 
> Actually quite interesting that with the New Architecture react-native also moved away from message passing: https://reactnative.dev/docs/the-new-architecture/why

@sliptype just double-checked an looks like solana works fine when newArch=false, so rolling forward with it!

PR #2509, opened 2023-01-04
use allSettled util instead of Promise.allSettled

amendelsohn:
### Description

https://stackoverflow.com/q/66575667

`Promise.allSettled` is undefined in the js runner in prod but not the one used in dev. The dev runner has extra features for debugging and has the function populated.

### Dragons

Do we want the eslint warning for web and stems also?

### How Has This Been Tested?

Simulator with staging target:  `npx react-native run-ios --simulator "iPhone 14 Pro" --configuration Staging`

eslint rule to prevent future failure

### Feature Flags ###

all failures were behind `OFFLINE_MODE_ENABLED`

### DIFF
--- a/packages/common/src/utils/allSettled.ts
+++ b/packages/common/src/utils/allSettled.ts
@@ -1,4 +1,5 @@
 export const allSettled =
+  // eslint-disable-next-line no-restricted-properties
   Promise.allSettled ||
   ((promises: any[]) =>
     Promise.all(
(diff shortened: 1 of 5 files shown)


raymondjacobson (approved):
Awesome find.

Re; your question -- definitely not stems b/c that's not on mobile. Web shouldn't be on mobile, but is still in the interim because we haven't finished the common migration. Could go either way there.

sliptype (approved):
Fantastic find! I don't think we need the eslint rule on web etc.

PR #2512, opened 2023-01-04
[C-1713] Remove UserBadges from desktop card component

Kyle-Shanks:
### Description
Small change to remove the user badges from the collection cards on the user profile
Screenshot 2023-01-04 at 1 06 58 PM

### Dragons

Are there any cards that still need this? The cards in search didn't seems to use this, but wanted to call this out

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A


audius-infra:
Preview this change https://demo.audius.co/kj-Remove-user-badges-from-desktop-card

sliptype (changes requested):
I think this might affect other usages of the Card component, like on the search results page we want the badges to be there

sliptype (approved):
Awesome!

PR #2513, opened 2023-01-04
Fix the base58 error on solana nft collection key

sddioulde:
### Description

Fix the base58 error on solana nft collection key.
As written in the comment above the fix, the types returned are sometimes strings, other types PublicKey even though the library type says that it should be PublicKey.

`toBase58` function does not exist on strings, and this is what caused the bug.

See screenshot below for types returned by the library.

Screen Shot 2023-01-04 at 4 40 38 PM

### Dragons

None

### How Has This Been Tested?

No longer fails with account that currently fails on staging.

N/A

### Feature Flags ###

None


jowlee (approved):
nice catch

PR #2517, opened 2023-01-05
[C-1757] Handle @audius/common absolute imports

dylanjeffers:
### Description

- Fixes @audius/common absolute imports compile issue (essentially absolute imports were being casted to "any")
- Fixes all @audius/common usages in web and native apps
- @amendelsohn check out the changes in various offline-mode sections
- @sliptype check out changes related to oauth events
- @nicoback got a small change related to share-to-ig/snap events

headsup @piazzatron @rickyrombo @sddioulde, this is finally fixed!!!

### Dragons

Lots of code changes, but mostly type change


packages/mobile/src/store/store.ts
```diff
@@ -92,7 +93,8 @@ if (__DEV__) {
 export const store = createStore(
   createRootReducer(),
   applyMiddleware(...middlewares)
-)
+) as unknown as Store<AppState> // need to explicitly type the store for offline-mode store reference
```
amendelsohn, line 96: Oh, nice. This will fix the selector types on `store.getState`?

Could call that out more generally too since we may use it outside of offline mode in the future.

I was going to tackle this problem in general when exploring the new react-query selectors.

amendelsohn (approved):
Awesome! Offline mode changes look good.

packages/mobile/src/store/store.ts
dylanjeffers: yeah it does, def a bit hacky doing it like this, so maybe we can think of a more standard way in the future

audius-infra:
Preview this change https://demo.audius.co/dj-c-1757-common-absolute-imports

audius-infra:
Preview this change https://demo.audius.co/dj-c-1757-common-absolute-imports

PR #2518, opened 2023-01-05
Fix local iOS dev/install

raymondjacobson:
### Description

```
cd packages/mobile/ios && pod deintegrate && pod cache clean --all && cd -
npm i && npm run ios:prod
```

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/ios/Podfile
+++ b/packages/mobile/ios/Podfile
@@ -51,6 +51,9 @@ target 'AudiusReactNative' do
     )
     __apply_Xcode_12_5_M1_post_install_workaround(installer)
     # flipper_post_install(installer)
+    installer.pods_project.build_configurations.each do |config|
+      config.build_settings["EXCLUDED_ARCHS[sdk=iphonesimulator*]"] = "arm64"
+    end
 
     # https://github.com/CocoaPods/CocoaPods/issues/11402#issuecomment-1201464693
     installer.pods_project.targets.each do |target|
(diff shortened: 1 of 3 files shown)


packages/mobile/ios/Podfile.lock
```diff
@@ -955,7 +955,7 @@ SPEC CHECKSUMS:
   boost: a7c83b31436843459a1961bfd74b96033dc77234
   BVLinearGradient: 34a999fda29036898a09c6a6b728b0b4189e1a44
   CocoaAsyncSocket: 065fd1e645c7abab64f7a6a2007a48038fdc6a99
-  DoubleConversion: 831926d9b8bf8166fd87886c4abab286c2422662
+  DoubleConversion: 5189b271737e1565bdce30deb4a08d647e3f5f54
```
dylanjeffers, line 958: oh looks like these got changed in the share to snapchat pr

packages/mobile/ios/Podfile
```diff
@@ -51,6 +51,9 @@ target 'AudiusReactNative' do
     )
     __apply_Xcode_12_5_M1_post_install_workaround(installer)
     # flipper_post_install(installer)
+    installer.pods_project.build_configurations.each do |config|
```
dylanjeffers, line 54: interesting, so i think this make sense, but do we want to apply this wholesale to every build config? in the case of LD linking failing, a way around it is to manually install the pod, basically. we had to do this for one of the pods above

dylanjeffers: looking at pbxcodeproj and seems like we actually have these defined: `				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = arm64;` i wonder if maybe this is a mistake? and in which case your fix might actually be the patch for it! thoughts?

dylanjeffers (approved):
nice finds

PR #2526, opened 2023-01-06
[PAY-808] [PAY-810] Float new playlists to top of library

piazzatron:
### Description

- New playlists start on top of playlist library in navbar
- Changed default sort order to recency in profile page
- Fixed issues with playlist creation:
Prior to this PR, creating a playlist would write out a playlist to your account's collections, which would get flushed to local storage. The playlist library wouldn't get updated until *next session* as part of the sign on sagas, which check for playlist library state differing from account collection state. This PR writes out the playlist library as part of the confirmation flow.

### Dragons

Playlist creation flow is a bit scary. But this seems to work

### How Has This Been Tested?

Web. Testing mobile

### Feature Flags ###

### DIFF
--- a/packages/web/src/common/store/cache/collections/sagas.js
+++ b/packages/web/src/common/store/cache/collections/sagas.js
@@ -33,6 +33,7 @@ import { fetchUsers } from 'common/store/cache/users/sagas'
 import * as confirmerActions from 'common/store/confirmer/actions'
 import { confirmTransaction } from 'common/store/confirmer/sagas'
 import * as signOnActions from 'common/store/pages/signon/actions'
+import { addPlaylistsNotInLibrary } from 'common/store/playlist-library/sagas'
 import { waitForWrite } from 'utils/sagaHelpers'
 
 import { reformat } from './utils'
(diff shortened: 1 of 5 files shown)


audius-infra:
Preview this change https://demo.audius.co/piazz-fix-playlist-ordering

audius-infra:
Preview this change https://demo.audius.co/piazz-fix-playlist-ordering

packages/web/src/common/store/cache/collections/sagas.js
```diff
@@ -234,6 +235,9 @@ function* confirmCreatePlaylist(uid, userId, formFields, source) {
           })
         )
 
+        // Write out the new playlist to the playlist library
+        yield call(addPlaylistsNotInLibrary)
```
dylanjeffers, line 239: just want to understand flow between this call, and the upload sagas one. will they not get called back-to-back?

dylanjeffers (approved):
sweet!

packages/web/src/common/store/cache/collections/sagas.js
piazzatron: you would think so, and I did as well - however, it looks like the upload sagas does the same work create playlist does (it calls createPlaylist, etc), so we need to treat them distinctly

PR #2527, opened 2023-01-06
Remove excluded archs & xcode 12.5 patch

raymondjacobson:
### Description

This removes the excluded archs now that upgrades to MacOS 13.1 seem to be able to run an arm64 simulator.
We should be careful merging this because it will make builds fails if you are even on 13.01!

Requires an `npm i`

### Dragons

### How Has This Been Tested?

Locally, failing on 12.6, working on 13.1

### Feature Flags ###


packages/mobile/ios/Podfile
```diff
@@ -49,7 +49,6 @@ target 'AudiusReactNative' do
       # necessary for Mac Catalyst builds
       :mac_catalyst_enabled => false
     )
-    __apply_Xcode_12_5_M1_post_install_workaround(installer)
```
rickyrombo, line 52: https://react-native-community.github.io/upgrade-helper/?from=0.66.5&to=0.70.6 it seems like it only applies to older Xcode, do we really want to remove?

packages/mobile/ios/AudiusReactNative.xcodeproj/project.pbxproj
```diff
@@ -453,7 +453,6 @@
 				CODE_SIGN_ENTITLEMENTS = "AudiusReactNative/Audius Music.entitlements";
 				CURRENT_PROJECT_VERSION = 1;
 				DEVELOPMENT_TEAM = LRFCG93S85;
-				EXCLUDED_ARCHS = "";
```
rickyrombo, line 456: Are we sure we want to remove these? @dylanjeffers already removed the arm64 exclusion

rickyrombo (commented):
Not sure if we really need this

packages/mobile/ios/Podfile
raymondjacobson: won't pod install exclude arm64 after you run it?

rickyrombo: I don't think so - that was a different couple lines that did it manually

dylanjeffers:
based on learnings, it looks like we have to exclude arm64 arch from simulator since some 3rd party pods only provide an x86 variant for simulator

PR #2541, opened 2023-01-07
[C-1779] Improve makeStyles performance

dylanjeffers:
### Description

Major overhaul to makeStyles, resulting in massive performance gains. A transition from feed to trending is 40% faster (**yes 40%!!**) in total cpu time.

#### Changes
- Compiles stylesheets for each theme for every component on app startup, limiting dynamic style compilation as much as possible
- Removes expensive objectHash and unoptimized local style cache
- Moves dynamic styles in memoized code blocks in various components

Result is smaller memory footprint (no object hashing + custom cache), and major in-app performance (95% of expensive, render-blocking style compilation is removed, as we now rely on pre-compiled styles). This also brings us much more inline with react-native style standards.

#### Other fixes
- moves android flipper file to the right location, fixing android flipper
- updates rive pods, which were missed in a merge

### Dragons

There may be visual regressions, as the new useStyles function looses "dynamic variables" capabilities. Those were migrated to memoized blocks and/or compiled as separate static styles, so big change for human error.

### Proof
Proof of 40% improvement: (the underline sections are from the legacy useStyles)
Screenshot 2023-01-06 at 10 28 20 PM


packages/mobile/src/styles/makeStyles.ts
```diff
@@ -1,87 +1,65 @@
...
-const styleCache = {}
-
-export const makeStyles = <
-  PropsT = undefined,
-  T extends NamedStyles<T> = NamedStyles<any>
->(
-  styles: Styles<T, PropsT>
-): UseStyles<T, PropsT> => {
-  return ((props?: PropsT): T => {
+export const makeStyles = <T extends NamedStyles<T> | NamedStyles<any>>(
```
raymondjacobson, line 25: My understanding is that this is the crux of the change -- we don't use hooks and instead use StyleSheet.create here, which I think our old application may have done everywhere throughout the components themselves.

Probably a good testament to revisiting our overall component structure and comparing that with some standard RN cookbook type examples to make sure we're not over engineering some other areas.

What a wonderful find you had here!! Probably couldn't have done it without Flipper -- kudos 👏

dylanjeffers: so actually we still use hooks @raymondjacobson, its just that we dont defer the compilation till inside the hook, we do the majority of the work beforehand (ie create the stylesheets beforehand)

raymondjacobson: Yes. all makes sense. I think maybe a side effect of trying to be fancy though? an extremely hard to maintain but very wet codebase would probably have defined all the styles statically at the top like this https://snack.expo.dev

sliptype (approved):
This is amazing! All looks good to me

fwiw we could have found this before flipper too, the xcode Instruments profiler gives us this same data

PR #2550, opened 2023-01-09
[C-1788][C-1784][C-1785] Fix offline queue loops and duplicates

amendelsohn:
### Description

Prevent infinite favorite/download loop in saga
Prevent duplicate downloadReasons in saved json which caused extra jobs to be spawned on startup.


sliptype (approved):

raymondjacobson:
> @raymondjacobson if you are trying this, you may want to delete and reinstall the app to clear out the duplicate download reasons from the json and ensure your queue is empty.

thank you!!!

PR #2552, opened 2023-01-10
[C-1790] Improve lottie colorize perf

dylanjeffers:
### Description

Improves colorize perf by computing lottie animations ahead-of-time, and only once, similar to our new styling system. This results in much better performance, as we render 20x lottie images on initial load, and many more throughout app lifecycle. In particular there is an expensive JSON.stringify(parse) situation that is heavy on main thread that this helps prevent


sliptype (approved):
Nice! Should we do the one in PullToRefresh too?

dylanjeffers:
> Nice! Should we do the one in PullToRefresh too?

Yess on it

PR #2554, opened 2023-01-10
[C-1791] Add fast-cache, safe-fast-cache

dylanjeffers:
### Description

Adds two new cache modes: fast-cache and safe-fast-cache to test out potential cache speed improvements

Fast-cache heuristics:
- Only overwrite existing entries that are old based on a cache timeout.

Safe-fast-cache heuristics:
- Only overwrite existing entries when there is a material difference between them

Fast-cache is 95% faster on average through the lifecycle of the app, which is pretty epic. Obviously this methodology can break pretty easily since it has a long cache cycle. Thinking about adding a configuration for this so we can test out different cache-time scenarios. The other option we have here is to add "replace" or a "merge" option for cache adds that should merge in with existing values. Note that CACHE/UPDATE handles most of the important cases here.

Safe-fast-cache is probably 60% faster and much safer since it actually compares incoming data and checks equality before updating state. This perf can probably improve with adjustments to the incoming api values, there are a lot of inconsistencies that result in an unnecessary overwrite. Some of this has already been done in the ResponseAdapter, but it appears this file doesn't apply to every datum?

### Dragons

Changing cache a lot. Things are behind feature flags, but i still could have messed up the core implementation in the process.

Additionally, I lack a ton of understanding about the cache in general, so im not sure if this the right spot for the changes, whether there are other places to improve, etc.


packages/web/src/common/store/cache/sagas.js
```diff
@@ -310,8 +325,35 @@ function* watchRemove() {
...
+  const fastCache = yield call(
+    remoteConfig.getFeatureEnabled,
+    FeatureFlags.FAST_CACHE
+  )
+  const safeFastCache = yield call(
+    remoteConfig.getFeatureEnabled,
+    FeatureFlags.SAFE_FAST_CACHE
+  )
+
+  console.log('hello', fastCache, safeFastCache)
```
raymondjacobson, line 340: sup

raymondjacobson (commented):
this is really interesting. i'm worried about undefined/hard to explain behaviors when things are more stale than we expect. it also would make more sense to me that we don't send network requests for things rather than just stopping them from updating the reducer. 

will give more thorough review

dylanjeffers:
> this is really interesting. i'm worried about undefined/hard to explain behaviors when things are more stale than we expect. it also would make more sense to me that we don't send network requests for things rather than just stopping them from updating the reducer. 
> 
> will give more thorough review 

Re requesting the right data, completely agree, but there are also cases we can't know ahead of time like user tracks, explore etc I think right?

audius-infra:
Preview this change https://demo.audius.co/dj-c-1791-fast-cache

packages/common/src/store/cache/reducer.ts
```diff
@@ -41,21 +58,22 @@ export const initialCacheState = {
...
+  _timestamp: _timestamp ?? Date.now()
 })
 
 // Unwraps a cache entry into its public metadata
 const unwrapEntry = (entry: { metadata: any }) => {
   if (entry && entry.metadata) {
     return entry.metadata
   }
-  return {}
+  return null
```
sliptype, line 76: This could cause runtime errors, in non ts files we could be expecting empty object

packages/common/src/store/cache/reducer.ts
```diff
@@ -130,59 +144,68 @@ export const mergeCustomizer = (
...
+        existing.blocknumber > entity.metadata.blocknumber
+      ) {
+        // do nothing
+      } else if (replace) {
+        newEntries[entity.id] = wrapEntry(entity.metadata)
+      } else if (
+        existing &&
+        _timestamp + DEFAULT_ENTRY_TTL > now &&
+        cacheType === 'fast'
+      ) {
```
sliptype, line 180: Maybe add a `// do nothing`

packages/common/src/store/cache/reducer.ts
```diff
@@ -130,59 +144,68 @@ export const mergeCustomizer = (
...
-            )
-          )
-        }
-
-        newUids[e.uid] = e.id
-        if (e.id in newSubscribers) {
-          newSubscribers[e.id].add(e.uid)
+    const now = Date.now()
+
+    for (let i = 0; i < entries.length; i++) {
```
sliptype, line 162: What's the reason for changing this to a for loop?

sliptype (approved):
Let's be very careful with enabling these, and test thoroughly on stage. I think this is a great step in the right direction but do worry about adding complexity to our already less than optimal cache layer. We should really prioritize setting up rtk-query

packages/common/src/store/cache/reducer.ts
dylanjeffers: so not having this resulting in error, since "existing" always returned true, which was not expected behavior i think based on reading through the saga. note this is just for unwrapping in the saga.

packages/common/src/store/cache/reducer.ts
dylanjeffers: ya!

packages/common/src/store/cache/reducer.ts
dylanjeffers: from what ive read its faster than a forEach, since we dont have to create function context for each one

PR #2555, opened 2023-01-10
Fix lineup loading after return to online

amendelsohn:
### Description

This doesn't work as is, but I think if I can get the saga to be called, it will fix the issue.
I can see SET_REACHABLE in the action log, but setupBackend() is not getting dispatched

### DIFF
--- a/packages/web/src/common/store/backend/sagas.ts
+++ b/packages/web/src/common/store/backend/sagas.ts
@@ -106,17 +106,25 @@ function* watchSetupBackend() {
   yield* takeEvery(backendActions.SETUP, setupBackend)
 }
 
-// Watch for changes to reachability and if not fully set up, re set-up the backend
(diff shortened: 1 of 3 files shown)


packages/web/src/common/store/backend/sagas.ts
```diff
@@ -106,17 +106,25 @@ function* watchSetupBackend() {
...
-    yield* call(waitForValue, (state) => getIsReachable(state))
     // Try to set up again, which should block further actions until completed
+    console.log('SetupBackendSaga action dispatched')
     yield* put(backendActions.setupBackend())
+  } else {
+    console.log('SetupBackendSaga already set up')
   }
 }
 
+function* watchSetReachable() {
```
rickyrombo, line 124: maybe add a console log here to ensure the `watch` is set?

PR #2557, opened 2023-01-10
Update the feature flag for rate cta drawer

Kyle-Shanks:
### Description
Update feature flag for Rate CTA Drawer to make sure that users are on the app version with the bug fixes

### Dragons

N/a

### How Has This Been Tested?

N/A

N/A

### Feature Flags ###

RATE_CTA_ENABLED_V2 -> Added
RATE_CTA_ENABLED -> Removed


sliptype (approved):
Nice! My bad for not calling this out when we made those fixes

I'm working on an idea right now to allow us to create optimizely audiences based on app version, so we can set the minimum version a feature is active on

PR #2560, opened 2023-01-10
[C-1764] Fix offline app initialization for offline mode

amendelsohn:
### Description

An error thrown in the wallet saga `fetchBalanceAsync` were causing all sagas to stop. After which point, no new actions were being processed.

FetchBalanceAsync should have been waiting for audiusBackendInstance to be setup, but was not.

We've split the reachability check into more states, adding an 'unconfirmed' state which can be used to optimistically assume reachability unless it is explicitly false. We use this for read calls that can be made without the full backend and will simply fail to respond. For writes requiring audiusBackend setup, we will use the more strict check that will only pass if reachability has been tested and confirmed true.

Changes:
1. Added third 'unconfirmed' reachability state
2. Added error logging to saga onError handler
3. Optimistically ApiClient.init() and setup backend
4. Retry setupBackend on every change to reachability if it hasn't been successfully set up yet
5. WaitForWrite now waits for confirmed reachability

### Dragons

Changes to app startup are always scary.

WaitForRead and WaitForWrite are intuitive, but do not necessarily cover all the cases. We may want to take another pass at those soon.

### How Has This Been Tested?

iOS Simulator

TODO:
- [ ] make sure we didn't break web

### Feature Flags ###

This is part of offline mode work, but notably not covered by the feature flag.

### DIFF
--- a/packages/mobile/src/store/store.ts
+++ b/packages/mobile/src/store/store.ts
@@ -81,7 +81,12 @@ const createRootReducer = () =>
     shareToStoryProgress
   })
 
-const sagaMiddleware = createSagaMiddleware({ context: storeContext })
+const sagaMiddleware = createSagaMiddleware({
+  context: storeContext,
+  onError: (e) => {
+    console.error('Caught Saga Error', e, e.stack)
+  }
+})
 
 const middlewares = [sagaMiddleware]
 
--- a/packages/web/src/common/store/backend/sagas.ts
+++ b/packages/web/src/common/store/backend/sagas.ts
@@ -56,7 +56,7 @@ export function* waitForReachability() {
 function* awaitReachability() {
   const isNativeMobile = yield* getContext('isNativeMobile')
   const isReachable = yield* select(getIsReachable)
-  if (isReachable || !isNativeMobile) return true
+  if (isReachable === true || !isNativeMobile) return true
   const { action } = yield* race({
     action: take(reachabilityActions.SET_REACHABLE),
     delay: delay(REACHABILITY_TIMEOUT_MS)
@@ -68,6 +68,11 @@ export function* setupBackend() {
   // Optimistically fetch account, then do it again later when we're sure we're connected
   // This ensures we always get the cached account when starting offline if available
   yield* put(accountActions.fetchLocalAccount())
+
+  // Init APICLient
+  const apiClient = yield* getContext('apiClient')
+  apiClient.init()
+
   const establishedReachability = yield* call(awaitReachability)
   // If we couldn't connect, show the error page
   // and just sit here waiting for reachability.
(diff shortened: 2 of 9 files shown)


packages/mobile/src/store/store.ts
```diff
@@ -81,7 +81,12 @@ const createRootReducer = () =>
     shareToStoryProgress
   })
 
-const sagaMiddleware = createSagaMiddleware({ context: storeContext })
+const sagaMiddleware = createSagaMiddleware({
+  context: storeContext,
+  onError: (e) => {
+    console.error('Caught Saga Error', e, e.stack)
```
raymondjacobson, line 87: we will get this in sentry now too. this is great!

packages/web/src/utils/sagaHelpers.ts
```diff
@@ -1,17 +1,19 @@
...
-} from 'common/store/backend/sagas'
+} from 'store/reachability/sagas'
 
 /**
  * Required for all writes
  */
 export function* waitForWrite() {
   yield* call(waitForBackendSetup)
   yield* call(waitForAccount)
+  yield* call(waitForConfirmedReachability)
```
raymondjacobson, line 16: seems reasonable for now. agreed to discuss at client sync.

packages/mobile/src/store/store.ts
amendelsohn: Sweet!! Is this a reasonable way to log it for readability in sentry?
I just wrote this for local debugging, but figured it's good to include

packages/web/src/common/store/wallet/sagas.ts
```diff
@@ -171,7 +171,7 @@ function* getWalletBalanceAndWallets() {
 }
 
 function* fetchBalanceAsync() {
-  yield* waitForRead()
+  yield* waitForWrite()
```
raymondjacobson, line 174: maybe a comment here to explain why since this is unintuitivie

raymondjacobson (approved):

packages/mobile/src/store/store.ts
raymondjacobson: this should be enough. they'll truncate our stack trace, but we'll learn quickly

sliptype (approved):
Makes sense to me, let's def discuss the WaitForRead/Write stuff!

packages/web/src/common/store/backend/sagas.ts
```diff
@@ -68,6 +61,11 @@ export function* setupBackend() {
   // Optimistically fetch account, then do it again later when we're sure we're connected
   // This ensures we always get the cached account when starting offline if available
   yield* put(accountActions.fetchLocalAccount())
+
+  // Init APICLient
+  const apiClient = yield* getContext('apiClient')
```
dylanjeffers, line 66: do we need to move this now that we are technically just going through with init?

dylanjeffers (approved):
lgtm!

packages/web/src/common/store/backend/sagas.ts
amendelsohn: I think the change is not required, but it works this way too

audius-infra:
Preview this change https://demo.audius.co/amendel-offline-lineups-with-dylan

PR #2561, opened 2023-01-10
[C-1397] Add appVersion to optimizely audience attributes

sliptype:
### Description

This allows us to apply feature flags to specific versions of the app, removing the need to rename feature flags before releasing a long running feature!

We can create audiences in Optimizely like so:
image
and then only roll out a feature to that audience.

The `version` attribute type required optimizely-sdk > 4.3.0. We have an outstanding snyk pr to upgrade to latest anyway so I upgraded to 4.9.2

### Dragons

Feature flag changes are dangerous

### How Has This Been Tested?

Tested on staging with different app versions

### Feature Flags ###

### DIFF
--- a/packages/common/src/services/remote-config/remote-config.ts
+++ b/packages/common/src/services/remote-config/remote-config.ts
@@ -43,6 +43,7 @@ export type RemoteConfigOptions<Client> = {
   setFeatureFlagSessionId: (id: number) => Promise<void>
   setLogLevel: () => void
   environment: Environment
+  appVersion: string
 }
 
 export const remoteConfig = <
@@ -60,7 +61,8 @@ export const remoteConfig = <
   getFeatureFlagSessionId,
   setFeatureFlagSessionId,
   setLogLevel,
-  environment
+  environment,
+  appVersion
 }: RemoteConfigOptions<Client>) => {
   const state: State = {
     didInitialize: false,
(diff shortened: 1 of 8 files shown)


packages/mobile/package.json
```diff
@@ -44,7 +44,7 @@
     "@hcaptcha/react-native-hcaptcha": "1.3.4",
     "@noble/hashes": "1.1.5",
     "@noble/secp256k1": "1.7.0",
-    "@optimizely/optimizely-sdk": "4.0.0",
+    "@optimizely/optimizely-sdk": "4.9.2",
```
dylanjeffers, line 47: nice

dylanjeffers, line 47: nice

packages/common/src/services/remote-config/remote-config.ts
```diff
@@ -219,8 +221,10 @@ export const remoteConfig = <
 
     try {
       const enabled = state.didInitialize
-        ? client.isFeatureEnabled(flag, id.toString(), { userId: id }) ??
-          defaultVal
+        ? client.isFeatureEnabled(flag, id.toString(), {
+            userId: id,
+            appVersion
```
dylanjeffers, line 226: woah so this is just an allowed value with optimizely in these later versions? clutch

dylanjeffers (approved):
love this

Kyle-Shanks (approved):
Nice

packages/common/src/services/remote-config/remote-config.ts
sliptype: We can actually pass any value, but yeah the new versions add a `version` type that can be matched on via semver

rickyrombo (approved):

piazzatron:
amazing

PR #2562, opened 2023-01-11
Do not write subscribe to discovery on auto-subscribe after follow

michellebrier:
### Description
Moved the subscribe on follow functionality to the indexer in discovery when indexing FOLLOW events [here](https://github.com/AudiusProject/audius-protocol/pull/4592). The indexer will validate the follow before creating both the follow and subscribe records. This means we don't need to issue a separate SUBSCRIBE event after the FOLLOW is confirmed in the client, as the time required to wait for the sequential relays to complete was affecting subscriptions writes.

Add an onFollow flag when processing the SET_NOTIFICATION_SUBSCRIPTION action and don't write to discovery when it's true. Right now we are still dual writing to both discovery and identity so the identity write completes as usual. We'll switch over to discovery after this and the backend change are deployed and the discovery subscriptions table has been backfilled.

### Dragons

### How Has This Been Tested?
Tested in web with breakpoints. Verified in DN that the indexer handled the auto subscribe/unsubscribe on follow/unfollow

Will query subscriptions tables in identity / discovery to ensure parity before switching over and deprecating identity subscriptions.

### Feature Flags ###

### DIFF
--- a/packages/web/src/pages/profile-page/ProfilePageProvider.tsx
+++ b/packages/web/src/pages/profile-page/ProfilePageProvider.tsx
@@ -239,7 +239,6 @@ class ProfilePage extends PureComponent<ProfilePageProps, ProfilePageState> {
     if (!profile) return
     const userId = profile.user_id
     this.props.onUnfollow(userId)
-    this.props.setNotificationSubscription(userId, false)
 
     if (this.props.account) {
       this.props.updateCurrentUserFollows(false)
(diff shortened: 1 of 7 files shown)


packages/web/src/pages/profile-page/ProfilePageProvider.tsx
```diff
@@ -238,7 +238,6 @@ class ProfilePage extends PureComponent<ProfilePageProps, ProfilePageState> {
     if (!profile) return
     const userId = profile.user_id
     this.props.onUnfollow(userId)
-    this.props.setNotificationSubscription(userId, false)
```
michellebrier, line 241: unfollow already dispatches a `SET_NOTIFICATION_SUBSCRIPTION` action. this was causing 2 unsubscribe requests to be issued to identity

audius-infra:
Preview this change https://demo.audius.co/mbrier/plat-572/client_subscribe_on_follow_em

audius-infra:
Preview this change https://demo.audius.co/mbrier/plat-572/client_subscribe_on_follow_em

sliptype (approved):
Looks good to me! Will the notification bell still be displayed as enabled immediately after following?

michellebrier:
> Looks good to me! Will the notification bell still be displayed as enabled immediately after following?

Yes!

PR #2564, opened 2023-01-11
[C-1726] useLocalImage

sliptype:
### Description

* Re-enables useLocalImage. `Audio` is now using `getImageSourceOptimistic` so the initial problem with image hooks returning null is resolved
* Adds `getLocalImageSource`. This is used In `Audio` to get the local source

Wanted to get your thoughts on the performance of `useLocalImage` in general. We are now doing 3 filesystem reads for every image load before starting to fetch the image remotely. Maybe we do only want to enable local images when reachability is false? Also this is another good argument for making image sizing explicit, because we would only need a single filesystem read

### Dragons

Touches all images

### How Has This Been Tested?

Tested that images, lock screen images, work online/offline. Felt like the perf is maybe slightly worse

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/audio/Audio.tsx
+++ b/packages/mobile/src/components/audio/Audio.tsx
@@ -141,6 +141,7 @@ export const Audio = () => {
   )
   const currentUserId = useSelector(getUserId)
   const isReachable = useSelector(getIsReachable)
+  const isNotReachable = isReachable === false
   const isOfflineModeEnabled = useIsOfflineModeEnabled()
 
   // Queue things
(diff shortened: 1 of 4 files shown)


amendelsohn (approved):
Looks good! Thank you!

Great call on the performance. It feels like using local ones ought to be a perf win, but given the way we're doing it right now, I agree that doing them only if unreachable might be preferable. It also removes the issue of stale local images, and other management concerns.

sliptype:
> Looks good! Thank you!
> 
> Great call on the performance. It feels like using local ones ought to be a perf win, but given the way we're doing it right now, I agree that doing them only if unreachable might be preferable. It also removes the issue of stale local images, and other management concerns.

Ok cool I'll add a check for reachability

dylanjeffers (approved):
assuming reachability check, looks great!

amendelsohn (approved):

PR #2565, opened 2023-01-11
[PAY-778] Update track page for premium tracks

sddioulde:
### Description

Premium tracks will have the corner tag based on access, special title based on premium conditions, button enabled or disabled based on access, and a special section based on premium conditions and access.

_Still to do in upcoming PR (so this one is not super large)_
- refresh/poll premium track after user has attempted unlock
- do not display premium (unlocked/how to unlock) section in track tile if current user is track owner (confirm with design if there needs to be sthg there at all)
- use confirmed design for unlocking state
- make sure follow from artist hover tile also triggers unlocking state

### Dragons

None

### How Has This Been Tested?

Local dapp v. stage.
Uploaded premium tracks, then tried diff users with/without access, and tried playing the track. Acts accordingly.

Nothing besides the send tip now having a new metric where `trackPage` is the source.
Added a linear to add metrics for unlock attempts.

### Feature Flags ###


packages/web/src/components/track/PremiumTrackSection.tsx
```diff
@@ -0,0 +1,371 @@
...
+      const url = `https://opensea.io/collection/${premiumConditions.nft_collection.slug}`
+      window.open(url, '_blank')
+    } else if (chain === Chain.Sol) {
+      const explorerUrl = `https://explorer.solana.com/address/${address}`
+      const url = externalLink ? new URL(externalLink).hostname : explorerUrl
+      window.open(url, '_blank')
+    }
+  }, [premiumConditions])
+
+  const renderLockedDescription = useCallback(() => {
```
sddioulde, line 113: thinking about separating this by gate type rather than by UI component...
maybe in another PR

packages/web/src/components/track/CardTitle.tsx
```diff
@@ -0,0 +1,81 @@
+import { FeatureFlags, PremiumConditions } from '@audius/common'
+import { IconCollectible, IconSpecialAccess } from '@audius/stems'
+import cn from 'classnames'
+
+import Tooltip from 'components/tooltip/Tooltip'
+import { useFlag } from 'hooks/useRemoteConfig'
+import HiddenTrackHeader from 'pages/track-page/components/HiddenTrackHeader'
+
+import styles from './GiantTrackTile.module.css'
```
rickyrombo, line 9: Nit: can we make each component manage its own styles?

packages/web/src/components/track/PlayPauseButton.tsx
```diff
@@ -0,0 +1,39 @@
+import { FeatureFlags } from '@audius/common'
+import { Button, ButtonType, IconPause, IconPlay } from '@audius/stems'
+
+import { useFlag } from 'hooks/useRemoteConfig'
+
+import styles from './GiantTrackTile.module.css'
```
rickyrombo, line 6: Same here

packages/web/src/components/track/PremiumTrackCornerTag.tsx
```diff
@@ -0,0 +1,28 @@
+import { FeatureFlags } from '@audius/common'
+import { IconLockedCornerTag, IconUnlockedCornerTag } from '@audius/stems'
+
+import { useFlag } from 'hooks/useRemoteConfig'
+
+import styles from './GiantTrackTile.module.css'
```
rickyrombo, line 6: And here

packages/web/src/components/track/PremiumTrackSection.tsx
```diff
@@ -0,0 +1,371 @@
...
+import { ReactComponent as IconExternalLink } from 'assets/img/iconExternalLink.svg'
+import { ReactComponent as IconVerifiedGreen } from 'assets/img/iconVerifiedGreen.svg'
+import FollowButton from 'components/follow-button/FollowButton'
+import { IconTip } from 'components/notification/Notification/components/icons'
+import UserBadges from 'components/user-badges/UserBadges'
+import { useFlag } from 'hooks/useRemoteConfig'
+import { AppState } from 'store/types'
+import { usePrevious } from 'react-use'
+
+import styles from './GiantTrackTile.module.css'
```
rickyrombo, line 29: And here

rickyrombo (approved):
Love that we're breaking components out of the single GiantTrackTile! Nice stuff

packages/web/src/components/track/CardTitle.tsx
sddioulde: was lazy lol. will keep track of this and hopefully get to it

PR #2566, opened 2023-01-12
Add explicit sizing to image fetching

sliptype:
### Description

Makes our image sizes explicit again so that we can use fast-image and potentially add a cache layer

### Dragons

*Is there anything the reviewer should be on the lookout for? Are there any dangerous changes?*

### How Has This Been Tested?

*Please describe the tests that you ran to verify your changes. Provide repro instructions & any configuration.*

### How will this change be monitored?

*For features that are critical or could fail silently please describe the monitoring/alerting being added.*

### Feature Flags ###

*Are all new features properly feature flagged? Describe added feature flags.*

PR #2568, opened 2023-01-12
[C-1801] Improve mobile image performance

dylanjeffers:
### Description

Image improvements using react-native-fast-image. In particular fixes issue with account profile header image glitching in and out as the page is changed

Id like to get this improvement in for track/collection tiles too, but want to double check it's compatibility with the useLocalImage stuff, and dont want to cause merge conflicts for yall.

I was considering adding a skeleton/opacity animation, but the images come in literally instantaneously so im thinking we just roll with no animation. If we do see issues though, happy to add one in


sliptype (approved):
Good stuff! I think we should add a loading state though, can we use `ImageLoader`? We need to keep in mind people on slow connections

Also we should be able to consolidate this once my pr gets in and use fast image everywhere! https://github.com/AudiusProject/audius-client/pull/2566

dylanjeffers:
> Good stuff! I think we should add a loading state though, can we use `ImageLoader`? We need to keep in mind people on slow connections
> 
> Also we should be able to consolidate this once my pr gets in and use fast image everywhere! #2566

Yeah that is a good point, i really just am not in love with adding a whole animated.value for so many dang images... i do think it has a "default animation" thats like an opacity thing, but prob not exactly what we want for our design? ill do a lil research on this?

Kyle-Shanks (approved):

raymondjacobson:
> > Good stuff! I think we should add a loading state though, can we use `ImageLoader`? We need to keep in mind people on slow connections
> > Also we should be able to consolidate this once my pr gets in and use fast image everywhere! #2566
> 
> Yeah that is a good point, i really just am not in love with adding a whole animated.value for so many dang images... i do think it has a "default animation" thats like an opacity thing, but prob not exactly what we want for our design? ill do a lil research on this?

I feel like animated.value is pretty darn cheap tho. agreed on the slow connections thing and us needing something. a white square/circle is pretty gross looking

raymondjacobson (approved):

dylanjeffers:
> > > Good stuff! I think we should add a loading state though, can we use `ImageLoader`? We need to keep in mind people on slow connections
> > > Also we should be able to consolidate this once my pr gets in and use fast image everywhere! #2566
> > 
> > 
> > Yeah that is a good point, i really just am not in love with adding a whole animated.value for so many dang images... i do think it has a "default animation" thats like an opacity thing, but prob not exactly what we want for our design? ill do a lil research on this?
> 
> I feel like animated.value is pretty darn cheap tho. agreed on the slow connections thing and us needing something. a white square/circle is pretty gross looking

Thanks for comments, continuing to tune the right balance, and testing on slower connection. From what i can tell though, on a fast connection, using animated value results in a slower experience for sure

dylanjeffers:
Handling this in another approach

PR #2569, opened 2023-01-12
update playlist_page route to accept permalink optionally

sabrina-kiam:
### Description
This adds an optional permalink argument to the PLAYLIST_PAGE route generation and branches off of whether a permalink is available or the playlist id and playlist name are available for legacy routing. 

I know we discussed passing in a playlist object, but that approach didn't work because not all calls to this had a playlist object to pass in. I tried another approach where it accepted an object of PlaylistPageArgs, but that required a lot of changes to all calls to this function and looked bulkier. This approach seemed the least invasive and easy to build off of but let me know what you think! 

### Testing
Tested locally pointed to stage data. Loading playlists worked fine. Creating one didn't work but I think that's because stage is messed up right now?

### DIFF
--- a/packages/web/src/utils/route.ts
+++ b/packages/web/src/utils/route.ts
@@ -274,19 +274,35 @@ export const fullAlbumPage = (handle: string, title: string, id: ID) => {
   return `${BASE_URL}${albumPage(handle, title, id)}`
 }
 
-
-export const playlistPage = (handle: string, playlistName?: string | null, playlistId?: ID | null, permalink?: string | null) => {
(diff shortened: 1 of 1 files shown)


dylanjeffers (approved):
The comments make a lot of sense, and this seems like the best approach given that. Yeah i think creating playlists is likely an issue on staging currently, though maybe try current staging separately from your changes to confirm?

audius-infra:
Preview this change https://demo.audius.co/sk-playlist-page-rt

PR #2571, opened 2023-01-12
[C-1813] Fix and simplify logic for offline lineup fetching

amendelsohn:
### Description

Ensure the lineup fetch is called for each switch from offline to online.

Added hooks for automatic callbacks on reachability state switch, similar to those we have for foreground and background.

Appears to fix the issue described in the ticket.

### Dragons

Maybe risk of race condition if we switch states too quickly, but we have debounce in reachability.

### How Has This Been Tested?

iOS sim, switching from online and offline states

### Feature Flags ###

`OFFLINE_MODE_ENABLED`


packages/mobile/src/hooks/useReachabilityState.ts
```diff
@@ -0,0 +1,45 @@
...
+  return useReachabilityState(onBecomeReachable, null)
+}
+
+export const useBecomeUnreachable = (
+  onBecomeUnreachable: OnBecomeUnreachable
+) => {
+  return useReachabilityState(null, onBecomeUnreachable)
+}
+
+export default useReachabilityState
```
Kyle-Shanks, line 45: is the default export needed here?

Kyle-Shanks (approved):
Nice

packages/mobile/src/hooks/useReachabilityState.ts
amendelsohn: it's not used, but I mirrored the structure in useAppState in case we want to use it later

PR #2573, opened 2023-01-12
[C-1815] Fix legacy image loading

sliptype:
### Description

We are relying on `onError` to be fired when react-native `Image` cannot load a source so that we can try the next source. Occasionally `onError` [does not seem to fire after already firing multiple times](https://github.com/facebook/react-native/issues/28611). When we need to try a long list of sources that means the image doesn't load occasionally. Because the legacy sources are at the end of the list of sources, we were seeing legacy images such as the Audius profile pic not loading

Instead of appending the legacy sources to the end of the list, now we interleave them so that the legacy path is tried for each CN after the new path is tried

### Dragons

### How Has This Been Tested?

### Feature Flags ###


packages/common/src/utils/array.ts
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * Interleave array elements
+ *
+ * interleave ([0, 2, 4, 6], [1, 3, 5]) // [ 0 1 2 3 4 5 6 ]
+ * interleave ([0, 2, 4], [1, 3, 5, 7]) // [ 0 1 2 3 4 5 7 ]
+ * interleave ([0, 2, 4], [])           // [ 0 2 4 ]
+ * interleave ([], [1, 3, 5, 7])        // [ 1 3 5 7 ]
+ * interleave ([], [])                  // [ ]
+ */
+export const interleave = <A, B>([x, ...xs]: A[], ys: B[] = []): (A | B)[] =>
```
Kyle-Shanks, line 11: ooh this is a neat one

Kyle-Shanks (approved):

packages/common/src/utils/array.ts
Kyle-Shanks: my initial thought what that this was also sorting the numbers so a comment there may be nice, but not needed for sure

sliptype: Oh yup the example is a little confusing, I can update in next pr!

raymondjacobson:
I think this change is right for now, but will also increase req. volume on the happy path probably. we should watch/resolve, especially w/ storage v2

PR #2588, opened 2023-01-13
[DMs] Add new message dialog and prompts

rickyrombo:
### Description

- Adds `ChatPrompt` component, which will display if there are no chats or if the user hasn't selected a message
- Adds `ChatComposeModal` (which is a misnomer, should rename to `NewChatModal` probably) to allow for searching for a user to send a message to
- Adds a web saga for routing the user to the new chat once created
- Adds a new type of user list to support the new modal, so that we can search for users

### Dragons

- Clicking a row on the search modal takes you to the user profile, probably not what we want here
  - Actually noticed a bug in the main userlist component where clicking just underneath the follow button (or in this case, message button) does nothing (the click handler is on the chip, not the whole row)
- Instead of spinner, went with disabling button for now.. todo fix
- Doesn't use proper compose icons... todo fix
- Need to figure out how to default user list to mutuals, and how to customize user list for whom you can message 🤔 
- Need to change chat creation to be deterministic based on members (right now, can create any number of chats with the same user)

### How Has This Been Tested?

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/mjp-dms-compose-new

packages/web/src/store/sagas.ts
```diff
@@ -150,6 +152,7 @@ export default function* rootSaga() {
     artistRecommendationsSagas(),
     buyAudioSagas(),
     changePasswordSagas(),
+    chatWebSagas(),
```
rickyrombo, line 155: I'm not in love with this name, or the splitting out of the sagas into two separate sagas. Couldn't think of a better way to handle the redirect after creating a chat though, since common doesn't know about the routing per client

packages/web/src/store/sagas.ts
```diff
@@ -165,6 +168,7 @@ export default function* rootSaga() {
     remixSettingsSagas(),
     repostPageSagas(),
     scrollLockSagas(),
+    searchPageSagas(),
```
rickyrombo, line 171: Eh, this could probably be called `searchUserListModalSagas` or something, but all the other user list sagas are called `xxxPageSagas`?

packages/web/src/components/user-list/components/UserList.tsx
```diff
@@ -57,16 +60,18 @@ const UserList = (props: UserListProps) => {
...
-                size='small'
-                following={user.does_current_user_follow}
-                onFollow={() => props.onFollow(user.user_id)}
-                onUnfollow={() => props.onUnfollow(user.user_id)}
-                showIcon
-                stopPropagation
-              />
-            ) : null}
+            {user.user_id !== props.userId
+              ? props.renderActionButton?.(user.user_id) || (
```
rickyrombo, line 64: Unfortunately, this approach doesn't change what the onclick does for the whole row (well, the user chip part of the row). That might need revisiting...

packages/web/src/pages/chat-page/components/ChatComposeModal.tsx
```diff
@@ -0,0 +1,107 @@
...
+          <Input placeholder={messages.searchUsers} onChange={handleChange} />
+        </div>
+        <div className={styles.results}>
+          <Scrollbar className={styles.scrollbar}>
+            <ConnectedUserList
+              tag={CHAT_COMPOSE_TAG}
+              stateSelector={getUserList}
+              onNavigateAway={handleClose}
+              beforeClickArtistName={handleClose}
+              renderActionButton={(userId: ID) => (
```
rickyrombo, line 91: Are there rerender concerns with this?

packages/web/src/pages/chat-page/components/ChatComposeModal.tsx
```diff
@@ -0,0 +1,107 @@
...
+export const ChatComposeModal = () => {
+  const dispatch = useDispatch()
+  const [isVisible, setIsVisible] = useModalState('ChatCompose')
+  const [pendingUserChat, setPendingUserChat] = useState<number>()
+  const [query, setQuery] = useState('')
+
+  const doSearch = useCallback(() => {
+    dispatch(searchUserListActions.setSearchQuery(query))
+    dispatch(userListActions.reset(SEARCH_USER_LIST_TAG))
+    dispatch(userListActions.loadMore(SEARCH_USER_LIST_TAG))
```
rickyrombo, line 48: Seems like the connected user list component should do this itself on initial render, but it wasn't working for me. Also I think this use case is different in that there's not really a need to paginate and keep loading more, but rather to keep requerying and replacing the existing content.

Wondering if it makes more sense to not use the connected version, but just the presentation version with custom logic? Or even ditch the idea of a user list modal for this case but render something similar, but different...

packages/web/src/common/store/user-list/search/sagas.ts
```diff
@@ -0,0 +1,46 @@
...
+    userIds: users.map((u) => (u as User).user_id),
+    hasMore: false
+  }
+}
+
+const userListSagas = UserListSagaFactory.createSagas({
+  tag: SEARCH_USER_LIST_TAG,
+  fetchUsers: searchUsers,
+  stateSelector: getUserList,
+  errorDispatcher: function* () {}
```
rickyrombo, line 41: TODO

audius-infra:
Preview this change https://demo.audius.co/mjp-dms-compose-new

audius-infra:
Preview this change https://demo.audius.co/mjp-dms-compose-new

packages/common/src/store/pages/chat/slice.ts
```diff
@@ -39,6 +39,14 @@ const slice = createSlice({
   name: 'application/pages/chat',
   initialState,
   reducers: {
+    createChat: (_state, _action: PayloadAction<{ userIds: ID[] }>) => {
+      // triggers saga
+    },
+    createChatSucceeded: (state, action: PayloadAction<{ chat: UserChat }>) => {
+      console.log(action)
```
piazzatron, line 46: log

packages/common/src/store/user-list/search/reducers.ts
```diff
@@ -0,0 +1,35 @@
...
+})
+
+const initialState = {
+  query: ''
+}
+
+const searchPageReducer = createReducer<SearchOwnState, SearchActions>(
+  initialState,
+  {
+    [actions.SET_SEARCH_QUERY](state: any, action: any) {
```
piazzatron, line 23: nit: any

piazzatron (approved):
Will do a more thorough review w comments later but only thing I'm noticing for now is that the search debounce is super fast

PR #2589, opened 2023-01-13
Premium track page - continued

sddioulde:
### Description

- refresh/poll premium track after user has attempted unlock
- make sure follow from artist hover tile also triggers unlocking state
- do not display premium (unlocked/how to unlock) section in track tile if current user is track owner (this changes after new design update)

### Dragons

some of this work needs to update with new design

### How Has This Been Tested?

local dapp v stage

n/a

### Feature Flags ###

existing feature flags


audius-infra:
Preview this change https://demo.audius.co/sd-pc-track-page-next

packages/common/src/store/premium-content/slice.ts
```diff
@@ -0,0 +1,59 @@
...
+    updatePremiumContentSignatures: (state, action: PayloadAction<UpdatePremiumContentSignaturesPayload>) => {
+      state.premiumTrackSignatureMap = {
+        ...state.premiumTrackSignatureMap,
+        ...action.payload
+      }
+    },
+    updatePremiumTrackStatus: (state, action: PayloadAction<UpdatePremiumTrackStatusPayload>) => {
+      state.status = action.payload.status
+    },
+    refreshPremiumTrack: (_, __: PayloadAction<RefreshPremiumTrackPayload>) => {
```
rickyrombo, line 45: ultra nitpicky can ignore, but in future fwiw I do `(_state, _action)`

packages/web/src/common/store/premiumContent/sagas.ts
```diff
@@ -287,8 +293,45 @@ function* updateCollectibleGatedTrackAccess(
...
+}: {
+  trackParams: TrackRouteParams
+  frequency: number
+}) {
+  const { trackId, slug, handle } = trackParams ?? {}
+  while (true) {
+    const premiumTrackSignatureMap = yield* select(getPremiumTrackSignatureMap)
+    if (trackId && premiumTrackSignatureMap[trackId]) {
+      yield* put(updatePremiumTrackStatus({ status: 'UNLOCKED' }))
+      break
```
rickyrombo, line 310: nit: to avoid a `while(true)`/break, could dupe the fetching logic right before the loop, and move this check to the `while` conditional, and then refetch at the end of the loop

rickyrombo (approved):

packages/web/src/common/store/premiumContent/sagas.ts
sddioulde: ah right 👍🏿

PR #2592, opened 2023-01-13
[C-1829] Fix android build due to gradle dep

dylanjeffers:
### Description

Fix android build due to missing react-native-gradle-plugin. Specifically installed the version required by react-native peerDep. will apply this to rc-stage as well!


sliptype (approved):
Thank you!

PR #2594, opened 2023-01-14
[C-1772] Fix share-drawer for tracks with missing images

dylanjeffers:
### Description

Fixes subtle bug where sharing a track with no image causes a bug with useShareStory hook. Im starting to think it might make sense to try and delay the rendering of these hooks until a user clicks on the "share" buttons, but also that would not fix this bug. just some thoughts to prevent too much code running in the simple rendering of the share drawer.

also i wonder if there is a way for us to have better typecheck here, since this passed even though source can be ` number` or `ImageSource[]`

### DIFF
--- a/packages/mobile/src/components/share-drawer/useShareToStory.tsx
+++ b/packages/mobile/src/components/share-drawer/useShareToStory.tsx
@@ -84,7 +84,7 @@ export const useShareToStory = ({
   }
 
   const trackImageUri =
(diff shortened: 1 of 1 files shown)


sliptype (approved):

sliptype:
There shouldn't be a way to create a track with no image from a user perspective right?

dylanjeffers:
> There shouldn't be a way to create a track with no image from a user perspective right?

Oh yeah good point, wait what's happening.. no large image?

PR #2595, opened 2023-01-14
[PAY-780] Premium content track tile - Web

sddioulde:
### Description

Add track tile for premium tracks

### Dragons

N/A

### How Has This Been Tested?

Local dapp vs stage for:
- collectible gated track
- tip gated track
- follow gated track

N/A

### Feature Flags ###

existing feature flags

### DIFF
--- a/packages/web/src/components/track/GiantTrackTile.module.css
+++ b/packages/web/src/components/track/GiantTrackTile.module.css
@@ -294,6 +294,7 @@
   top: 0;
   /* may need design to reproduce a more pixel perfect icon to avoid the -1px below */
   left: -1px;
+  z-index: 3;
 }
 
 .bottomSection {
--- a/packages/web/src/components/track/desktop/ConnectedTrackTile.tsx
+++ b/packages/web/src/components/track/desktop/ConnectedTrackTile.tsx
@@ -59,7 +59,7 @@ const { getTrack } = cacheTracksSelectors
 const { getUserFromTrack } = cacheUsersSelectors
 const { saveTrack, unsaveTrack, repostTrack, undoRepostTrack } =
   tracksSocialActions
-const getUserHandle = accountSelectors.getUserHandle
+const { getUserHandle, getUserId } = accountSelectors
 
 type OwnProps = {
   uid: UID
@@ -87,6 +87,7 @@ const ConnectedTrackTile = memo(
     size,
     track,
     user,
+    currentUserId,
     ordered,
     showArtistPick,
     goToRoute,
@@ -112,7 +113,11 @@ const ConnectedTrackTile = memo(
     const {
       is_delete,
       is_unlisted: isUnlisted,
+      is_premium: isPremium,
+      premium_conditions: premiumConditions,
+      premium_content_signature: premiumContentSignature,
       track_id: trackId,
+      owner_id: ownerId,
       title,
       permalink,
       repost_count,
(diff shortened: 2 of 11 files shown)


packages/web/src/components/track/desktop/ConnectedTrackTile.tsx
```diff
@@ -140,6 +143,8 @@ const ConnectedTrackTile = memo(
     const isTrackPlaying = isActive && isPlaying
     const isOwner = handle === userHandle
     const isArtistPick = showArtistPick && _artist_pick === trackId
+    const doesUserHaveAccess =
+      !isPremium || isOwner || !!premiumContentSignature
```
sddioulde, line 147: shouldn't even need isOwner here as backend should generate signature if user is owner

packages/web/src/pages/track-page/components/desktop/TrackPage.tsx
```diff
@@ -117,7 +117,8 @@ const TrackPage = ({
   const isSaved = heroTrack?.has_current_user_saved ?? false
   const isReposted = heroTrack?.has_current_user_reposted ?? false
   const loading = !heroTrack
-  const doesUserHaveAccess = isOwner || !!heroTrack?.premium_content_signature
+  const doesUserHaveAccess =
+    !heroTrack?.is_premium || isOwner || !!heroTrack?.premium_content_signature
```
sddioulde, line 121: shouldn't even need isOwner here as backend should generate signature if user is owner

packages/web/src/components/track/desktop/TrackTile.tsx
```diff
@@ -55,6 +75,195 @@ const RankAndIndexIndicator = ({
...
+  isOwner: boolean
+  isDarkMode?: boolean
+  isMatrixMode: boolean
+  showIconButtons?: boolean
+  onClickRepost: (e?: MouseEvent) => void
+  onClickFavorite: (e?: MouseEvent) => void
+  onClickShare: (e?: MouseEvent) => void
+}
+
+const BottomRow = ({
```
sddioulde, line 138: maybe should also move this to other file

packages/web/src/components/track/desktop/TrackTile.tsx
```diff
@@ -55,6 +75,195 @@ const RankAndIndexIndicator = ({
   )
 }
 
+const PremiumContentLabel = ({
```
sddioulde, line 78: going to move this to other file

packages/web/src/components/track/types.ts
```diff
@@ -199,6 +210,9 @@ export type DesktopTrackTileProps = {
 
   /** The relative link of the track */
   permalink: string
+
+  /** Whether the bottom bar can be overridden */
+  canOverrideBottomBar?: boolean
```
rickyrombo, line 215: Why do we need this?

packages/web/src/components/track/GiantTrackTile.module.css
```diff
@@ -294,6 +294,7 @@
   top: 0;
   /* may need design to reproduce a more pixel perfect icon to avoid the -1px below */
   left: -1px;
+  z-index: 3;
```
rickyrombo, line 297: I think we might have a z-index tracker somewhere, not sure

rickyrombo (approved):

packages/web/src/components/track/types.ts
sddioulde: because PlaylistTile uses TrackTile and it should not show the 'locked' bottom bar. same for FinishPage during upload, where it should show a progress bar

packages/web/src/components/track/GiantTrackTile.module.css
sddioulde: oh yeah. adding it there too thanks

PR #2598, opened 2023-01-17
[C-1831] Update common/stems/web to react 18

dylanjeffers:
### Description

Upgrades the rest of our monorepo to react-18

### Dragons

A few dependencies had to be updated that had major version changes. I made sure to follow the upgrade guides, the only library that had consumer-level changes was react-parallax.

Removed @audius/sdk from sub-packages that don't need it, a cleanup that should ideally result in faster install times.

### How Has This Been Tested?
Running smoothly on web and mobile


audius-infra:
Preview this change https://demo.audius.co/dj-c-1831-react-18-types

packages/common/src/store/index.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable import/export */
```
amendelsohn, line 1: this seems like a standard export for us. Do we want to update the rule?

packages/stems/src/hooks/useMediaQueryListener.ts
```diff
@@ -7,7 +7,7 @@ export const useMediaQueryListener = (mediaQuery: string) => {
   // Updates whether there is a match or not
   // when the media query status changes
   const listener = useCallback(
-    matcher => setIsMatch(matcher.matches),
+    (matcher: any) => setIsMatch(matcher.matches),
```
amendelsohn, line 10: do we have a better type?

packages/web/src/hooks/useImageSize.test.tsx
```diff
@@ -11,7 +11,7 @@ const TestComponent = (
   const dispatch = () => {}
   // @ts-ignore
   const image = useImageSize({ ...props, dispatch })
-  return <div>{image ?? 'nothing'}</div>
+  return <div>{image() ?? 'nothing'}</div>
```
amendelsohn, line 14: is this a bugfix? seems like the `useImageSize` overloads returning different but similar types could get confusing

packages/web/src/index.tsx
```diff
@@ -13,4 +12,8 @@ import './index.css'
 // when running in dev mode.
 import Root from './root'
 
-ReactDOM.render(<Root />, document.getElementById('root'))
+const container = document.getElementById('root')
+if (container) {
```
amendelsohn, line 16: Is falsy container a realistic case, or just to help type checking?
I assume if we skip the root render, the site doesn't load at all

packages/web/src/pages/feed-page/FeedPageProvider.d.ts
```diff
@@ -0,0 +1,10 @@
+import { FeedPageContentProps } from './types'
```
amendelsohn, line 1: do these d.ts files just add top-level types for our old js files?

amendelsohn (approved):
WOW!!

Can't say I understood every change in here, particularly the parllax stuff, but it all looks reasonable to me. Thank you for bringing us up to date!

packages/common/src/store/index.ts
dylanjeffers: so weirdly enough i could not for the life of me figure out why the errors were happening. it was saying stuff like duplicate import of `ADD` `DELETE` which are the action types, but then when i went to the index for that store slice, it did `export * as blah`, so i don't know how those actions were also being reexported....

packages/stems/src/hooks/useMediaQueryListener.ts
dylanjeffers: id love too, but i tried many things and nothing worked. it appears we are using it in a way that's not according to the types...

packages/web/src/hooks/useImageSize.test.tsx
dylanjeffers: yeah i think this could be an error for sure, but also its just in a test so im kinda cool with it. looking at the code it was like, if dispatch is passed in, return a function, so potentially this is right, but also didn't care that much given test!

packages/web/src/index.tsx
dylanjeffers: for us it's impossible unless we update the index.html file and remove the "root" element, was just a type thing

packages/stems/src/hooks/useMediaQueryListener.ts
amendelsohn: oh, awesome lol

packages/web/src/hooks/useImageSize.test.tsx
amendelsohn: yeah more concerned with the types being confusing for actual usages. If this bug can slip by in a test, we might make the same mistake for realsies

packages/web/src/pages/feed-page/FeedPageProvider.d.ts
dylanjeffers: exactly, the new upgrades complained about types that weren't being caught before cause i think older versions of react would just throw it's hands up and say `(props: any) => any` before, whereas now it is asking to actually type the js files

amendelsohn: Perfect! Love this change. Someday we'll be 100% typescript, but until that day, this is close enough

packages/web/src/hooks/useImageSize.test.tsx
dylanjeffers: yeah good call, let me dig into the intracies of this function rq

amendelsohn: Sweet. I'm assuming the reason for the function is when you want the calculation of the value deferred as late as possible, but idk

dylanjeffers: @amendelsohn updated!

audius-infra:
Preview this change https://demo.audius.co/dj-c-1831-react-18-types

PR #2600, opened 2023-01-17
Improve useLocalImage perf

sliptype:
### Description

Immediately return empty array if reachable, this avoids the overhead of awaiting the async function

### Dragons

### How Has This Been Tested?

### Feature Flags ###


dylanjeffers (approved):
sweet, thanks! so for reachable = true, isn't it the case that we still get loading: false, loading: true, loading: false flow still?

sliptype:
> sweet, thanks! so for reachable = true, isn't it the case that we still get loading: false, loading: true, loading: false flow still?

Your comment made me realize that this logic was entirely backwards. We should be short-circuiting when reachable. In the case we are reachable loading will always be false and return instantly. In the case we are not reachable, the async check will run and should go from loading: true -> loading: false

PR #2605, opened 2023-01-17
[PAY-819] Fix track update cid bug

sddioulde:
### Description

Track updates/edits currently do not send over the track cid as part of the metadata, therefore causing DNs to set the track_cid value to null for the updated track. This fixes it.

### Dragons

None

### How Has This Been Tested?

Local dapp vs prod. Updated track and cid remained intact.

N/A

### Feature Flags ###

None

### DIFF
--- a/packages/common/src/schemas/index.ts
+++ b/packages/common/src/schemas/index.ts
@@ -47,7 +47,7 @@ const trackMetadataSchema = {
 
 export const newTrackMetadata = (fields, validate = false) => {
   const validFields = validate
-    ? pick(fields, Object.keys(trackMetadataSchema).concat(['track_id']))
+    ? pick(fields, Object.keys(trackMetadataSchema).concat(['track_id', 'track_cid']))
     : fields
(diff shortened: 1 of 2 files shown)


packages/common/src/schemas/index.ts
```diff
@@ -47,7 +47,7 @@ const trackMetadataSchema = {
 
 export const newTrackMetadata = (fields, validate = false) => {
   const validFields = validate
-    ? pick(fields, Object.keys(trackMetadataSchema).concat(['track_id']))
+    ? pick(fields, Object.keys(trackMetadataSchema).concat(['track_id', 'track_cid']))
```
dylanjeffers, line 50: nice, would we want to just add this to trackMetadataSchema?

dylanjeffers (approved):
nice fix

packages/common/src/schemas/index.ts
sddioulde: that's what im wondering too

PR #2606, opened 2023-01-18
Add proper fix for track cid bug

sddioulde:
### Description

proper fix. tested and it still works

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/common/src/schemas/index.ts
+++ b/packages/common/src/schemas/index.ts
@@ -3,6 +3,7 @@
 import { pick } from 'lodash'
 
 const trackMetadataSchema = {
+  track_cid: null,
   owner_id: null,
   title: null,
   length: null,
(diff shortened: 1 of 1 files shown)


packages/common/src/schemas/index.ts
```diff
@@ -47,7 +48,7 @@ const trackMetadataSchema = {
 
 export const newTrackMetadata = (fields, validate = false) => {
   const validFields = validate
-    ? pick(fields, Object.keys(trackMetadataSchema).concat(['track_id', 'track_cid']))
+    ? pick(fields, Object.keys(trackMetadataSchema).concat(['track_id']))
```
dylanjeffers, line 51: so do we need to concat track_id even since that does not matter?

dylanjeffers (approved):
love it!

packages/common/src/schemas/index.ts
sddioulde: i have not double checked whether it matters and just wanted a quick fix track cid fix without changing what was there prior

PR #2608, opened 2023-01-18
[C-1690] Sync newly favorited/unfavorited collections

amendelsohn:
### Description

During sync process on session start:
1. Fetch list of user's favorited collections
2. Ensure those collections are downloaded
3. Remove downloaded collections that are no longer favorited
4. Enqueue download for favorited collections that are not downloaded

Moves startSync to saga

### Dragons

Easy to miss steps here.

### How Has This Been Tested?

iOS Simulator

TODO: add metrics for sync stats

### Feature Flags ###

`OFFLINE_MODE_ENABLED`


amendelsohn:
image

Getting these type errors since the upgrade I think. Doesn't seem like it's from this change, but it's breaking the build.

packages/mobile/src/store/offline-downloads/sagas.ts
```diff
@@ -86,8 +108,71 @@ function* watchClearOfflineDownloads() {
...
+
+    const isFavoritesDownloadEnabled =
+      offlineCollectionsState[DOWNLOAD_REASON_FAVORITES]
+    if (isFavoritesDownloadEnabled) {
+      // Individual tracks
+      yield* call(syncFavoritedTracks)
+
+      // Favorited collections
+      for (const collectionId of accountCollectionIds) {
+        yield* put(fetchCollection(collectionId))
```
dylanjeffers, line 141: so just want to make sure the impact here isn't too great on startup if we are making alot of requests?

dylanjeffers (approved):
all looks clean, just curious about the number of requests and potential perf impacts

packages/mobile/src/store/offline-downloads/sagas.ts
amendelsohn: Yeah I was thinking about how to mitigate that. Might be okay to add a short delay here between requests so they never stack up?

dylanjeffers: yeah i think potentially we just see what the perf looks like? so in this case, can we just wait until the user navigates to the collections to fire this off?

dylanjeffers: i guess this is only a one-time thing though? or are we continuously looking for updates?

amendelsohn: Just once per session. All we need is to generally keep your offline content up to date. If we miss something, we'll do it next session. Thought about adding some sort of loop in case someone leaves the app open for months, but I'm not so worried about that edge case

amendelsohn: Yeah sounds good. Any tips for measuring perf impact with flipper? I can try an account with several collections and see if it scores lower on the time to response test.

Impact of the sync process should be:
- 1 request per collection the user has owned/favorited
- delete operations for removed tracks
- any missing tracks enqueued for download
- overhead of calculating what's missing in the above

it's linear wrt the number and length of collections, and the largest component (downloads) is deferred to the queue

amendelsohn: https://linear.app/audius/issue/C-1855/ensure-sync-doesnt-cause-performance-hit-on-app-load

Making a follow-up ticket so I can get this merged and tested sooner.

dylanjeffers: yeah with flipper, one potential easy win is to try the react perf plugin, the one that gives lighthouse like scores. we can try that together when this is played

PR #2617, opened 2023-01-18
[C-1810] Block more screens when offline

amendelsohn:
### Description

Shows the offline indicator instead of failing to show content that requires connectivity.

Recommend review with hide whitespace changes.

### Dragons

Might be better off with a blanket solution for all screens. We could pass in an `offline={true}` prop for those that don't require it.

### DIFF
--- a/packages/mobile/src/screens/chats-screen/ChatsScreen.tsx
+++ b/packages/mobile/src/screens/chats-screen/ChatsScreen.tsx
@@ -4,7 +4,7 @@ import { chatActions, chatSelectors } from '@audius/common'
 import { View } from 'react-native'
 import { useDispatch, useSelector } from 'react-redux'
 
-import { Screen, Text } from 'app/components/core'
+import { Screen, ScreenContent, Text } from 'app/components/core'
 
 const { getChats, getChatMessages } = chatSelectors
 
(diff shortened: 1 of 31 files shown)


packages/mobile/src/screens/chats-screen/ChatsScreen.tsx
```diff
@@ -24,18 +24,20 @@ export const ChatsScreen = () => {
...
-          return (
-            <View key={message.message_id}>
-              <Text>{message.sender_user_id} says:</Text>
-              <Text>{message.message}</Text>
-              <Text>{message.created_at}</Text>
-            </View>
-          )
-        })}
-      </View>
+      <ScreenContent>
```
dylanjeffers, line 27: yeah could we just make the offline thing an option for screen? and default it to "not offline capable", and just the few screens that need it, we go `<Screen allowOffline />` or something?

dylanjeffers (commented):
If we are in a big rush, maybe go forward with current impl, but imo moving this to screen would result in way fewer changes and less cognitive complexity potentially?

packages/mobile/src/screens/chats-screen/ChatsScreen.tsx
amendelsohn: Yes, I think that's worth doing. I'll circle back to this tonight and do it that way.

dylanjeffers (approved):
based on chats, the move is to wrap all screen "content" in "ScreenContent"!

PR #2618, opened 2023-01-18
[C-1832] Fix shuffle for new track player implementation

Kyle-Shanks:
### Description
Remove a lot of unnecessary code and add a check to make shuffle work a lot better with the new queue implementation for track player

### Dragons

Sometimes the next queue track will play for a split second before the next shuffle track plays because of on track change handlers and dispatching redux actions

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A


dylanjeffers (approved):
heck yeah, awesome when simplifying code results in a fix!

PR #2619, opened 2023-01-18
[PAY-845] Sort rewards tiles by completed/disbursed

dharit-tan:
### Description
Completed rewards that are ready to claim float to the top of the rewards tiles. Rewards that are already disbursed sink to the bottom. Respects optimistic claiming. Both mobile and web.

### How Has This Been Tested?

Tested locally on web and ios sim. Also tested that aggregate rewards eg. invites function correctly.

### DIFF
--- a/packages/common/src/utils/challenges.ts
+++ b/packages/common/src/utils/challenges.ts
@@ -155,11 +155,14 @@ export const sortChallenges = (
 
     if (userChallenge1?.state === 'disbursed') {
       return 1
-    } else if (userChallenge1?.state === 'completed') {
+    }
+    if (userChallenge1?.state === 'completed') {
       return -1
-    } else if (userChallenge2?.state === 'disbursed') {
+    }
+    if (userChallenge2?.state === 'disbursed') {
       return -1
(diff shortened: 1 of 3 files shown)


audius-infra:
Preview this change https://demo.audius.co/rt-PAY-845

packages/common/src/utils/challenges.ts
```diff
@@ -145,3 +145,23 @@ export const challengeRewardsConfig: Record<
...
+export const sortChallenges = (
+  userChallenges: Partial<Record<ChallengeRewardID, OptimisticUserChallenge>>
+): ((id1: ChallengeRewardID, id2: ChallengeRewardID) => number) => {
+  return (id1, id2) => {
+    const userChallenge1 = userChallenges[id1]
+    const userChallenge2 = userChallenges[id2]
+
+    if (userChallenge1?.state === 'disbursed') {
+      return 1
+    } else if (userChallenge1?.state === 'completed') {
```
sddioulde, line 158: nit: since we have returns, dont need `else if`, can just use `if`

sddioulde (approved):

PR #2623, opened 2023-01-18
[C-1863] filter out missing tracks in offline lineup

amendelsohn:
### Description

Lineup was crashing due to sending `undefined` tracks in the `fetchLineupMetadatasSucceeded` action.


nicoback2 (approved):
LGTM

PR #2624, opened 2023-01-19
[PAY-835] Use proper icons for DMs

rickyrombo:
### Description

Adds the compose, send, and plus icons for DMs and consumes

### Dragons

### How Has This Been Tested?

### Feature Flags ###


Kyle-Shanks (approved):
Looks good. There might be a slight issue with the fill of the SVGs. I've noticed in the past if the SVGs aren't filled black, they don't respect the styles for button transitions and things like that, but if you didn't notice an issue in testing then it should be fine

rickyrombo:
> Looks good. There might be a slight issue with the fill of the SVGs. I've noticed in the past if the SVGs aren't filled black, they don't respect the styles for button transitions and things like that, but if you didn't notice an issue in testing then it should be fine

Hmm we should figure out how to deal with that, since the new pattern going forward is to have all SVGs red so we know when they're unstyled (see product design slack)

PR #2625, opened 2023-01-19
[C-1811] Fetch feed on reconnect while in background

amendelsohn:
### Description

`ApiClient.getSocialFeed` was returning `[]` when the request failed. This caused the lineup to load "successfully" with 0 entries, meaning we display the suggested follows instead of feed.

Lineup now also re-requests on reconnect.

### Dragons

This requires a change to the ApiClient return pattern. Imo, `null` is a much more useful response for a failed request than `[]` or `{}` which both lie to you.

Question for reviewers: Should we change the other ApiClient methods to match? This would keep consistency and prevent such bugs in the future, but comes at a higher risk of failure now.

### How Has This Been Tested?

Launch app offline
Background (but do not close) the app
Turn wifi back on
Open the app
Feed now loads properly where it would display the suggested first follows before

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/lineup/Lineup.tsx
+++ b/packages/mobile/src/components/lineup/Lineup.tsx
@@ -14,6 +14,7 @@ import {
   TrackTile,
   LineupTileSkeleton
 } from 'app/components/lineup-tile'
+import { useBecomeReachable } from 'app/hooks/useReachabilityState'
 import { useScrollToTop } from 'app/hooks/useScrollToTop'
 
 import { Delineator } from './Delineator'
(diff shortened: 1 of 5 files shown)


dylanjeffers (approved):
nice find. i would say let's make a ticket about this general problem for the other methods, since there is def a chance for failure. this should probably also apply to audiusbackend/libs calls too?

packages/mobile/src/components/lineup/Lineup.tsx
```diff
@@ -320,6 +321,35 @@ export const Lineup = ({
     extraFetchOptions
   ])
 
+  useBecomeReachable(
+    useCallback(() => {
```
dylanjeffers, line 325: is there a way that we can do this?
```
useBecomeReachable(handleLoadMore)
```

audius-infra:
Preview this change https://demo.audius.co/amendel-offline-background-reconnect

PR #2626, opened 2023-01-19
[C-1838] Remove global progress, improve Audio performance

dylanjeffers:
### Description

- Removes global.progress usage throughout mobile app
- Removes tons of unecessary rerenders over the course of an app lifecycle by removing the `useProgress` hook in `Audio`. This was churning through 5 rerenders a second whether or not music was playing, yikes!
- Drastically improves TrackingBar performance by removing the 5 rerenders every second to a single rerender per track (same major issue as audio, but at least it only churned when music was playing)

### Dragons

Touches Audio.tsx. I think this should improve it overall since it stabilizes it's rerenders.

### DIFF
--- a/packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
+++ b/packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
@@ -115,7 +115,6 @@ export const NowPlayingDrawer = memo(function NowPlayingDrawer(
   const styles = useStyles()
 
   const { isOpen, onOpen, onClose } = useDrawer('NowPlaying')
-  console.log('isOpen?', isOpen)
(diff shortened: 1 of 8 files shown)


packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
```diff
@@ -114,6 +115,7 @@ export const NowPlayingDrawer = memo(function NowPlayingDrawer(
   const styles = useStyles()
 
   const { isOpen, onOpen, onClose } = useDrawer('NowPlaying')
+  console.log('isOpen?', isOpen)
```
Kyle-Shanks, line 118: log

Kyle-Shanks (approved):
amazing changes dude

PR #2627, opened 2023-01-19
[C-1866] Fix favorites toggle

amendelsohn:
### Description

Recent changes broke the favorites toggle download. Fixed and separated out the `downloadAllFavorites` flow for readability.

PR #2628, opened 2023-01-19
[C-1867] track collection fav/unfavorited states separately

amendelsohn:
### Description

In order to know whether to delete a collection and its tracks when we toggle the allFavorites toggle off, we need to know whether the collection was downloaded as part of favorites or because it was downloaded individually.

Collections downloaded individually should persist offline even after allFavorites download is turned off.

We already store this for tracks (see DownloadReason), but this should fix it for collections.

### Dragons

It's a lot of code written late at night.
Relies on https://github.com/AudiusProject/audius-client/pull/2623 and https://github.com/AudiusProject/audius-client/pull/2627

### How Has This Been Tested?

### Feature Flags ###


packages/mobile/src/hooks/useLoadOfflineTracks.ts
```diff
@@ -57,9 +57,17 @@ export const useLoadOfflineData = () => {
     }[] = []
     for (const collectionId of offlineCollections) {
       try {
-        dispatch(addCollection(collectionId))
-        if (collectionId === DOWNLOAD_REASON_FAVORITES) continue
+        if (collectionId === DOWNLOAD_REASON_FAVORITES) {
+          dispatch(addCollection({ collectionId, isFavoritesDownload: false }))
```
nicoback2, line 61: I'm probably just reading the code wrong, but should `isFavoritesDownload` be set to `true` here?

packages/mobile/src/hooks/useLoadOfflineTracks.ts
```diff
@@ -57,9 +57,17 @@ export const useLoadOfflineData = () => {
...
+          continue
```
nicoback2, line 62: Personally prefer an `else` instead of `continue` for readability

packages/mobile/src/store/offline-downloads/sagas.ts
```diff
@@ -89,7 +89,7 @@ export function* downloadSavedCollection(
     }
   }))
   if (!tracksForDownload) return
-  downloadCollectionById(action.collectionId, false)
+  downloadCollection(collection, true)
```
nicoback2, line 92:
```suggestion
  downloadCollection(collection, /* isFavoritesDownload */ true)
```

packages/mobile/src/store/offline-downloads/slice.ts
```diff
@@ -23,10 +26,16 @@ export enum OfflineTrackDownloadStatus {
   ERROR = 'ERROR'
 }
 
+type CollectionDownloadPayload = {
+  collectionId: string
+  isFavoritesDownload: boolean
```
nicoback2, line 31: Prob worth a comment here explaining what `isFavoritesDownload` means / what the two different types of downloaded collections are (i.e. from favoriting the collection and having "Download All Favorites" on vs. from directly downloading the collection)

packages/mobile/src/services/offline-downloader/offline-storage.ts
```diff
@@ -105,7 +106,8 @@ export const purgeDownloadedCollection = async (collectionId: string) => {
   const collectionDir = getLocalCollectionDir(collectionId)
   if (!(await exists(collectionDir))) return
   await RNFS.unlink(collectionDir)
-  store.dispatch(removeCollection(collectionId))
+  store.dispatch(removeCollection({ collectionId, isFavoritesDownload: true }))
+  store.dispatch(removeCollection({ collectionId, isFavoritesDownload: false }))
```
nicoback2, line 110: This may create a new entry in either `state.favoritedCollections` or `state.collections` in the offline downloads slice, should be fine but not sure if you intended that

nicoback2 (approved):

packages/mobile/src/hooks/useLoadOfflineTracks.ts
nicoback2: or I guess the "Favorites" collection is not considered a Favorites download?

packages/mobile/src/services/offline-downloader/offline-storage.ts
amendelsohn: yeah that's fine as long as its value is set to false

PR #2632, opened 2023-01-19
Add @tsconfig/react-native to mobile tsconfig

dylanjeffers:
### Description

Adds standardized @tsconfig/react-native to mobile tsconfig. This has made me realize that it would be great to move all web deps to common so we can finally flip on `strict: true` cause there are tons of subtle ts errors out there in mobile land currently

### Dragons
Anyone know whats up with the podfile changing a lot? no sure if i should include my pod change?


sliptype (approved):
Love it!

Yeah I got that podfile change locally too, let's commit it

We need to finish the common migration for sure, it's a big undertaking. Let's knock out pieces between features

PR #2639, opened 2023-01-20
[C-1755] Fix repeat jobs and sync inefficiencies

amendelsohn:
### Description

Fixes several cases where downloads would be queued multiple times or sync would do unnecessary work.

1. Collections were always syncing every track after the last refactor
2. Downloads were queuing twice for favorited collections
3. Tracks removed from collections were being wrongly queued for download
4. Tracks were being queued for download twice in collection sync

This (mostly) fixes the issue with "infinite" loading spinner on downloads. It was not infinite, but tracks that are invalid downloads (e.g. missing files or unavailable) would fail 3 times per associated job.

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/services/offline-downloader/offline-sync.ts
+++ b/packages/mobile/src/services/offline-downloader/offline-sync.ts
@@ -23,7 +23,6 @@ import {
   downloadCollectionCoverArt,
   downloadTrackCoverArt,
   DOWNLOAD_REASON_FAVORITES,
-  removeCollectionDownload,
   removeDownloadedCollectionFromFavorites
 } from './offline-downloader'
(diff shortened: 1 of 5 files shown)


packages/mobile/src/services/offline-downloader/offline-sync.ts
```diff
@@ -170,7 +169,11 @@ export const syncCollectionTracks = async (
     updatedCollection
   )
 
-  downloadCollection(updatedCollectionWithArt)
+  downloadCollection(
+    updatedCollectionWithArt,
```
raymondjacobson, line 173: in future probably good to have these args be spread objects so it's less work to refactor

raymondjacobson (approved):

PR #2641, opened 2023-01-20
[C-1864] Fix orphaned tracks after cancellation

amendelsohn:
### Description

Race condition was causing cancellation to happen but the job could still be requeued or succeed after it should have cancelled.

Fix is to check state at the very end just to make sure we get it right.

### Dragons

### How Has This Been Tested?

### Feature Flags ###


dylanjeffers (approved):
lgtm

packages/mobile/src/services/offline-downloader/offline-downloader.ts
```diff
@@ -390,36 +394,42 @@ export const batchRemoveTrackDownload = async (
...
       }
-    } catch (e) {
-      console.debug(
-        `failed to remove track ${trackId} from collection ${downloadReason.collection_id}`
-      )
+      await writeTrackJson(trackIdStr, trackToWrite)
     }
-  })
+  } catch (e) {
+    console.debug(
```
raymondjacobson, line 429: maybe worth console erroring?

raymondjacobson (approved):

PR #2643, opened 2023-01-20
[C-1862] Cache user images on sign in

sliptype:
### Description

On sign in, cache user profile picture and cover photo using the `Image` cache so that they are available offline

Interesting thing is that the profile picture wasn't manually being cached, but because it was always rendered/fetched in an online session before going offline it was always cached by `Image`. So technically caching the profile picture on sign in isn't necessary, but if we were to ever change the UI to not show the profile picture on initial app load it would not be available in the cache. I could be convinced to remove the explicit caching though, wdyt?

### Dragons

### How Has This Been Tested?

### Feature Flags ###


amendelsohn (approved):
Oh, neat! Didn't know that's how it works. I like keeping the explicit cache.

PR #2644, opened 2023-01-20
[C-1856] Upgrade react-native to 0.71

dylanjeffers:
### Description

Upgrades react-native to newly released 0.71. With this comes an awesome new flex-gap property, that should simplify our styling dev, so we aren't having to put "margin-bottom/top" on every child element.

### Dragons

Since it is a project-wide change, we should make sure everything continues to to work as expected.
The amount of changes feels wrong, so going to look into package-lock changes
-- just tried removing lock and node_mods, and got same thing, so i think this is due to core deps changing

Lets wait to merge until monday, so this isn't included in release branch

### How Has This Been Tested?

Tested android and ios


packages/mobile/package.json
```diff
@@ -169,10 +169,11 @@
     "ios-deploy": "1.11.4"
   },
   "devDependencies": {
-    "@babel/core": "7.15.5",
+    "@babel/core": "^7.20.0",
+    "@babel/preset-env": "^7.20.0",
     "@babel/plugin-transform-react-jsx": "7.16.7",
-    "@babel/runtime": "7.15.4",
-    "@react-native-community/eslint-config": "0.0.5",
+    "@babel/runtime": "^7.20.0",
```
sliptype, line 175: Pin pls

packages/mobile/patches/react-native-code-push+7.0.5.patch
```diff
@@ -0,0 +1,46 @@
+diff --git a/node_modules/react-native-code-push/android/codepush.gradle b/node_modules/react-native-code-push/android/codepush.gradle
+index 22be273..468fdb1 100644
+--- a/node_modules/react-native-code-push/android/codepush.gradle
++++ b/node_modules/react-native-code-push/android/codepush.gradle
+@@ -25,6 +25,17 @@ gradle.projectsEvaluated {
+     }
+     
+     android.applicationVariants.all { variant ->
++        // skip debug build variant
```
sliptype, line 9: Why is this patch needed?

sliptype (approved):

packages/mobile/patches/react-native-code-push+7.0.5.patch
dylanjeffers: https://github.com/microsoft/react-native-code-push/issues/2418

dylanjeffers: will add this as comment

PR #2649, opened 2023-01-20
Memoize sorted challenges

dharit-tan:
### Description
Memoize the sorted challenges, and use not-optimistic user challenges on web to avoid tiles teleporting abruptly.

### Dragons

### How Has This Been Tested?

Local web and ios sim.

### Feature Flags ###

### DIFF
--- a/packages/common/src/utils/challenges.ts
+++ b/packages/common/src/utils/challenges.ts
@@ -1,4 +1,8 @@
-import { ChallengeRewardID, OptimisticUserChallenge } from '../models'
+import {
+  ChallengeRewardID,
+  UserChallenge,
+  OptimisticUserChallenge
+} from '../models'
 
 import { formatNumberCommas } from './formatUtil'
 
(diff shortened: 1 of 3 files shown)


packages/common/src/utils/challenges.ts
```diff
@@ -145,26 +145,3 @@ export const challengeRewardsConfig: Record<
     id: 'trending-underground'
   }
 }
-
-export const sortChallenges = (
```
rickyrombo, line 149: I think we can leave this here and share it? Why did it move?

Also, maybe we can rename this to `makeChallengeSortComparator` or something since it's not actually sorting

packages/web/src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
```diff
@@ -164,12 +164,37 @@ const useRewardIds = (
   return filteredRewards
 }
 
+const sortChallenges = (
+  userChallenges: Record<string, UserChallenge>
+): ((id1: ChallengeRewardID, id2: ChallengeRewardID) => number) => {
+  return (id1, id2) => {
+    if (Object.keys(userChallenges).length === 0) {
```
rickyrombo, line 171: Is this for an empty object check? Any reason the other use doesn't have this check?

Also, for simplicity, maybe this could be checks like `if (!userChallenge1)` in the lines below?

rickyrombo (approved):

packages/common/src/utils/challenges.ts
dharit-tan: Names of members to compare changes between using optimistic vs unoptimistic challenges!

packages/web/src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
dharit-tan: Yep empty object, your suggestion makes sense. And good call I should probably add it to the other one as well.

packages/common/src/utils/challenges.ts
rickyrombo: ... wat? I don't know what this means 😅

packages/web/src/pages/audio-rewards-page/ChallengeRewardsTile.tsx
rickyrombo: if both are the same, we can probably keep it in utils then right?

packages/common/src/utils/challenges.ts
dharit-tan: In mobile I want to use the optimistic challenges in the comparator, but in web I want to use the unoptimistic. The object types and members are different (ie. is_disbursed in unoptimistic vs. state === 'disbursed' in optimistic). Figured it was easiest to handle each case individually.

dharit-tan:
Sorry had to force-push, was getting lint errors not related to my code.

packages/common/src/utils/challenges.ts
rickyrombo: ahhhhhhh I see it now

rickyrombo: yeah this is kinda annoying. I do like keeping this in utils either way even if it means two separate functions - maybe `makeOptimisticChallengeComparator` and `makeUserChallengeComparator`?

rickyrombo:
> Sorry had to force-push, was getting lint errors not related to my code.

You can always merge master! :)

packages/common/src/utils/challenges.ts
dharit-tan: Fair enough!

audius-infra:
Preview this change https://demo.audius.co/rt-sort-challenge

PR #2651, opened 2023-01-21
[C-1884] Improve top supporters fetching and loading

dylanjeffers:
### Description

Improves top-supporters/supporting flow by moving fetching logic from sagas to top-supporters components, and removing CLS by using `top_supporters` user value to crate a loading skeleton that exactly matches loaded-in top-supporters component.

### Dragons

There are potential issues with this related to the other ways top-supporters are used, ie in tipping flow etc. we should make sure there are no regressions on those areas.

### DIFF
--- a/packages/mobile/src/screens/notifications-screen/Notification/ProfilePictureListSkeleton.tsx
+++ b/packages/mobile/src/screens/notifications-screen/Notification/ProfilePictureListSkeleton.tsx
@@ -1,0 +1,42 @@
+import { times } from 'lodash'
+import { View } from 'react-native'
+
+import { makeStyles } from 'app/styles'
+
+const useStyles = makeStyles(({ spacing, palette, typography }) => ({
+  root: {
+    flexDirection: 'row',
+    marginRight: spacing(6)
+  },
+  image: {
+    marginRight: spacing(-2),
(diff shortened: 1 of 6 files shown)


packages/mobile/src/screens/notifications-screen/Notification/ProfilePictureListSkeleton.tsx
```diff
@@ -0,0 +1,42 @@
...
+) => {
+  const { count, limit } = props
+  const styles = useStyles()
+  return (
+    <View style={styles.root}>
+      {times(Math.min(count, limit)).map((index) => (
+        <View key={index} style={styles.image} />
+      ))}
+      {count >= limit ? <View style={styles.extra} /> : null}
+      <View></View>
```
raymondjacobson, line 39: is this view to balance out the flex?

raymondjacobson (approved):

packages/mobile/src/screens/notifications-screen/Notification/ProfilePictureListSkeleton.tsx
dylanjeffers: no i think a mistake actually, good catch

PR #2654, opened 2023-01-22
[C-1876] Hide download indicator on deleted tracks

sliptype:
### Description

I spent some time trying to filter out deleted tracks from the download queue, but because we are only pulling the `favorites` endpoint we don't have the data to know if the track is deleted. Fetching that data would require extra calls for each batch of tracks in `fetchAllFavoritedTracks` etc and seemed like it would add quite a bit of complexity

This is the simple solution, the download fails silently in the background and the deleted track doesn't show up when offline

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/track-list/TrackListItem.tsx
+++ b/packages/mobile/src/components/track-list/TrackListItem.tsx
@@ -240,9 +240,11 @@ export const TrackListItem = ({
               </Text>
             </View>
 
-            <View style={styles.downloadIndicator}>
-              <DownloadStatusIndicator trackId={track_id} size={18} />
-            </View>
+            {!isDeleted && (
+              <View style={styles.downloadIndicator}>
+                <DownloadStatusIndicator trackId={track_id} size={18} />
+              </View>
+            )}
           </View>
           <Text numberOfLines={1} style={styles.artistName}>
             {name}


packages/mobile/src/components/track-list/TrackListItem.tsx
```diff
@@ -240,9 +240,11 @@ export const TrackListItem = ({
               </Text>
             </View>
 
-            <View style={styles.downloadIndicator}>
-              <DownloadStatusIndicator trackId={track_id} size={18} />
-            </View>
+            {!isDeleted && (
```
amendelsohn, line 243: Can you apply the same logic in LineupTileStats?

Alternatively, move the check into DownloadStatusIndicator having it fetch the track or passing it in as a prop.

amendelsohn (approved):
Ah, yeah. This works, thank you!

packages/mobile/src/components/track-list/TrackListItem.tsx
sliptype: I don't believe there is a way to see a deleted track in a lineup, or is there? We only show them in favorites and playlists so that they don't just dissapear when deleted

amendelsohn: Ah, that's a good point! This is fine then :)

PR #2661, opened 2023-01-23
[QA-308] Fix explore page crash

amendelsohn:
### Description

We filter nullable, but don't allow the absence of the cached collection to make it through the map safely.

### Dragons

Could be hiding bugs with loading and cacheing the collections. Not sure why this started happening now.

### DIFF
--- a/packages/common/src/store/pages/explore/selectors.ts
+++ b/packages/common/src/store/pages/explore/selectors.ts
@@ -13,7 +13,7 @@ export const getExplorePlaylists = createSelector(
   (state: CommonState) => state.pages.explore.playlists,
   (state: CommonState) => state.collections.entries,
   (playlists, collections) =>
-    playlists.map((id) => collections[id].metadata).filter(removeNullable)
+    playlists.map((id) => collections[id]?.metadata).filter(removeNullable)
 )
 
 export const getExploreArtists = createSelector(


dylanjeffers (changes requested):
so actually looks like a potential bug with explore playlists? im not seeing them getting cached correctly, ie they come from api endpoint, but then our selector is unable to grab them. ive been tracking it down to the `retreiveCollections` function, which i beleive was updated recently, maybe @sabrina-kiam? im thinking ideally we fix that issue cause this `entity[id].metadata` is everywhere

sliptype (approved):
The crash seemed to me like it could be related to the playlist url changes that got merged, but if this null check fixes it then I'm not sure what change caused it

dylanjeffers:
> The crash seemed to me like it could be related to the playlist url changes that got merged, but if this null check fixes it then I'm not sure what change caused it

Yeah i think it's due to playlist url change, or something around there! this fixes the crash, but doesnt fix the real issue, which is that we are getting no explore playlists on the screen (mobile and web, check out staging rn for example). so im down to merge this if we want, but there are like 100's of usages where we dont check ?. for metadata so it feels like a bandaid, is all im sayin

amendelsohn:
> > The crash seemed to me like it could be related to the playlist url changes that got merged, but if this null check fixes it then I'm not sure what change caused it
> 
> Yeah i think it's due to playlist url change, or something around there! this fixes the crash, but doesnt fix the real issue, which is that we are getting no explore playlists on the screen (mobile and web, check out staging rn for example). so im down to merge this if we want, but there are like 100's of usages where we dont check ?. for metadata so it feels like a bandaid, is all im sayin

Yeah, totally. There's a reason it broke now and this wasn't the change that did it. When I tested it locally, I got playlists on the explore tab, but I'm happy to wait for the proper fix if that's in the works.

dylanjeffers:
> > > The crash seemed to me like it could be related to the playlist url changes that got merged, but if this null check fixes it then I'm not sure what change caused it
> > 
> > 
> > Yeah i think it's due to playlist url change, or something around there! this fixes the crash, but doesnt fix the real issue, which is that we are getting no explore playlists on the screen (mobile and web, check out staging rn for example). so im down to merge this if we want, but there are like 100's of usages where we dont check ?. for metadata so it feels like a bandaid, is all im sayin
> 
> Yeah, totally. There's a reason it broke now and this wasn't the change that did it. When I tested it locally, I got playlists on the explore tab, but I'm happy to wait for the proper fix if that's in the works.

Ah, so they eventually come down, just there is that in-between time? i see, in that case your change makes sense then. did you see the behavior on staging web though, cause in that case looks like nothing is coming through.

dylanjeffers (approved):
approved if we wanna roll through with this

amendelsohn:
I only saw it working in prod. Mobile staging loaded nothing, but I wasn't sure if staging was set up with explore data.

amendelsohn:
We solved this a different way

PR #2667, opened 2023-01-23
[C-1854] Prevent crash on settings page

sliptype:
### Description

Wait to dispatch `fetchServices` until the `fetchedServices` state is updated

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/web/src/components/service-selection/SelectedServices.tsx
+++ b/packages/web/src/components/service-selection/SelectedServices.tsx
@@ -1,6 +1,7 @@
 import { useState, useEffect, useCallback } from 'react'
 
 import cn from 'classnames'
+import { isEqual } from 'lodash'
 import { useDispatch } from 'react-redux'
 
 import { ReactComponent as IconInfo } from 'assets/img/iconInfo.svg'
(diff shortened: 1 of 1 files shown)


packages/web/src/components/service-selection/SelectedServices.tsx
```diff
@@ -53,8 +53,10 @@ export const SelectedServices = (props: SelectedServicesProps) => {
 
   useEffect(() => {
     if (!fetchedServices && (!services || services.length === 0)) {
-      setFetchedServices(true)
-      dispatch(fetchServices())
+      setFetchedServices(() => {
```
amendelsohn, line 56: woah didn't know you could do that!

amendelsohn (approved):
Nice, this seems like a better fix.

Thoughts on adding the null coalesce to the selector as well? I could see and argument for either keeping a louder failure or preventing the crash.

sliptype:
> Nice, this seems like a better fix.
> 
> Thoughts on adding the null coalesce to the selector as well? I could see and argument for either keeping a louder failure or preventing the crash.

Your comment made me realize that an even better fix is to provide an equality function to useSelector because `getSelectedServices` always returns a new array and causes a rerender

Not really sure how null coalescing would be a solution here, what do you mean?

PR #2677, opened 2023-01-24
[C-1923] Implement fast profile images

dylanjeffers:
### Description

Uses fast-image for profile images, drastically improving various loading states and performance around the app

### Dragons

Offline-mode prefetch user image might be an issue?


sliptype (approved):
Awesome! Are we confident that we caught all the cases this time 😅

dylanjeffers:
> Awesome! Are we confident that we caught all the cases this time 😅

Yes i believe so! we have the rest of the week to catch too, will make sure i do full run before next release

PR #2678, opened 2023-01-24
[C-1890] Improve artist-recommendations fetching/loading

dylanjeffers:
### Description

Improves mobile performance by only fetching artist-recommendations when a user is followed, rather than every time a profile is requested.


rickyrombo (approved):
How fast does the related artists load on Mobile? I think generally the job takes a bit if it needs refreshing, so I opted to prefetch

Kyle-Shanks (approved):

packages/web/src/common/store/profile/sagas.js
```diff
@@ -324,20 +324,22 @@ function* fetchProfileAsync(action) {
...
+        )
+      }
     }
 
-    // Delay so the page can load before we fetch mutual followers
-    yield delay(2000)
-
     if (!isNativeMobile) {
+      // Delay so the page can load before we fetch mutual followers
+      yield delay(2000)
```
rickyrombo, line 341: ugh this is so jank 😂 - i wonder if there's a better way now?

dylanjeffers:
> How fast does the related artists load on Mobile? I think generally the job takes a bit if it needs refreshing, so I opted to prefetch

From my testing it seemed quite quick!

packages/web/src/common/store/profile/sagas.js
dylanjeffers: So this is mobile web so will continue to ignore lmao!!

rickyrombo: it's not desktop web too??

audius-infra:
Preview this change https://demo.audius.co/dj-c-1890-artist-recomendations

PR #2680, opened 2023-01-24
[C-1799] Add slug support for collection deep-links

dylanjeffers:
### Description

Updates mobile deep links to support playlist slug format

### Dragons

Need to confirm if we also need to decostruct the "slug" to handle + id and pass all three along to fetch request.

### DIFF
--- a/packages/mobile/src/screens/collection-screen/CollectionScreen.tsx
+++ b/packages/mobile/src/screens/collection-screen/CollectionScreen.tsx
@@ -68,32 +68,12 @@ const useStyles = makeStyles(({ spacing }) => ({
  */
 export const CollectionScreen = () => {
   const { params } = useRoute<'Collection'>()
+  const { id, searchCollection, slug, collectionType } = params ?? {}
   const dispatch = useDispatch()
 
-  // params is incorrectly typed and can sometimes be undefined
-  const {
-    id: idParam,
-    searchCollection,
-    collectionName,
-    collectionType
-  } = params ?? {}
-
-  const id = useMemo(() => {
-    if (collectionName) {
-      // Use collectionName from params if provided
-      // This is to support deep linking
-      // TODO: update this when collections are updated to use slug url format
-      // https://linear.app/audius/issue/C-1198/update-mobile-deep-linking-to-support-collection-slug-url-format
-      const nameParts = collectionName.split('-')
-      const collectionId = parseInt(nameParts[nameParts.length - 1], 10)
-      return collectionId as number
-    }
-    return idParam as number
-  }, [collectionName, idParam])
-
   const handleFetchCollection = useCallback(() => {
-    dispatch(fetchCollection(id))
-  }, [dispatch, id])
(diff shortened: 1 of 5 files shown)


packages/mobile/src/screens/collection-screen/CollectionScreen.tsx
```diff
@@ -68,32 +68,12 @@ const useStyles = makeStyles(({ spacing }) => ({
...
-      const collectionId = parseInt(nameParts[nameParts.length - 1], 10)
-      return collectionId as number
-    }
-    return idParam as number
-  }, [collectionName, idParam])
-
   const handleFetchCollection = useCallback(() => {
-    dispatch(fetchCollection(id))
-  }, [dispatch, id])
+    dispatch(fetchCollection(id, slug && decodeURIComponent(slug)))
```
sabrina-kiam, line 75: why do we need to decode the slug here?

packages/mobile/src/screens/collection-screen/CollectionScreen.tsx
```diff
@@ -68,32 +68,12 @@ const useStyles = makeStyles(({ spacing }) => ({
...
-  // params is incorrectly typed and can sometimes be undefined
-  const {
-    id: idParam,
-    searchCollection,
-    collectionName,
-    collectionType
-  } = params ?? {}
-
-  const id = useMemo(() => {
-    if (collectionName) {
```
sabrina-kiam, line 82: curious if you think we should keep this around for that edge case you mentioned earlier where playlists in local storage might not have permalinks yet?

sabrina-kiam (commented):
looks great!! thanks so much for doing this. i had a few questions we can discuss and then hopefully we can test everything soon

packages/mobile/src/screens/collection-screen/CollectionScreen.tsx
dylanjeffers: Good question, so for urls that have special characters, they are typically encoded, so decoding here to get the right special characters back

packages/mobile/src/screens/collection-screen/CollectionScreen.tsx
dylanjeffers: Yeah exactly what I was thinking as well, unless it's the case that the actions/sagas we are running take the permalink and extract handle id as a backup? Guessing that's not the case and I should send all three along in the fetch action?

sabrina-kiam: yeah that is not the case, so we'd have to send all 3 along in fetch 😬 yea i think we'd want to pass along in fetch? also realizing this is the screen used for albums too probably, and albums use the original routing and expect the name to contain id and name so we still need this for sure

packages/mobile/src/screens/collection-screen/CollectionScreen.tsx
sabrina-kiam: gotcha ty!

PR #2694, opened 2023-01-26
Improve offline mode react performance

sliptype:
### Description

Improves selectors used in `DownloadStatusIndicator` and `Audio` to minimize rerenders while offline mode is on.
0 rerenders on feed screen while offline mode is on!

Will follow up with a pr to address excess image fetching and rn queue performance

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/audio/Audio.tsx
+++ b/packages/mobile/src/components/audio/Audio.tsx
@@ -15,7 +15,8 @@ import {
   encodeHashId,
   Genre,
   tracksSocialActions,
-  SquareSizes
+  SquareSizes,
+  shallowCompare
 } from '@audius/common'
 import { isEqual } from 'lodash'
 import queue from 'react-native-job-queue'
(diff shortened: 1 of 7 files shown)


packages/mobile/src/components/audio/Audio.tsx
```diff
@@ -117,24 +118,40 @@ export const Audio = () => {
...
 
   // Queue Things
   const queueIndex = useSelector(getIndex)
   const queueShuffle = useSelector(getShuffle)
   const queueOrder = useSelector(getOrder)
   const queueTrackUids = queueOrder.map((trackData) => trackData.uid)
-  const queueTrackMap = useSelector((state) =>
-    getTracks(state, { uids: queueTrackUids })
+  const queueTrackIds = queueOrder.map((trackData) => trackData.id)
+  const queueTrackMap = useSelector(
```
sliptype, line 128: `useProxySelector` was still rerendering, not sure why

amendelsohn (approved):
This is great!

Wondering if we should audit usage of the other selectors. With all these keyed objects, I feel like there are other places I'm selecting the whole object where I shouldn't. Better to make these where we pass in the ids, right?

sliptype:
> This is great!
> 
> Wondering if we should audit usage of the other selectors. With all these keyed objects, I feel like there are other places I'm selecting the whole object where I shouldn't. Better to make these where we pass in the ids, right?

Yes! I'm going through them now

packages/mobile/src/screens/favorites-screen/AlbumsTab.tsx
```diff
@@ -27,10 +27,10 @@ export const AlbumsTab = () => {
   const [filterValue, setFilterValue] = useState('')
   const isReachable = useSelector(getIsReachable)
   const isOfflineModeEnabled = useIsOfflineModeEnabled()
-  const offlineDownloadStatus = useSelector(getOfflineDownloadStatus)
   const userAlbums = useProxySelector(
-    (state: CommonState) =>
-      getAccountCollections(state, filterValue).filter((collection) => {
+    (state: AppState) => {
+      const offlineDownloadStatus = getOfflineDownloadStatus(state)
```
amendelsohn, line 32: Ah, nice. So this rolls it all into one dependency?

How does this work if the status changes? I assume the hook runs but the component only rerenders if the result is different?

amendelsohn (approved):

packages/mobile/src/screens/favorites-screen/AlbumsTab.tsx
sliptype: Exactly! Any time the store state changes _all_ selectors in useSelectors are rerun and the result compared by default with reference equality

useProxySelector does this slightly more efficiently by tracking which parts of the state got accessed and then only rerendering if those parts changed

amendelsohn: Perfect. That's super cool. So we should be combining our complex selectors whenever possible.

sliptype: It's definitely a tradeoff, running a more complex computation every time the state changes vs rerendering a component more often. Usually rendering react components is wayyy more expensive though. Also there is some overhead to useProxySelector but I'm not sure how big

PR #2696, opened 2023-01-26
[C-1891] Restart app on saga error, Refactor mobile toasts

dylanjeffers:
### Description

Restarts the app on saga error, to ensure users don't get into funky states where nothing works. Ideally this never actually occurs, but it will also help us understand when there are issues.

Refactors toast logic to use redux, which allows us to dispatch toasts from sagas. This is also more in-line with web.


sliptype (approved):
NIce! Kinda pains me to remove a context in favor of sagas but it was necessary

dylanjeffers:
> NIce! Kinda pains me to remove a context in favor of sagas but it was necessary

So luckily no sagas involved just slice, but yeah needed to dispatch in a saga lol

PR #2697, opened 2023-01-26
Fix lineups not loading in Offline Mode (C-1945, C-1882, C-1803)

nicoback2:
### Description
After the perf improvements, the `useReachabilityState` callbacks were not being fired on every render anymore. This is a good thing, but it ended up revealing issues with how/when we were loading the lineup on the favorited tracks tab and all the favorited collection screens.

If you were offline and then loaded any of these screens, the lineup simply would not load. This is because the call to fetch local tracks was ONLY triggered when going from online -> offline. So this PR makes sure that we fetch local tracks on first render as well.

However this was not the only fix needed. We were also not updating our `state.pages.collection` (https://github.com/AudiusProject/audius-client/blob/main/packages/common/src/store/pages/collection/selectors.ts#L2) slice when navigating to a collection screen while offline, so the collection id passed to `useOfflineCollectionLineup` was always empty. I put in a quick fix for this but we should probably add logic to update that part of the store correctly when loading the offline lineup.

The last fix implemented here was that we were filtering out tracks needlessly from offline collection lineups because their downloaded tracks' data did not yet have the collection as a reason for download.

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/hooks/useLoadOfflineTracks.ts
+++ b/packages/mobile/src/hooks/useLoadOfflineTracks.ts
@@ -1,12 +1,11 @@
 import { useCallback } from 'react'
 
-import {
+import type {
   CollectionMetadata,
   Track,
   UserMetadata,
   lineupActions,
   UserTrackMetadata,
-  collectionPageActions,
   SmartCollectionVariant
 } from '@audius/common'
 import {
@@ -141,7 +140,11 @@ export const useLoadOfflineData = () => {
  * @param lineupActions the actions instance for the lineup
  */
 export const useOfflineCollectionLineup = (
-  collectionId: typeof DOWNLOAD_REASON_FAVORITES | number | SmartCollectionVariant | null,
+  collectionId:
+    | typeof DOWNLOAD_REASON_FAVORITES
+    | number
+    | SmartCollectionVariant
+    | null,
   fetchOnlineContent: () => void,
   lineupActions: lineupActions.LineupActions
 ) => {
--- a/packages/mobile/src/hooks/useLoadOfflineTracks.ts
+++ b/packages/mobile/src/hooks/useLoadOfflineTracks.ts
@@ -226,4 +226,5 @@ export const useOfflineCollectionLineup = (
   ])
 
   useReachabilityState(fetchOnlineContent, fetchLocalContent)
+  return fetchLocalContent
 }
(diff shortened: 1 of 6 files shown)


packages/mobile/src/screens/collection-screen/CollectionScreen.tsx
```diff
@@ -262,6 +262,7 @@ const CollectionScreenComponent = (props: CollectionScreenComponentProps) => {
             hasReposted={has_current_user_reposted}
             hasSaved={has_current_user_saved}
             isAlbum={is_album}
+            collectionId={playlist_id}
```
nicoback2, line 265: this is the quick fix mentioned in the pr description

dylanjeffers (approved):
Wow super clean! Imo long term it would be great for us to now have to worry about the two cases from the fetching selecting perspective, maybe when we get the persist stuff going

packages/mobile/src/hooks/useLoadOfflineTracks.ts
```diff
@@ -153,12 +153,19 @@ export const useOfflineCollectionLineup = (
...
       const lineupTracks = Object.values(offlineTracks)
-        .filter((track) =>
-          track.offline?.reasons_for_download.some(
-            (reason) => reason.collection_id === collectionId.toString()
-          )
+        .filter(
+          (track) =>
+            track.offline?.reasons_for_download.some(
+              (reason) => reason.collection_id === collectionId.toString()
+            ) || collectionTrackIds.has(track.track_id)
```
nicoback2, line 168: reasoning: a track in a collection may be downloaded from the favorited tracks page, but not yet have the collection added as a second reason for download (since the download queue hasn't gotten there yet).

nicoback2: sadly this doesnt work the other way around- we only know to include a downloaded track in the favorited tracks tab if it has the favorited tracks collection as a reason for download. cant fix this until we have a way of keeping the full list of favorited track metadata before they are all downloaded - i.e. can do what we do here on line 156 but for the favorited tracks "collection".

amendelsohn: Great call. Sad we don't have a track list for favorites yet.

packages/mobile/src/hooks/useLoadOfflineTracks.ts
```diff
@@ -174,7 +181,7 @@ export const useOfflineCollectionLineup = (
       store.dispatch(cacheActions.add(Kind.TRACKS, cacheTracks, false, true))
 
       if (collectionId === DOWNLOAD_REASON_FAVORITES) {
-        // Reorder lineup tracks accorinding to favorite time
```
amendelsohn, line 177: 💀

amendelsohn (approved):
Amazing finds! Totally missed the collection page loading.

PR #2699, opened 2023-01-27
Fix cache types

raymondjacobson:
### Description

These types are hard, but this fixes them partially.
Getting cache entries in bulk were lying that they were Record<string, User|Track|Collection>, but it's really a partial since for whatever reason the value may not actually end up in the cache.

This deserves a bit better of a cleanup, but hopefully this can prevent some errors like missing user_ids on undefined in downstream components, etc.

I think we may consider filtering undefined's out when coming back from the cache too, but that also probably leads to some undefined behaviors and I felt that doing them in the downstream selectors / components was safer.

### Dragons

### How Has This Been Tested?

Ran all type checks & ran app locally web & mobile

### Feature Flags ###

### DIFF
--- a/packages/common/src/store/account/selectors.ts
+++ b/packages/common/src/store/account/selectors.ts
@@ -1,5 +1,6 @@
 import { createSelector } from 'reselect'
 
+import { Collection } from 'models/Collection'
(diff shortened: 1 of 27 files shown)


dylanjeffers (approved):
wow amazing changes

PR #2700, opened 2023-01-27
Add toggle for message notifs on web + mobile

michellebrier:
### Description
Behind existing `CHAT_ENABLED` feature flag

### How Has This Been Tested?
Manually verified flag showed/hid message toggle, verified could toggle on and off and persist with corresponding backend changes, verified bulk toggle on/off worked.
image
image


sliptype (approved):
Nice!

PR #2691, opened 2023-01-26
[C-1754] Fix modal popups not closing

rickyrombo:
### Description

Fixes the issue where the popups wouldn't dismiss when clicking the modal contents, without reintroducing the $AUDIO transfer modal bug.

Will rebase later...

### Dragons

*Is there anything the reviewer should be on the lookout for? Are there any dangerous changes?*

### How Has This Been Tested?

*Please describe the tests that you ran to verify your changes. Provide repro instructions & any configuration.*

### How will this change be monitored?

*For features that are critical or could fail silently please describe the monitoring/alerting being added.*

### Feature Flags ###

*Are all new features properly feature flagged? Describe added feature flags.*

PR #2706, opened 2023-01-27
[DMs] Redo create chat modal

rickyrombo:
### Description

- Removes the existing "UserList" method of doing the `CreateChatModal`
- Adds new component `SearchUsersModal` with its own store and sagas, to be used to search for users. It allows for a default list of users to be sent in as a prop and paginated using a callback function
- Changes `ChatCompose` to `CreateChat` for the modal name
- Uses the new `SearchUsersModal` as the `CreateChatModal`

### Dragons

### How Has This Been Tested?

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/mjp-dms-compose-modal

audius-infra:
Preview this change https://demo.audius.co/mjp-dms-compose-modal

packages/common/src/store/user-list/index.ts
```diff
@@ -37,7 +37,3 @@ export * from './followers/types'
 export * as favoritesUserListActions from './favorites/actions'
 export * as favoritesUserListSelectors from './favorites/selectors'
 export * from './favorites/types'
-
-export * as searchUserListActions from './search/actions'
```
rickyrombo, line 41: In theory, I guess I could keep the same store structure as though it were a userList and have searchUsersModal use those actions/sagas. I made my own but what do you think @dylanjeffers ?

piazzatron (approved):
looks great. I noticed a few little things in testing, u may know these already, may require design followup:

- Dismiss on clickOutside seems broken for this modal
- We probably want to show a loading spinner while waiting for a search to return
- We probably want some sort of zero state if there are no users for a search 
- Clicking on the user's row triggered a crash once, and now it seems to work. Missed the stack trace unfortunately

rickyrombo:
> * Dismiss on clickOutside seems broken for this modal

Appears to be broken on stage for all modals, will look into ASAP (probably related to #2691 )

> * We probably want to show a loading spinner while waiting for a search to return

Adding!

> * We probably want some sort of zero state if there are no users for a search

Will add!

> * Clicking on the user's row triggered a crash once, and now it seems to work. Missed the stack trace unfortunately

This is probably due to trying to navigate to a chat before the DB has it, known issue, will fix potentially with websockets

audius-infra:
Preview this change https://demo.audius.co/mjp-dms-compose-modal

PR #2713, opened 2023-01-27
[C-1969 C-1967] Don't re-favorite playlist on download

dylanjeffers:
### Description

Catches case where we'd re-favorite a playlist when downloading. This logic is needed for cases where we try to download a collection that we havent favorited.

### DIFF
--- a/packages/mobile/src/services/offline-downloader/offline-downloader.ts
+++ b/packages/mobile/src/services/offline-downloader/offline-downloader.ts
@@ -130,9 +130,12 @@ export const downloadCollection = async (
   )
 
   if (!user) return
-  store.dispatch(
-    saveCollection(collection.playlist_id, FavoriteSource.OFFLINE_DOWNLOAD)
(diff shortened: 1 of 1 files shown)


amendelsohn (approved):

raymondjacobson (approved):
we check also for tracks elsewhere i assume?

sliptype (approved):

sliptype:
> we check also for tracks elsewhere i assume?

We don't favorite tracks because the only way to download tracks is if they are already favorited or are in a favorited collection, right @amendelsohn ?

amendelsohn:
> > we check also for tracks elsewhere i assume?
> 
> We don't favorite tracks because the only way to download tracks is if they are already favorited or are in a favorited collection, right @amendelsohn ?

Correct. This is only an issue for collections. We have logic to download tracks if they're added to favorites, but we never need to favorite a track automatically.

PR #2720, opened 2023-01-27
[C-1973] Improve offline image downloading

sliptype:
### Description

Improves offline image downloading in the following ways:
* Uses the new method of fetching images via `useContentNodeImage`, this is what the rest of the app is using
  * No longer races multiple requests at once to figure out the best url and then download the winner
  * No longer tries the entire network of content nodes if it fails to find an image
* Only downloads/stores the 1000x1000px image to avoid downloading 3 separate files
* Updates useLocalImage to always use the 1000x1000px image

This results in a pretty big cpu usage reduction. Before when an image request failed the js thread would max out, now the cpu usage doesn't spike at all!

### Dragons

Images affected, potential side effects due to no longer populating `_cover_art_sizes` on the track/collection object. I think everything should be fine though because we only use the metadata from disk when we are offline and we use useLocalImage in those cases

### How Has This Been Tested?

### Feature Flags ###


packages/mobile/src/services/offline-downloader/offline-downloader.ts
```diff
@@ -440,33 +445,57 @@ export const removeTrackDownload = async ({
...
-  const coverArtUris = Object.values(track._cover_art_sizes)
-  await Promise.all(
-    coverArtUris.map(async (coverArtUri) => {
-      const destination = getLocalTrackCoverArtDestination(
-        track.track_id.toString(),
-        coverArtUri
-      )
-      await downloadIfNotExists(coverArtUri, destination)
+const downloadCoverArt =
+  <T extends UserTrackMetadata | UserCollectionMetadata>(
```
amendelsohn, line 449: nice!

packages/mobile/src/services/offline-downloader/offline-storage.ts
```diff
@@ -142,8 +139,8 @@ export const getLocalTrackCoverArtDestination = (
   return path.join(getLocalTrackDir(trackId), getArtFileNameFromUri(uri))
 }
 
-export const getLocalTrackCoverArtPath = (trackId: string, size: string) => {
-  return path.join(getLocalTrackDir(trackId), `${size}.jpg`)
+export const getLocalTrackCoverArtPath = (trackId: string) => {
+  return path.join(getLocalTrackDir(trackId), '1000x1000.jpg')
```
amendelsohn, line 143: should we make this from the size enum or something? It seems pretty unlikely to change though.

amendelsohn: at least a const in this file would be nice actually

amendelsohn (approved):
Wow this is so much simpler! I love it!!

PR #2727, opened 2023-01-30
WIP: Use permalink in frontend routing to fetch playlists

sabrina-kiam:
### Description
Replace front end routing to playlists with new `:handle/playlist/:slug` format. This updates all calls to PlaylistPage() route with a permalink to fetch a playlist via permalink instead of playlist id. 

Note: some things still broken rn
- have to update the use of `this.state.playlistId` throughout CollectionPageProvider since that lots of things still depend on this (editing, sharing, etc a playlist).
- have to solidify the flow for when a playlist is created or edited
- 
### Dragons
This is a big frontend change and will require thorough testing regarding playlists. We'll want to check:
- searching and selecting playlists
- creating playlists
- editing/sharing/liking playlists
- trending playlists/explore page playlists
- user playlists
- playlists loaded from the cache vs retrieved from backend

### How Has This Been Tested?
still being tested against the above cases

### DIFF
--- a/packages/common/src/models/Collection.ts
+++ b/packages/common/src/models/Collection.ts
@@ -49,6 +49,7 @@ export type CollectionMetadata = {
   playlist_image_multihash?: string
   playlist_image_sizes_multihash?: string
   offline?: OfflineCollectionMetadata
+  permalink: string
 }
 
 // This is available on mobile for offline tracks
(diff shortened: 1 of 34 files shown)


sabrina-kiam:
Okay this should be in a reasonable state! 

A few things to note:
- Creating a playlist is still rusty. I will update entity manager so that we write legacy routes (`handle/playlist/playlist-name-id`) to the playlist routes table so that on creation we can immediately route to something. As of now, it creates a playlist and you can view the temporary playlist. However if you refresh or try to go to that permalink directly, it fails to fetch the playlist
- I found a couple playlists that i was unable to go to through search for some reason and couldn't nail down why (`/[Jun28android/playlist/this-boy-renamed-297](http://localhost:3001/Jun28android/playlist/this-boy-renamed-297)` and `[AndyB/playlist/test-playlist.222-868](http://localhost:3001/AndyB/playlist/test-playlist.222-868)`)
- Will also put up a separate diff for updating album page route calls in the front end

PR #2728, opened 2023-01-30
[C-1986] Use RNFetchBlob for offline mode

sliptype:
### Description

Use RNFetchBlob for all offline mode operations. This should improve performance over RNFS because the encoding is done natively

Also a couple small image fixes in here

### Dragons

In `listTracks` and `getOfflineCollections` we are now using `ls` instead of `readDir` which means we can't check whether or not an item is a file/directory. I believe there should only ever be directories here so it should be ok, @amendelsohn can you confirm?

We could also use `lstat` to get more info about the item, similar to `readDir`, but it is slower than `ls`

### How Has This Been Tested?

Tested downloading, loading, and playing back offline tracks

### Feature Flags ###


amendelsohn (approved):
Looks great!

PR #2737, opened 2023-01-31
[C-1938] Remove tracks within collections when toggling off favorites

raymondjacobson:
### Description

When toggling off favorites download, we don't clear all the tracks that could have been downloaded within collections.
This makes that possible, but it's still very slow. Needs something like https://linear.app/audius/issue/C-1974/batch-cancel-downloads-instead-of-1-by-1 to make that better.

### Dragons

### How Has This Been Tested?

Before change:
Downloaded all faves => turned off downloads => some tracks remain downloaded inside collections

After change:
Downloaded all faves => turned off downloads => empty

### Feature Flags ###


dylanjeffers (approved):
yeah looks good. agreed that a single batchRemove would be swell, been thinking about it, and it seems like it would be hard with all the trackJson changes etc that need to happen per delete

PR #2738, opened 2023-01-31
[C-1937, C-1964] Add pagination to the mobile favorites screen and fix the title width for deleted track list items

Kyle-Shanks:
### Description
* Update mobile favorites page to use the sagas in common that were set up for pagination on web to allow for paginated requests
* Add spinner at the bottom when we are fetching more so that the user know that there are more tracks in the list
* Update the component that we have the layout change callback on to fix the deleted track text disappearing bug

### Dragons

This is a sizable update to how favorites are loaded to there is some risk of introducing weird errors states

### How Has This Been Tested?

Lots of manual testing

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/mobile/src/components/track-list/TrackList.tsx
+++ b/packages/mobile/src/components/track-list/TrackList.tsx
@@ -125,13 +125,14 @@ export const TrackList = ({
   const renderTrack: FlatListProps<TrackMetadata>['renderItem'] = ({
     item,
     index
-  }) =>
-    renderDraggableTrack({
+  }) => {
+    return renderDraggableTrack({
       item,
       index,
       drag: () => {},
       isActive: false
     }) as ReactElement
+  }
 
   if (showSkeleton)
     return (
--- a/packages/mobile/src/components/track-list/TrackListItem.tsx
+++ b/packages/mobile/src/components/track-list/TrackListItem.tsx
@@ -222,11 +222,11 @@ export const TrackListItem = ({
         ) : null}
         {isReorderable && <IconDrag style={styles.dragIcon} />}
         <View style={styles.nameArtistContainer}>
-          <View style={styles.topLine}>
-            <View
-              style={styles.trackTitle}
-              onLayout={(e) => setTitleWidth(e.nativeEvent.layout.width)}
-            >
+          <View
+            style={styles.topLine}
+            onLayout={(e) => setTitleWidth(e.nativeEvent.layout.width)}
+          >
+            <View style={styles.trackTitle}>
               <Text
                 numberOfLines={1}
                 style={[styles.trackTitleText, { maxWidth: titleMaxWidth }]}
--- a/packages/mobile/src/screens/favorites-screen/TracksTab.tsx
+++ b/packages/mobile/src/screens/favorites-screen/TracksTab.tsx
@@ -1,7 +1,8 @@
-import { useCallback, useState } from 'react'
+import { useCallback, useEffect, useMemo, useState } from 'react'
 
 import type { ID, UID } from '@audius/common'
 import {
+  Kind,
   useProxySelector,
   savedPageActions,
   playerSelectors,
@@ -16,10 +17,12 @@ import {
   reachabilitySelectors
 } from '@audius/common'
 import { useFocusEffect } from '@react-navigation/native'
+import { debounce } from 'lodash'
 import { useDispatch, useSelector } from 'react-redux'
 
 import { Tile, VirtualizedScrollView } from 'app/components/core'
 import { EmptyTileCTA } from 'app/components/empty-tile-cta'
+import LoadingSpinner from 'app/components/loading-spinner'
 import { TrackList } from 'app/components/track-list'
 import type { TrackMetadata } from 'app/components/track-list/types'
 import { WithLoader } from 'app/components/with-loader/WithLoader'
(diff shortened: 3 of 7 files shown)


packages/mobile/src/screens/favorites-screen/TracksTab.tsx
```diff
@@ -64,16 +74,42 @@ const useStyles = makeStyles(({ palette, spacing }) => ({
...
+
 export const TracksTab = () => {
   const dispatch = useDispatch()
   const styles = useStyles()
   const isReachable = useSelector(getIsReachable)
   const isOfflineModeEnabled = useIsOfflineModeEnabled()
 
-  const handleFetchSavesOnline = useCallback(() => {
-    dispatch(fetchSaves())
+  const [focusFetch, setFocusFetch] = useState(false)
```
sliptype, line 85: A more descriptive name here would be good here, maybe `hasFetchedOnFocus`

packages/mobile/src/components/track-list/TrackList.tsx
```diff
@@ -125,13 +125,18 @@ export const TrackList = ({
...
-      isActive: false
-    }) as ReactElement
+  }) => {
+    if (item.kind === Kind.EMPTY) {
+      return renderSkeletonTrack({ index })
+    } else {
+      return renderDraggableTrack({
+        item,
+        index,
+        drag: () => {},
```
dylanjeffers, line 135: Is it possible to not pass a method here?

packages/mobile/src/components/track-list/TrackList.tsx
```diff
@@ -125,13 +125,18 @@ export const TrackList = ({
...
+        isActive: false
+      }) as ReactElement
```
dylanjeffers, line 137: Is there a way to have the types work without casting?

packages/mobile/src/components/track-list/TrackListItem.tsx
```diff
@@ -222,11 +222,11 @@ export const TrackListItem = ({
...
         {isReorderable && <IconDrag style={styles.dragIcon} />}
         <View style={styles.nameArtistContainer}>
-          <View style={styles.topLine}>
-            <View
-              style={styles.trackTitle}
-              onLayout={(e) => setTitleWidth(e.nativeEvent.layout.width)}
-            >
+          <View
+            style={styles.topLine}
+            onLayout={(e) => setTitleWidth(e.nativeEvent.layout.width)}
```
dylanjeffers, line 227: Wondering is we can avoid the layout calculation for performance?

sliptype (approved):
Awesome! Glad we were able to use the existing common sagas for this 🔥

packages/mobile/src/screens/favorites-screen/TracksTab.tsx
```diff
@@ -93,13 +129,55 @@ export const TracksTab = () => {
...
+    filterValue,
+    isFetchingMore,
+    isLoading,
+    isOfflineModeEnabled,
+    isReachable,
+    tracks.length
+  ])
+
+  useEffect(() => {
+    const allTracks = tracks.every((track) => track.kind === Kind.TRACKS)
```
dylanjeffers, line 161: This does feel like it would be expensive as the list gets long? What do you think?

packages/mobile/src/screens/favorites-screen/TracksTab.tsx
```diff
@@ -146,47 +224,60 @@ export const TracksTab = () => {
...
+        )
+      ) : (
+        <>
+          <OfflineContentBanner />
+          <FilterInput
+            value={filterValue}
+            placeholder={messages.inputPlaceholder}
+            onChangeText={setFilterValue}
+          />
+          <WithLoader loading={initialFetch}>
```
dylanjeffers, line 247: Personally would like to see skeletons render initially, would that be hard?

packages/mobile/src/screens/favorites-screen/TracksTab.tsx
```diff
@@ -146,47 +224,60 @@ export const TracksTab = () => {
...
-                  showDivider
-                  togglePlay={togglePlay}
-                  trackItemAction='save'
-                  tracks={tracks.filter(filterTrack)}
-                  hideArt
+          <EmptyTileCTA message={messages.emptyTabText} />
+        )
+      ) : (
+        <>
+          <OfflineContentBanner />
```
dylanjeffers, line 241: Can all this be part of the flat list header? Also I wonder if we can remove the virtualized scroll view then

packages/mobile/src/screens/favorites-screen/TracksTab.tsx
```diff
@@ -146,47 +224,60 @@ export const TracksTab = () => {
...
+                    togglePlay={togglePlay}
+                    trackItemAction='save'
+                    tracks={filteredTracks}
+                    onEndReachedThreshold={1.5}
+                    onEndReached={handleMoreFetchSaves}
+                    hideArt
+                  />
+                </Tile>
+              ) : null}
+              {isFetchingMore ? (
```
dylanjeffers, line 268: Can this be added to flat list footer?

dylanjeffers (approved):
Awesome stuff, added some extra comments that are probably off the optical variety

packages/mobile/src/screens/favorites-screen/TracksTab.tsx
Kyle-Shanks: simple equality check so doesn't seem too bad. I think that it may run more often than it needs to, but would wait to see if we have perf issues before trying to fix this

packages/mobile/src/screens/favorites-screen/TracksTab.tsx
Kyle-Shanks: More of a product question I suppose. Probably not too difficult, but I was trying to not change things too much

dylanjeffers: Yeah for this go around no issues, just wanted to air me thoughts in case you felt similar

Kyle-Shanks: Yea I like the idea. I'll add an issue in linear to update this later

PR #2740, opened 2023-01-31
Improve useFetchAllFavoritedTracks fetch

dylanjeffers:
### Description

Uses new api that has a limit of 1000, so we dont need to loop. not sure if we need to do the data munging part with with the decodeHashId?

### DIFF
--- a/packages/mobile/src/hooks/useFetchAllFavoritedTracks.ts
+++ b/packages/mobile/src/hooks/useFetchAllFavoritedTracks.ts
@@ -1,6 +1,6 @@
 import { useMemo } from 'react'
 
-import type { APIFavorite, ID } from '@audius/common'
+import type { APIFavorite } from '@audius/common'
 import {
   decodeHashId,
   encodeHashId,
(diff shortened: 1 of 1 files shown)


packages/mobile/src/hooks/useFetchAllFavoritedTracks.ts
```diff
@@ -18,36 +18,31 @@ const { getLocalSaves } = savedPageSelectors
...
-        offset
-      }
-    )
-    const { data: result } = await fetch(url).then((response) =>
-      response.json()
-    )
+  const url = apiClient.makeUrl(
+    `/users/${encodeHashId(currentUserId)}/favorites`,
+    {
+      user_id: currentUserId
```
raymondjacobson, line 33: I think we need to specify limit=10000 here

packages/mobile/src/hooks/useFetchAllFavoritedTracks.ts
```diff
@@ -18,36 +18,32 @@ const { getLocalSaves } = savedPageSelectors
...
-          trackId: decodeHashId(trackSave.favorite_item_id),
-          favoriteCreatedAt: trackSave.created_at
-        }))
-    )
-  }
-  return tracksAndTimestamps
+  return result
+    .filter((trackSave) => trackSave.favorite_type === 'SaveType.track')
+    .map((trackSave: APIFavorite) => ({
+      trackId: decodeHashId(trackSave.favorite_item_id) as number,
```
amendelsohn, line 44: yeah, I think we still need this

packages/mobile/src/hooks/useFetchAllFavoritedTracks.ts
```diff
@@ -18,36 +18,32 @@ const { getLocalSaves } = savedPageSelectors
...
+    created_at: string
+  }[]
+}
+
 export const fetchAllFavoritedTracks = async (currentUserId: number) => {
-  let tracksAndTimestamps: { trackId: ID; favoriteCreatedAt: string }[] = []
-  let loadMore = true
-  let offset = 0
-  // TODO: store results in state to avoid duplicate fetching
-  while (loadMore) {
```
amendelsohn, line 26: I assume we're okay with a hard cap of 10k downloaded favorites? We could keep the while loop but it won't come up much

sliptype (approved):

amendelsohn (approved):

packages/mobile/src/hooks/useFetchAllFavoritedTracks.ts
dylanjeffers: yeah i think lets just cap it!

PR #2751, opened 2023-02-01
[C-1987][C-1936][C-2009] Queue for collection downloads

amendelsohn:
### Description

Changes:
1. Queue for collections download
2. Added download status for collections
3. Collection tracks process in order one by one
4. Separated DownloadStatusIndicator into Collection vs Tracks components
5. Fixed tracks list skeleton count for collection details page
6. Cleaned up selectors for offline mode
7. Added new queued state to collection header

Notable change is the decision to include a component-specific selector in CollectionDownloadStatusIndicator.

### Dragons

Nope


packages/mobile/src/services/offline-downloader/offline-download-queue.ts
```diff
@@ -79,13 +138,52 @@ export const startDownloadWorker = async () => {
...
+}
+
 export const cancelQueuedDownloads = async (
   payloadsToCancel: TrackDownloadWorkerPayload[]
 ) => {
   const payloadsToCancelById = groupBy(
     payloadsToCancel,
     (payload) => payload.trackId
   )
+  queue.stop()
```
raymondjacobson, line 186: why move this up? just to make sure we don't accidentally push something in after we've canceled?

packages/mobile/src/services/offline-downloader/offline-downloader.ts
```diff
@@ -102,74 +100,102 @@ export const downloadAllFavorites = async () => {
...
+  const collectionFromApi = (
+    await apiClient.getPlaylist({
+      playlistId: collectionId,
+      currentUserId
+    })
+  )[0]
+
+  if (!collectionFromApi) {
+    throw failJob(
+      `collection to download not found on discovery - ${collectionFromApi}`
```
raymondjacobson, line 167: heads up, i am working on a PR to throw custom typed error messages here, so we can do different things on different kinds of errors

packages/mobile/src/store/offline-downloads/slice.ts
```diff
@@ -124,19 +156,24 @@ const slice = createSlice({
 })
 
 export const {
+  // TODO: don't name these the same thing
```
raymondjacobson, line 159: wdym?

raymondjacobson (approved):
🏆

packages/mobile/src/services/offline-downloader/offline-download-queue.ts
```diff
@@ -79,13 +138,52 @@ export const startDownloadWorker = async () => {
   queue.start()
 }
 
+export const cancelQueuedCollectionDownloads = async (
```
raymondjacobson, line 141: is this called anywhere?

packages/mobile/src/hooks/useLoadOfflineTracks.ts
```diff
@@ -57,11 +58,16 @@ export const useLoadOfflineData = () => {
     for (const collectionId of offlineCollections) {
       try {
         if (collectionId === DOWNLOAD_REASON_FAVORITES) {
-          dispatch(addCollection({ collectionId, isFavoritesDownload: false }))
+          dispatch(
+            completeCollectionDownload({
+              collectionId,
+              isFavoritesDownload: false
```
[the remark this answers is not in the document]
amendelsohn: this is a special case because the favorites collection itself is not a collection within favorites, so we track it at the top level with collections that were individually marked for download. Kinda confusing though =\ Adding a comment at least

packages/mobile/src/services/offline-downloader/offline-download-queue.ts
amendelsohn: it's so we don't let the workers pick up new jobs while we're deciding which ones to cancel (the inactive ones still in the queue)

packages/mobile/src/store/offline-downloads/slice.ts
amendelsohn: startDownload and batchStartDownload don't actually use the same states. They should be named differently so they don't sound like the single vs batch of the same action.

packages/mobile/src/services/offline-downloader/offline-download-queue.ts
amendelsohn: ooh yep we missed that. Added the call in removeCollectionDownload

PR #2754, opened 2023-02-01
Fix playlist top endpoint with user id

jowlee:
### Description
Updates the client's request to /v1/full/playlists/top to include the encoded user id

NOTE: Pending libs update and version update

### Dragons

### How Has This Been Tested?
Ran against staging and verified that the encoded user_id was a part of the request to the endpoint

### Feature Flags ###

### DIFF
--- a/packages/common/src/services/explore/Explore.ts
+++ b/packages/common/src/services/explore/Explore.ts
@@ -215,7 +215,8 @@ export class Explore {
   async getTopCollections(
     type?: 'playlist' | 'album',
     followeesOnly?: boolean,
-    limit = 20
+    limit = 20,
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://demo.audius.co/jowlee-fix-top-playlist

dylanjeffers (approved):
looks great. curious of reasoning for this, does it exclude artist that person already follows or something?

jowlee:
> looks great. curious of reasoning for this, does it exclude artist that person already follows or something?

it's just so we can migrate to using the URL params instead of x-header which is a legacy way of passing user id. 
This endpoint was just inconsistent

PR #2756, opened 2023-02-01
[PAY-868] Handle follow and tip send states across

sddioulde:
### Description
Make sure follow / unfollow across app updates premium track access accordingly.
Same for tipping.

This is to show correct locked/unlocking/unlocked states for premium tracks across entire app as user may follow/unfollow or send tip from multiple places in app.

### Dragons

n/a

### How Has This Been Tested?

local dapp v stage

n/a

### Feature Flags ###

existing feature flags

### DIFF
--- a/packages/common/src/store/tipping/slice.ts
+++ b/packages/common/src/store/tipping/slice.ts
@@ -156,6 +156,9 @@ const slice = createSlice({
     },
     setShowTip: (state, action: PayloadAction<{ show: boolean }>) => {
       state.showTip = action.payload.show
+    },
+    refreshTipGatedTracks: (_state, _action: PayloadAction<{ userId: ID }>) => {
+      // triggers saga
     }
   }
 })
@@ -178,7 +181,8 @@ export const {
   fetchRecentTips,
   fetchUserSupporter,
   setTipToDisplay,
-  setShowTip
+  setShowTip,
+  refreshTipGatedTracks
 } = slice.actions
 
 export const actions = slice.actions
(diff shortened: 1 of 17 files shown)


audius-infra:
Preview this change https://demo.audius.co/sd-pc-follow-gate

packages/common/src/store/tipping/slice.ts
```diff
@@ -156,6 +156,9 @@ const slice = createSlice({
     },
     setShowTip: (state, action: PayloadAction<{ show: boolean }>) => {
       state.showTip = action.payload.show
+    },
+    refreshTipGatedTracks: (_, __: PayloadAction<{ userId: ID }>) => {
```
rickyrombo, line 160: suuuper nit, but i like to name these `_state` and `_action` for clarity later

packages/web/src/common/store/premiumContent/sagas.ts
```diff
@@ -329,14 +336,102 @@ function* pollPremiumTrack({
...
+  yield* put(updatePremiumTrackStatuses(statusMap))
+
+  const trackIds: ID[] = []
+  const metadatas: { id: ID; metadata: Partial<TrackMetadata> }[] = []
+  Object.keys(statusMap).forEach((trackId) => {
+    const id = parseInt(trackId)
+    trackIds.push(id)
+    metadatas.push({ id, metadata: { premium_content_signature: null } })
+  })
+  yield* put(removePremiumContentSignatures({ trackIds }))
```
rickyrombo, line 419: Feels like it would be easy at some point to introduce a bug by doing one of these actions without the other - I wonder if `removePremiumContentSignature` should have the side-effect of making sure that the cached metadata _also_ removes the premium content signature? Wdyt?

rickyrombo (approved):

dharit-tan (approved):

packages/web/src/common/store/premiumContent/sagas.ts
sddioulde: agreed and handled

PR #2757, opened 2023-02-01
C-1958 Show # of downloaded tracks in collection card instead of actual # of tracks in collection when offline

nicoback2:
### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/screens/favorites-screen/AlbumsTab.tsx
+++ b/packages/mobile/src/screens/favorites-screen/AlbumsTab.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react'
+import { useMemo, useState } from 'react'
 
 import {
   useProxySelector,
@@ -22,6 +22,7 @@ import { FilterInput } from './FilterInput'
 import { NoTracksPlaceholder } from './NoTracksPlaceholder'
 import { OfflineContentBanner } from './OfflineContentBanner'
 import { getAccountCollections } from './selectors'
+import { buildCollectionIdsToNumPlayableTracksMap } from './utils'
 
 const { getIsReachable } = reachabilitySelectors
 
@@ -35,6 +36,18 @@ export const AlbumsTab = () => {
   const isReachable = useSelector(getIsReachable)
   const isOfflineModeEnabled = useIsOfflineModeEnabled()
   const isDoneLoadingFromDisk = useSelector(getIsDoneLoadingFromDisk)
+
+  const offlineDownloadStatus = useProxySelector(
+    (state: AppState) => {
+      if (isDoneLoadingFromDisk && isOfflineModeEnabled && !isReachable) {
+        return getOfflineDownloadStatus(state)
+      }
+      // We don't need offline download status when we're not offline. This saves us rerenders while we're downloading things and updating the offline download slice.
+      return undefined
+    },
+    [isReachable, isOfflineModeEnabled, isDoneLoadingFromDisk]
+  )
+
   const userAlbums = useProxySelector(
     (state: AppState) => {
       if (isOfflineModeEnabled && !isReachable) {
@@ -42,14 +55,13 @@ export const AlbumsTab = () => {
           return []
         }
       }
-      const offlineDownloadStatus = getOfflineDownloadStatus(state)
       return getAccountCollections(state, filterValue).filter((collection) => {
         if (!collection.is_album) {
           return false
         }
         if (isOfflineModeEnabled && !isReachable) {
           const trackIds =
-            collection?.playlist_contents?.track_ids?.map(
+            collection.playlist_contents.track_ids.map(
               (trackData) => trackData.track
             ) ?? []
 
@@ -59,8 +71,9 @@ export const AlbumsTab = () => {
             (trackIds.length === 0 ||
               trackIds.some((t) => {
                 return (
+                  offlineDownloadStatus &&
                   offlineDownloadStatus[t.toString()] ===
-                  OfflineDownloadStatus.SUCCESS
+                    OfflineDownloadStatus.SUCCESS
                 )
               }))
           )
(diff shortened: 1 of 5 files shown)


packages/mobile/src/screens/favorites-screen/PlaylistsTab.tsx
```diff
@@ -60,18 +70,33 @@ export const PlaylistsTab = () => {
...
+      filterValue,
+      isReachable,
+      isOfflineModeEnabled,
+      offlineDownloadStatus,
+      isDoneLoadingFromDisk
+    ]
   )
 
+  const numPlayableTracksMap = useMemo(() => {
+    return buildCollectionIdsToNumPlayableTracksMap(
```
nicoback2, line 93: This solution feels very jank but it's the best I could come up with given how the code is structured, and while keeping CollectionList and its children purely presentational. Very open to other solutions if this is too cumbersome!

packages/mobile/src/screens/favorites-screen/PlaylistsTab.tsx
```diff
@@ -35,6 +36,16 @@ export const PlaylistsTab = () => {
   const isOfflineModeEnabled = useIsOfflineModeEnabled()
   const isReachable = useSelector(getIsReachable)
   const isDoneLoadingFromDisk = useSelector(getIsDoneLoadingFromDisk)
+  const offlineDownloadStatus = useProxySelector(
+    (state: AppState) => {
+      if (isDoneLoadingFromDisk && isOfflineModeEnabled && !isReachable) {
+        return getOfflineDownloadStatus(state)
```
nicoback2, line 42: I'm using `offlineDownloadStatus` as the source of truth, but it seems to be very error prone right now... like it doesn't always match up with what is ultimately shown on the collection screen, which ultimately uses `offlineTracks`. So maybe it makes sense more sense to check `offlineTracks` instead of `offlineDownloadStatus` to be safe? On the other hand, if `offlineDownloadStatus` is indeed always supposed to match `offlineTracks` (i.e. set of tracks with status SUCCESS equals set of tracks in `offlineTracks`), it's fine to keep it this way. Wdyt?

packages/mobile/src/screens/favorites-screen/AlbumsTab.tsx
```diff
@@ -55,18 +67,33 @@ export const AlbumsTab = () => {
...
+    [
+      filterValue,
+      isReachable,
+      isOfflineModeEnabled,
+      offlineDownloadStatus,
+      isDoneLoadingFromDisk
+    ]
   )
 
+  const numPlayableTracksMap = useMemo(() => {
```
dylanjeffers, line 89: could this just be trackCount, and just pass that into the collectionList? it seems the collection list does the manual work of extracting the num, but could be cleaner to just pass a number into the list?

packages/mobile/src/screens/favorites-screen/PlaylistsTab.tsx
dylanjeffers: @amendelsohn i think you might have some ideas here given our discussion about source of truth between cached tracks and offlineTracks?

packages/mobile/src/screens/favorites-screen/PlaylistsTab.tsx
```diff
@@ -35,6 +36,16 @@ export const PlaylistsTab = () => {
   const isOfflineModeEnabled = useIsOfflineModeEnabled()
   const isReachable = useSelector(getIsReachable)
   const isDoneLoadingFromDisk = useSelector(getIsDoneLoadingFromDisk)
+  const offlineDownloadStatus = useProxySelector(
```
dylanjeffers, line 39: based on convos with @sliptype, i think the best setup for playlistsTab would be for its only job to be getting collectionIds that we want to render, online or offline, and pass that into the card list, and then each playlist card would be in charge of getting the tracks, online or offline, which we can then derive the count etc. @nicoback curious how you feel about that as the paradigm here? would it be too much of a refactor?

dylanjeffers (approved):
approving cause i dont want to block you @nicoback, but im personally hoping we can consider a refactor for these components in general based on the comment below, let me know what youre thinking

packages/mobile/src/screens/favorites-screen/PlaylistsTab.tsx
dylanjeffers: yeah i think this feeling is potentially resolved with the ideas below?

packages/mobile/src/screens/favorites-screen/AlbumsTab.tsx
nicoback2: @dylanjeffers  well the collection list takes a list of collections, so we can't just pass a single number (since each collection has a different number of tracks). but i feel like im misunderstanding what you're saying?

dylanjeffers: No worries, maybe a quick huddle?

packages/mobile/src/screens/favorites-screen/PlaylistsTab.tsx
nicoback2: it'd be a bit of a refactor just because collection card list is used in several places elsewhere in the code (like 10 places), not just the Playlists/Albums tabs. But I'm down to do it, it wouldn't be too bad. cc @sliptype

nicoback2: Like def agree that is the way to go lol. This code is pretty suboptimal.

packages/mobile/src/screens/favorites-screen/PlaylistsTab.tsx
nicoback2: ooh interested (also add offlineDownloadStatus to that list if we're just talking source of truth for whether a track should be shown when offline or not)

packages/mobile/src/screens/favorites-screen/AlbumsTab.tsx
nicoback2: 💯, pretty sure i had the same thought when writing this lol

packages/mobile/src/screens/favorites-screen/PlaylistsTab.tsx
nicoback2: Ooh gotcha. Yeah thinking about it tho, i think it wouldn't take too much time. @sliptype if you aren't working on this, i can pick it up first thing tomorrow and timebox to half an hour or so

sliptype: I haven't gotten a chance to look at the CardList yet, got a couple of other refactor to knock out first. Yeah @nicoback definitely feel free to look at it if you want, but maybe let's see if there is anything else more pressing on the backlog first

nicoback2: oh yeah there are definitely more pressing things. ill leave this open and come back to it after hitting more high pri items then

PR #2759, opened 2023-02-02
C-2002 Un-download deleted playlists on sync

nicoback2:
### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/services/offline-downloader/offline-sync.ts
+++ b/packages/mobile/src/services/offline-downloader/offline-sync.ts
@@ -22,6 +22,7 @@ import {
   downloadCollectionCoverArt,
   downloadTrackCoverArt,
   DOWNLOAD_REASON_FAVORITES,
+  removeCollectionDownload,
   removeDownloadedCollectionFromFavorites
 } from './offline-downloader'
 import { purgeDownloadedTrack, writeTrackJson } from './offline-storage'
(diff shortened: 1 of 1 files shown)


packages/mobile/src/services/offline-downloader/offline-sync.ts
```diff
@@ -160,14 +161,29 @@ export const syncCollectionTracks = async (
...
+
+  const downloadedCollectionTrackIds =
+    offlineCollection.tracks
+      ?.map((track) => track.track_id)
+      ?.filter((trackId) => downloadedTracks[trackId] != null) ?? []
+
+  if (updatedCollection.is_delete) {
+    removeCollectionDownload(
+      collectionId,
+      downloadedCollectionTrackIds.map((trackId) => ({
```
nicoback2, line 174: @dylanjeffers @amendelsohn , not sure if it made more sense to use `downloadedCollectionTrackIds` here or `downloadedOrQueuedCollectionTrackIds`. (This is for the `tracksForDownload` param for `removeCollectionDownload`.)

dylanjeffers: good q, def defer to @amendelsohn on this one

amendelsohn: Either one is likely fine, but we have a check in `downloadCollection` for is_delete against the latest version from discovery, so I think it's fine to just do the downloaded ones and not the queued ones. `removeCollectionDownload` does cancel the queued job, so it should work fine either way. The advantage of doing it now is you don't have to wait for the queue to drain, but it will cost more up front.

Overall, my instinct is to use `downloadedCollectionTrackIds`, but I do think either would be functional.

amendelsohn (approved):

PR #2760, opened 2023-02-02
Refactor lineup hooks

sliptype:
### Description

* Refactor `useOfflineCollectionLineup` and `useOfflineFavoritesLineup` into `useCollectionLineup` and `useFavoritesLineup` at the component level
* Still passing in the function the fetch the data from remote because we need it in TracksTab to handle all the pagination
   Might make sense to push all the pagination and filter logic into `useFavoritesLineup` but wanted to keep this pr fairly small. @dylanjeffers what do you think about this?

### Dragons

relatively safe

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/screens/collection-screen/useCollectionLineup.ts
+++ b/packages/mobile/src/screens/collection-screen/useCollectionLineup.ts
@@ -1,0 +1,114 @@
+import { useCallback } from 'react'
+
+import type { SmartCollectionVariant } from '@audius/common'
+import {
+  useProxySelector,
+  Kind,
+  makeUid,
+  cacheActions,
+  cacheCollectionsSelectors,
+  collectionPageLineupActions,
+  collectionPageSelectors,
+  lineupSelectors,
+  reachabilitySelectors
+} from '@audius/common'
+import moment from 'moment'
+import { useDispatch, useSelector } from 'react-redux'
+
+import { useIsOfflineModeEnabled } from 'app/hooks/useIsOfflineModeEnabled'
+import { useReachabilityEffect } from 'app/hooks/useReachabilityEffect'
+import { store } from 'app/store'
+import { getOfflineTracks } from 'app/store/offline-downloads/selectors'
+
+const { getCollection } = cacheCollectionsSelectors
+const { getCollectionTracksLineup } = collectionPageSelectors
+const { makeGetTableMetadatas } = lineupSelectors
+const { getIsReachable } = reachabilitySelectors
+
+const getTracksLineup = makeGetTableMetadatas(getCollectionTracksLineup)
+
+/**
+ * Returns a collection lineup, supports boths online and offline
+ * @param collectionId the numeric collection id
+ */
+export const useCollectionLineup = (
+  collectionId: number | SmartCollectionVariant | null,
+  fetchLineup: () => void
+) => {
+  const dispatch = useDispatch()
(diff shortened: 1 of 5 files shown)


packages/mobile/src/screens/collection-screen/useCollectionLineup.ts
```diff
@@ -0,0 +1,114 @@
...
+) => {
+  const dispatch = useDispatch()
+  const isOfflineModeEnabled = useIsOfflineModeEnabled()
+  const offlineTracks = useSelector(getOfflineTracks)
+  const isReachable = useSelector(getIsReachable)
+  const collection = useSelector((state) => {
+    return getCollection(state, { id: collectionId as number })
+  })
+  const collectionTracks = useSelector(getCollectionTracksLineup)
+  const collectionTrackUidMap = collectionTracks.entries.reduce(
```
dylanjeffers, line 46: would it be more performant to make the trackUidMap in the useSelector? maybe need an equality check there or somethnig?

packages/mobile/src/screens/favorites-screen/TracksTab.tsx
```diff
@@ -90,27 +89,25 @@ export const TracksTab = () => {
...
-      fetchSavesOnline()
+      fetchSaves()
     }
-  }, [isReachable, fetchSavesOnline])
+  }, [isReachable, fetchSaves])
 
-  const savedTrackUids: string[] = useSelector((state) => {
-    return getSavedTracksLineup(state).entries.map(({ uid }) => uid)
-  }, isEqual)
+  const { entries } = useFavoritesLineup(fetchSaves)
```
dylanjeffers, line 109: so if im understanding, the fetchSaves whether called in this method or above will ultimately update the entries we select and return in this hook? makes sense but yeah a bit annoying we have to do a `useEffect` here as well. so could we potentially just rerun the `fetchSaves` call in the `useFavoritesLineup` when fetchSaves itself is a new function? maybe that's even more complicated. tldr im fine with this, but just asking qs

dylanjeffers (approved):
i like these improvements a lot. given the track-list fetch function with the useEffect, all good not adding it into the hook itself for now

packages/mobile/src/screens/collection-screen/useCollectionLineup.ts
sliptype: Yeah probably, this isn't causing rerenders but probably expensive to compute on every render

packages/mobile/src/screens/favorites-screen/TracksTab.tsx
sliptype: Yeah that's the path I was going down but it means pulling all the pagination and filtering into the hook as well. So then the hook would accept the filterValue, and return the lineup along with a fetchMore function

sliptype: Otherwise there would still be a fetchMoreSaves in the component, so the fetch logic would be split into two places

sliptype: Rerunning fetchSaves when it changes becomes complicated bc we usually only want to run it on reachability change. I think passing in filterValue and moving all the pagination and filtering into the hook would be cleaner. I'll think on it some more

PR #2764, opened 2023-02-02
[C-1974] Add batch remove downloads

dylanjeffers:
### Description

Adds batch remove downloads flow
Interested to see what yall think regarding the architecture. In this case, remove-downloads-button dispatches an action, and that action runs a saga which then batch updates the redux store, which triggers more sagas to save to disk. With this architecture we have a potentially simpler data flow.

TODO: need to fit in the queue cancel. would like to talk about that a bit to make sure we are cancelling the correct jobs

### DIFF
--- a/packages/mobile/src/store/offline-downloads/sagas/removeAllDownloadedFavoritesSaga.ts
+++ b/packages/mobile/src/store/offline-downloads/sagas/removeAllDownloadedFavoritesSaga.ts
@@ -43,14 +43,71 @@ export function* watchRemoveAllDownloadedFavorites() {
 }
 
 function* removeAllDownloadedFavoritesWorker() {
-  const tracksToDequeue = yield* call(removeAllFavoritesTracks)
-  const collectionsToDequeue = yield* call(removeAllFavoritesCollections)
+  const { collectionsToRemove, collectionsToUpdate, collectionsToDequeue } =
+    yield* call(removeFavoritedCollections)
+
+  const { tracksToRemove, tracksToUpdate, tracksToDequeue } = yield* call(
+    removeFavoritedTracks
+  )
+
+  if (collectionsToRemove.length > 0) {
+    yield* put(
+      removeCollectionDownloads({ collectionIds: collectionsToRemove })
+    )
+  }
+  if (collectionsToUpdate.length > 0) {
+    yield* put(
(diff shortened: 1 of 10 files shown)


packages/mobile/src/store/offline-downloads/sagas/removeAllDownloadedFavoritesSaga.ts
```diff
@@ -0,0 +1,127 @@
...
+function* removeAllFavoritesTracks() {
+  const offlineTracks = yield* select(getOfflineTracks)
+  const offlineTrackList = Object.keys(offlineTracks).map(
+    (offlineTrackId) => offlineTracks[offlineTrackId]
+  )
+
+  const tracksToRemove: ID[] = []
+  const reasonsToUpdate: TrackReasonsToUpdate[] = []
+
+  // TODO: double check if "favorites" collection is in here
```
sliptype, line 49: Is this done?

packages/mobile/src/store/offline-downloads/sagas/removeAllDownloadedFavoritesSaga.ts
```diff
@@ -0,0 +1,127 @@
...
+  const offlineFavoritedCollections = yield* select(
+    getOfflineFavoritedCollections
+  )
+  const favoritedCollectionIds = new Set([
+    'favorites',
+    ...Object.keys(offlineFavoritedCollections)
+  ])
+
+  offlineTrackList.forEach((offlineTrack) => {
```
sliptype, line 58: Apparently `for ... of` is more performant than `.forEach`

sliptype (approved):
I like this architecture a lot! Also having the sagas in separate named files is so nice

packages/mobile/src/store/offline-downloads/sagas/removeAllDownloadedFavoritesSaga.ts
dylanjeffers: yes thanks

packages/mobile/src/store/offline-downloads/sagas/removeAllDownloadedFavoritesSaga.ts
dylanjeffers: down with that

PR #2766, opened 2023-02-02
[C-2015] Debounce the callbacks for the download toggles for favorites and collections

Kyle-Shanks:
### Description

Debounced the callback passed to the download switches for favorites and collections to try to prevent bad states when flipping the toggle rapidly.

Choose a slightly longer debounce time to handle slight lag when pressing the toggle, but willing to update the time to whatever we think makes sense here

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/mobile/src/screens/collection-screen/CollectionHeader.tsx
+++ b/packages/mobile/src/screens/collection-screen/CollectionHeader.tsx
@@ -9,13 +9,13 @@ import {
   collectionsSocialActions,
   Variant
 } from '@audius/common'
-import { debounce } from 'lodash'
 import { View } from 'react-native'
 import { useSelector, useDispatch } from 'react-redux'
 
 import { Switch, Text } from 'app/components/core'
 import { getCollectionDownloadStatus } from 'app/components/offline-downloads/CollectionDownloadStatusIndicator'
 import { DownloadStatusIndicator } from 'app/components/offline-downloads/DownloadStatusIndicator'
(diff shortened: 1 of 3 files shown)


packages/mobile/src/screens/collection-screen/CollectionHeader.tsx
```diff
@@ -207,7 +209,7 @@ const OfflineCollectionHeader = (props: OfflineCollectionHeaderProps) => {
       </View>
       <View style={styles.headerRight}>
         <Switch
-          value={isMarkedForDownload}
+          defaultValue={isMarkedForDownload}
```
Kyle-Shanks, line 212: updated this to defaultValue so that the rapid presses of the toggle would display the proper value instead of flipping back to the real value and looking buggy

dylanjeffers (dismissed):
just for fun, can we use https://github.com/streamich/react-use/blob/master/docs/useDebounce.md ?

Kyle-Shanks:
> just for fun, can we use https://github.com/streamich/react-use/blob/master/docs/useDebounce.md ?

Seems like using this would require us to create state to track the downloadEnabled status that the toggle passes back, the debounced state, and add some effects just for listening to the state update and doing the same function call. Not difficult, but is that worth doing/is there a reason to it that way?

Kyle-Shanks:
Updated to add and use the useDebouncedCallback hook

sliptype (approved):

PR #2768, opened 2023-02-02
[C-1970] Stop downloading after going offline

nicoback2:
### Description
See comments!

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/offline-downloads/OfflineDownloader.tsx
+++ b/packages/mobile/src/components/offline-downloads/OfflineDownloader.tsx
@@ -38,7 +38,6 @@ export const OfflineDownloader = () => {
 
   const isReachable = useSelector(getIsReachable)
 
-  // Move this logic to watch set (un)rechable sagas?:
   useEffect(() => {
     if (!initialized) return
     const isQueueRunning = queue.isRunning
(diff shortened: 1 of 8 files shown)


packages/mobile/src/store/offline-downloads/sagas.ts
```diff
@@ -208,15 +209,15 @@ export function* startSync() {
 }
 
 export function* handleSetReachable() {
-  yield* call(setPlayCounterWorker, playCounterWorker)
```
nicoback2, line 211: These deletions fix the issue because setPlayCounterWorker calls `queue.start()`, so we were starting the queue again after stopping it in a different part of the code.

nicoback2: But I'm unsure why this code for the play counter worker was here in the first place? We don't do this blockedPlayCounterWorker thing for the other workers, and I couldn't tell what made it necessary for the play counter worker.

packages/mobile/src/components/offline-downloads/OfflineDownloader.tsx
```diff
@@ -29,6 +29,7 @@ export const OfflineDownloader = () => {
 
   const isReachable = useSelector(getIsReachable)
 
+  // Move this logic to watch set (un)rechable sagas?:
```
nicoback2, line 32: Part of the reason this happened was because this "go offline effect" logic existed in two separate places, here and the sagas file. So I think I'm gonna move the logic in this effect to the handleSet(Un)reachable sagas unless y'all think it works better here!

nicoback2:
Found a more comprehensive solution. Closing this, ignore!

PR #2769, opened 2023-02-03
[C-2012] [C-1999] [C-1734] Fix collection image loading

sliptype:
### Description

* Add backgroundColor when displaying the fallback image for tracks and collections. It was invisible because it was white on white
* Fix loading of collection images on web profile and mobile favorites
  * In the saga responsible for fetching the collection images, we were mutating the collection returned from the selector. This meant that the selector at the profile page level (and others) were not running because the updated collection cache entries were equal to the ones already existing in the store
  * I believe this bug was revealed because we reduced redundant rerenders using better selector equality checks https://github.com/AudiusProject/audius-client/commit/82c465dd19980a8bce47730b06e0c4a50df40c0f

### Dragons

Are we mutating selected values anywhere else and putting them back in the cache??

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/hooks/useContentNodeImage.ts
+++ b/packages/mobile/src/hooks/useContentNodeImage.ts
@@ -10,6 +10,7 @@ import { audiusBackendInstance } from 'app/services/audius-backend-instance'
 export type ContentNodeImageSource = {
   source: ImageSourcePropType
   handleError: () => void
+  isFallbackImage: boolean
 }
 
 export const isImageUriSource = (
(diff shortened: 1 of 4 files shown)


packages/mobile/src/hooks/useContentNodeImage.ts
```diff
@@ -163,27 +165,25 @@ export const useContentNodeImage = ({
     }
   }, [imageSourceIndex, imageSources])
 
+  const showFallbackImage = useMemo(() => {
+    return !user || !cid || failedToLoad
+  }, [failedToLoad, user, cid])
+
   const source = useMemo(() => {
-    if (!user || !cid || failedToLoad) {
+    if (showFallbackImage) {
```
raymondjacobson, line 173: nice

raymondjacobson (approved):

PR #2770, opened 2023-02-03
[C-1941] Refresh queue when download toggle is flipped

raymondjacobson:
### Description

This was kinda tricky!
Architecturally, it's simple, refresh the queue stuff when the toggle is flipped to fix the bug -- this is because we use UIDs to check whether the queue should change. We need to keep the UIDs the same to allow playback to work when reachability state changes.

There was no exposure to the collection that is playing from the queue, so I had to add that, and then do some changes within Audio.tsx to respect toggles to the switch contextually based on what you're listening to.

Could be more elegant, but I think not terrible given the way this comes together.

### Dragons

### How Has This Been Tested?

Scrubber too fast on short song I played, but I promise both played during this repro!

https://user-images.githubusercontent.com/2731362/216489719-28d91a26-91d7-4f6b-9bb6-b6add1d03d19.mp4

### Feature Flags ###

### DIFF
--- a/packages/common/src/store/queue/selectors.ts
+++ b/packages/common/src/store/queue/selectors.ts
@@ -1,6 +1,7 @@
 import { createSelector } from 'reselect'
 
 import { UID } from '../../models'
+import { Uid } from '../../utils/uid'
 import { cacheUsersSelectors, cacheTracksSelectors } from '../cache'
 import { playerSelectors } from '../player'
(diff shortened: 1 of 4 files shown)


Kyle-Shanks (approved):

audius-infra:
Preview this change https://demo.audius.co/rj-c-1941

PR #2772, opened 2023-02-03
Improve download toggle, download progress and fetch

raymondjacobson:
### Description

Commits can be reviewed in isolation:
- [C-2026] Make it so we download all tracks from favorites + from collections within favorites in one fell swoop so download progress shows X/Y and Y is the complete number of tracks. Otherwise, you get Y in two parts
- [C-2028 & C-2027] Same fix: Add another state to consider downloads abandoned. Problem is that tracks that enter ERROR state get swept from the downloadTracks so your downloadprogress Y number goes down and the download spinner thinks you've never finished
- [C-2029] Quick little optimize to download art + track at same time.

### Dragons

### How Has This Been Tested?

Simulator

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/services/offline-downloader/offline-downloader.ts
+++ b/packages/mobile/src/services/offline-downloader/offline-downloader.ts
@@ -353,9 +353,11 @@ export const downloadTrack = async (trackForDownload: TrackForDownload) => {
       return
     }
 
-    await downloadTrackCoverArt(trackFromApi)
+    await Promise.all([
+      downloadTrackCoverArt(trackFromApi),
+      tryDownloadTrackFromEachCreatorNode(trackFromApi)
+    ])
 
-    await tryDownloadTrackFromEachCreatorNode(trackFromApi)
     const now = Date.now()
     const trackToWrite: Track & UserTrackMetadata = {
       ...trackFromApi,
--- a/packages/mobile/src/services/offline-downloader/offline-downloader.ts
+++ b/packages/mobile/src/services/offline-downloader/offline-downloader.ts
@@ -24,7 +24,8 @@ import { getAccountCollections } from 'app/screens/favorites-screen/selectors'
 import { store } from 'app/store'
 import {
   getOfflineCollections,
-  getOfflineFavoritedCollections
+  getOfflineFavoritedCollections,
+  getTrackOfflineDownloadStatus
 } from 'app/store/offline-downloads/selectors'
 import {
   actions as offlineDownloadsActions,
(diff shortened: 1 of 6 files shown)


packages/mobile/src/services/offline-downloader/offline-downloader.ts
```diff
@@ -335,9 +353,11 @@ export const downloadTrack = async (trackForDownload: TrackForDownload) => {
       return
     }
 
-    await downloadTrackCoverArt(trackFromApi)
+    await Promise.all([
```
dylanjeffers, line 356: This makes sense for speed, but does it mean we have more goin on in background? Prob not a big deal either way?

packages/mobile/src/store/offline-downloads/slice.ts
```diff
@@ -98,7 +106,11 @@ const slice = createSlice({
       state.downloadStatus[trackId] = OfflineDownloadStatus.ERROR
     },
     removeDownload: (state, { payload: trackId }: PayloadAction<string>) => {
-      delete state.downloadStatus[trackId]
+      if (state.downloadStatus[trackId] === OfflineDownloadStatus.ABANDONED) {
```
dylanjeffers, line 109: So in this case delete doesn't fully delete status so we still have same denominator? When it it's tried to be removed again though we delete? Little confused!

dylanjeffers (approved):
Nice ideas here, it all makes sense to me,

packages/mobile/src/services/offline-downloader/offline-downloader.ts
raymondjacobson: js & react native should be really good at handling this, lots of idle IO time, especially as the art will come back before the track in 99% of cases so the parsing of the response happens while the other request is just idling

packages/mobile/src/store/offline-downloads/slice.ts
raymondjacobson: I didn't love this part a super ton, but basically we issue the same `removeDownload` calls from the worker erroring as well as when we actually try to delete/purge things. I figured that this was sorta ok. As in, the first pass removing puts things from ERROR to ABANDONED and then the second pass would actually delete them

:/ do you think this would be better as a separate action? I don't want to refactor a ton, but that might be better?

dylanjeffers: All good! I agree it's a bit confusing but just wanted to understand intention for now

raymondjacobson: i fix better

PR #2773, opened 2023-02-03
C-2022 C-1970 Fix for: When offline and not fully downloaded, queued jobs keep running and fail

nicoback2:
### Description
Before: If you went offline after starting a download, or starting the app offline while you still had tracks to download, the queued jobs would keep running and fail.

After: The queue is stopped if you're offline.

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/services/offline-downloader/workers/playCounterWorker.ts
+++ b/packages/mobile/src/services/offline-downloader/workers/playCounterWorker.ts
@@ -3,6 +3,8 @@ import queue, { Worker } from 'react-native-job-queue'
 
 import { store } from 'app/store'
 
+import { startQueueIfOnline } from './utils'
+
 const { recordListen } = tracksSocialActions
 
 export const PLAY_COUNTER_WORKER = 'play_counter_worker'
@@ -20,7 +22,7 @@ export const setPlayCounterWorker = async (
   queue.stop()
   queue.removeWorker(PLAY_COUNTER_WORKER)
   queue.addWorker(worker)
-  await queue.start()
+  await startQueueIfOnline()
 }
 
 export const playCounterWorker = new Worker(PLAY_COUNTER_WORKER, countPlay)
--- a/packages/mobile/src/services/offline-downloader/workers/playCounterWorker.ts
+++ b/packages/mobile/src/services/offline-downloader/workers/playCounterWorker.ts
@@ -1,9 +1,7 @@
 import { tracksSocialActions } from '@audius/common'
-import queue, { Worker } from 'react-native-job-queue'
+import { Worker } from 'react-native-job-queue'
 
 import { store } from 'app/store'
-
-import { startQueueIfOnline } from './utils'
 
 const { recordListen } = tracksSocialActions
 
@@ -14,15 +12,6 @@ export type PlayCountWorkerPayload = { trackId: number }
 const countPlay = async (payload: PlayCountWorkerPayload) => {
   const { trackId } = payload
   store.dispatch(recordListen(trackId))
-}
-
-export const setPlayCounterWorker = async (
-  worker: Worker<PlayCountWorkerPayload>
-) => {
-  queue.stop()
-  queue.removeWorker(PLAY_COUNTER_WORKER)
-  queue.addWorker(worker)
-  await startQueueIfOnline()
 }
 
 export const playCounterWorker = new Worker(PLAY_COUNTER_WORKER, countPlay)
--- a/packages/mobile/src/services/offline-downloader/workers/playCounterWorker.ts
+++ b/packages/mobile/src/services/offline-downloader/workers/playCounterWorker.ts
@@ -26,10 +26,3 @@ export const setPlayCounterWorker = async (
 }
 
 export const playCounterWorker = new Worker(PLAY_COUNTER_WORKER, countPlay)
-export const blockedPlayCounterWorker = new Worker(
-  PLAY_COUNTER_WORKER,
-  countPlay,
-  {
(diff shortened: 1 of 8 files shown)


amendelsohn:
> After: The queue is stopped if you're offline.

❤️

packages/mobile/src/services/offline-downloader/workers/utils.ts
```diff
@@ -0,0 +1,14 @@
+import { reachabilitySelectors } from '@audius/common'
+import queue from 'react-native-job-queue'
+
+import { store } from 'app/store'
+
+const getIsReachable = reachabilitySelectors.getIsReachable
+
+export const startQueueIfOnline = async () => {
```
nicoback2, line 8: can make this a takeLatest saga, thoughts?

packages/mobile/src/store/offline-downloads/sagas.ts
```diff
@@ -216,6 +217,7 @@ export function* watchSetReachable() {
 }
 
 export function* handleSetUnreachable() {
+  // Why do we do this?
   yield* call(setPlayCounterWorker, blockedPlayCounterWorker)
```
amendelsohn, line 221: I think you won't need this if the queue is properly stopped while offline.
This was a hack to keep a playCounterWorker attached to the queue that won't process any tracks. The blockedPlayCounterWorker has concurrency 0 so it won't pick up tracks.

packages/mobile/src/services/offline-downloader/workers/playCounterWorker.ts
```diff
@@ -20,7 +22,7 @@ export const setPlayCounterWorker = async (
   queue.stop()
   queue.removeWorker(PLAY_COUNTER_WORKER)
   queue.addWorker(worker)
-  await queue.start()
+  await startQueueIfOnline()
```
amendelsohn, line 25: should be able to remove `setPlayCounterWorker`

packages/mobile/src/services/offline-downloader/workers/utils.ts
dylanjeffers: yeah good q, what do we think? it seems in some cases we want the queue to start after we stopped it to do some queue removal methods, and other times we probably want to ensure it starts/stops due to reachability changes?

packages/mobile/src/store/offline-downloads/sagas.ts
```diff
@@ -208,6 +208,7 @@ export function* startSync() {
 }
 
 export function* handleSetReachable() {
+  // Why do we do this?
```
dylanjeffers, line 211: @amendelsohn ^

packages/mobile/src/services/offline-downloader/workers/trackDownloadWorker.ts
```diff
@@ -38,14 +40,23 @@ const onFailure = async (
...
-      queue.start()
+      startQueueIfOnline()
       break
     }
     default:
       break
   }
 }
 
+export const setTrackDownloadWorker = async (
```
amendelsohn, line 51: this doesn't appear to be used

dylanjeffers (approved):
This all makes a lot of sense. i agree it might make sense to have a saga that checks for reachability changes and starts/stops based on that, as well as the individual start/stop checks?

packages/mobile/src/store/offline-downloads/sagas.ts
nicoback2: Discussed @ daily sync! im currently testing to see if we can remove this without repercussions

packages/mobile/src/services/offline-downloader/workers/utils.ts
```diff
@@ -0,0 +1,14 @@
...
+  const state = store.getState()
+  const reachable = getIsReachable(state)
+  if (reachable) {
+    return queue.start()
```
amendelsohn, line 12: Because this can fail when we're not online, we could miss a window where we intended to start the queue but skipped it. I think we may need something to start the queue whenever we become reachable. This could also be a problem though, since some operations want to stop the queue temporarily. We might need some sort of mutex?

packages/mobile/src/services/offline-downloader/workers/utils.ts
amendelsohn: Yes, exactly.

I do think we need to be careful to respect the temporary queue.stop() ... queue.start() pattern we use while doing cancel operations

packages/mobile/src/services/offline-downloader/workers/utils.ts
nicoback2: @amendelsohn  we do have already an effect to start the queue whenever we become reachable 
https://github.com/AudiusProject/audius-client/blob/nkang--queue/packages/mobile/src/components/offline-downloads/OfflineDownloader.tsx#L41-L50

packages/mobile/src/services/offline-downloader/workers/trackDownloadWorker.ts
nicoback2: oh shoot!

packages/mobile/src/services/offline-downloader/workers/utils.ts
amendelsohn: Ah, you're right. And I'm slowly convincing myself that the race condition is invalid... not sure

nicoback2:
ok ive spent hella long testing this and it seems to be all good. going to merge as is and we can always make improvements later.

PR #2775, opened 2023-02-03
[C-2034] Cancel all download jobs on log out

amendelsohn:
### Description

Cancel all download worker jobs on log out.

Questions:
Should we cancel play count jobs? I think these are okay, but we might record them for the wrong user once we're back online. Maybe need to tag the job with the user id? This is also an edge case so maybe okay to drop them and cancel the jobs.


packages/mobile/src/services/offline-downloader/offline-download-queue.ts
```diff
@@ -224,3 +224,19 @@ export const cancelQueuedDownloads = async (
...
+  jobs.forEach((rawJob) => {
+    const { workerName, id } = rawJob
+    if (
+      workerName === TRACK_DOWNLOAD_WORKER ||
+      workerName === COLLECTION_DOWNLOAD_WORKER
+    ) {
+      queue.cancelJob(id)
+    }
+  })
+  // TODO: call startIfOffline
```
amendelsohn, line 240: should be start if online. Waiting on merge from https://github.com/AudiusProject/audius-client/pull/2773

nicoback2: merged!

sliptype (approved):
Looks good to me, should we address that todo?

amendelsohn:
> Looks good to me, should we address that todo?

Yep it's on the way

PR #2783, opened 2023-02-06
Fix remix

sddioulde:
### Description

2 issues:
- using a setter incorrectly, passing in function instead of boolean, as if it were from useState
- wrong function name which wasn't caught due to lack of TypeScript usage

### Dragons

### How Has This Been Tested?

local dapp vs staging

n/a

### Feature Flags ###

n/a


dylanjeffers (approved):
Sweet

PR #2784, opened 2023-02-06
[C-2011] move whole tracks out of offline slice

amendelsohn:
### Description

Changes to the way we store offline tracks in redux.
1. offlineMetadata (`track.offline`) is stored in the offlineDownloads slice
2. tracks are stored in the tracks cache
3. `getOfflineTracks` selector merges the offlineMetadata from the offline slice with the full track from the cache and returns the track with offline populated

Other changes:
1. Renamed actions for clarity and consistency
2. Updated usages of `getOfflineTracks` to use the more specific selectors and respect the new return type

TODO:
- [x] Ensure the user is added to the users cache whenever we read from offline
- [x] Merge the user from cache whenever we write track to disk

### Dragons

Cache track is not necessarily the same version we have on disk as it is shared with non-offline_mode use cases


packages/mobile/src/store/offline-downloads/slice.ts
```diff
@@ -188,43 +193,83 @@ const slice = createSlice({
...
+      state,
+      action: PayloadAction<TrackOfflineMetadataPayload[]>
+    ) => {
+      const { payload: trackOfflineMetadatas } = action
+      trackOfflineMetadatas.forEach((trackOfflineMetadata) => {
+        const { trackId, offlineMetadata } = trackOfflineMetadata
+        state.offlineTrackMetadata[trackId] = offlineMetadata
+      })
+    },
+    batchAddTrackDownloadReason: (
```
amendelsohn, line 206: @dylanjeffers these are placeholders as we currently use `batchSetTrackOfflineMetadata`, but I know you're waiting on these for your changes. Maybe we can sync on the best structure

dylanjeffers: sweet, yeah i like batchSetTrackOfflineMetadata

amendelsohn: do you want me to remove the others? we don't use the add/remove download reason anymore in my change

sliptype (approved):
Awesome work! Looks good to me

are the todos in the pr description done?

PR #2791, opened 2023-02-07
[C-2079] Fix mobile sign up analytics

sliptype:
### Description

The way we were firing sign up complete analytics in a useEffect in the SignOnScreen did not always trigger because the screen was removed as soon as an account existed (in RootScreen) 

This pr moves the analytics and notification reminder into a mobile specific saga that watches SIGN_UP_SUCCEEDED. We've followed this pattern for other mobile specific sagas but open to thoughts/comments. Also @dylanjeffers I followed the pattern you did for offline mode sagas

Edit: I think sign up needs an audit / some love in general

### Dragons

Specifically did not make any changes that affect web. This should be fairly safe

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/store/sign-up/sagas/watchSignUpSucceededSaga.ts
+++ b/packages/mobile/src/store/sign-up/sagas/watchSignUpSucceededSaga.ts
@@ -1,0 +1,38 @@
+import { make } from 'common/store/analytics/actions'
+import { SIGN_UP_SUCCEEDED } from 'common/store/pages/signon/actions'
+import {
+  getEmailField,
+  getHandleField
+} from 'common/store/pages/signon/selectors'
+import { takeEvery, put, select } from 'typed-redux-saga'
+
+import { remindUserToTurnOnNotifications } from 'app/components/notification-reminder/NotificationReminder'
+import { EventNames } from 'app/types/analytics'
+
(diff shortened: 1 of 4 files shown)


packages/mobile/src/store/sign-up/sagas/watchSignUpSucceededSaga.ts
```diff
@@ -0,0 +1,38 @@
...
+  )
+
+  yield put(
+    make(EventNames.CREATE_ACCOUNT_FINISH, {
+      emailAddress: emailField.value,
+      handle: handleField.value
+    })
+  )
+
+  remindUserToTurnOnNotifications(put)
```
dylanjeffers, line 37: do we want to yield call this? lol interesting we pass it put!

dylanjeffers (approved):
overall love the refactor!

packages/mobile/src/store/sign-up/sagas/watchSignUpSucceededSaga.ts
sliptype: I don't think we need to, just need to put this action in a best effort fashion. Nothing else is waiting on this saga

Yeah lol, it accepted dispatch from the component but `put` works the same!

PR #2793, opened 2023-02-07
Fix audio tx modal where metadata is null

dharit-tan:
### Description
In some rare cases, the tx metadata doesn't exist on identity (usually due to bugs). This fixes client so that it doesn't crash when this occurs.

### Dragons
Fallout from this fix: the modal will start at 580px then suddenly shrink when this case occurs. I think this is ok to ignore though as this is a rare case.

### How Has This Been Tested?

Local web client.

### Feature Flags ###

### DIFF
--- a/packages/web/src/common/store/pages/audio-transactions/sagas.ts
+++ b/packages/web/src/common/store/pages/audio-transactions/sagas.ts
@@ -8,6 +8,7 @@ import {
   transactionDetailsActions,
   getContext
 } from '@audius/common'
+import type { InAppAudioPurchaseMetadata } from '@audius/common'
 import { AudiusLibs, full } from '@audius/sdk'
 import { call, takeLatest, put } from 'typed-redux-saga'
 
@@ -151,15 +152,18 @@ function* fetchTransactionMetadata() {
     fetchAudioTransactionMetadata.type,
     function* (action: ReturnType<typeof fetchAudioTransactionMetadata>) {
       const { txDetails } = action.payload
+      if (txDetails.transactionType !== TransactionType.PURCHASE) {
+        return
+      }
       yield* call(waitForLibsInit)
       const libs: AudiusLibs = yield* call(audiusBackendInstance.getAudiusLibs)
-      const response = (yield* call(
+      const response: InAppAudioPurchaseMetadata = yield* call(
         [
           libs.identityService!,
           libs.identityService!.getUserBankTransactionMetadata
         ],
         txDetails.signature
-      )) as any[]
+      )
       yield put(
         fetchTransactionDetailsSucceeded({
           transactionId: txDetails.signature,
@@ -167,7 +171,7 @@ function* fetchTransactionMetadata() {
             ...txDetails,
             // If metadata does not exist on identity, mark as null to indicate
             // that fetch was attempted but failed.
-            metadata: response[0]?.metadata ?? null
+            metadata: response ?? null
           }
         })
       )
--- a/packages/web/src/common/store/pages/audio-transactions/sagas.ts
+++ b/packages/web/src/common/store/pages/audio-transactions/sagas.ts
@@ -6,7 +6,8 @@ import {
   formatDate,
   StringAudio,
   transactionDetailsActions,
-  getContext
+  getContext,
+  Nullable
 } from '@audius/common'
 import type { InAppAudioPurchaseMetadata } from '@audius/common'
 import { AudiusLibs, full } from '@audius/sdk'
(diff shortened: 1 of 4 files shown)


packages/web/src/components/transaction-details-modal/components/TransactionDetailsContent.tsx
```diff
@@ -97,6 +97,10 @@ const UserDetails = ({ userId }: UserDetailsProps) => {
 const dateAndMetadataBlocks = (transactionDetails: TransactionDetails) => {
   switch (transactionDetails.transactionType) {
     case TransactionType.PURCHASE: {
+      // Rare case: metadata will be null if failed to fetch from identity.
+      if (transactionDetails.metadata === null) {
+        return
+      }
       return (
         <>
           <TransactionPurchaseMetadata
```
piazzatron, line 106: Should be able to take out the `!` on the line below now

packages/web/src/common/store/pages/audio-transactions/sagas.ts
```diff
@@ -153,19 +153,21 @@ function* fetchTransactionMetadata() {
...
+      )) as any[]
       yield put(
         fetchTransactionDetailsSucceeded({
           transactionId: txDetails.signature,
           transactionDetails: {
             ...txDetails,
-            metadata: (response as any[])[0].metadata
+            // If metadata does not exist on identity, mark as null to indicate
+            // that fetch was attempted but failed.
+            metadata: response.length === 0 ? null : response[0].metadata
```
piazzatron, line 170: nit: `response[0]?.metadata ?? null` maybe cleaner

piazzatron (approved):

packages/web/src/components/transaction-details-modal/components/TransactionDetailsContent.tsx
dharit-tan: lint was complaining cuz i think it could technically still be undefined.

piazzatron: above check as `if (!transactionDetails.metadata)` might do it then

audius-infra:
Preview this change https://demo.audius.co/rt-fix-tx

packages/web/src/common/store/pages/audio-transactions/sagas.ts
rickyrombo: can it just be undefined? do we need to explicitly have it be null?

packages/web/src/common/store/pages/audio-transactions/sagas.ts
```diff
@@ -153,19 +153,21 @@ function* fetchTransactionMetadata() {
...
       const libs: AudiusLibs = yield* call(audiusBackendInstance.getAudiusLibs)
-      const response = yield* call(
+      const response = (yield* call(
         [
           libs.identityService!,
           libs.identityService!.getUserBankTransactionMetadata
         ],
         txDetails.signature
-      )
+      )) as any[]
```
rickyrombo, line 162: I must have missed this `any` before! hopefully this helps to fix? https://github.com/AudiusProject/audius-protocol/pull/4731

packages/web/src/components/transaction-details-modal/components/TransactionDetailsContent.tsx
```diff
@@ -97,11 +97,13 @@ const UserDetails = ({ userId }: UserDetailsProps) => {
...
+      // Rare case: metadata will be null if failed to fetch from identity.
+      if (!transactionDetails.metadata) {
+        return
+      }
       return (
         <>
-          <TransactionPurchaseMetadata
-            metadata={transactionDetails.metadata!}
-          />
+          <TransactionPurchaseMetadata metadata={transactionDetails.metadata} />
```
rickyrombo, line 106: Feels like it should still render the TransactionPurchaseMetadata, which should take in `InAppPurchaseMetadata | undefined` and appropriately default to empty/unknown values for everything.

dharit-tan: Good call.

packages/web/src/common/store/pages/audio-transactions/sagas.ts
dharit-tan: I'd still have to cast though right? Bc type in identity != type in client, we'd have to convert all uses of this type in client to the identity in type to avoid cast unless I'm mistaken?

rickyrombo: As long as the types are equivalent, it should be ok

rickyrombo (approved):

audius-infra:
Preview this change https://demo.audius.co/rt-fix-tx

PR #2795, opened 2023-02-08
Decode artist pick IDs and fix artist pick reads from discovery

michellebrier:
### Description
Noticed I didn't encode artist pick track IDs in the user API response... added in discovery in https://github.com/AudiusProject/audius-protocol/pull/4725; these are the accompanying client changes. Will remove the conditional and 'number' type when all discovery nodes have updated.

Also noticed reading artist pick from discovery wasn't working when viewing other people's profiles. This was because we had added a line manually pruning the artist pick field when fetching and caching the user's metadata sometime between now and when I initially added the flag in. Does not affect viewing one's own profile because that info is fetched and cached in getAccount in AudiusBackend.ts.

Going to hotfix this change onto the 1.5.9 release so that we can turn on reads then delete the legacy identity path for artist pick in the next client release.

### Dragons

### How Has This Been Tested?
Tested reads and writes with and without the backend changes encoding the IDs.

### Feature Flags ###

### DIFF
--- a/packages/common/src/models/User.ts
+++ b/packages/common/src/models/User.ts
@@ -10,8 +10,8 @@ import { Timestamped } from './Timestamped'
 
 export type UserMetadata = {
   album_count: number
-  artist_pick_track_id: number | null
-  bio: string | null
+  artist_pick_track_id: Nullable<number>
+  bio: Nullable<string>
   cover_photo: Nullable<CID>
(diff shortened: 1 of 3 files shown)


dylanjeffers (approved):
All makes sense, thank you for the great description and comments

PR #2803, opened 2023-02-08
[C-2083] Add offline sync to queue

dylanjeffers:
### Description

Improves offline sync process by adding collection syncs to the queue. This is necessary to prevent fetching all the collections on app load each time. Instead we add those processes to the queue, spacing out the requests. Moves all the collection code from "updateStaleOfflineDataSaga" into the "syncCollectionWorker"

Also adds a new collection endpoint `getCollectionMetadata` which does not include all the track metadata, saving potentially a ton of network bandwidth for large collections.

### Dragons

There is a really good callout about tracks that are deemed unplayable, ie deleted/unlisted/gated, and that if we simply remove them, each collection sync would re-add that track to the queue, which in turn will fail on it continuously. not something we need to handle strictly here though.


packages/mobile/src/store/offline-downloads/sagas/downloadQueueSagas/syncCollectionWorker.ts
```diff
@@ -0,0 +1,222 @@
...
+  startCollectionSync
+} from '../../slice'
+
+const { SET_UNREACHABLE } = reachabilityActions
+const { getUserId } = accountSelectors
+const { getCollection } = cacheCollectionsSelectors
+
+const isTrackFavoriteReason = (downloadReason: DownloadReason) =>
+  downloadReason.is_from_favorites &&
+  downloadReason.collection_id === DOWNLOAD_REASON_FAVORITES
```
Kyle-Shanks, line 39: should this be `||`? I may be wrong and would defer to @amendelsohn but i thought that favorites would not have `is_from_favorites` set to true so this would never return true?

amendelsohn: a favorited track will be `{ is_from_favorites: true, collection_id: 'favorites' }` so this is correct

Kyle-Shanks: yes i see below now. thanks 👍

Kyle-Shanks (approved):
great changes, nice

packages/common/src/services/audius-api-client/AudiusAPIClient.ts
```diff
@@ -1047,6 +1060,19 @@ export class AudiusAPIClient {
...
 
+  async getFavorites({ currentUserId, limit }: GetFavoritesArgs) {
+    this._assertInitialized()
+    const encodedUserId = encodeHashId(currentUserId)
+    const params = { user_id: encodedUserId, limit }
+    const response = await this._getResponse<APIResponse<APIFavorite[]>>(
+      FULL_ENDPOINT_MAP.userFavorites(encodedUserId),
+      params
+    )
+    if (!response) return []
```
amendelsohn, line 1071: I think we need to return null here. Explanation below

packages/common/src/services/audius-api-client/AudiusAPIClient.ts
```diff
@@ -1170,6 +1196,26 @@ export class AudiusAPIClient {
     return adapted
   }
 
+  async getCollectionMetadata({
+    collectionId,
+    currentUserId
+  }: GetCollectionMetadataArgs) {
+    this._assertInitialized()
+    const encodedCurrentUserId = encodeHashId(currentUserId)
+    const encodedCollectionId = this._encodeOrThrow(collectionId)
```
amendelsohn, line 1205: is there a reason to use encode vs encodeOrThrow? Odd to have both here

packages/mobile/src/store/offline-downloads/sagas/downloadQueueSagas/downloadTrackWorker.ts
```diff
@@ -99,11 +98,5 @@ function* downloadTrackAsync(
     return OfflineDownloadStatus.ERROR
   }
 
-  const verified = yield* call(verifyTrack, trackId.toString(), true)
```
amendelsohn, line 102: is there a reason we don't want to do this anymore?

packages/mobile/src/store/offline-downloads/sagas/downloadQueueSagas/syncCollectionWorker.ts
```diff
@@ -0,0 +1,222 @@
...
+  }
+}
+
+function* syncFavoritesCollection() {
+  const currentUserId = yield* select(getUserId)
+  const apiClient = yield* getContext('apiClient')
+  if (!currentUserId) return CollectionSyncStatus.ERROR
+  const offlineTrackMetadata = yield* select(getOfflineTrackMetadata)
+
+  const latestFavoritedTracks = yield* call(
```
amendelsohn, line 85: It's important that we return null if the getFavorites call fails or gives no response and that we stop the favorites sync in that case. Otherwise if the API call fails, we'll get `latestFavoritedTracks = []` and trackIdsToRemove will be all downloaded tracks. The `[]` return value needs to be reserved for the case where we got a successful response but the user has no favorited tracks on discovery.

amendelsohn: I think you've handled this properly in syncCollection as we return null from the APIClient and also have checks for empty data.

amendelsohn (approved):
Awesome work! Just fix the `[]` bug and I think we'll be set.

packages/common/src/services/audius-api-client/AudiusAPIClient.ts
dylanjeffers: yeah no idea, just copied from the method below for playlists. ill just do encode honestly

packages/mobile/src/store/offline-downloads/sagas/downloadQueueSagas/downloadTrackWorker.ts
dylanjeffers: yeah i thought that there was no way our methods above adding content to the filesystem would run without throwing if there was an issue with validity? basically thinking verifyTrack makes the most sense on the rehydrate phase

packages/mobile/src/store/offline-downloads/sagas/downloadQueueSagas/syncCollectionWorker.ts
dylanjeffers: yes wow great find! updating

PR #2807, opened 2023-02-09
[C-1984] Don't use nativeDriver for slider and tracking bar anims

sliptype:
### Description
image

* Don't use nativeDriver for slider and tracking bar anims
  Both were crashing but only the error for tracking bar was logged because the app crashed afterwards
  The serialized animations for an hour long track have over 300k properties and hermes has an upper bound
  https://reactnative.dev/docs/animations#using-the-native-driver

### Dragons

Slightly less performant, but I checked and there was not a significant difference in cpu usage or app responsiveness

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/now-playing-drawer/TrackingBar.tsx
+++ b/packages/mobile/src/components/now-playing-drawer/TrackingBar.tsx
@@ -59,7 +59,10 @@ export const TrackingBar = (props: TrackingBarProps) => {
     currentAnimation.current = Animated.timing(translateXAnimation.current, {
       toValue: 1,
       duration: timeRemaining * 1000,
-      useNativeDriver: true
+      // Can't use native driver because this animation is potentially hours long,
+      // and would have to be serialized into an array to be passed to the native layer.
+      // The array exceeds the number of properties allowed in hermes
+      useNativeDriver: false
     })
 
     currentAnimation.current.start()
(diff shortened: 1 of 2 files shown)


packages/mobile/src/components/now-playing-drawer/TrackingBar.tsx
```diff
@@ -59,7 +59,10 @@ export const TrackingBar = (props: TrackingBarProps) => {
     currentAnimation.current = Animated.timing(translateXAnimation.current, {
       toValue: 1,
       duration: timeRemaining * 1000,
-      useNativeDriver: true
+      // Can't use native driver because this animation is potentially hours long,
+      // and would have to be serialized into an array to be passed to the native layer.
+      // The array exceeds the number of properties allowed in hermes
+      useNativeDriver: false
```
raymondjacobson, line 65: damn this is WILD. maybe we can do something based on track length? regardless. holy hell great work.

raymondjacobson (approved):

packages/mobile/src/components/now-playing-drawer/TrackingBar.tsx
sliptype: Maybe but I think serializing even a 3 minute animation is potentially expensive. I wonder if reanimated would have the same problem

dylanjeffers (approved):
Wow awesome find here! It does indeed make sense why native would matter here, good stuff

dylanjeffers (commented):
Would we also want to update the the useNativeDriver in the slider? Seems like we wouldn't want to send that massive data over as well?

sliptype:
> Would we also want to update the the useNativeDriver in the slider? Seems like we wouldn't want to send that massive data over as well?

Yeah I did both, or did I miss another case of this somewhere?

PR #2813, opened 2023-02-09
Fix onPressRail animation error

dylanjeffers:
### Description

Fixes issue where pressing the slider rail throws error, this is due to us changing the animation type for `useNativeDriver: false`


sliptype (approved):
Good catch!

PR #2814, opened 2023-02-09
[C-2067] Add 'play-count' queue

amendelsohn:
### Description

Add play-count type to offline queue.
This was removed in the queue refactor.

### Dragons

1. Do we need error handling? Is best-effort good enough?
2. Is the same queue state logic sufficient for this? Process while online, stop while offline.

### How Has This Been Tested?

- [ ] TODO - untested


dylanjeffers (approved):
I kinda like fire and forget, so this looks good to me, assuming your testing works!

PR #2816, opened 2023-02-10
[C-2040] Improve duration counter accuracy/performance

dylanjeffers:
### Description

Adds interesting perf improvement to the position-timestamp on the now-playing drawer. 

By directly manipulating the text using `setNativeProps` we can avoid triggering a rerender every second as the track is playing, and can drastically improve perf when dragging scrubber, which was causing 40+ rerenders a second.

Also improves scrubber perf by memorizing the panHandlers

### DIFF
--- a/packages/mobile/src/components/scrubber/PositionTimestamp.tsx
+++ b/packages/mobile/src/components/scrubber/PositionTimestamp.tsx
@@ -1,50 +1,35 @@
-import { useEffect } from 'react'
+import { forwardRef } from 'react'
 
-import type { Nullable } from '@audius/common'
-import { formatSeconds } from '@audius/common'
-import { useStreamPosition } from 'react-native-google-cast'
-import { useProgress } from 'react-native-track-player'
+import type { Text as TextInputProps } from 'react-native'
+import { TextInput, View } from 'react-native'
 
-import { Text } from 'app/components/core'
 import { makeStyles } from 'app/styles'
 
-type PositionTimestampProps = {
-  dragSeconds: Nullable<string>
-  setDragSeconds: (seconds: Nullable<string>) => void
-}
+type PositionTimestampProps = Partial<TextInputProps>
 
-const useStyles = makeStyles(({ palette }) => ({
+const useStyles = makeStyles(({ palette, typography }) => ({
   timestamp: {
     width: 50,
     color: palette.neutral,
-    fontSize: 12,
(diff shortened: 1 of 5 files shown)


packages/mobile/src/components/scrubber/PositionTimestamp.tsx
```diff
@@ -1,50 +1,35 @@
...
-      {dragSeconds || formatSeconds(position)}
-    </Text>
-  )
-}
+export const PositionTimestamp = forwardRef<TextInput, PositionTimestampProps>(
+  (props, ref) => {
+    const styles = useStyles()
+    return (
+      <View pointerEvents='none'>
+        <TextInput
```
dylanjeffers, line 26: Note, the hack is using text input since you can't directly change text component children

packages/mobile/src/components/scrubber/usePosition.ts
```diff
@@ -0,0 +1,57 @@
...
+    const updatePosition = () => {
+      const timeout = setTimeout(() => {
+        positionRef.current += 1
+
+        positionElementRef.current?.setNativeProps({
+          text: formatSeconds(positionRef.current)
+        })
+
+        if (positionRef.current <= duration) {
+          currentTimeout = updatePosition()
```
sliptype, line 35: Could you maybe add a comment about how outerTimeout and currentTimeout interact, it's kinda confusing

packages/mobile/src/components/scrubber/PositionTimestamp.tsx
sliptype: Does the styling still look right? And the input isn't editable right?

sliptype (approved):
So cool!

packages/mobile/src/components/scrubber/PositionTimestamp.tsx
dylanjeffers: Yeah it looks the same from what I can tell, and yes the view with pointerEvents none prevents its interaction

sliptype: Check android 👀

packages/mobile/src/components/scrubber/usePosition.ts
dylanjeffers: yeah totally agreed

PR #2819, opened 2023-02-10
[PAY-779][PAY-781] Add premium track page and tile on mobile - Part 2

sddioulde:
### Description

Part 2 of the work that adds premium track page and tile to mobile.
https://github.com/AudiusProject/audius-client/pull/2818 will merge into this one and this one will merge into https://github.com/AudiusProject/audius-client/pull/2820.

### Dragons

n/a

### How Has This Been Tested?

local ios dapp against stage

### How will this change be monitored?

n/a

### Feature Flags ###

existing premium content feature flags

PR #2820, opened 2023-02-10
[PAY-779][PAY-781] Add premium track page and tile on mobile - Part 1

sddioulde:
### Description

Broke up the many changes into 3 PRs.
This one is Part 1 of the work that adds premium track page and tile to mobile.
https://github.com/AudiusProject/audius-client/pull/2819 will merge into this one.

### Dragons

n/a

### How Has This Been Tested?

local ios dapp against stage

n/a

### Feature Flags ###


packages/common/src/hooks/usePremiumContentAccess.ts
```diff
@@ -0,0 +1,60 @@
...
+  return { isUserAccessTBD, doesUserHaveAccess }
+}
+
+export const useSpecialAccessEntity = (premiumConditions: Nullable<PremiumConditions>) => {
+  const { follow_user_id: followUserId, tip_user_id: tipUserId } =
+    premiumConditions ?? {}
+
+  const users = useSelector((state: CommonState) =>
+    getUsers(state, {
+      ids: [followUserId, tipUserId].filter((id): id is number => !!id)
```
dharit-tan, line 47: Not quite following this, my understanding is `id is number` proves that `id` is of type `number`? Can't just use `removeNullable` or something?
is it bc `followerUserId` comes from the backend and we don't know the type yet?

packages/mobile/src/components/details-tile/DetailsTile.tsx
```diff
@@ -179,8 +188,16 @@ export const DetailsTile = ({
   saveCount,
   headerText,
   title,
-  user
+  user,
+  track
 }: DetailsTileProps) => {
+  const isPremiumContentEnabled = useIsPremiumContentEnabled()
+  const { doesUserHaveAccess } = usePremiumContentAccess(
+    track ? (track as unknown as Track) : null
```
dharit-tan, line 196: Why do we need the cast here?

packages/mobile/src/components/details-tile/DetailsTileHasAccess.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+const DetailsTileUnlockedSection = ({
+  premiumConditions
+}: {
+  premiumConditions: PremiumConditions
+}) => {
+  const styles = useStyles()
+  const neutral = useColor('neutral')
+  const { followee, tippedUser } = useSpecialAccessEntity(premiumConditions)
+  console.log('premiumConditions is', premiumConditions, followee, tippedUser)
```
dharit-tan, line 77: Intentional?

packages/mobile/src/components/details-tile/DetailsTileHasAccess.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+  const renderDescription = useCallback(() => {
+    if (premiumConditions.nft_collection) {
+      return (
+        <View>
+          <Text style={styles.description}>yolo</Text>
```
dharit-tan, line 83: hahaha

packages/mobile/src/components/details-tile/DetailsTileHasAccess.tsx
```diff
@@ -0,0 +1,160 @@
...
+        <View style={styles.descriptionContainer}>
+          <Text>
+            <Text style={styles.description}>
+              {messages.unlockedTipGatedPrefix}
+            </Text>
+            <Text style={styles.description}>{tippedUser.name}</Text>
+            <UserBadges
+              badgeSize={16}
+              user={tippedUser}
+              nameStyle={styles.description}
```
dharit-tan, line 105: does `nameStyle` do anything if `hideName` is true?

dharit-tan: fwiw I've been able to get the name to look how I want using `nameStyle` recently, instead of including the name separately as `Text`. Remember there being some issues on web tho.

packages/mobile/src/components/details-tile/DetailsTilePremiumAccess.tsx
```diff
@@ -0,0 +1,60 @@
+import { useMemo } from 'react'
```
dharit-tan, line 1: Not seeing where this component is used in `DetailsTile.tsx`, assuming it's in a later change.

packages/mobile/src/screens/track-screen/PremiumTrackCornerTag.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const styles = useStyles()
+
+  if (isOwner) {
+    if (premiumConditions.nft_collection) {
+      return <IconCollectibleCornerTag style={styles.icon} />
+    }
+    return <IconSpecialAccessCornerTag style={styles.icon} />
+  }
+
+  if (doesUserHaveAccess) {
```
dharit-tan, line 34: Question for product - if the user has unlocked a track, do we not want to show them something to let them know that this is a special thing?

dharit-tan (approved):

packages/common/src/hooks/usePremiumContentAccess.ts
sddioulde: it's a neat little trick that says that the ids that satisfy the filter condition should be considered as number. otherwise TS may complain and still think that the id, even after filter, could be undefined/null

packages/mobile/src/components/details-tile/DetailsTileHasAccess.tsx
sddioulde: removed in subsequent PRs, thanks.

packages/mobile/src/components/details-tile/DetailsTile.tsx
sddioulde: because it could alse be a `SearchTrack`, but we only care about certain fields, which both types have

packages/mobile/src/components/details-tile/DetailsTileHasAccess.tsx
sddioulde: lmao, removed in subsequent PRs. will double check these funny strings before merging, but they should all be gone

packages/mobile/src/screens/track-screen/PremiumTrackCornerTag.tsx
sddioulde: we show them a label on the top right hand side of the tile, or in the big tile on track screen. they'll know it's a collectible gated or special access track

packages/mobile/src/components/details-tile/DetailsTileHasAccess.tsx
sddioulde: ah nice. will look into this

packages/mobile/src/components/details-tile/DetailsTile.tsx
piazzatron, line 196: agreed w reed re this cast – the `as unknown as x` cast is a bit of a smell, any way we can do this without this cast?

packages/common/src/hooks/usePremiumContentAccess.ts
piazzatron: does our `removeNullable` type helper not work?

packages/mobile/src/store/sagas.ts
```diff
@@ -103,6 +104,9 @@ export default function* rootSaga() {
     // Tipping
     ...tippingSagas(),
 
+    // Premium content
+    ...premiumContentSagas(),
```
piazzatron, line 108: How come we have premiumContentSagas in both mobile + common sagas?

packages/common/src/hooks/usePremiumContentAccess.ts
```diff
@@ -0,0 +1,60 @@
...
+      !!track.premium_content_signature ||
+      !!(trackId && premiumTrackSignatureMap[trackId])
+    const isCollectibleGated = !!track.premium_conditions?.nft_collection
+    const isSignatureToBeFetched =
+      isCollectibleGated &&
+      !!trackId &&
+      premiumTrackSignatureMap[trackId] === undefined
+
+    return {
+      isUserAccessTBD: !hasPremiumContentSignature && isSignatureToBeFetched,
```
piazzatron, line 33: are there any ways we could model this state more accurately? i.e. are there impossible combinations of `isUserAccessTBD` and `doesUserHaveAccess` (tbd: true & haveAccess: true?). could we model this with a string union something like `accessState: unknown | accessible | denied`?

packages/common/src/hooks/usePremiumContentAccess.ts
```diff
@@ -0,0 +1,60 @@
...
+    })
+  )
+  const followee = useMemo(
+    () => (followUserId ? users[followUserId] : null),
+    [users, followUserId]
+  )
+  const tippedUser = useMemo(
+    () => (tipUserId ? users[tipUserId] : null),
+    [users, tipUserId]
+  )
```
piazzatron, line 57: appreciate that we're trying to be careful with `useMemo` but if we're just accessing a map here not really necessary

piazzatron (approved):

packages/mobile/src/components/details-tile/DetailsTile.tsx
```diff
@@ -236,6 +266,7 @@ export const DetailsTile = ({
   return (
     <Tile styles={{ root: styles.root, content: styles.tileContent }}>
       <View style={styles.topContent}>
+        {renderCornerTag?.()}
```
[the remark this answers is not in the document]
sddioulde: optional

packages/mobile/src/components/details-tile/DetailsTile.tsx
sddioulde: agreed with yall, but i dont think it's a big deal here. We know the track is Track or SearchTrack, so if it exists we can just use the cast here. would spend more time refactoring this but i dont think it's hi-pri at this point...

packages/mobile/src/store/sagas.ts
sddioulde: not sure wym. followed the chatSagas pattern here. wasn't aware that there is redundancy?

PR #2822, opened 2023-02-10
[C-2087] Add stale track queue worker

dylanjeffers:
### Description

Adds a stale track queue worker, to ensure we don't eat up the main thread potentially making tons of requests for new track metadata.

Updates when the staleTrackSaga runs, which is now a part of the sync saga, to prevent double dispatches of adding items to the queue. now we merge in the stale track actions with the other sync actions.


sliptype (approved):
Awesome! So we are still potentially gonna queue up a ton of tracks when things become stale right? Should we add a limit?

dylanjeffers:
Yeah I love the idea of a limit on top of the queuing. What we thinking @amendelsohn @raymondjacobson ?

amendelsohn:
I like the idea of limiting it and/or bumping the stale time to 30d instead of 7

raymondjacobson:
> I like the idea of limiting it and/or bumping the stale time to 30d instead of 7

seems fine

PR #2824, opened 2023-02-11
[C-2107] Add local saves to offline queue

dylanjeffers:
### Description

Ensures local saves are enqueued for download when toggling favorites downloads on

### DIFF
--- a/packages/mobile/src/store/offline-downloads/sagas/requestDownloadAllFavoritesSaga.ts
+++ b/packages/mobile/src/store/offline-downloads/sagas/requestDownloadAllFavoritesSaga.ts
@@ -1,4 +1,8 @@
-import { accountSelectors, getContext } from '@audius/common'
+import {
+  accountSelectors,
+  getContext,
+  savedPageSelectors
+} from '@audius/common'
 import { takeEvery, select, call, put } from 'typed-redux-saga'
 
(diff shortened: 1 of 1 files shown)


packages/mobile/src/store/offline-downloads/sagas/requestDownloadAllFavoritesSaga.ts
```diff
@@ -25,31 +31,49 @@ function* downloadAllFavorites() {
...
+  const trackReasonsForDownload = [
+    { is_from_favorites: true, collection_id: DOWNLOAD_REASON_FAVORITES }
+  ]
+
+  // Add local saves
+  const localSaves = yield* select(getLocalSaves)
+  const localSaveIds = Object.keys(localSaves).map((id) => parseInt(id, 10))
+  const now = Date.now()
+
+  const localSavesToAdd: OfflineItem[] = localSaveIds.map((id) => ({
```
sliptype, line 43: Can this be combined with the prev map?

sliptype (approved):
Nice!

packages/mobile/src/store/offline-downloads/sagas/requestDownloadAllFavoritesSaga.ts
dylanjeffers: yeah def, will do all in one shot

PR #2826, opened 2023-02-11
[C-2108] Improve local image fetch

dylanjeffers:
### Description

Improves local image fetch by using track/collection status to determine if image uri is available


sliptype (approved):
Yessss, good stuff

amendelsohn (approved):
Sweet

PR #2827, opened 2023-02-11
[C-2102] Improve online/offline transitions

dylanjeffers:
### Description

Improves component animations/transitions when going from online to offline and vis versa. happy to post a video if thatll make it more clear

### DIFF
--- a/packages/mobile/src/screens/favorites-screen/OfflineContentBanner.tsx
+++ b/packages/mobile/src/screens/favorites-screen/OfflineContentBanner.tsx
@@ -1,5 +1,6 @@
 import { reachabilitySelectors } from '@audius/common'
 import { View } from 'react-native'
+import Animated, { SlideOutUp } from 'react-native-reanimated'
 import { useSelector } from 'react-redux'
 
 import IconNoWifi from 'app/assets/images/iconNoWifiSmall.svg'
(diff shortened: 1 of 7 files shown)


packages/mobile/src/screens/favorites-screen/OfflineContentBanner.tsx
```diff
@@ -43,7 +44,7 @@ export const OfflineContentBanner = () => {
   const isReachable = useSelector(getIsReachable)
   if (!isOfflineModeEnabled || isReachable) return null
   return (
-    <View>
+    <Animated.View exiting={SlideOutUp.duration(1000)}>
```
amendelsohn, line 47: should it slide in too?

amendelsohn (approved):
Oh that's super easy. Very cool!

packages/mobile/src/screens/favorites-screen/OfflineContentBanner.tsx
dylanjeffers: So when experimenting it slides in due to the surrounding layout transition, but needed this for the disappearance to look smooth. lets def play with it on device when up to make sure it looks clean

amendelsohn: ah, okay cool

PR #2829, opened 2023-02-13
Set offline mode flag default true

amendelsohn:
Prepare flag for release
default: true is necessary for offline cases


sliptype (approved):

amendelsohn:
cherry-picked to release

PR #2830, opened 2023-02-13
[C-2109] Fix favorited collection not downloading

dylanjeffers:
### Description

Fixes issue where favoriting a collection was not triggering download. this was due to missing `yield*` before `takeEvery`.-
- Adds missing watchUnfavoriteCollectionSaga, potentially this should just replace the "RemoveFavoritedDownloadedCollectionDrawer" flow since that is disabled now... @amendelsohn 
- Fixes formatting of "favorited_at"
- Improves sync process by not only checking if new tracks have been added to collection, but also checks to see if some tracks were not initiated for download before. I was noticing this was happening, potentially due to us queuing up a download with a stale cached playlist. Not exactly how we get into this place, but i think it is a good addition to always check offilne status in addition of tracks in addition to the list of tracks from current collection.


packages/mobile/src/store/offline-downloads/sagas/requestDownloadCollectionSaga.ts
```diff
@@ -20,6 +20,7 @@ export function* requestDownloadCollectionSaga() {
 
 function* downloadCollection(action: CollectionAction) {
   const { collectionId } = action.payload
+  // TODO should we fetch from api instead?
```
amendelsohn, line 23: I think yes

amendelsohn: should we make a shared method that does the logic below and reuse it in `requestDownloadFavoritedCollectionSaga`?

packages/mobile/src/store/offline-downloads/sagas/watchSaveTrackSaga.ts
```diff
@@ -15,28 +15,29 @@ function* downloadSavedTrack(
...
-          id: trackId,
-          metadata: {
-            favorite_created_at: moment().format('YYYY-MM-DD HH:mm:ss'),
-            reasons_for_download: [
-              {
-                is_from_favorites: true,
-                collection_id: DOWNLOAD_REASON_FAVORITES
-              }
-            ]
+  if (isFavoritesDownloadEnabled) {
```
amendelsohn, line 22: negate to reduce nesting?

amendelsohn (approved):
Looks good!

Not sure about the RemoveFavoritedDownloadedCollectionDrawer. Is that case no longer possible?

packages/mobile/src/store/offline-downloads/sagas/watchSaveTrackSaga.ts
dylanjeffers: im down, but in this case its really just an else if at the end so i almost saw it as more clear to "if this case run this small block"?

packages/mobile/src/store/offline-downloads/sagas/requestDownloadCollectionSaga.ts
dylanjeffers: yes i agree, a little worried about the differences with the download favorites i guess? but suppose we can just add an arg there?

amendelsohn: hmm let's just keep it separate then. No need to overdo reusability

packages/mobile/src/store/offline-downloads/sagas/watchSaveTrackSaga.ts
dylanjeffers: yeah i usually will negate if the next block is long and it continues, but since this is the end, i changed it, not sure why but lol my thoughts

PR #2831, opened 2023-02-13
[C-2104] Fix favorites pagination when coming online

sliptype:
### Description

* Move the setting of ApiClient isReachable into a saga that listens for reachability actions
* The ApiClient reachability state was getting out of sync with the actual reachability state

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/utils/reachability.ts
+++ b/packages/mobile/src/utils/reachability.ts
@@ -6,7 +6,6 @@ import { AppState } from 'react-native'
 
 import { env } from 'app/services/env'
 import { dispatch } from 'app/store'
-import { storeContext } from 'app/store/storeContext'
 
 const REACHABILITY_URL = env.REACHABILITY_URL
 
--- a/packages/mobile/src/utils/reachability.ts
+++ b/packages/mobile/src/utils/reachability.ts
@@ -65,7 +65,6 @@ const updateReachability = async (netInfoState: NetInfoState) => {
     // Supercede the setUnreachable debounce if necessary
     setUnreachable(false)
     dispatch(reachabilityActions.setReachable())
-    storeContext.apiClient.setIsReachable(true)
   }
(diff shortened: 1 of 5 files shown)


packages/mobile/src/utils/reachability.ts
```diff
@@ -65,7 +64,6 @@ const updateReachability = async (netInfoState: NetInfoState) => {
     // Supercede the setUnreachable debounce if necessary
     setUnreachable(false)
     dispatch(reachabilityActions.setReachable())
-    storeContext.apiClient.setIsReachable(true)
```
amendelsohn, line 68: So this wasn't working? Or it just didn't always get called?

amendelsohn (approved):
Wow good find!

Kyle-Shanks (approved):

packages/mobile/src/utils/reachability.ts
dylanjeffers: oh wow epic find! this might fix another issue we were seeing with download errors when going offline

amendelsohn: and also possibly the empty favorites lineup while starting offline

dylanjeffers (approved):
epic, this should also solve another issue!!! great work

packages/mobile/src/utils/reachability.ts
sliptype: So this was definitely causing an issue when manually setting the reachability via an action. It also seems to fix the issue with favorites pagination but I'm not entirely certain how the reachability was getting out of sync in that case, because this is the only place we dispatch the action

Gonna dig a bit more before merging this

audius-infra:
Preview this change https://demo.audius.co/sk-c-2104-favorites-pagination-offline

PR #2833, opened 2023-02-13
[C-2095] Improve download cancel

dylanjeffers:
### Description

Fixes issue where going offline resulted in next 1/3 tracks in queue erroring out. This is due to the APIClient setting "isReachable" before the reachability redux action, resulting in jobs finishing in error state before "cancel" race call has a chance to actually cancel the job. @sliptype work should probably solve this, but added another safeguard with the "abortOnUnrecahable" option, which just allows the fetch request to happen and stall, allowing the cancel task to run properly.

Additionally, cleans up the download tasks by moving all relevant code into the worker files, and in particular improves the `yield* all` calls. Also adds a cool "cancelable download task" using the redux-saga "CANCEL" option. this is helpful for cleaning up the potentially long track mp3 download.


packages/mobile/src/store/offline-downloads/sagas/downloadQueueSagas/workers/downloadFile.ts
```diff
@@ -0,0 +1,10 @@
+import { CANCEL } from 'redux-saga'
+import RNFetchBlob from 'rn-fetch-blob'
+
+// Downloads file from uri to destination, with saga cancellation
+export function downloadFile(uri: string, destination: string) {
+  const { fetch } = RNFetchBlob.config({ path: destination })
+  const fetchTask = fetch('GET', uri)
+  fetchTask[CANCEL] = fetchTask.cancel
```
dylanjeffers, line 8: https://redux-saga.js.org/docs/api#canceltask

packages/mobile/src/services/offline-downloader/offline-downloader.ts
```diff
@@ -1,124 +0,0 @@
-import path from 'path'
```
amendelsohn, line 1: 🔥

packages/mobile/src/store/offline-downloads/sagas/downloadQueueSagas/workers/downloadFile.ts
amendelsohn: very cool

packages/mobile/src/store/offline-downloads/sagas/downloadQueueSagas/workers/downloadTrackWorker.ts
```diff
@@ -70,7 +77,8 @@ function* downloadTrackAsync(
 
   const track = yield* call([apiClient, apiClient.getTrack], {
     id: trackId,
-    currentUserId
+    currentUserId,
+    abortOnUnreachable: false
```
amendelsohn, line 81: So what actually happens when we go unreachable is we don't fail here, but the saga will be cancelled soon after?

packages/mobile/src/store/offline-downloads/sagas/downloadQueueSagas/workers/downloadTrackWorker.ts
```diff
@@ -81,22 +89,78 @@ function* downloadTrackAsync(
...
+  return await RNFetchBlob.fs.writeFile(
+    trackMetadataPath,
+    JSON.stringify(trackMetadata)
+  )
+}
+
+async function removeDownloadedTrack(trackId: ID) {
+  const trackDirectory = getLocalTrackDir(trackId.toString())
+  return await RNFetchBlob.fs.unlink(trackDirectory)
+}
```
amendelsohn, line 166: Is this all just copied? Any notable changes?

packages/mobile/src/store/offline-downloads/sagas/redownloadStaleTracksSaga.ts
```diff
@@ -4,7 +4,7 @@ import { select } from 'typed-redux-saga'
 import { getOfflineTrackMetadata } from '../selectors'
 import type { OfflineItem } from '../slice'
 
-const STALE_DURATION_TRACKS = moment.duration(1, 'second')
```
amendelsohn, line 7: whoops!

amendelsohn: Also wondering if this should be longer? What's the cost of sync for a large library?

packages/mobile/src/store/offline-downloads/sagas/downloadQueueSagas/workers/downloadTrackWorker.ts
amendelsohn: Maybe worth a comment here about why this flag is important

amendelsohn (approved):
Wow! Thanks for doing the cleanup along with this.

packages/mobile/src/store/offline-downloads/sagas/downloadQueueSagas/workers/downloadTrackWorker.ts
dylanjeffers: yeah exactly, cause what was happening is that it would try and immediately return null, which results in us calling ERROR in the job. def will add a comment!

packages/mobile/src/store/offline-downloads/sagas/downloadQueueSagas/workers/downloadTrackWorker.ts
dylanjeffers: its mostly copied with some changes to prevent extra checks on file existence, which should result in speed improvements since half the promises are removed

packages/mobile/src/store/offline-downloads/sagas/redownloadStaleTracksSaga.ts
dylanjeffers: so i have batched to only 50? so itll be 50 per session ad infinum probably. and keep in mind we only redownload if created_at changes, so these fly by pretty quick. but def up for whatever you think.

amendelsohn: Yeah that sounds good. Wondering if it's worth sorting to start with the most stale first? If the queue is just that fast though it's probably irrelevant.

If weekly session count is less than CollectionSize / 50 then we may never catch up. Not sure fairness is helpful there though. There may be legal requirements for how often we have to guarantee check each track, so maybe a follow-up could be doing 50/session plus anything over 30d gets sync'd immediately. Idk

dylanjeffers: ah yeah that is smart about most stale! love that idea, will add that now, and then we can follow up with any more things we thing makes sense

PR #2835, opened 2023-02-13
[C-2113 C-2114] Add useThrottledCallback, fix indicators

dylanjeffers:
### Description

Adds useThrottledCallback, different from useDebouncedCallback, to ensure the first call is fast, and subsequent ones are not called until after wait. also improves the types and api so it's identical to `useCallback`

Also fixes download indicators when offline


packages/mobile/src/hooks/useDebouncedCallback.ts
```diff
@@ -1,18 +0,0 @@
-import { useCallback, useRef } from 'react'
```
Kyle-Shanks, line 1: My first instinct was to ask to keep this bc it could be useful, but this is fine, I'm just sad to see my baby go haha

Kyle-Shanks (approved):
Nice this is a better fix for sure

packages/mobile/src/hooks/useDebouncedCallback.ts
dylanjeffers: i know! it was ill inspired by your code and just relied on lodash to do the main work. i realized at some point that "debounced" means delay calling, whereas "throttled" means call right away and delay before next, so this change is mostly just a rename

Kyle-Shanks: yup makes a lot more sense for the use case here, good call

PR #2836, opened 2023-02-13
[C-2089] Distinguish between errored and abandoned

dylanjeffers:
### Description

Distinguishes between errored and abandoned, to make it clearer to us whether a track/collection is not downloadable vs an error in downloading. For now we don't need to do anything with this information, but setting it up now so we can add some tooling around handling these cases separately in the future, ie re-trying errored.

Also attempts to fix drawer glitch and improves collection header loading states.

TODO: Adding abandoned actions, one moment...


packages/mobile/src/utils/trackUtils.ts
```diff
@@ -1,13 +1,10 @@
...
-    Track,
-    'is_available' | 'is_delete' | 'is_invalid' | 'is_unlisted' | 'owner_id'
-  >,
-  currentUserId: number
-) =>
-  track.is_available &&
-  !track.is_delete &&
-  !track.is_invalid &&
-  !(track.is_unlisted && track.owner_id !== currentUserId)
+export const isTrackValid = (track: UserTrackMetadata, currentUserId: ID) => {
```
amendelsohn, line 3: minor, but imo isValid feels pretty generic and to me sounds more like a check for illegal values. Not sure the best name, but I think isAvailableForPlay was actually closer

amendelsohn (approved):
Nice!

sliptype (approved):
Makes sense!

packages/mobile/src/utils/trackUtils.ts
dylanjeffers: yeah i like that, how about "downloadable"?

PR #2840, opened 2023-02-14
[C-2123] Fix collection lineup when coming online

sliptype:
### Description

Simple fix but I tried a lot of different things to get to this place lol. There still seems to be some race case that is resulting in the lineup being empty occasionally, but it was happening so infrequently that it was difficult to pin down

I think ideally we wouldn't be resetting the collection. This is important now because the getCollectionTracks saga waits for the collection to exist, and without the reset it uses the previous collection. https://github.com/AudiusProject/audius-client/blob/main/packages/web/src/common/store/pages/collection/lineups/sagas.js#L34 But if we just defer the fetchLineupMetadatas until after the collection has been fetched, we wouldn't need to reset

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/lineup/Lineup.tsx
+++ b/packages/mobile/src/components/lineup/Lineup.tsx
@@ -334,7 +334,7 @@ export const Lineup = ({
 
   useReachableEffect(
     useCallback(() => {
-      if (entries.length > 0 || status === Status.LOADING) return
+      if (status === Status.LOADING) return
       handleLoadMore(true)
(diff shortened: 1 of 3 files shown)


audius-infra:
Preview this change https://demo.audius.co/sk-c-2123-collection-coming-online

Kyle-Shanks (approved):

audius-infra:
Preview this change https://demo.audius.co/sk-c-2123-collection-coming-online

PR #2841, opened 2023-02-14
[PAY-888][PAY-909][PAY-869] Premium content features and fixes

sddioulde:
### Description

Multiple fixes and small additional features. See commits.

- need to skip inaccessible premium tracks on mobile on next and previous

### Dragons

n/a

### How Has This Been Tested?

local web and mobile apps vs stage and prod

n/a

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/sd-pc-fixes

packages/web/src/components/modal-radio/ModalRadioItem.tsx
```diff
@@ -10,24 +10,29 @@ type ModalRadioItemProps = {
...
   return (
-    <div
-      className={cn(
-        styles.radioItem,
-        { [styles.selected]: selected, [styles.disabled]: disabled },
-        className
-      )}
-    >
-      <div className={styles.radioItemContent} onClick={onClick}>
+    <div className={cn(styles.radioItem, className)}>
```
rickyrombo, line 25: I still think we should make use of labels/radio inputs and the radio group component, if you get a free cycle.

packages/web/src/common/store/pages/track/sagas.js
```diff
@@ -172,7 +173,7 @@ function* watchFetchTrack() {
           yield put(pushRoute(NOT_FOUND_PAGE))
           return
         }
-      } else {
+      } else if (!skipSideEffects) {
```
rickyrombo, line 176: Can you explain a bit what was happening here? I'm curious

packages/web/src/common/store/queue/sagas.ts
```diff
@@ -302,8 +305,21 @@ export function* watchNext() {
...
+    const {
+      track_id: trackId,
+      is_premium: isPremium,
+      premium_content_signature: premiumContentSignature
+    } = track ?? {}
+    const hasPremiumContentSignature =
+      !!premiumContentSignature ||
+      !!(trackId && premiumTrackSignatureMap[trackId])
+    const doesUserHaveAccess = !isPremium || hasPremiumContentSignature
+    // Skip deleted or owner deactivated locked premium track
```
rickyrombo: nit, `// Skip deleted, owner deactivated, or locked premium tracks`

rickyrombo (approved):

packages/web/src/common/store/pages/track/sagas.js
sddioulde: - when you unlock, say, a follow gated track, all follow gated tracks by that user will be unlocked for you.
- as we poll these tracks for their premium content signatures for the user, these "side effects" would be called.
- these side effects set the track id, which we don't want as they sometimes update the page with the wrong track's data, causing a very bad looking bug. e.g. tracks X and Y are follow gated, im on track X's page, but now i keep seeing the track title flicker between X and Y
- the side effects also add the track to lineup, which we also don't care about in this case

rickyrombo: It sounds like what you want to do is call `retrieveTracks`? Is there a reason we call `fetchTrack` in the trackPage sagas instead?

audius-infra:
Preview this change https://demo.audius.co/sd-pc-fixes

packages/web/src/common/store/pages/track/sagas.js
sddioulde: hmm, maybe i do want retrieveTracks, double-checking. thanks

audius-infra:
Preview this change https://demo.audius.co/sd-pc-fixes

packages/web/src/common/store/pages/track/sagas.js
sddioulde: adding this to my upcoming items. will def look at it. thx again.
going to merge this for now

audius-infra:
Preview this change https://demo.audius.co/sd-pc-fixes

PR #2848, opened 2023-02-14
[C-2126] Fetch whole account for cache after confirmed update

amendelsohn:
### Description

Cacheing the account after confirmer comes back with a confirmed update was not writing the full version of the account to local storage. Here we change to get the updated account from the api then use the same process as on sign in to update the cached account. This ensures parity in the shape of the data in local storage.

### Dragons

We're dispatching fetchAccountSucceeded an extra time here without fetchAccountRequested first. Should we add that in the confirmer side?

### DIFF
--- a/packages/web/src/common/store/account/sagas.js
+++ b/packages/web/src/common/store/account/sagas.js
@@ -246,7 +246,7 @@ export function* fetchLocalAccountAsync() {
   }
 }
 
-function* cacheAccount(account) {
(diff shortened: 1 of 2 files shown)


dylanjeffers (approved):
looks good, but check out my pr for this, cause i think we can get away with just removing the  local storage call!

sliptype (approved):
This seems super reasonable. I don't think we need the extra fetchAccountRequested tbh

PR #2849, opened 2023-02-14
[C-2126] Fix empty collections on first session

dylanjeffers:
### Description

Realized that we can probably remove the call to local-storage here, since it gets picked up whenever there is a cache/add/update to a user, it will merge that data with existing stuff, preserving the full account data.


amendelsohn (approved):
Will this break if we have other updates to the user?
I understand the is_mobile_user thing is just for rewards on the back end, but say they update their description or whatever?

dylanjeffers:
> Will this break if we have other updates to the user? I understand the is_mobile_user thing is just for rewards on the back end, but say they update their description or whatever?

so there is a cache/update right after this for the user, which will include the metadata change, and then there is a watcher that will sync this data to local storage, from what i can tell

PR #2850, opened 2023-02-14
Add wifi only download option

dylanjeffers:
### Description

Adds option for wifi only downloads to prevent audius eating up bandwidth when downloading on cellular

### TODO

We need a new download indicator option, maybe a "paused" icon to alert the user that the downloads are paused while they are still online.


amendelsohn:
@dylanjeffers I remade this on [ amendel-cellular-download-option](https://github.com/AudiusProject/audius-client/tree/amendel-cellular-download-option)

PR #2852, opened 2023-02-14
[C-2132] unlink is safe on ios but not android

amendelsohn:
### Description

We were getting "error deleting file" from `RNFetchBlob.unlink` on Android.
Looks like iOS handles it silently and Android throws.

### Dragons

This fixes the crash, but it also looks like `unlink` does not actually delete the files from disk on android... we can follow up with a separate change for that.


dylanjeffers (approved):
ah great catches!

PR #2854, opened 2023-02-14
[C-2134 C-2030] Fix queueStatus on reachability, theme

dylanjeffers:
### Description

Fixes critical issue where going offline then going online leaves queueStatus in "PAUSED" state resulting in invalid state when actually starting to download.

Fixes offline dark/matrix mode icons

Improves slice/worker types


amendelsohn (approved):
Niice

PR #2855, opened 2023-02-14
Ensure queue always starts back up

amendelsohn:
### Description

Good to be safe.
Companion PR to https://github.com/AudiusProject/audius-client/pull/2854


dylanjeffers (approved):
yeah that seems like the move!

PR #2858, opened 2023-02-15
[C-2112] Rename various offline state elements

dylanjeffers:
### Description

Renames some of the state to more accurately represent each piece. This needs to be done now since otherwise itll be a pain to rename once it's live.

#### Main renames:
downloadQueue -> offlineQueue since it does all things offline related, not just download. (also renamed various sagas for this)
requestDownloadQueueItem -> requestProcessNextJob
start/success/cancel/error/abort Download -> start.../abortJob
addOfflineItems -> addOfflineEntries to use same language we already use in audius-client
OfflineItem -> OfflineJob to clarify between entries and jobs

Also fixes issue with filter color


amendelsohn (commented):
Love these changes! It'll be great for clarity whenever we have to revisit this.

amendelsohn (approved):

PR #2863, opened 2023-02-15
[C-2141] Fix missing lineup when coming back online

dylanjeffers:
### Description

Fixes missing lineup when coming back online
This improves general cancellation flow for fetchLineupMetadatas saga

Also fixes issue where collection header text says "downloading" when partially downloaded and offline


amendelsohn (approved):
Incredible fix. How did you find this?

PR #2877, opened 2023-02-16
Migrate search-users saga + deps to common

dharit-tan:
### Description
I didn't want to add a copy of the saga needed for searching users for DMs to mobile, so I moved the existing one in web to common, which involved moving a few dependencies in caching as well.

Note: I did not move `waitForRead` cuz there were a lot of usages, so right now that code is duplicated between web and common. Can try to migrate that too if preferred.

### Dragons

### How Has This Been Tested?

Tested locally on both web and ios. Checked that searching still worked for example.

### Feature Flags ###


packages/web/src/store/sagas.ts
```diff
@@ -176,7 +176,7 @@ export default function* rootSaga() {
     toastSagas(),
     shareSoundToTikTokModalSagas(),
     smartCollectionPageSagas(),
-    searchUsersModalSagas(),
+    searchUsersSagas(),
```
rickyrombo, line 179: nit: I liked `searchUsersModalSagas` to differentiate from regular search

rickyrombo (approved):
Heroic! 🦸

packages/web/src/store/sagas.ts
dharit-tan: I actually was just gonna push a commit that renames everything to omit 'Modal' lol. But I see your point.. tryna think of something else that better describes the functionality.. but if not will fall back to modal.

dharit-tan: you know what modal is fine.

rickyrombo: this actually would be an ideal case for not using sagas at all and adopting the new RTK hotness when that gets merged: https://github.com/AudiusProject/audius-client/pull/2828

audius-infra:
Preview this change https://demo.audius.co/rt-PAY-895

audius-infra:
Preview this change https://demo.audius.co/rt-PAY-895

dylanjeffers:
Awesome work btw

PR #2879, opened 2023-02-16
[C-1301] TikTok native auth

sliptype:
### Description

* Use tiktok loginkit to perform native auth when tiktok app is installed. Fall back to webview auth if app is not installed
* Currently can only test if only one audius app installed on device, if multiple it will redirect to the wrong one potentially
* Android also doesn't seem to be working because of the TikTok developer app set up. Going to address this next
* Disable caching of tiktok credentials in local storage, this was useful for Share sound to tiktok but it breaks multiple sign ups on the same device
* Fix capitalization of TikTokAuthButton

### Dragons

### How Has This Been Tested?

### Feature Flags ###


packages/mobile/patches/react-native-tiktok+1.1.1.patch
```diff
@@ -1662,6 +1662,19 @@ index 627b86d..0000000
...
+diff --git a/node_modules/react-native-tiktok/ios/Tiktok.swift b/node_modules/react-native-tiktok/ios/Tiktok.swift
+index ce31978..af2093e 100644
+--- a/node_modules/react-native-tiktok/ios/Tiktok.swift
++++ b/node_modules/react-native-tiktok/ios/Tiktok.swift
+@@ -7,7 +7,7 @@ class Tiktok: UIViewController {
+     
+   @objc
+   func auth(_ callback: @escaping RCTResponseSenderBlock) {
+-    let scopes = ["user.info.basic,video.list"] // list your scopes
++    let scopes = ["user.info.basic"] // list your scopes
```
nicoback2, line 1674: OOC what's this for?

packages/mobile/src/hooks/useTikTokAuth.ts
```diff
@@ -1,37 +1,110 @@
...
+        accessToken: access_token,
+        openId: open_id,
+        expiresIn: expires_in
       })
-    },
+    }
+
+    // Needed for Android
+    tikTokEvents.addListener('onAuthCompleted', (resp) => {
+      handleTikTokAuth(resp.code, !!resp.status, resp.status)
```
nicoback2, line 98: Just to confirm -- `!!resp.status` tells us if there is an error and `resp.status` is the error message? So it's empty if the thing is successful?

packages/mobile/ios/AudiusReactNative/AppDelegate.mm
```diff
@@ -263,4 +263,20 @@ - (Class)getModuleClassFromName:(const char *)name
 
 #endif
 
+- (BOOL)application:(UIApplication *)application openURL:(NSURL *)url sourceApplication:(NSString *)sourceApplication annotation:(id)annotation
+{
+    if ([[TikTokOpenSDKApplicationDelegate sharedInstance] application:application openURL:url sourceApplication:sourceApplication annotation:annotation]) {
+        return YES;
+    }
+    return NO;
+}
```
nicoback2, line 272: You can delete this - the logic is already in the existing `openURL` function above (I added it there for the Share Video to TikTok)

packages/mobile/ios/AudiusReactNative/AppDelegate.mm
```diff
@@ -263,4 +263,20 @@ - (Class)getModuleClassFromName:(const char *)name
...
+
+- (BOOL)application:(UIApplication *)application handleOpenURL:(NSURL *)url
```
nicoback2, line 274: I think you can delete this too but not sure. I didn't include it for Share Video to TikTok and everything seems to work

nicoback2 (approved):
Niceee! Approving to unblock but had a couple Qs + there's a block you added in AppDelegate that isn't needed

packages/mobile/src/hooks/useTikTokAuth.ts
sliptype: Yes, the status is `0` if successful and otherwise is a code representing the error https://developers.tiktok.com/doc/getting-started-android-handling-errors

This is only for android

packages/mobile/ios/AudiusReactNative/AppDelegate.mm
sliptype: Ok I'll try removing it, it wasn't working before I added it but I think I changed some other things at the same time

nicoback2: Ooh gotcha

packages/mobile/patches/react-native-tiktok+1.1.1.patch
sliptype: The scopes list is hardcoded into `react-native-tiktok` and we don't need `video.list`. The other option is to add it to our scopes and get a re-review, but I figured this was easier / more correct

dylanjeffers (changes requested):
sorry! but just double checking with all the lock files?

sliptype:
> sorry! but just double checking with all the lock files?

Thanks for catching! Don't need the root or web changes

dylanjeffers (approved):
yaya

PR #2883, opened 2023-02-17
[C-2179] Show solana-phone wallet connect option on saga

dylanjeffers:
### Description

Checks react-native android Model constant to determine if we are on saga phone, and displays solana-mobile wallet connect option instead of phantom.

- Also improves general ui aligning to spec
- Removes unnecessary feature flag

### How was this tested
Confirmed using saga device. It connected to sol-flare automatically and went through full connect process. Using react-native constants here is what sol dev team recommended also


packages/mobile/src/services/audio-player/NativeMobileAudio.ts
```diff
@@ -18,4 +18,7 @@ export class NativeMobileAudio {
   getDuration = () => 0
   onBufferingChange = () => {}
   onError = () => {}
+  setPlaybackRate = () => {}
```
raymondjacobson, line 21: great catch

sliptype (approved):

raymondjacobson (approved):

PR #2888, opened 2023-02-17
[C-1896] Add playback rate drawer and controls for mobile

Kyle-Shanks:
### Description
* Add PlaybackRateDrawer
* Make updates to audio file to change the rate of the TrackPlayer based on the playbackRate and if the track is a podcast or not
* Add PlaybackRateButton for the NowPlayingDrawer AudioControls component
* Updated the scrubber and position timestamp to respect the different playback speeds
* playback rate is persisted using the local storage functions passed to the CommonStoreContext set up in the web PR 

### Dragons

There was one instance where the rate of the TrackPlayer was resetting on its own, but i can't get repro consistently so I'll figure that out later if it becomes an issue

### How Has This Been Tested?

Lots of manual testing

N/A

### Feature Flags ###

Controls are behind the PODCAST_CONTROL_UPDATES_ENABLED feature flag


packages/mobile/src/services/audio-player/NativeMobileAudio.ts
```diff
@@ -13,15 +13,12 @@ export class NativeMobileAudio {
...
-  setPlaybackRate = () => null
+  setPlaybackRate = () => {}
   isBuffering = () => false
   getPosition = async () => 0
   getDuration = () => 0
   getPlaybackRate = () => '1x' as PlaybackRate
   getAudioPlaybackRate = () => 1.0
   onBufferingChange = () => {}
   onError = () => {}
-  setPlaybackRate = () => {}
```
Kyle-Shanks, line 24: these were duplicates so I removed them

packages/mobile/src/components/core/IconButton.tsx
```diff
@@ -37,6 +40,7 @@ const defaultHitSlop = { top: 12, right: 12, bottom: 12, left: 12 }
  * The default size is 18x18 but this can be overridden via styles.icon
  */
 export const IconButton = ({
+  autoWidth = false,
```
dylanjeffers, line 43: cool updates here! just want to make sure this leads to consistent results with the other icon buttons?

packages/mobile/src/components/now-playing-drawer/AudioControls.tsx
```diff
@@ -32,6 +42,18 @@ const createStyles = () =>
       display: 'flex',
       alignItems: 'center'
     },
+    emptyPlaceholder: {
+      minHeight: 24,
```
dylanjeffers, line 46: spacing for all these?

dylanjeffers: oh looks like this is a createStyles legacy thing, would you mind updating to useStyles format?

packages/mobile/src/components/now-playing-drawer/AudioControls.tsx
```diff
@@ -99,8 +124,14 @@ export const AudioControls = ({
     dispatch(repeat({ mode }))
   }, [dispatch, repeatMode])
 
+  const onPressPlaybackRate = useCallback(() => {
```
dylanjeffers, line 127: handlePressPlaybackRate?

dylanjeffers: would it make sense to just add this to the playbackratebutton itself?

packages/mobile/src/components/scrubber/Slider.tsx
```diff
@@ -142,11 +146,12 @@ export const Slider = memo((props: SliderProps) => {
   useEffect(getRailPageX, [railRef])
 
   const currentAnimation = useRef<Animated.CompositeAnimation>()
+  const playbackRate = useSelector(getPlaybackRate)
   const play = useCallback(
     (timeRemaining: number) => {
       currentAnimation.current = Animated.timing(translationAnim, {
         toValue: railWidth,
-        duration: timeRemaining,
+        duration: timeRemaining / parseFloat(playbackRate),
```
dylanjeffers, line 154: i am getting a little worried about the parseFloats, instead of it being a number?

packages/mobile/src/components/scrubber/usePosition.ts
```diff
@@ -11,6 +12,7 @@ export const usePosition = (
 ) => {
   const positionRef = useRef(0)
   const positionElementRef = useRef<TextInput>(null)
+  const progress = useProgress(200)
```
dylanjeffers, line 15: so the main idea here was to avoid using useProgress cause it causes rerender every 200 ms! lmk how you want to handle this, im thinking best way is to just update the setTimeout call to trigger depending on rate, it was 1000 for 1 second, but we can make it 15000 etc for whatever the playback rate is (would require a bit of math to relate to rate)

dylanjeffers (dismissed):
Awesome awesome work here. everything looks super clean.

Main blocking callouts are:
1. Lets update the svgs to have #000 for fill, so we can apply a themeable fill to it. lmk if you're not aware of this pattern cause it's pretty random
2. Let's see if we can avoid the useProgress. This is def quite a hacky situation where we are relying on timeouts, not useProgress, so if you feel extra strong about this, happy to consider. but for most efficient rerender's the timeout approach can be adjusted with the rate i beleive
3. We need to apply same technique to the TrackingBar too!

packages/mobile/src/components/scrubber/Slider.tsx
Kyle-Shanks: I can add a function that reads from a playback rate map as a middle ground so that parseFloat is not used everywhere

packages/mobile/src/components/core/IconButton.tsx
Kyle-Shanks: Yea this should only affect buttons that set this to true so it should be ok

packages/mobile/src/components/scrubber/usePosition.ts
Kyle-Shanks: thinking about this one, wouldn't updating the timeout to use a faster rate cause rerenders at that faster rate anyway? I can update the timeout function to read from TrackPlayer at its own rate and then update the rate to match the playbackRate, but want to make sure that it would actually save on rerenders

audius-infra:
Preview this change https://demo.audius.co/kj-Add-mobile-playback-rate-controls

packages/mobile/src/components/core/IconButton.tsx
dylanjeffers: ah right of course!

dylanjeffers: oh one thing actually, so we have `fullWidth` on buttons that might do the same thing, can we use that prop name for consistency?

packages/common/src/store/player/sagas.ts
```diff
@@ -9,9 +10,18 @@ import { PlaybackRate, PLAYBACK_RATE_LS_KEY } from './types'
  * Sets the playback rate from local storage when the app loads
  */
 function* setInitialPlaybackRate() {
+  const getFeatureEnabled = yield* getContext('getFeatureEnabled')
   const getLocalStorageItem = yield* getContext('getLocalStorageItem')
+  const isPodcastFeaturesEnabled =
+    ((yield* call(
```
dylanjeffers, line 16: if you merge with main, a fix was added so you dont need to do any of this casting or ?? stuff, thank heavens

packages/mobile/src/components/now-playing-drawer/PlaybackRateButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+  '1x': Icon1x,
+  '1.1x': Icon1_1x,
+  '1.2x': Icon1_2x,
+  '1.5x': Icon1_5x,
+  '2x': Icon2x,
+  '2.5x': Icon2_5x,
+  '3x': Icon3x
+}
+
+export const PlaybackRateButton = (props: PlaybackRateButtonProps) => {
```
dylanjeffers, line 33: such a clean file, great work

packages/mobile/src/components/playback-rate-drawer/PlaybackRateDrawer.tsx
```diff
@@ -0,0 +1,85 @@
...
+  '1.2x',
+  '1.5x',
+  '2x',
+  '2.5x',
+  '3x'
+]
+
+const useStyles = makeStyles(({ palette, spacing }) => ({
+  title: {
+    display: 'flex',
```
dylanjeffers, line 34: display: flex be default :)

dylanjeffers (approved):
looks great!! lets chat real quick about the usePosition hook, too

PR #2889, opened 2023-02-18
[C-2155 C-2127] Fix duplicate feed fetch

dylanjeffers:
### Description

Fixes issue where lineups double fetch on load. Other than improving perf, this could also help mitigate scrolling issues we were seeing in lineups due to out-of-order pagination.

The fix essentially is to add an option to reachability effect to not call on first render

### DIFF
--- a/packages/mobile/src/components/lineup/Lineup.tsx
+++ b/packages/mobile/src/components/lineup/Lineup.tsx
@@ -338,7 +338,8 @@ export const Lineup = ({
     useCallback(() => {
       if (status === Status.LOADING || !inView) return
       handleLoadMore(true)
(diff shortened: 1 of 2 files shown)


raymondjacobson (approved):
Makes sense. When would we want this to run on the first render? (ie should this just not be the default?)

dylanjeffers:
> Makes sense. When would we want this to run on the first render? (ie should this just not be the default?)

So i think most of the time we want this on the first render, cause it's essentially a useEffect + responds to reachability. it's just for lineup there is an annoying edge-case where for first render, we only want to fetch if "selfLoad" is true, which is handled by a different effect.

PR #2890, opened 2023-02-18
[C-2169] Fix track-player hot reload

dylanjeffers:
### Description

- Fixes issue where track player triggered app-ending bug when hot-reload/metro reload was triggered.
- Fixes issue where app-startup results in multiple warnings that "track player is not setup" due to effects running before async setup is finished.
- Fixes perf issue where we Promise.all'd non-promise methods.


packages/mobile/src/components/audio/Audio.tsx
```diff
@@ -407,62 +406,57 @@ export const Audio = () => {
...
+        url = apiClient.makeUrl(
+          `/tracks/${encodeHashId(track.track_id)}/stream`
+        )
+      } else {
+        isM3u8 = true
+        const ownerGateways = audiusBackendInstance.getCreatorNodeIPFSGateways(
+          trackOwner.creator_node_endpoint
+        )
+        url = hlsUtils.generateM3U8Variants({
+          segments: track?.track_segments ?? [],
```
sddioulde, line 431: nit: seems like you don't need the `?` as you haven't used it in the lines of code above that use `track`

packages/mobile/src/components/audio/Audio.tsx
```diff
@@ -407,62 +406,57 @@ export const Audio = () => {
...
+          cid: track ? track.cover_art_sizes || track.cover_art : null,
+          user: trackOwner,
+          size: SquareSizes.SIZE_1000_BY_1000,
+          localSource: localTrackImageSource
+        })?.uri ?? DEFAULT_IMAGE_URL
+
+      return {
+        url,
+        type: isM3u8 ? TrackType.HLS : TrackType.Default,
+        title: track?.title,
```
sddioulde, line 452: nit: same comment about `?`

packages/mobile/src/components/audio/Audio.tsx
sddioulde: unless the lines above are the ones that need the `?` to check for null/undefined?

sddioulde (approved):
lgtm

PR #2898, opened 2023-02-21
[C-2196] Implement modernized cache

dylanjeffers:
### Description

Introduces a few updates and improvements to the current caching architecture by relying on some modern patterns.
- Implements slices for each cache
- Implements `createEntityAdapter` for each cache to ensure our crud methods use productionalized patterns
- Major readability improvements for redux-debugger, as we no longer have `CACHE/ADD, CACHE/ADD_SUCCESSFUL` for every datatype, we now have `tracks/addTracks`, `tracks/updateTrack`, `tracks/removeTrack` etc for each type.
- Removes subscribe/unsubscribe logic that wasn't properly implemented, leading to unecssary cache changes
- Moves timestamps out of metadata level, resulting in much faster updates (as each update sometimes was just a timestamp change, which still resulted in full reselect)
- Reduces reliance on uids, which are brittle and have the potential to totally pollute cache, and also resulted in tons of cache updates.
- Adds typescript
- Gears up for future cache normalization improvements + flows (by replacing reformat with normalizr, and adding reducers to automagically update/add tracks/users when collections are added
- Gears up to further denormalizing our store (remix_of in track has reference to full user! we can easily describe schemas with normalizr to handle this moving forward
- Gears up to making fetch flow easier to follow, ideally through `createAsyncAction` thunks

### Dragons

Pretty much touches every flow, and given a lot is in js, there will definetely be missing pieces still. Still wanted to get this out in pr form to show that these changes are possible and most of the way there though!

### How Has This Been Tested?

Running locally. A lot of broken functionality that needs to be parsed through, but main flows are working.


packages/stems/src/components/Modal/Modal.tsx
```diff
@@ -132,13 +132,13 @@ export const Modal = forwardRef<HTMLDivElement, ModalProps>(function Modal(
         headerContainerClassName ||
         showDismissButton
       ) {
-        console.warn(
-          'Header and title-related props of `Modal` have been deprecated. Use the `ModalHeader` sub-component instead.'
-        )
+        // console.warn(
```
sliptype, line 135: Do we really want to comment this out?

packages/web/src/common/store/account/sagas.js
```diff
@@ -295,9 +290,10 @@ function* associateTwitterAccount(action) {
     const { verified } = profile
     if (!account.is_verified && verified) {
       yield put(
-        cacheActions.update(Kind.USERS, [
-          { id: userId, metadata: { is_verified: true } }
-        ])
+        usersActions.updateUser({
+          id: userId,
+          changes: { is_verified: true }
```
sliptype, line 295: Nice

packages/web/src/common/store/cache/sagas.js
```diff
@@ -146,14 +137,6 @@ function* retrieveFromSourceThenCache({
   idField,
   uids
 }) {
-  if (shouldSetLoading) {
```
sliptype, line 149: Do we not need to set loading states anymore?

packages/web/src/common/store/cache/sagas.js
```diff
@@ -170,188 +153,34 @@ function* retrieveFromSourceThenCache({
...
 
 const sagas = () => {
-  return [
-    initializeCacheType,
-    watchAdd,
-    watchUnsubscribe,
-    watchUnsubscribeSucceeded,
-    watchRemove
-  ]
+  return []
```
sliptype, line 183: Is this right?

packages/web/src/common/store/lineup/sagas.js
```diff
@@ -386,58 +388,9 @@ function* reset(
   sourceSelector,
   action
 ) {
-  const lineup = yield select(lineupSelector)
-  // Remove this lineup as a subscriber from all of its tracks and collections.
-  const subscriptionsToRemove = {} // keyed by kind
-  const source = sourceSelector ? yield select(sourceSelector) : lineupPrefix
-
-  for (const entry of lineup.entries) {
```
sliptype, line 394: So we aren't ever pruning the cache now, right?

sliptype (commented):
This looks awesome! A lot of scary changes though. Would love to get piazza and/or ray's thoughts on it too

PR #2899, opened 2023-02-22
[C-2164] add retry and requeue logic to download track jobs

amendelsohn:
### Description

1. DownloadTrackWorker retries the download up to 3 times internally before putting down the job
2. If the worker fails max times, re-enqueue the track to be tried again later
3. If the re-queue count hits 3, abandon the job

*Note:* The job is still abandoned immediately if the track is unavailable for download or the job is otherwise invalid.

### How Has This Been Tested?

Set `downloadTrackAsync` to fail 75% of the time. Tracks were succeeding at a high rate, and failed tracks would be processed again at the back of the queue. We successfully downloaded nearly all tracks even while the internal job was usually failing.

### DIFF
--- a/packages/mobile/src/store/offline-downloads/sagas/offlineQueueSagas/workers/downloadTrackWorker.ts
+++ b/packages/mobile/src/store/offline-downloads/sagas/offlineQueueSagas/workers/downloadTrackWorker.ts
@@ -50,7 +50,7 @@ function* shouldAbortDownload(trackId: ID) {
   }
 }
 
-export function* downloadTrackWorker(trackId: ID, requeueCount: number) {
+export function* downloadTrackWorker(trackId: ID, requeueCount?: number) {
   const queueItem: OfflineJob = { type: 'track', id: trackId, requeueCount }
   track(
     make({ eventName: EventNames.OFFLINE_MODE_DOWNLOAD_START, ...queueItem })
@@ -82,7 +82,7 @@ export function* downloadTrackWorker(trackId: ID, requeueCount: number) {
         })
       )
       yield* call(removeDownloadedTrack, trackId)
-      if (requeueCount < MAX_REQUEUE_COUNT - 1) {
+      if ((requeueCount ?? 0) < MAX_REQUEUE_COUNT - 1) {
         yield* put(errorJob(queueItem))
       } else {
         yield* put(abandonJob(queueItem))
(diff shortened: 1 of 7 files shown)


packages/mobile/src/store/offline-downloads/slice.ts
```diff
@@ -315,6 +319,11 @@ const slice = createSlice({
         state.collectionStatus[id] = OfflineDownloadStatus.ERROR
       } else if (type === 'track') {
         state.trackStatus[id] = OfflineDownloadStatus.ERROR
+        // re-queue the track
+        state.offlineQueue.push({
+          ...action.payload,
+          requeueCount: action.payload.requeueCount + 1
+        })
```
amendelsohn, line 326: Feels like the check for max_retry_count could happen here instead. Thoughts?

The reason I put it in the worker is so that we can use the proper `abandonJob` action, but I could also see setting it to `ABANDONED` state being a correct result of calling `errorJob`

packages/mobile/src/store/offline-downloads/sagas/offlineQueueSagas/workers/downloadTrackWorker.ts
```diff
@@ -47,54 +50,67 @@ function* shouldAbortDownload(trackId: ID) {
...
-  } else if (jobResult === OfflineDownloadStatus.SUCCESS) {
-    track(
-      make({
-        eventName: EventNames.OFFLINE_MODE_DOWNLOAD_SUCCESS,
-        ...queueItem
-      })
-    )
-    yield* put(completeJob({ ...queueItem, completedAt: Date.now() }))
-    yield* put(requestProcessNextJob())
+  for (let retryCount = 0; retryCount < MAX_RETRY_COUNT; retryCount++) {
```
dylanjeffers, line 60: so is it possible to do retries just for the jobResult? i feel like abort and cancel happen when they should, like when we abort, we should not retry. maybe we can huddle about this?

packages/mobile/src/store/offline-downloads/slice.ts
dylanjeffers: yeah i agree with your thinking, could also maybe happen in the "processNextJob" main function?

dylanjeffers (approved):
I do think this is good as is, but it might be nice to consider where we put the retries, and where we decide to enque/set-state etc?

packages/mobile/src/store/offline-downloads/sagas/offlineQueueSagas/workers/downloadTrackWorker.ts
amendelsohn: Yeah I think you're right. I didn't figure out the structure in the moment, but I can see it now.

PR #2903, opened 2023-02-22
[C-2171] Fix remove download performance

dylanjeffers:
### Description

Fixes slowness due to redDirRec and non-concurrent removal
Also fixes remove analytics to only be called when we actually remove

### DIFF
--- a/packages/mobile/src/services/offline-downloader/offline-storage.ts
+++ b/packages/mobile/src/services/offline-downloader/offline-storage.ts
@@ -13,17 +13,13 @@ import { allSettled } from '@audius/common'
 import RNFetchBlob from 'rn-fetch-blob'
 
 const {
-  fs: { dirs, exists, ls, lstat, mkdir, readFile, unlink, writeFile }
+  fs: { dirs, exists, ls, mkdir, readFile, unlink, writeFile }
 } = RNFetchBlob
 
 export type OfflineCollection = Collection & { user: UserMetadata }
 
 export const downloadsRoot = path.join(dirs.DocumentDir, 'downloads')
 const IMAGE_FILENAME = '1000x1000.jpg'
-
-export const getPathFromRoot = (string: string) => {
-  return string.replace(downloadsRoot, '~')
-}
 
 export const getLocalCollectionsRoot = () => {
   return path.join(downloadsRoot, `/collections`)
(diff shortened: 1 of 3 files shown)


packages/mobile/src/store/offline-downloads/sagas/watchRemoveOfflineItems.ts
```diff
@@ -24,23 +29,36 @@ function* deleteItemsFromDisk(action: RemoveOfflineEntriesAction) {
...
+  )
+}
+
+function* removeItemFromDisk(
+  item: OfflineEntry,
+  trackStatus: TrackStatus,
+  collectionStatus: CollectionStatus
+) {
+  if (item.type === 'collection' && !collectionStatus[item.id]) {
+    track(make({ eventName: EventNames.OFFLINE_MODE_REMOVE_ITEM, ...item }))
```
raymondjacobson, line 45: this feels like it could be a lot of events. amplitude react native will batch them, with configurable concurrency but maybe we should do one event w more metadata?

packages/mobile/src/store/offline-downloads/sagas/watchRemoveOfflineItems.ts
```diff
@@ -24,23 +29,36 @@ function* deleteItemsFromDisk(action: RemoveOfflineEntriesAction) {
...
-    } else if (item.type === 'track' && !trackStatus[item.id]) {
-      const trackDirectory = yield* call(getLocalTrackDir, item.id.toString())
-      const exists = yield* call(RNFetchBlob.fs.exists, trackDirectory)
-      if (exists) {
-        yield* call(RNFetchBlob.fs.unlink, trackDirectory)
-      }
+    const trackDirectory = yield* call(getLocalTrackDir, item.id.toString())
+    const exists = yield* call(RNFetchBlob.fs.exists, trackDirectory)
+    if (exists) {
+      console.log('removing track!')
```
raymondjacobson, line 60: log!

raymondjacobson (approved):
lgtm

packages/mobile/src/store/offline-downloads/sagas/watchRemoveOfflineItems.ts
dylanjeffers: I def agree. Ik not too sure why this data would be useful even... Maybe can chat with Andrew about it

amendelsohn: Hmm yeah I don't think we need to track that. If we do, we should do the count of times the toggle is turned off and/or the total number of items to be deleted.

packages/mobile/src/services/offline-downloader/offline-storage.ts
```diff
@@ -194,43 +190,13 @@ export const verifyCollection = async (
   return booleanResults.every((result) => result)
 }
 
-export const purgeAllDownloads = async (withLogs?: boolean) => {
+export const purgeAllDownloads = async () => {
   if (await exists(downloadsRoot)) {
-    if (withLogs) {
-      console.log(`Before purge:`)
-    }
-    await readDirRec(downloadsRoot)
```
amendelsohn, line 202: ohh yeah good catch

amendelsohn (approved):
Thank you!

PR #2904, opened 2023-02-22
[C-2198] Fix runtime error due to missing drawer data

dylanjeffers:
### Description

Fixes issue where drawer data is removed when visibility is false, which broke because we still kept drawer around for 100ms longer. realizing none of this is needed now, it appears the background fixes for the drawer were enough here, and this was overkill.

### DIFF
--- a/packages/mobile/src/app/Drawers.tsx
+++ b/packages/mobile/src/app/Drawers.tsx
@@ -1,5 +1,4 @@
 import type { ComponentType } from 'react'
-import { useState, useEffect } from 'react'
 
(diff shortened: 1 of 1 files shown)


raymondjacobson (approved):
seems reasonable but definitely should be tested on the apps a lot

dylanjeffers:
> seems reasonable but definitely should be tested on the apps a lot

Yes definitely agree, tested android and ios sims, which both were getting the glitch originally anyway, so i feel confident. but will double check on rc prod too

PR #2905, opened 2023-02-22
[C-2130] Fix race cond in lineup init

raymondjacobson:
### Description

Rough bug. Turns out the way we initialize lineup reducers is bad.
Cannot read order of undefined comes from the fact that an action gets dispatched against feed when you repost *any* track https://github.com/AudiusProject/audius-client/blob/61d832b5e78ecfffd5f26c9bdcd96758ac5b61b1/packages/web/src/common/store/pages/profile/lineups/feed/sagas.js#L119 and that hits an `undefined` value for the feed lineup slice.

This removes the lazy init and normalizes a few things.

### Dragons

This is a deep cut. Major lineup change (but should be for the better)

### How Has This Been Tested?

Locally vs staging. Bug no longer reproable

### Feature Flags ###

### DIFF
--- a/packages/common/src/store/pages/trending/reducer.ts
+++ b/packages/common/src/store/pages/trending/reducer.ts
@@ -15,7 +15,12 @@ import { GENRES } from 'utils/genres'
 
 import { TimeRange } from '../../../models'
 
-import { trendingWeek, trendingMonth, trendingAllTime } from './lineup/reducer'
+import {
+  trendingWeek,
+  trendingMonth,
+  trendingAllTime,
+  makeInitialState
+} from './lineup/reducer'
 
 const urlParams = new URLSearchParams(window.location.search)
 const genre = urlParams.get('genre')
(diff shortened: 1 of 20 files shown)


packages/common/src/store/pages/trending/reducer.ts
```diff
@@ -57,17 +65,7 @@ const trendingAllTimeReducer = asLineup(
   trendingAllTime
 )
 
-const reducer = (state, action) => {
```
dylanjeffers, line 60: gotcha so this is the part that is replaced with not-lazy version to ensure each lineup is ready to go when a repost action happens?

dylanjeffers (approved):
nice changes! its good that our lineups have all the right stuff from the beginning probably anyway?

raymondjacobson:
> nice changes! its good that our lineups have all the right stuff from the beginning probably anyway?

yeah, definitely. i'm not sure why it wasn't done this way to start.

audius-infra:
Preview this change https://demo.audius.co/rj-c-2130

raymondjacobson:
> > nice changes! its good that our lineups have all the right stuff from the beginning probably anyway?
> 
> yeah, definitely. i'm not sure why it wasn't done this way to start. 

https://github.com/AudiusProject/audius-client-old/commit/463b09cd5d5f380de1cf54161fe421bff857fc1b

Here's the origin

packages/common/src/store/pages/trending/reducer.ts
raymondjacobson: yeyo

PR #2906, opened 2023-02-22
[C-2200] Fix profile tabs loading/empty states

dylanjeffers:
### Description

Fixes a few issues with profile tab loading and empty states.
- Fixes case where album/playlist-count == 0 wasn't showing empty tile
- Fixes case where ListEmptyComponent was used instead of LineupEmptyComponent, resulting in empty lineups rendering nothing
- Fixes case where RepostsTab was always lazy, resulting in issues for non-artists. Fixed by passing `lazy: true` to reposts only for non-artists


raymondjacobson (approved):
seems reasonable to me!

PR #2907, opened 2023-02-22
[PAY-920] Chat tails, unread indicator

dharit-tan:
### Description

- Chat tails
- Unread messages indicator
- Fetch more messages when user scrolls to top.

Need some guidance on this before merging - see comments.

### Dragons

- There is a white line in the chat tail where the svg path and rect are separated.

### How Has This Been Tested?

Android emulator

Screenshot_1677046448

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
+++ b/packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
@@ -15,8 +15,7 @@ import { useThemePalette } from 'app/utils/theme'
 const { getUserId } = accountSelectors
 
 const messages = {
-  newMessage: 'New Message',
-  s: 's'
+  newMessage: 'New Message'
 }
 
 const useStyles = makeStyles(({ spacing, palette, typography }) => ({
@@ -82,18 +81,21 @@ const useStyles = makeStyles(({ spacing, palette, typography }) => ({
   },
   tail: {
     display: 'flex',
-    position: 'relative',
-    marginTop: -12,
-    zIndex: -1
+    position: 'absolute',
+    zIndex: -1,
+    bottom: 47
   },
   tailIsAuthor: {
-    marginRight: -12
+    right: -spacing(3)
   },
   tailOtherUser: {
-    marginLeft: -12,
+    left: -spacing(3),
     transform: [{ scaleX: -1 }]
   }
 }))
+
+const pluralize = (message: string, shouldPluralize: boolean) =>
+  message + (shouldPluralize ? 's' : '')
 
 type ChatMessageListItemProps = {
   message: ChatMessage
@@ -123,7 +125,7 @@ export const ChatMessageListItem = ({
             {message.message}
           </Text>
         </View>
-        {hasTail && (
+        {hasTail ? (
           <>
             <View
               style={[
@@ -139,19 +141,18 @@ export const ChatMessageListItem = ({
               </Text>
             </View>
           </>
-        )}
+        ) : null}
       </View>
 
-      {isEarliestUnread && (
+      {isEarliestUnread ? (
         <View style={styles.unreadTagContainer} key='unreadTag'>
           <View style={styles.unreadSeparator} />
           <Text style={styles.unreadTag}>
-            {unreadCount} {messages.newMessage}
-            {unreadCount > 1 && messages.s}
+            {unreadCount} {pluralize(messages.newMessage, unreadCount > 1)}
           </Text>
           <View style={styles.unreadSeparator} />
         </View>
-      )}
+      ) : null}
     </>
   )
 }
--- a/packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
+++ b/packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
@@ -14,10 +14,6 @@ import { useThemePalette } from 'app/utils/theme'
 
 const { getUserId } = accountSelectors
 
-const messages = {
-  newMessage: 'New Message'
-}
-
 const useStyles = makeStyles(({ spacing, palette, typography }) => ({
   rootOtherUser: {
     display: 'flex',
@@ -56,29 +52,6 @@ const useStyles = makeStyles(({ spacing, palette, typography }) => ({
     fontSize: typography.fontSize.xs,
     color: palette.neutralLight2
   },
-  unreadTagContainer: {
-    marginVertical: spacing(6),
-    flexGrow: 1,
-    display: 'flex',
-    flexDirection: 'row',
-    alignItems: 'center'
-  },
-  unreadSeparator: {
-    height: 1,
-    backgroundColor: palette.neutralLight5,
-    flexGrow: 1
-  },
-  unreadTag: {
-    color: palette.white,
-    fontSize: typography.fontSize.xxs,
-    fontFamily: typography.fontByWeight.bold,
-    textTransform: 'uppercase',
-    letterSpacing: 0.5,
-    backgroundColor: palette.neutralLight5,
-    paddingHorizontal: spacing(2),
-    paddingVertical: spacing(1),
-    borderRadius: spacing(0.5)
-  },
   tail: {
     display: 'flex',
     position: 'absolute',
@@ -94,20 +67,15 @@ const useStyles = makeStyles(({ spacing, palette, typography }) => ({
   }
 }))
 
-const pluralize = (message: string, shouldPluralize: boolean) =>
-  message + (shouldPluralize ? 's' : '')
-
 type ChatMessageListItemProps = {
   message: ChatMessage
   hasTail: boolean
-  isEarliestUnread: boolean
   unreadCount: number
 }
 
 export const ChatMessageListItem = ({
   message,
   hasTail,
-  isEarliestUnread,
   unreadCount
 }: ChatMessageListItemProps) => {
   const styles = useStyles()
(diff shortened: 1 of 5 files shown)


packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
```diff
@@ -1,17 +1,48 @@
...
+import { useThemePalette } from 'app/utils/theme'
 
 const { getUserId } = accountSelectors
 
+const messages = {
+  newMessage: 'New Message',
+  s: 's'
+}
+
+const ChatTail = ({ fill }) => (
```
dharit-tan, line 25: Having trouble when importing from svg file, this is the only way I've got it to work so far. Any ideas on what's going wrong here?

packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
```diff
@@ -45,34 +76,114 @@ const useStyles = makeStyles(({ spacing, palette, typography }) => ({
...
+    position: 'relative',
+    marginRight: -14,
+    marginTop: -12
+  },
+  tailOtherUser: {
+    display: 'flex',
+    position: 'relative',
+    marginLeft: -14,
+    marginTop: -12,
+    transform: [{ scaleX: -1 }]
```
dharit-tan, line 133: Thoughts on relative vs absolute positioning? Both work, I understand the differences in theory but not sure which to use in practice - seems like both are valid here.

dharit-tan: Eg. which would be better if the user has zoomed in somehow on their phone, or enlarged text?

audius-infra:
Preview this change https://demo.audius.co/rt-PAY-920

packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
dharit-tan: Was importing incorrectly!

audius-infra:
Preview this change https://demo.audius.co/rt-PAY-920

audius-infra:
Preview this change https://demo.audius.co/rt-PAY-920

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -92,22 +101,83 @@ export const ChatScreen = () => {
...
+        dispatch(sendMessage({ chatId, message }))
+        setValue('')
+        setIconOpacity(ICON_BLUR)
+      }
+    },
+    [chatId, setValue, dispatch]
+  )
+
+  const handleScrollToTop = () => {
+    dispatch(setActiveChat({ chatId: null }))
```
piazzatron, line 170: Why do we set active chat to null here?

packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
piazzatron: Personally not a huge fan of the negative margins thing - relative or absolute positioning could work here IMO. I tend to opt for absolute positioning if you know you want to take an object out of the DOM hierarchy, but sometimes it can be trickier to find the correct coordinates with absolute, in which case relative can be useful.

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -92,22 +101,83 @@ export const ChatScreen = () => {
...
-  const [otherUser] = useSelector((state) => getOtherChatUsers(state, chatId))
+  useEffect(() => {
+    if (chatId && status === Status.IDLE) {
+      // Initial fetch
+      dispatch(fetchMoreMessages({ chatId }))
+      dispatch(setActiveChat({ chatId }))
+    }
+  }, [dispatch, chatId, status, summary])
+
+  const earliestUnreadIndex = useMemo(() => {
```
piazzatron, line 141: Possible perf thing to flag, this is going to run whenever we add a new message and will iterate over the entire messages array, and then a second time with the indexOf. We could probably use findIndex to cut down one of these calls, which would also stop iterating over the array once the condition is met. Do we need to even compute this if we're in an active chat, though?

piazzatron (approved):
Looking good, mostly just nits + one perf thing!

packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
sddioulde: it's a case by case basis for me between absolute and relative. e.g. if i know an item, say, a track tile corner tag is supposed to be at the top left, easier to do absolute with top: 0 and left: 0. there are other cases where you care more about the relative positioning. whatever is simplest for the given situation

packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
```diff
@@ -45,34 +56,102 @@ const useStyles = makeStyles(({ spacing, palette, typography }) => ({
...
+    letterSpacing: 0.5,
+    backgroundColor: palette.neutralLight5,
+    paddingHorizontal: spacing(2),
+    paddingVertical: spacing(1),
+    borderRadius: spacing(0.5)
+  },
+  tail: {
+    display: 'flex',
+    position: 'relative',
+    marginTop: -12,
```
sddioulde, line 86: nit: can still use `-spacing(3)` here and in below negative numbers

sddioulde (approved):

packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
dharit-tan: nice thx guys

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
dharit-tan: Good call!

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -92,22 +101,81 @@ export const ChatScreen = () => {
...
+    if (chatId && status === Status.IDLE) {
+      // Initial fetch
+      dispatch(fetchMoreMessages({ chatId }))
+    }
+  }, [dispatch, chatId, status, summary])
 
-  const [otherUser] = useSelector((state) => getOtherChatUsers(state, chatId))
+  useEffect(() => {
+    // Update chatFrozenRef when entering a new chat screen
+    if (chat && chatId !== chatFrozenRef.current?.chat_id) {
```
dharit-tan, line 139: @rickyrombo I moved the ref setting up bc I think we actually want to make sure we don't call `findIndex` with previous chat params. Seems to work on testing, does this make sense?

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
dharit-tan: Actually unfortunately I think we do need to recalc every time chatMessages changes, bc if we send a message or a new message comes in, the unread tag will move if earliestUnreadIndex doesn't update.

packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
rickyrombo: agree with the above, def case by case, and both would work here

packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
```diff
@@ -45,34 +55,104 @@ const useStyles = makeStyles(({ spacing, palette, typography }) => ({
...
+  tailIsAuthor: {
+    right: -spacing(3)
+  },
+  tailOtherUser: {
+    left: -spacing(3),
+    transform: [{ scaleX: -1 }]
   }
 }))
 
+const pluralize = (message: string, shouldPluralize: boolean) =>
```
rickyrombo, line 97: Can't wait until we just have all the i18n tools! https://react.i18next.com/misc/using-with-icu-format#plural

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -137,14 +204,18 @@ export const ChatScreen = () => {
...
-                renderItem={({ item, index }) => {
-                  return (
-                    <ChatMessageListItem
-                      message={item}
-                      hasTail={hasTail(item, chatMessages[index - 1])}
-                    />
-                  )
-                }}
+                renderItem={({ item, index }) => (
+                  <ChatMessageListItem
```
rickyrombo, line 208: For consistency w/ web, could the separator live in a fragment as a sibling to the `ChatMessageListem` here rather than inside the `ChatMessageListItem` component? or does that not work? (I can also update the web side to match)

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
rickyrombo: Yeah this is fine, especially as you called out we want the index to change when the user sends/receives a new message so using the effect/state etc is all moot... sorry for the wrong advice there

rickyrombo (approved):

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
dharit-tan: can do

PR #2911, opened 2023-02-22
Fix lineup showing empty tile before fetch

dylanjeffers:
### Description

Cases where swiping to a lineup shows empty tile before the initial fetch.


amendelsohn (approved):
Ah, good idea!

PR #2912, opened 2023-02-22
[C-2207] Fix scrubber timestamp sync issues

dylanjeffers:
### Description

Fixes two issues with timestamp sync:
1. When touching the scrubber handle and not dragging, the timestamp does not continue, since "isInteracting" was not properly being set to false
2. When going to background on android, `setTimeouts` are paused, resulting in inconsistent timer.

### DIFF
--- a/packages/mobile/src/components/scrubber/usePosition.ts
+++ b/packages/mobile/src/components/scrubber/usePosition.ts
@@ -1,7 +1,10 @@
 import { useCallback, useEffect, useRef } from 'react'
 
 import { formatSeconds } from '@audius/common'
+import { useAppState } from '@react-native-community/hooks'
 import type { TextInput } from 'react-native'
+import TrackPlayer from 'react-native-track-player'
+import { useAsync } from 'react-use'
 
 export const usePosition = (
   mediaKey: string,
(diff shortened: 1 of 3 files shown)


packages/mobile/src/components/scrubber/usePosition.ts
```diff
@@ -47,6 +50,16 @@ export const usePosition = (
     }
   }, [isPlaying, isInteracting, duration])
 
+  // Android pauses timeouts when in background, so we use app state
+  // to trigger a position coercion
+  const appState = useAppState()
+
+  useAsync(async () => {
```
sliptype, line 57: Could we `useEnterForeground` instead?

sliptype (approved):
lgtm!

packages/mobile/src/components/scrubber/usePosition.ts
dylanjeffers: yeeee

dylanjeffers: interesting, so looks like useEnterForeground's deps and rerenders are a bit off, since itll trigger everytime incoming function is remade... we might want to add a dep array to that impl, or pass useCallbacks into it? wonder what you think is best here...

dylanjeffers: going to merge for now, but im keen to update useEnterForeground etc to be top perf, and then integrate that here.

PR #2915, opened 2023-02-23
[C-2202] Sync legacy search history

dylanjeffers:
### Description

Syncs legacy search history with current persisted search history. We can prob just have this running for the next 2 or 3 verisons and get most people off old system, and then remove this


packages/mobile/src/store/search/searchSagas.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  if (!Array.isArray(legacySearchHistory)) {
+    yield* call([AsyncStorage, AsyncStorage.removeItem], SEARCH_HISTORY_KEY)
+    return
+  }
+
+  const persistedSearchHistory = yield* select(getSearchHistory)
+
+  const mergedSearchHistory = uniq([
+    ...persistedSearchHistory,
```
sliptype, line 30: Doesn't really matter, but was it possible to have multiple identical search history items previously? Because this would remove that and make them all unique right

sliptype (approved):
Nice

packages/mobile/src/store/search/searchSagas.ts
dylanjeffers: yeah good call, i looked at current implementation, and any new item that's the same as a previous one will filter out the old one, and put latest first, so i think this accomplishes that, since persisted queries will be latest, and we filter out the legacy ones if duplicate!

PR #2923, opened 2023-02-23
Update pod installation to work with react-native 0.71

sliptype:
### Description

* Update podfile min ios version to 14 so that rive builds. See https://github.com/rive-app/rive-react-native/issues/153
  This means rive may not work on older versions of ios
* Update Podfile.lock

These are the steps I took to successfully install pods:

```
npm run clean
rbenv install 2.7.6
ruby -v // to check you are using correct ruby version
rbenv rehash
sudo gem install bundler:1.17.3
bundle install
pod repo update
pod install
```

rbenv docs: https://github.com/rbenv/rbenv

These threads were both helpful:
* https://github.com/facebook/react-native/issues/34608#issuecomment-1319114711
* https://stackoverflow.com/questions/37914702/how-to-fix-your-ruby-version-is-2-3-0-but-your-gemfile-specified-2-2-5-while

### Dragons

rive may not work on version of ios older than 14

### How Has This Been Tested?

### Feature Flags ###


dylanjeffers (approved):
looks great, wondering whats up with the fastlane issue?

dylanjeffers (changes requested):
fwiw i was able to get this working without any pbxproj changes, which i think might be messing up the circleci build? anyway going to throw up a pr just to see if it builds and we can chat about what the move is!

dylanjeffers:
for ref https://github.com/AudiusProject/audius-client/pull/2924

dylanjeffers:
We went ahead with the fix to just the hermes version, instead of also updating PromisesObjC since this was resulting in install issues with version of ruby managed by homebrew instead of rbenv. Since getting everyone on rbenv will be a bit of work, was thinking we go with the simple changes first, and then figure out a good way to get everyone on rbenv?

PR #2924, opened 2023-02-24
Fix podfile, specify ruby-version

dylanjeffers:
### Description

Attempt at fixing podfile, and specifies ruby-version. Ideally we have some docs in root readme on how to install ruby using rbenv etc. happy to add that if this ends up building correctly


dylanjeffers:
so updates on this one. it appears to work, but on my end im getting an error with the launching metro terminal from `npm run ios:stage`, it requires me to run `npm start` beforehand to set that up

sliptype:
> so updates on this one. it appears to work, but on my end im getting an error with the launching metro terminal from `npm run ios:stage`, it requires me to run `npm start` beforehand to set that up

Interesting, can we update the scripts in package.json to include opening the metro terminal?

sliptype (approved):

PR #2927, opened 2023-02-25
[C-2218] Add repo init script to install nvm and rbenv

dylanjeffers:
### Description

Adds init.sh script to aid users in ensuring they have the correct external dependencies before installing and running audius-client

### Dragons

Not sure of all the contexts one might run this command. I just tested macos. From what i can tell this script is fairly universal however.


rickyrombo (approved):
might be good to add this as an npm script or something

also if it's cheap, maybe this is a preinstall script? though if it were in a hot path like that it'd probably be best to be just a check rather than also the script to fix idk

dylanjeffers:
> might be good to add this as an npm script or something
> 
> also if it's cheap, maybe this is a preinstall script? though if it were in a hot path like that it'd probably be best to be just a check rather than also the script to fix idk

I would love to add it as a preinstall, but i think catch-22 with nvm check needing to install right npm?

rickyrombo:
> > might be good to add this as an npm script or something
> > also if it's cheap, maybe this is a preinstall script? though if it were in a hot path like that it'd probably be best to be just a check rather than also the script to fix idk
> 
> I would love to add it as a preinstall, but i think catch-22 with nvm check needing to install right npm?

good point, though personally I think the preinstall script should just inform you you need to run the script to get in a good state, rather than do the installs for you automatically (just do the checks but not the installs).

dylanjeffers:
> 



> > > might be good to add this as an npm script or something
> > > also if it's cheap, maybe this is a preinstall script? though if it were in a hot path like that it'd probably be best to be just a check rather than also the script to fix idk
> > 
> > 
> > I would love to add it as a preinstall, but i think catch-22 with nvm check needing to install right npm?
> 
> good point, though personally I think the preinstall script should just inform you you need to run the script to get in a good state, rather than do the installs for you automatically (just do the checks but not the installs).

I like that idea too, on one hand i wanted a single command folks could run to get everything dev-wise set up, and on the other just verify the right versions. maybe we can drop the install steps here and just do a check like ur saying? personally i dont like how we have the notion doc separate from this repo

README.md
```diff
@@ -33,6 +33,21 @@
 
 This repo is maintained using [`lerna`](https://github.com/lerna). After cloning run:
 
+```bash
+./scrits/init.sh
+```
+This will ensure youre running with the correct nvm and nbenv configurations, and install them if not.
```
raymondjacobson, line 39: rbenv

scripts/init.sh
```diff
@@ -0,0 +1,42 @@
...
+      echo $nvm_shell_config >>~/.zshrc
+    elif [ -n "$BASH_VERSION" ]; then
+      echo $nvm_shell_config >>~/.bashrc
+    fi
+  fi
+
+  nvm install
+}
+
+ensure_ruby() {
```
raymondjacobson, line 22: this will only work on macos b/c of homebrew. readme should probably say something about this or we should also consider into a linux solution

scripts/init.sh
```diff
@@ -0,0 +1,42 @@
...
+
+  rbenv install -s
+  echo "Now using ruby v$(rbenv version-name)"
+
+  if ! command -v pod &>/dev/null; then
+    gem install cocoapods
+  fi
+}
+
+ensure_nvm
```
raymondjacobson, line 41: probably ensure_node and ensure_ruby or ensure_nvm and ensure_rbenv

raymondjacobson (approved):

scripts/init.sh
dylanjeffers: sgsg

dylanjeffers:
btw yall i think i might go the preinstall route and just simply check versions and let folks figure out how they would like install them for now. this will at a minimum get us somewhere on the right path, and then we can consider a general script, cause im realizing it is actually not as clean as i originally thought (setting nvm in the script does not carry over to original shell for example)

scripts/init.sh
raymondjacobson: i guess pods don't work on linux, but running web/desktop should be good

dylanjeffers: yeah given all the complexity here im thinking just a script to check versions for now may be the move?

rickyrombo (approved):

PR #2933, opened 2023-02-27
[PAY-921] Scroll to earliest unread chat

dharit-tan:
### Description
Scrolls to earliest unread chat.

There may be some issues with this when there are many messages (>200), but it works perfectly apart from that. Making a note to revisit this in https://linear.app/audius/issue/PAY-975/[mobile]-misc-todos.

### Dragons

We're using the flatlist error handler to accomplish this which seems a bit jank, but from perusing stackoverflow it looks like this is the accepted way to do this.

### How Has This Been Tested?

Android emulator / ios simulator against staging.

### Feature Flags ###


packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -198,6 +200,38 @@ export const ChatScreen = () => {
     [chatId, setInputMessage, dispatch]
   )
 
+  useEffect(() => {
+    console.log('useEffect for scroll', earliestUnreadIndex)
```
rickyrombo, line 204: console.log

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -260,8 +305,10 @@ export const ChatScreen = () => {
                   </Fragment>
                 )}
                 inverted
-                onEndReached={handleScrollToTop}
+                initialNumToRender={chatMessages?.length ?? 0}
```
rickyrombo, line 308: I don't think we need this anymore

rickyrombo (approved):

PR #2936, opened 2023-02-27
[PLAT-702] Add repost-of-repost notification

dylanjeffers:
### Description

adds repost-of-repost notification ui for web/mobile

### Dragons

There is still a bit to figure out regarding `onClick` for the notif tiles. I think it's possible for us to render a user-list like we do with reposts etc... it would probably be more consistent. But for now, we simply just go to the content.

### DIFF
--- a/packages/common/src/store/notifications/types.ts
+++ b/packages/common/src/store/notifications/types.ts
@@ -39,6 +39,7 @@ export enum NotificationType {
   Follow = 'Follow',
   Favorite = 'Favorite',
   Repost = 'Repost',
+  RepostOfRepost = 'RepostOfRepost',
   Milestone = 'Milestone',
   RemixCreate = 'RemixCreate',
   RemixCosign = 'RemixCosign',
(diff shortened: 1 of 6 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-plat-702-repost-of-repost

packages/common/src/store/notifications/types.ts
```diff
@@ -179,6 +180,36 @@ export type RepostPushNotification = {
...
+  entityType: Entity.Playlist | Entity.Album | Entity.Track
+}
+
+export type RepostOfRepostPushNotification = {
+  blocknumber: number
+  entityId: ID
+  initiator: ID
+  timestamp: string
+  type:
+    | PushNotificationType.RepostAlbum
```
sabrina-kiam, line 196: should there be a new PushNotificationType for repost of repost or we can reuse repost types here?

sabrina-kiam (approved):
code looks great thanks dylan! is it possible to see a screenshot of the notification? i wasn't able to create one on the demo site

also i think julian may have clarified what we want the onClick to be within the tile in the figma mockups if this answers the dragons you mentioned?
Screenshot 2023-03-01 at 12 16 47 PM

packages/common/src/store/notifications/types.ts
dylanjeffers: Yeah good q, in this case i think it might be nice to use the same repost type cause imo they should be in lockstep?

dylanjeffers:
> code looks great thanks dylan! is it possible to see a screenshot of the notification? i wasn't able to create one on the demo site
> 
> also i think julian may have clarified what we want the onClick to be within the tile in the figma mockups if this answers the dragons you mentioned? Screenshot 2023-03-01 at 12 16 47 PM

image

image

In this case there is "repost of track" since im guessing we can repost collections too? but anyway i think we can always update this stuff.

regarding onClick, im going to circle back with ux on one thing there, cause i dont think itll be extra work to have a notif-user-list basically

packages/common/src/store/notifications/types.ts
sabrina-kiam: i think it may be cleaner to have a separate repost of repost type but i trust whatever you decide! i'm not familiar with this code and what `PushNotificationType` is used for. also what's lockstep?

sabrina-kiam:
yup we can repost collections. screenshots look great thanks dylan!!

dylanjeffers:
> oh actually one more question do you know what happens if the list of users is really long? like if the circles go past the width of the screen?

yes it should do the +x image thats in the mockup! its using a core component for that so assuming it works :D

PR #2937, opened 2023-02-27
[PLAT-733] Add favorite of repost notification

dylanjeffers:
### Description

Adds favorite of repost notification


packages/common/src/store/notifications/types.ts
```diff
@@ -179,6 +181,66 @@ export type RepostPushNotification = {
...
+  entityType: Entity.Playlist | Entity.Album | Entity.Track
+}
+
+export type FavoriteOfRepostPushNotification = {
+  blocknumber: number
+  entityId: ID
+  initiator: ID
+  timestamp: string
+  type:
+    | PushNotificationType.RepostAlbum
```
sabrina-kiam, line 227: since this is favorite of repost, not sure if this should be `PushNotificationType.SaveAlbum` or it's own type `PushNotificationType.SaveOfFavoriteAlbum`?? or maybe it is supposed to be repost here too? i'm not sure what this type is used for so i trust whatever you decide

packages/web/src/components/notification/Notification/FavoritedOfRepostNotification.tsx
```diff
@@ -0,0 +1,102 @@
+import { MouseEventHandler, useCallback } from 'react'
```
sabrina-kiam, line 1: nit - can we rename this file to `FavoriteOfRepostNotification` to match the mobile file name?

packages/web/src/components/notification/Notification/RepostOfRepostNotification.tsx
```diff
@@ -0,0 +1,106 @@
...
+          totalUserCount={userIds.length}
+          stopPropagation
+        />
+      </NotificationHeader>
+      <NotificationBody>
+        <UserNameLink user={firstUser} notification={notification} />{' '}
+        {otherUsersCount > 0 ? (
+          <OthersLink othersCount={otherUsersCount} onClick={handleClick} />
+        ) : null}{' '}
+        {messages.reposted} {entityType.toLowerCase()}{' '}
```
sabrina-kiam, line 100: tiny nit: i think you may need to either hard code a {' '} space here or add a space to the messages.reposted value

dylanjeffers: just double checked and i think we are good?

PR #2938, opened 2023-02-28
Send save_of_repost metadata on item save

sabrina-kiam:
### Description
This sends save of repost metadata to get processed by entity manager any time a track or collection gets reposted. Pretty much mimics what we've done for repost of a repost

### How Has This Been Tested?

Ran client locally against staging and printed metadata in audius backend endpoints to confirm the correct metadata is passed.

### DIFF
--- a/packages/common/src/services/audius-backend/AudiusBackend.ts
+++ b/packages/common/src/services/audius-backend/AudiusBackend.ts
@@ -1806,9 +1806,15 @@ export const audiusBackend = ({
   }
 
   // Favoriting a track
-  async function saveTrack(trackId: ID) {
+  async function saveTrack(
+    trackId: ID,
+    metadata?: { is_save_of_repost: boolean }
+  ) {
     try {
-      return await audiusLibs.EntityManager.saveTrack(trackId)
+      return await audiusLibs.EntityManager.saveTrack(
+        trackId,
+        JSON.stringify(metadata)
+      )
     } catch (err) {
       console.log(getErrorMessage(err))
       throw err
@@ -1829,9 +1835,15 @@ export const audiusBackend = ({
   }
 
   // Favorite a playlist
-  async function saveCollection(playlistId: ID) {
+  async function saveCollection(
+    playlistId: ID,
+    metadata?: { is_save_of_repost: boolean }
+  ) {
     try {
-      return await audiusLibs.EntityManager.savePlaylist(playlistId)
+      return await audiusLibs.EntityManager.savePlaylist(
+        playlistId,
+        JSON.stringify(metadata)
+      )
     } catch (err) {
       console.log(getErrorMessage(err))
       throw err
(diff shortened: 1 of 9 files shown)


audius-infra:
Preview this change https://demo.audius.co/sk-save-repost

packages/common/src/services/audius-backend/AudiusBackend.ts
```diff
@@ -1806,9 +1806,15 @@ export const audiusBackend = ({
...
-  async function saveTrack(trackId: ID) {
+  async function saveTrack(
+    trackId: ID,
+    metadata?: { is_save_of_repost: boolean }
+  ) {
     try {
-      return await audiusLibs.EntityManager.saveTrack(trackId)
+      return await audiusLibs.EntityManager.saveTrack(
+        trackId,
+        JSON.stringify(metadata)
```
dylanjeffers, line 1816: just curious if we need to json.stringify here. interesting that function doesn't do it for us? i guess would require libs update otherwise?

packages/web/src/common/store/social/collections/sagas.ts
```diff
@@ -334,10 +334,16 @@ export function* saveCollectionAsync(
   })
   yield* put(event)
 
+  const saveMetadata = action.isFeed
+    ? // If we're on the feed, and the content
+      // being saved is a repost
+      { is_save_of_repost: collection.followee_reposts.length !== 0 }
+    : { is_save_of_repost: false }
```
dylanjeffers, line 341: can this just be undefined?

packages/web/src/common/store/social/tracks/sagas.ts
```diff
@@ -342,7 +342,12 @@ export function* saveTrackAsync(
   })
   yield* put(event)
 
-  yield* call(confirmSaveTrack, action.trackId, user)
+  const saveMetadata = action.isFeed
+    ? // If we're on the feed, and the content
+      // being saved is a repost
+      { is_save_of_repost: track.followee_reposts.length !== 0 }
+    : { is_save_of_repost: false }
```
dylanjeffers, line 349: and same here, it prob doesn't make a diff but curious your thinking.

dylanjeffers (approved):
looks great!

packages/common/src/services/audius-backend/AudiusBackend.ts
sabrina-kiam: yea that function requires a string, and all social functions in entity manager libs have the same arguments, so expecting string inputs was cleaner than expecting diff dicts

packages/web/src/common/store/social/collections/sagas.ts
sabrina-kiam: yea could be undefined! i figured it's best to avoid passing undefined's when possible, and passing in no metadata to the backend defaults to is_save_of_repost being false anyways, so making the default explicit seemed cleanest. if you think otherwise and prefer undefined tho lmk!

PR #2940, opened 2023-02-28
[C-2209] Improve collection-screen perf

dylanjeffers:
### Description

Improves collection-screen perf by improving data selectors and refactoring rendered track-list. The main issue was `getLineupTableMetadatas` selects every track by uid, resulting in a very slow select for lineup when the cache is large. Now we limit the selections to just what is needed.

### DIFF
--- a/packages/mobile/src/screens/collection-screen/useFetchCollectionLineup.ts
+++ b/packages/mobile/src/screens/collection-screen/useFetchCollectionLineup.ts
@@ -4,16 +4,13 @@ import type { SmartCollectionVariant } from '@audius/common'
 import {
   Uid,
   areSetsEqual,
-  useProxySelector,
   Kind,
   makeUid,
   cacheActions,
   cacheCollectionsSelectors,
   collectionPageLineupActions,
   collectionPageSelectors,
-  queueSelectors,
-  lineupSelectors,
-  reachabilitySelectors
+  queueSelectors
 } from '@audius/common'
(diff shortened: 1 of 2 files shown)


packages/mobile/src/screens/collection-screen/useFetchCollectionLineup.ts
```diff
@@ -24,17 +21,13 @@ import { getOfflineTrackIds } from 'app/store/offline-downloads/selectors'
...
-const { getIsReachable } = reachabilitySelectors
-
-const getTracksLineup = makeGetTableMetadatas(getCollectionTracksLineup)
 
 /**
  * Returns a collection lineup, supports boths online and offline
  * @param collectionId the numeric collection id
  */
-export const useCollectionLineup = (
+export const useFetchCollectionLineup = (
```
sliptype, line 30: This is kinda strange because it no longer follows the same pattern as `useFavoritesLineup`. Could the new selectors be incorporated into `useCollectionLineup`?

sliptype (approved):
Looks great! Just one question about the lineup hook

packages/mobile/src/screens/collection-screen/useFetchCollectionLineup.ts
dylanjeffers: i think it def can! but then it feels a bit weird that the hook just magically knows what we need? thoughts?

PR #2941, opened 2023-02-28
[PAY-922] Mobile chat drawers and profile screen buttons

dharit-tan:
### Description
- Profile screen top right icon is kebab, opens ProfileActionsDrawer (share profile, block messages)
- Block messages via ProfileActionsDrawer -> BlockMessagesDrawer
- Chat screen top right icon is kebab, opens ChatActionsDrawer (visit profile, block messages)
- Mark chat as read when navigating away from chat screen.

Todo:
- Blocked/unblocked variants for BlockMessagesDrawer
- Destructive Block button doesn't fully match design yet (default state is red instead of transparent): https://www.figma.com/file/AUCQosYyHmbyYLWIC6GZLi?node-id=486:57949#380892935
- Could use ActionDrawer for ProfileActionsDrawer and ChatActionsDrawer - I gave it a try but couldn't get it right off the bat, will fast-follow later. Also requires modifying types in Common just to pass the modal name which seems silly.

### How Has This Been Tested?

Android emulator against stage

Screenshot_1677614811
Screenshot_1677615426
Screenshot_1677615432

### Feature Flags ###


packages/mobile/src/components/block-messages/BlockMessages.tsx
```diff
@@ -0,0 +1,163 @@
...
+  title: 'Are you sure?',
+  confirmText1: 'Are you sure you want to block ',
+  confirmText2: ' from sending messages to your inbox?',
+  info: 'This will not affect their ability to view your profile or interact with your content.',
+  blockUser: 'Block User',
+  cancel: 'Cancel'
+}
+
+const useStyles = makeStyles(({ spacing, typography, palette }) => ({
+  drawer: {
```
rickyrombo, line 30: nit: Seems like everything inside the drawer (except the title) is 16px apart - could we we use `gap: spacing(4)` instead of margins on each thing individually? (and then `marginBottom: spacing(2)` on the title container)

packages/mobile/src/components/block-messages/BlockMessages.tsx
```diff
@@ -0,0 +1,163 @@
...
+    marginBottom: spacing(6),
+    alignSelf: 'center'
+  },
+  title: {
+    fontSize: typography.fontSize.xl,
+    fontFamily: typography.fontByWeight.heavy,
+    color: palette.neutralLight2,
+    textTransform: 'uppercase',
+    lineHeight: typography.fontSize.xl * 1.25,
+    textAlign: 'center'
```
rickyrombo, line 46: I don't think we still need this?

packages/mobile/src/components/block-messages/BlockMessages.tsx
```diff
@@ -0,0 +1,163 @@
...
+        <View style={styles.titleContainer}>
+          <IconBlockMessages fill={neutralLight2} />
+          <Text style={styles.title}>{messages.title}</Text>
+        </View>
+        <Text style={styles.confirm}>
+          {messages.confirmText1}
+          {user?.name}
+          {messages.confirmText2}
+        </Text>
+        <View style={styles.infoContainer}>
```
rickyrombo, line 131: super nit: I've been seeing this component (the info icon + some text in a box) in some PremiumContent stuff (at least on web) - might be useful to ask if this is going to become a reusable thing, and break it out to its own component if so

packages/mobile/src/components/chat-actions-drawer/ChatActionsDrawer.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+import { Text } from 'app/components/core'
+import { NativeDrawer } from 'app/components/drawer'
+import { useNavigation } from 'app/hooks/useNavigation'
+import type { AppState } from 'app/store'
+import { getData } from 'app/store/drawers/selectors'
+import { setVisibility } from 'app/store/drawers/slice'
+import { makeStyles } from 'app/styles'
+
+const CHAT_ACTIONS_MODAL_NAME = 'ChatActions'
```
rickyrombo, line 12: no action necessary on this as I know it's already an established pattern, but it does concern me slightly that we have the overloading of the word `Action` for `ActionDrawer` 😬

packages/mobile/src/components/chat-actions-drawer/ChatActionsDrawer.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+const useStyles = makeStyles(({ spacing, typography, palette }) => ({
+  drawer: {
+    marginVertical: spacing(7),
+    alignItems: 'center'
+  },
+  text: {
+    fontSize: 21,
+    lineHeight: spacing(6.5),
+    letterSpacing: 0.233333,
```
rickyrombo, line 27: Huh. Saw this matches figma, but just noting that this seems odd... 🤔

Flagging for @piazzatron these are the kinda weird styles that make me raise an eyebrow and pray they don't render funny

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -260,6 +263,33 @@ export const ChatScreen = () => {
...
+    dispatch(
+      setVisibility({
+        drawer: 'ChatActions',
+        visible: true,
+        data: { userId: otherUser.user_id }
+      })
+    )
+  }
+
+  const isLoading = status === Status.LOADING && chatMessages?.length === 0
```
rickyrombo, line 285: Oooh clever!!

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -279,11 +309,11 @@ export const ChatScreen = () => {
             )
           : messages.title
       }
-      topbarRight={null}
+      topbarRight={topBarRight}
     >
       <ScreenContent>
         <View style={styles.rootContainer}>
-          {status === Status.SUCCESS ? (
+          {isLoading ? (
```
rickyrombo, line 316: This seems backwards to me?

packages/mobile/src/screens/profile-screen/MessageButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { useDispatch } from 'react-redux'
+
+import IconMessage from 'app/assets/images/iconMessage.svg'
+import { Button } from 'app/components/core'
+import { makeStyles } from 'app/styles'
+
+const { createChat } = chatActions
+
+const messages = {
+  chat: 'Chat'
```
rickyrombo, line 14: nit: this should probably be "Message" not "Chat"

rickyrombo (approved):
I think there might be some backwards boolean logic? Otherwise looks good for now until we can do another cycle and convert some of these to ActionDrawers

packages/mobile/src/components/block-messages/BlockMessages.tsx
dharit-tan: dude HELL yes

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
dharit-tan: good catch!

PR #2942, opened 2023-02-28
[PAY-972] Handle duplicate wallet connections properly

piazzatron:
### Description

We would only show the error if you'd previously connected a wallet, on the `WalletsTable`
Only shown here previously:
Screen Shot 2023-02-28 at 3 34 14 PM

Now also showing here:

Screen Shot 2023-02-28 at 3 41 41 PM

### Dragons

[] Need to make sure this works on mobile
[] Doublechecking that this text goes away when a wallet *is* successfully connected

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/web/src/pages/audio-rewards-page/components/WalletsTable.tsx
+++ b/packages/web/src/pages/audio-rewards-page/components/WalletsTable.tsx
@@ -139,19 +139,16 @@ type WalletsTableProps = {
   className?: string
   hasActions?: boolean
   hideCollectibles?: boolean
-  suppressError?: boolean
 }
 
 const WalletsTable = ({
   hasActions = false,
   className,
-  hideCollectibles,
-  suppressError
+  hideCollectibles
 }: WalletsTableProps) => {
   const {
     status,
     confirmingWallet,
-    errorMessage,
     connectedEthWallets: ethWallets,
     connectedSolWallets: solWallets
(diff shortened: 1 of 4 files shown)


audius-infra:
Preview this change https://demo.audius.co/piazz-wallet-duplicates

packages/web/src/pages/audio-rewards-page/components/WalletsTable.tsx
```diff
@@ -251,7 +253,9 @@ const WalletsTable = ({
           isConfirmRemoving={false}
         />
       )}
-      {errorMessage && <div className={styles.error}>{errorMessage}</div>}
+      {!suppressError && errorMessage ? (
```
dylanjeffers, line 256: should we just remove this here, or is it needed for another component usage?

dylanjeffers (approved):
make sense, just wondering about if we need the duplicate or not

packages/web/src/pages/audio-rewards-page/components/WalletsTable.tsx
piazzatron: Ah I think you're right and we can remove it actually. It's used in the AudioBreakdownModal but don't think we care

piazzatron: fixed. good call

PR #2944, opened 2023-02-28
[C-2205] Wait for user to follow in sign up

sliptype:
### Description

We were rendering `HomeStack` as soon as `hasAccount` was true, meaning that if the account finished creating while the user was still on the screen to pick first follows it would automatically switch to the feed screen
* Add a couple new actions to track sign up progress on mobile (`startedSignOnProcess` was not used)
* Add a selector `getHasCompletedAccount` to designate that the user has an account and they have completed the sign up flow
* Ensure the Rate and Notification drawers don't get shown until the user has a completed account

### Dragons

Sign up changes! Dangerous

### How Has This Been Tested?

Tested:
* Sign up flow with user taking long time to follower accounts
* Sign up flow with user quickly following accounts
* Sign in

### Feature Flags ###

### DIFF
--- a/packages/web/src/common/store/pages/signon/selectors.js
+++ b/packages/web/src/common/store/pages/signon/selectors.js
@@ -18,8 +18,10 @@ export const getToastText = (state) => state.signOn.toastText
 export const getRouteOnCompletion = (state) => state.signOn.routeOnCompletion
 export const getRouteOnExit = (state) => state.signOn.routeOnExit
 export const getAccountReady = (state) => state.signOn.accountReady
-export const getStartedSignOnProcess = (state) =>
-  state.signOn.startedSignOnProcess
+export const getStartedSignUpProcess = (state) =>
+  state.signOn.startedSignUpProcess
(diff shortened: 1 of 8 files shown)


raymondjacobson (approved):
makes sense

packages/web/src/common/store/pages/signon/selectors.js
```diff
@@ -31,6 +34,16 @@ export const getSuggestedFollowIds = (state) => {
   return categories[selectedCategory] || []
 }
 
+export const getHasCompletedAccount = createSelector(
```
dylanjeffers, line 37: nice

dylanjeffers (approved):
def all makes sense to me, potentially could have a sign-up status that maps to the various phases better, but two bools is fine too.

sliptype:
> def all makes sense to me, potentially could have a sign-up status that maps to the various phases better, but two bools is fine too.

Yeah we have the page in redux but it only gets set in web. I was thinking through updating the redux page when navigating but there were a lot of cases to catch so I kept it simple for now

PR #2953, opened 2023-03-01
[C-2228] Fix link position in Hyperlink

sliptype:
### Description

* Fix positioning of links in Hyperlink component

This code is very hacky because we need to render the links in a separate stacking context to allow touches to pass through the bio text to scroll the header.

The problem was a race condition between the `linkRefs` ref being set and the useEffect that measures the positions of those elements. Was able to fix by not using a ref for `linkRefs` and tightening up the deps of `renderLink`. Also needed to iterate over `links` instead of `linkRefs` because `linkRefs` potentially exist before the layout of the element occurs, resulting in no position when we measure the element

Simulator Screen Shot - iPhone 14 Pro - 2023-03-01 at 16 22 11

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/core/Hyperlink.tsx
+++ b/packages/mobile/src/components/core/Hyperlink.tsx
@@ -55,7 +55,7 @@ export const Hyperlink = (props: HyperlinkProps) => {
   const styles = useStyles()
 
   const linkContainerRef = useRef<View>(null)
-  const linkRefs = useRef<Record<number, Text>>({})
+  const [linkRefs, setLinkRefs] = useState<Record<number, Text>>({})
   const [links, setLinks] = useState<Record<number, PositionedLink>>({})
   const [linkLayouts, setLinkLayouts] = useState<
     Record<number, LayoutRectangle>
@@ -68,13 +68,16 @@ export const Hyperlink = (props: HyperlinkProps) => {
    * android doesn't return the correct layout for nested text elements
    * */
   useEffect(() => {
-    Object.entries(linkRefs.current).forEach(([index, ref]) => {
-      ref.measureInWindow((x, y, width, height) => {
-        setLinkLayouts((linkLayouts) => ({
-          ...linkLayouts,
-          [index]: { x, y, width, height }
-        }))
-      })
+    Object.entries(links).forEach(([index]) => {
+      const linkRef = linkRefs[index]
+      if (linkRef) {
+        linkRef.measureInWindow((x, y, width, height) => {
+          setLinkLayouts((linkLayouts) => ({
+            ...linkLayouts,
+            [index]: { x, y, width, height }
+          }))
+        })
+      }
     })
 
     if (linkContainerRef.current) {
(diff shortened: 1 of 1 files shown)


packages/mobile/src/components/core/Hyperlink.tsx
```diff
@@ -68,47 +68,55 @@ export const Hyperlink = (props: HyperlinkProps) => {
    * android doesn't return the correct layout for nested text elements
    * */
   useEffect(() => {
```
dylanjeffers, line 70: would useLayoutEffect help? prob not cause you actually want it running after the first paint?

packages/mobile/src/components/core/Hyperlink.tsx
```diff
@@ -68,47 +68,55 @@ export const Hyperlink = (props: HyperlinkProps) => {
...
-        setLinkLayouts((linkLayouts) => ({
-          ...linkLayouts,
-          [index]: { x, y, width, height }
-        }))
-      })
+    Object.entries(links).forEach(([index]) => {
+      const linkRef = linkRefs[index]
+      if (linkRef) {
+        linkRef.measureInWindow((x, y, width, height) => {
+          setLinkLayouts((linkLayouts) => ({
```
dylanjeffers, line 75: does multiple sets in a for loop like this get merged into a single one in the end? otherwise wondering if we do map and then a set after?

packages/mobile/src/components/core/Hyperlink.tsx
```diff
@@ -68,47 +68,55 @@ export const Hyperlink = (props: HyperlinkProps) => {
...
+  }, [links, linkRefs, linkContainerRef])
 
   const handlePress = useOnOpenLink(source)
 
   const renderLink = useCallback(
     (text, match, index) => (
       <View
         onLayout={(e) => {
-          setLinks({
+          setLinks((links) => ({
```
dylanjeffers, line 96: callback the onLayout do anything?

packages/mobile/src/components/core/Hyperlink.tsx
```diff
@@ -68,47 +68,55 @@ export const Hyperlink = (props: HyperlinkProps) => {
...
               text,
               match
             }
-          })
+          }))
         }}
         ref={(el) => {
-          if (linkRefs.current && el) {
-            linkRefs.current[index] = el
+          if (el) {
```
dylanjeffers, line 105: ah quite cool, also callbacking do anything?

dylanjeffers (approved):
this is such a cool change, great job figuring out

packages/mobile/src/components/core/Hyperlink.tsx
sliptype: I think fine in this case because basically a leaf component that is rerendering anyway

packages/mobile/src/components/core/Hyperlink.tsx
sliptype: Same as above

PR #2954, opened 2023-03-01
[C-2232] Fix android share-to-twitter url

dylanjeffers:
### Description

Fixis android share-to-twitter by replacing /share url with /intent/tweet url. It turns out this works for ios and web too, so consolidated our twitterShareUrl functions in common. 
### Dragons

Not sure if this was tried before or if there is a fatal flaw with /intent/tweet? docs here https://developer.twitter.com/en/docs/twitter-for-websites/tweet-button/overview and issue that i referenced: https://stackoverflow.com/questions/6208363/sharing-a-url-with-a-query-string-on-twitter


sliptype (approved):
Looks great!

audius-infra:
Preview this change https://demo.audius.co/dj-c-2232-fix-android-twitter-share

PR #2957, opened 2023-03-02
[C-2194] Disabled the repost and favorite buttons in the NowPlayingDrawer for track owner

Kyle-Shanks:
### Description
* Disabled the repost and favorite buttons in the NowPlayingDrawer for the track owner
  * Display a toast as well
* Add queue removal logic for TrackPlayer to prevent the track reset bug

### Dragons

Did some testing, but there may we some weird edge cases with the new TrackPlayer queue removal logic

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/mobile/src/components/audio/Audio.tsx
+++ b/packages/mobile/src/components/audio/Audio.tsx
@@ -20,7 +20,7 @@ import {
   shallowCompare,
   savedPageTracksLineupActions
 } from '@audius/common'
-import { isEqual } from 'lodash'
+import { isEqual, range } from 'lodash'
 import TrackPlayer, {
   AppKilledPlaybackBehavior,
   Capability,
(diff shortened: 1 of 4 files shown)


packages/mobile/src/components/audio/Audio.tsx
```diff
@@ -444,10 +444,28 @@ export const Audio = () => {
...
+    const isQueueRemoval =
+      refUids.length > 0 &&
+      isEqual(refUids.slice(0, queueTrackUids.length), queueTrackUids)
+
+    if (isQueueRemoval) {
+      // NOTE: There might be a case where we are trying to remove the currently playing track.
+      // Shouldn't be possible, but need to keep an eye out for that
+      const startingRemovalIndex = queueTrackUids.length
+      const removalLength = refUids.length - queueTrackUids.length
+      const removalIndexArray = [...Array(removalLength).keys()].map(
```
sliptype, line 462: Might be easier to use lodash `range` here!

sliptype (approved):
Awesome! 

I assume you made the prop `isOwner` instead of `isDisabled` because the button isn't really disabled since it shows a toast? I do wonder if there could be other cases where we want this same behavior but for a different reason. Like premium content or something could prevent saves/reposts based on some criteria. But I think fine for now!

PR #2973, opened 2023-03-06
[PAY-955][PAY-983][PAY-992] Add UI for unsupported collections and other fixes

sddioulde:
### Description

Add web and native mobile UI for unsupported collections.

Also fix small premium content label bug on native mobile. (not super related but wanted to get this change in also)

### Dragons

### How Has This Been Tested?

local web dapp and mobile app vs stage

### Feature Flags ###

existing


packages/web/src/components/data-entry/DropdownInput.module.css
```diff
@@ -230,3 +233,11 @@
   line-height: 28px;
   height: 32px;
 }
+
+.select .footer:global(.ant-select-item-option) {
+  cursor: default;
+  margin: 0 -8px;
+  padding: 0px !important;
```
sddioulde, line 240: smell here, but antd styling is weird. don't want to spend too much time on this at this particular time

packages/mobile/src/screens/edit-track-screen/screens/ListSelectionScreen.tsx
```diff
@@ -44,6 +45,7 @@ const useStyles = makeStyles(({ spacing, typography }) => ({
     justifyContent: 'space-between'
   },
   content: { flex: 1 },
+  noFlex: { flex: undefined },
```
dharit-tan, line 48: Not quite following why no flex is needed when there's a footer, but I trust.

packages/mobile/src/screens/edit-track-screen/screens/RemixSettingsScreen.tsx
```diff
@@ -233,20 +218,22 @@ export const RemixSettingsScreen = () => {
...
-                <Text>{messages.changeAvailbilitySuffix}</Text>
-              </View>
-            </View>
-          )}
+          {isPremium ? (
+            <HelpCallout
+              style={styles.changeAvailability}
+              content={
+                <View style={styles.changeAvailabilityText}>
+                  <Text>{messages.changeAvailbilityPrefix}</Text>
```
dharit-tan, line 226: This seems to indicate that availability needs to be changed only when it's not public, but I don't see logic for that.

packages/web/src/components/track/PremiumContentLabel.tsx
```diff
@@ -29,21 +29,7 @@ export const PremiumContentLabel = ({
     return null
   }
 
-  if (isOwner) {
```
dharit-tan, line 32: Are we no longer showing the labels if `isOwner` or do they appear somewhere else?

dharit-tan (approved):

dharit-tan:
Need to run prettier!

packages/mobile/src/screens/edit-track-screen/screens/ListSelectionScreen.tsx
sddioulde: good question. Without this line, the footer gets stuck to the bottom of the screen even if there is only one or a few nft collections in the list. we want the footer to follow right after the list.

packages/web/src/components/track/PremiumContentLabel.tsx
sddioulde: we are. the logic is refactored and that path would hit the below block

packages/mobile/src/screens/edit-track-screen/screens/RemixSettingsScreen.tsx
sddioulde: this shows info that says that we are only able to mark as remix if the track is not premium

audius-infra:
Preview this change https://demo.audius.co/sd-unsupported

PR #2983, opened 2023-03-07
[PAY-994] Change gated content feature flag names

sddioulde:
### Description

Change feature flag names so that old clients won't accidentally get partial features.
The old feature flags are still kept in optimizely until we determine later that clients don't have old app versions.

### Dragons

make sure that all old feature flags should have been renamed. this should be the case

### How Has This Been Tested?

local web and mobile apps

### Feature Flags ###

the new feature flags themselves


audius-infra:
Preview this change https://demo.audius.co/sd-gated-ff

dharit-tan (approved):

PR #2991, opened 2023-03-07
[PAY-1019] Fix premium content tags overlap track title

dharit-tan:
### Description
Fixes UI bug where premium content tags were overlapping track tiles by truncating them earlier.
Bonus: Fixes style for ellipses.

https://www.figma.com/file/OBKeQ86g7vyzSnrn4ziLuz/NFT-Gated-Content?node-id=996-22293&t=QwvwSrYISQKCaZTq-0

### Dragons

### How Has This Been Tested?

Local web.

Screenshot 2023-03-07 at 7 05 47 PM copy

(Before ellipses style fix)
Screenshot 2023-03-07 at 6 56 26 PM copy

### Feature Flags ###

### DIFF
--- a/packages/web/src/components/track/desktop/TrackTile.module.css
+++ b/packages/web/src/components/track/desktop/TrackTile.module.css
@@ -147,6 +147,10 @@
   height: 20px;
 }
 
+.withPremium {
+  margin-right: 188px;
+}
+
 .title {
   font-style: normal;
   font-weight: var(--font-bold);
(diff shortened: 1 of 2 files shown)


rickyrombo (approved):
~~This works, but if we had more time I'd like to see us restructure a bit so that other tags at the top right are accounted for with relative widths (ie. special access etc might not be 188px exactly)~~

~~Ideally this could be done without absolute positioning~~

eh, this is actually more what the design looks like I guess. 🤷

dharit-tan:
> ~This works, but if we had more time I'd like to see us restructure a bit so that other tags at the top right are accounted for with relative widths (ie. special access etc might not be 188px exactly)~
> 
> ~Ideally this could be done without absolute positioning~
> 
> eh, this is actually more what the design looks like I guess. 🤷

Nah I hear you, just wanted a quick fix without getting into measuring and stuff for now.

PR #2993, opened 2023-03-08
Fix lineups with premium content infinite loading for signed out users

rickyrombo:
### Description

We're not TBD on user access when a user isn't signed in, so `isUserAccessTBD` should be false.

### Dragons

~~Might flicker premium tracks as "locked" until the account is set in the store for tracks that the user _does_ have access to? Will test~~

This seems to be a non issue- lineups are generally loaded after accounts are. Also would only happen on very initial load.

### How Has This Been Tested?

Tested using local prod build against some of the accounts we were seeing infinite load on LP

### Feature Flags ###

### DIFF
--- a/packages/common/src/hooks/usePremiumContent.ts
+++ b/packages/common/src/hooks/usePremiumContent.ts
@@ -1,7 +1,9 @@
-import { Chain, PremiumConditions, Track } from 'models'
 import { useMemo } from 'react'
 
 import { useSelector } from 'react-redux'
+
+import { Chain, PremiumConditions, Track } from 'models'
+import { getAccountUser } from 'store/account/selectors'
 import { cacheTracksSelectors, cacheUsersSelectors } from 'store/cache'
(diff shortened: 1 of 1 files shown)


packages/common/src/hooks/usePremiumContent.ts
```diff
@@ -28,20 +32,26 @@ export const usePremiumContentAccess = (track: Nullable<Partial<Track>>) => {
     const isSignatureToBeFetched =
       isCollectibleGated &&
       !!trackId &&
-      premiumTrackSignatureMap[trackId] === undefined
+      premiumTrackSignatureMap[trackId] === undefined &&
+      !!user // We're only fetching a sig if the user is logged in
```
rickyrombo, line 36: This is the only real change, the rest is pretty much just imports and autoformatting by prettier

piazzatron (approved):
makes sense!! nice

PR #3001, opened 2023-03-09
Remove upgradeToCreator

theoilie:
### Description
`upgradeToCreator()` has not existed in libs for a long time, so this PR removes the wallet connection logic that tries to call `upgradeToCreator()`

### Dragons
Most users should already have `metadata_multihash` and `creator_node_endpoint` (including new users - I tested that), and Entity Manager is handling them now (tagged Isaac for visibility), so this path was likely almost never called. [Sentry logs](https://audius.sentry.io/issues/3966211875/?project=1457231&query=is%3Aunresolved+upgradeToCreator&referrer=issue-stream&statsPeriod=14d) show one error in the past 14 days for this function (and the error was because the function doesn't exist in libs anyway so there's no point in trying to call it in client)

### How Has This Been Tested?
Created a new user, verified it has `metadata_multihash` and `creator_node_endpoint` at http://audius-protocol-discovery-provider-1:5000/v1/full/users/handle/<handle>, and successfully connected a wallet

Monitor for `CONNECT_WALLET_ASSOCIATION_ERROR` errors

### Feature Flags ###
None

### DIFF
--- a/packages/common/src/services/audius-backend/AudiusBackend.ts
+++ b/packages/common/src/services/audius-backend/AudiusBackend.ts
@@ -1223,14 +1223,6 @@ export const audiusBackend = ({
     }
   }
 
-  /**
-   * Upgrades a user to a creator
-   * @param {string} newCreatorNodeEndpoint will follow the structure 'cn1,cn2,cn3'
-   */
-  async function upgradeToCreator(newCreatorNodeEndpoint: string) {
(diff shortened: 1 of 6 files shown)


audius-infra:
Preview this change https://demo.audius.co/theo-remove-upgradeToCreator

audius-infra:
Preview this change https://demo.audius.co/theo-remove-upgradeToCreator

rickyrombo (approved):

PR #3013, opened 2023-03-09
Abort jobs on clearOfflineDownloads

amendelsohn:
### Description

Fix issue where a job could sneak through after removing all downloads on logout or using the new settings button.

### Dragons

Careful not to break existing job flows and accidentally cancel stuff.

### DIFF
--- a/packages/mobile/src/store/offline-downloads/sagas/offlineQueueSagas/workers/downloadCollectionWorker.ts
+++ b/packages/mobile/src/store/offline-downloads/sagas/offlineQueueSagas/workers/downloadCollectionWorker.ts
@@ -33,6 +33,7 @@ import {
 } from '../../../slice'
 import { isCollectionDownloadable } from '../../utils/isCollectionDownloadable'
 import { retryOfflineJob } from '../../utils/retryOfflineJob'
+import { shouldAbortJob } from '../../utils/shouldAbortJob'
 import { shouldCancelJob } from '../../utils/shouldCancelJob'
 
 import { downloadFile } from './downloadFile'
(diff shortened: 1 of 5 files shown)


packages/mobile/src/store/offline-downloads/sagas/offlineQueueSagas/workers/downloadCollectionWorker.ts
```diff
@@ -58,18 +59,19 @@ export function* downloadCollectionWorker(collectionId: CollectionId) {
...
-  const { jobResult, cancel, abort } = yield* race({
+  const { jobResult, cancel, abort, abortJob } = yield* race({
     jobResult: retryOfflineJob(
       MAX_RETRY_COUNT,
       1000,
       downloadCollectionAsync,
       collectionId
     ),
     abort: call(shouldAbortDownload, collectionId),
+    abortJob: call(shouldAbortJob),
```
dylanjeffers, line 70: so would we not want this to be a part of "abort" as well?

dylanjeffers (approved):
looks great but imo would be cool to combine the shouldAbortJob with shouldAbortDownload? maybe call the whole thing abortJob?

packages/mobile/src/store/offline-downloads/sagas/offlineQueueSagas/workers/downloadCollectionWorker.ts
amendelsohn: we take different action afterwards depending on whether it was cancelled or aborted, so I don't think that'll work

amendelsohn: yeah I was trying to think of the best way to combine them

PR #3014, opened 2023-03-10
Fix fetchNotifications fallback

dylanjeffers:
### Description

Fixes case where there may be no validNotifications, resulting in a runtime error. this kinda seems like a bug typescript should be able to catch, since we are indexing into an array. maybe we can check if there is a typescript option to prevent this in the future?

### DIFF
--- a/packages/web/src/common/store/notifications/fetchNotifications.ts
+++ b/packages/web/src/common/store/notifications/fetchNotifications.ts
@@ -96,7 +96,7 @@ export function* fetchNotifications(config: FetchNotificationsParams) {
       if (invalidNotifications.length !== 0) {
         const newLimit = limit - validNotifications.length
         const newTimestamp = moment(
-          validNotifications[validNotifications.length - 1].timestamp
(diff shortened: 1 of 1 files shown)


sliptype (approved):
Is this the best way to fix this? If there are no valid notifications we are passing `undefined` to moment. Could we just check the length of validNotifications before?

dylanjeffers:
> Is this the best way to fix this? If there are no valid notifications we are passing `undefined` to moment. Could we just check the length of validNotifications before?

happy to change it, but i checked that moment(undefined) is moment() -> now which is what we want. its convenient, but maybe not clear?

sliptype:
> > Is this the best way to fix this? If there are no valid notifications we are passing `undefined` to moment. Could we just check the length of validNotifications before?
> 
> happy to change it, but i checked that moment(undefined) is moment() -> now which is what we want. its convenient, but maybe not clear?

Ah ok that seems fine

audius-infra:
Preview this change https://demo.audius.co/dj-fix-fetch-notifications-fallback

PR #3005, opened 2023-03-09
Remove identity subscriptions reads/writes

michellebrier:
### Description
Subscriptions have been migrated to discovery. Remove legacy identity paths. Also remove the subscribe on follow feature flag as it seems to have been fully rolled out for some time now.
Also, in discovery, entity manager automatically subscribes on follow, so only set `action.update = true` and send the relay when toggling the subscribe button (not when toggling follow/unfollow).

DO NOT MERGE until `read_subscribers_from_discovery_enabled` has been turned on.

### Dragons

*Is there anything the reviewer should be on the lookout for? Are there any dangerous changes?*

### How Has This Been Tested?

*Please describe the tests that you ran to verify your changes. Provide repro instructions & any configuration.*

### How will this change be monitored?

*For features that are critical or could fail silently please describe the monitoring/alerting being added.*

### Feature Flags ###
Remove `auto_subscribe_on_follow` and `read_subscribers_from_discovery_enabled`.

*Are all new features properly feature flagged? Describe added feature flags.*

PR #3015, opened 2023-03-10
Read subscribers from bulk endpoint

michellebrier:
### Description
Bulk endpoint retrieves only subscriber ids (no other user info) and has no pagination limits. The current path using /user/id/subscribers is buggy because of the pagination defaults.

Need this code change to be in before turning on the `read_subscriptions_from_discovery` flag. After the flag is on, can deploy these PRs https://github.com/AudiusProject/audius-client/pull/3005 and https://github.com/AudiusProject/audius-protocol/pull/4909 removing the legacy identity subscriptions path.

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/common/src/services/audius-backend/AudiusBackend.ts
+++ b/packages/common/src/services/audius-backend/AudiusBackend.ts
@@ -2834,9 +2834,7 @@ export const audiusBackend = ({
         const encodedUserId = encodeHashId(userId)
         const bulkResp: { user_id: string; subscriber_ids: string[] }[] =
           await audiusLibs.User.bulkGetUserSubscribers([encodedUserId])
-        const encodedSubscriberIds = bulkResp.find(
-          (user) => user.user_id === encodedUserId
-        )!.subscriber_ids
+        const encodedSubscriberIds = bulkResp[0].subscriber_ids
         const subscriberIds = encodedSubscriberIds.map((id) => decodeHashId(id))
         return subscriberIds.includes(account.user_id)
       } catch (e) {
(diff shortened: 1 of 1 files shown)


packages/common/src/services/audius-backend/AudiusBackend.ts
```diff
@@ -2832,12 +2831,13 @@ export const audiusBackend = ({
...
       // Read subscribers from discovery
       try {
-        const subscribers: APIUser[] = await audiusLibs.User.getUserSubscribers(
-          encodeHashId(userId)
-        )
-        const subscriberIds = subscribers.map((subscriber) =>
-          decodeHashId(subscriber.id)
-        )
+        const encodedUserId = encodeHashId(userId)
+        const bulkResp: { user_id: string; subscriber_ids: string[] }[] =
```
dylanjeffers, line 2835: So does this return an array of same length as requested user ids?

packages/common/src/services/audius-backend/AudiusBackend.ts
```diff
@@ -2832,12 +2831,13 @@ export const audiusBackend = ({
...
+          await audiusLibs.User.bulkGetUserSubscribers([encodedUserId])
+        const encodedSubscriberIds = bulkResp.find(
+          (user) => user.user_id === encodedUserId
+        )!.subscriber_ids
```
dylanjeffers, line 2839: Ideally we don't use !, but if it's always the user in the first spot of the array, can maybe just destructure instead of using find?

dylanjeffers (approved):
Looks great, we might want to handle cases where values are missing instead of using ! though, lmk what you think

packages/common/src/services/audius-backend/AudiusBackend.ts
michellebrier: yes is should -- there's a catch in case of an error though

PR #3018, opened 2023-03-10
[C-2276] Fix upload track store updates

dylanjeffers:
### Description

Fixes issue where uploading a track does not add uploaded track to cache, or update user track_count, or update user's track lineup. This fixes the case where going to your profile after uploading a track wasnt showing the track.


packages/web/src/common/store/pages/profile/lineups/tracks/watchUploadTracksSaga.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+function* addUploadedTrackToLineup(action: UploadTracksSucceededAction) {
+  const { id, trackMetadatas } = action
+  const accountHandle = yield* select(getUserHandle)
+  if (!id || !trackMetadatas || !accountHandle) return
+
+  // We will only be adding single track uploads since multi-adds for lineups
+  // are cumbersome. When we want to suport multi-track upload on mobile or
+  // properly cache desktop profile lineups, we should revisit this.
```
Kyle-Shanks, line 28: 👍

Kyle-Shanks (approved):

raymondjacobson (approved):
Seems good as a near-term fix. Ideally architecturally would be handled better, do you have thoughts on how that could work?

Maybe we should just re-fetch the account and feed after the upload?

dylanjeffers:
> Seems good as a near-term fix. Ideally architecturally would be handled better, do you have thoughts on how that could work?
> 
> Maybe we should just re-fetch the account and feed after the upload?

Im open to anything, i based this off what we do when we favorite a track, adding it to favorite lineup etc. Do we want to do something different in this case? i do like idea of simply refetching account + lineups

audius-infra:
Preview this change https://demo.audius.co/dj-c-2276-fix-upload-track-store-updates

PR #3019, opened 2023-03-10
Add playback position logic on web for podcasts

Kyle-Shanks:
### Description
Adding playback position logic on web for podcasts. Does not use redux persist yet, but will be updated in the future.

* Add common sagas to store playback position info in local storage
* Update player sagas in web to listen for certain events and update the playback position
* Update track page to display podcast playback specific information and update the play button based on that
* Update lineup/track tile to display playback information for podcasts
* Update overflow menus to display playback options like marking as played/unplayed

Screenshot 2023-03-10 at 2 45 55 PM
Screenshot 2023-03-10 at 2 46 06 PM
Screenshot 2023-03-10 at 2 46 28 PM

### Dragons

N/A

### How Has This Been Tested?

Lots of manual testing

N/A

### Feature Flags ###

All changes should be behind the PODCAST_CONTROLS_ENABLED flag


packages/common/src/store/playback-position/sagas.ts
```diff
@@ -0,0 +1,57 @@
...
+    JSON.parse(localStorageState)
+  yield* put(setPlaybackPositionState({ playbackPositionState }))
+}
+
+function* watchTrackPositionUpdate() {
+  const setLocalStorageItem = yield* getContext('setLocalStorageItem')
+  yield* takeEvery(
+    [setTrackPosition.type, clearTrackPosition.type],
+    function* (
+      _action:
```
dylanjeffers, line 42: since we dont use the action, i think we can just have a function with no args?

packages/common/src/store/playback-position/slice.ts
```diff
@@ -0,0 +1,61 @@
...
+const initialState: PlaybackPositionState = {
+  trackPositions: {}
+}
+
+const slice = createSlice({
+  name: 'playback-position',
+  initialState,
+  reducers: {
+    // NOTE: This should only be called when seeding the initial state from local storage
+    setPlaybackPositionState: (
```
dylanjeffers, line 29: maybe using "initialize" instead of "set" can help invoke this is a one-time deal?

packages/common/src/store/playback-position/types.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export type PlaybackStatus = 'IN_PROGRESS' | 'COMPLETED'
+
+export type PlaybackPositionInfo = {
+  status: PlaybackStatus
+  playbackPosition: number
+}
+
+export type PlaybackPositionState = {
+  trackPositions: Record<ID, PlaybackPositionInfo>
```
dylanjeffers, line 13: so one thing i learned with Record, is that trackPositions[1] will always return PlaybackPositionInfo, which is technically not true, so maybe we should do 
```
trackPositions: {[Key in ID]?: PlaybackPositionInfo}
```

packages/common/src/store/reducers.ts
```diff
@@ -55,6 +55,8 @@ import trendingPlaylists from './pages/trending-playlists/slice'
 import trendingUnderground from './pages/trending-underground/slice'
 import trending from './pages/trending/reducer'
 import { TrendingPageState } from './pages/trending/types'
+import { PlaybackPositionState } from './playback-position'
+import playbackPosition from './playback-position/slice'
```
dylanjeffers, line 59: Im not opposed to this as the slice name, just wondering what we might want to call this? im thinking a little into the future, how we it might be more that just "position", it could be playback rate, playback timestamp, number, status, maybe play-count or something? just wanted to flag the idea on what would best encapsulate that.... maybe just "PlaybackState"?

packages/web/src/common/store/player/sagas.ts
```diff
@@ -183,6 +191,33 @@ export function* watchPlay() {
...
         const playbackRate = yield* select(getPlaybackRate)
         audioPlayer.setPlaybackRate(playbackRate)
+
+        const isNewPodcastControlsEnabled = yield* call(
+          getFeatureEnabled,
+          FeatureFlags.PODCAST_CONTROL_UPDATES_ENABLED
+        )
+        if (isNewPodcastControlsEnabled) {
+          // Set playback position for track to in progress if not already tracked
+          const trackPositions = yield* select(getTrackPositions)
```
dylanjeffers, line 201: maybe `const trackPlaybackInfo = yield* select(getTrackPosition, {id: trackId})`

packages/web/src/common/store/player/sagas.ts
```diff
@@ -183,6 +191,33 @@ export function* watchPlay() {
...
+              setTrackPosition({
+                trackId,
+                positionInfo: {
+                  status: 'IN_PROGRESS',
+                  playbackPosition: 0
+                }
+              })
+            )
+          } else {
+            // TODO: This looks bad, fix later
```
dylanjeffers, line 213: just wondering what ur thinkin here!

packages/web/src/common/store/player/sagas.ts
```diff
@@ -399,7 +434,51 @@ function* recordListenWorker() {
       if (trackId) yield* put(recordListen(trackId))
       lastSeenPlayCounter = playCounter
     }
-    yield* delay(RECORD_LISTEN_INTERVAL)
+
+    if (position) yield* delay(RECORD_LISTEN_INTERVAL)
```
dylanjeffers, line 438: just a check that position can equal 0, so want to double check if that case matters here

packages/web/src/common/store/player/sagas.ts
```diff
@@ -399,7 +434,51 @@ function* recordListenWorker() {
...
+  const duration = await audioPlayer.getDuration()
+  return {
+    position,
+    duration
+  }
+}
+
+/**
+ * Poll for tracking the playback position of tracks
+ */
```
dylanjeffers, line 453: record makes me think "tracking", so maybe "save"? what you think?

packages/web/src/common/store/player/sagas.ts
```diff
@@ -399,7 +434,51 @@ function* recordListenWorker() {
...
+        setTrackPosition({
+          trackId,
+          positionInfo: {
+            status: isComplete ? 'COMPLETED' : 'IN_PROGRESS',
+            playbackPosition: position
+          }
+        })
+      )
+    }
+    yield* delay(RECORD_PLAYBACK_POSITION_INTERVAL)
```
dylanjeffers, line 481: so does this loop even when tracks aren't playing technically? i think action channels might be the best option here, so we only need to run this while a track is actually playing? lmk what you think and maybe us + @amendelsohn can think of what that type of saga would look like.

packages/web/src/components/track/GiantTrackTile.js
```diff
@@ -303,22 +314,66 @@ class GiantTrackTile extends PureComponent {
     )
   }
 
+  renderProgressInfo() {
```
dylanjeffers, line 317: it might be nice to isolate this component so we can also isolate it's selectors for positionInfo etc? i know a lot of these components aren't really set up for this though. another win is that we can write this isolated component in typescript too. lmk and i can help set that up since i have experience adding small ts components in a large web js component

packages/web/src/components/track/GiantTrackTile.js
```diff
@@ -303,22 +314,66 @@ class GiantTrackTile extends PureComponent {
...
+      <>
+        <RepostFavoritesStats
+          isUnlisted={isUnlisted}
+          repostCount={repostCount}
+          saveCount={saveCount}
+          onClickReposts={onClickReposts}
+          onClickFavorites={onClickFavorites}
+        />
+        {isPodcast && isNewPodcastControlsEnabled
+          ? this.renderListenCount()
```
dylanjeffers, line 374: then one nice addition is that we can just render <Component /> here instead of renderFn

packages/web/src/components/track/PlayPauseButton.tsx
```diff
@@ -1,33 +1,62 @@
...
 
 type PlayPauseButtonProps = {
   doesUserHaveAccess: boolean
   playing: boolean
+  playbackStatus?: PlaybackStatus | null
   onPlay: () => void
 }
 
+const messages = {
+  play: 'PLAY',
```
dylanjeffers, line 17: nice. small nit which was there before, is that we could do `play` as the string and do `textTransform: uppercase`, that is technically more accessible, since screen reader would just read "P""L""A""Y"

packages/web/src/components/track/desktop/TrackTile.tsx
```diff
@@ -106,8 +113,51 @@ const TrackTile = memo(
...
+      FeatureFlags.PODCAST_CONTROL_UPDATES_ENABLED
+    )
+    const trackPositionInfo = useSelector((state: CommonState) =>
+      getTrackPosition(state, { trackId })
+    )
 
     const hasOrdering = order !== undefined
+    const isPodcast = genre === Genre.PODCASTS
+
+    const getDurationText = () => {
```
dylanjeffers, line 126: same potential idea here to add components, so these large components dont get too crazy with dependinces, hooks, and render fn size

packages/web/src/pages/track-page/components/desktop/TrackPage.tsx
```diff
@@ -150,13 +154,17 @@ const TrackPage = ({
...
+    getTrackPosition(state, { trackId: heroTrack?.track_id })
+  )
 
   const renderGiantTrackTile = () => (
     <GiantTrackTile
       loading={loading}
       playing={heroPlaying}
       trackTitle={defaults.title}
       trackId={defaults.trackId}
+      playbackPositionInfo={trackPositionInfo}
```
dylanjeffers, line 167: just to close the loop on this one, ideally we dont have to continue to add to a components props as new features get added, ya know?

dylanjeffers (approved):
such a huge lift!!!! it all looks great, just some architectural ideas to potentially lead to more isolated and easier to maintain code.

i think biggest thing to think about going forward is the saga architecture for how we listen/poll/loop to update track position.

packages/common/src/store/playback-position/types.ts
Kyle-Shanks: Oh interesting, good tip here

packages/common/src/store/reducers.ts
Kyle-Shanks: funny enough i called it playback state first and then changed it when I had to make the PlaybackStateState type lol

packages/web/src/common/store/player/sagas.ts
Kyle-Shanks: Can't remember why this was added so I'll remove the check

packages/web/src/components/track/PlayPauseButton.tsx
Kyle-Shanks: 🧠

PR #3020, opened 2023-03-10
[PAY-907] Mobile chat reactions

dharit-tan:
### Description
Implements chat reactions on mobile using a Portal that overlays a view over the entire screen in order to dim the background. Then displays the selected message as well as the reactions popup. Clicking anywhere outside of those two components closes the popup.

Notes:
- Known issue - reaction on message disappears when user presses message, this is to prevent having 2 overlapping reactions in the portal version of the message. Need to figure out a cleaner solution here.
- Feels like we should be able to `align-items` and have padding in the portal root view, but haven't gotten it to work so using `Dimensions` to set widths instead.

### Dragons

### How Has This Been Tested?

Local ios against stage.

https://user-images.githubusercontent.com/3893871/225518416-6dfbfe7c-f77b-42dc-9cf7-90ad288c9155.mp4

### Feature Flags ###


audius-infra:

audius-infra:

audius-infra:

audius-infra:

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -189,10 +198,20 @@ export const ChatScreen = () => {
   const { params } = useRoute<'Chat'>()
   const { chatId } = params
   const url = `/chat/${encodeUrlName(chatId ?? '')}`
-  const [iconOpacity, setIconOpacity] = useState(ICON_BLUR)
-  const [inputMessage, setInputMessage] = useState('')
+  const [iconOpacity, setIconOpacity] = useState<number>(ICON_BLUR)
```
dharit-tan, line 201: Open to ideas on how to simplify this

packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
```diff
@@ -64,53 +75,174 @@ const useStyles = makeStyles(({ spacing, palette, typography }) => ({
...
+            // increases as index decreases
+            x:
+              reactionPosition.x -
+              (reactions.length - 1 - index) * REACTION_SPACE_BETWEEN
+          }}
+          isAuthor={isAuthor}
+        />
+      )
+    })
+  } else {
```
piazzatron, line 155: nit: don't need `else`

packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
```diff
@@ -64,53 +75,174 @@ const useStyles = makeStyles(({ spacing, palette, typography }) => ({
...
-            {message.message}
-          </Text>
-        </View>
-        {hasTail ? (
-          <>
+    const userId = useSelector(getUserId)
+    const senderUserId = decodeHashId(message.sender_user_id)
+    const isAuthor = senderUserId === userId
+    // Offset of reactions in relation to message body parent.
+    const [reactionPosition, setReactionPosition] = useState<{
```
piazzatron, line 182: Could this be a ref?

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -189,10 +198,20 @@ export const ChatScreen = () => {
...
+  const [inputMessage, setInputMessage] = useState<string>('')
+  const [shouldShowPopup, setShouldShowPopup] = useState<boolean>(false)
+  const [reactionY, setReactionY] = useState<number>(0)
```
piazzatron, line 204: definitely supporting pulling these into their own hook, as well as converting to refs when possible. also shouldn't have to manually type these calls to useState, TS should be smart enough to infer from the starting value in most cases (unless it's null)

packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
```diff
@@ -64,53 +75,174 @@ const useStyles = makeStyles(({ spacing, palette, typography }) => ({
...
+    width: spacing(8),
+    position: 'absolute',
+    shadowColor: 'black',
+    shadowOffset: { width: 0, height: 3 },
+    shadowOpacity: 0.2,
+    shadowRadius: 5
   }
 }))
 
+const ChatReaction = ({ reaction, reactionPosition, isAuthor }) => {
```
piazzatron, line 102: Should we type this?

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -308,9 +339,80 @@ export const ChatScreen = () => {
...
+    }
+
+    // Measure position of selected message to create a copy of it on top
+    // of the dimmed background inside the portal.
+    const { top: messageY, height: messageHeight } = await new Promise<{
+      top: number
+      height: number
+    }>((resolve) => {
+      popupViewRef.measureInWindow((x, y, width, height) => {
+        resolve({ top: y, height })
```
piazzatron, line 397: nit: could avoid the outer destructuring and just call the values what we want here

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -308,9 +339,80 @@ export const ChatScreen = () => {
...
+    // Determine whether to show reactions popup above or below highlighted message.
+    const spaceAboveMessage = messageY - rootY
+    const shouldShowReactionsAbove =
+      spaceAboveMessage > REACTION_CONTAINER_HEIGHT
+    const reactionY = shouldShowReactionsAbove
+      ? messageY - REACTION_CONTAINER_HEIGHT - rootY + spacing(3.5)
+      : messageY + messageHeight - rootY + spacing(2)
+    setReactionY(reactionY)
+    setMessageY(messageY - rootY)
+    setShouldShowPopup(true)
```
piazzatron, line 410: Since many of these values change together, I would support grouping them in an object-based useStates as it makes sense. React docs do advocate for grouping values that change together in useState: https://reactjs.org/docs/hooks-faq.html#should-i-use-one-or-many-state-variables

cc @rickyrombo

piazzatron (approved):

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
dharit-tan: Some do change together like messageY and reactionY but rootY and composeY are independent from the rest and each other.. I'm liking the ref idea instead bc none of these should trigger a re-render.

packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
dharit-tan: nice

dharit-tan: Dang actually ref is giving me issues where it renders with the old values, gonna have to use state.

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -302,9 +323,75 @@ export const ChatScreen = () => {
...
+      popupChatMessage,
+      chatMessages[index - 1]
+    )
+    const senderUserId = decodeHashId(popupChatMessage.sender_user_id)
+    reactionInfo.current.isAuthor = senderUserId === userId
+    const selectedReactionValue = popupChatMessage.reactions?.find(
+      (r) => r.user_id === encodeHashId(userId)
+    )?.reaction
+    if (selectedReactionValue) {
+      setSelectedReaction(selectedReactionValue)
```
rickyrombo, line 375: This doesn't feel like it should be a state member of the `ChatScreen`? Maybe a state member of the `ReactionPopup` with an initial value set from here instead?

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -341,11 +442,18 @@ export const ChatScreen = () => {
                   keyExtractor={(message) => message.message_id}
                   renderItem={({ item, index }) => (
                     <>
-                      <ChatMessageListItem
-                        message={item}
-                        hasTail={hasTail(item, chatMessages[index - 1])}
-                        unreadCount={unreadCount}
-                      />
+                      <View>
```
rickyrombo, line 445: Just curious, why is this wrapped in a View now :eyes:

packages/mobile/src/screens/chat-screen/ReactionPopup.tsx
```diff
@@ -0,0 +1,180 @@
...
+                left: !reactionInfo.isAuthor ? spacing(6) : undefined
+              }
+            ]}
+          />
+        </Animated.View>
+        <Animated.View
+          style={[
+            styles.reactionsContainer,
+            {
+              top: reactionInfo.reactionTop,
```
rickyrombo, line 153: Can't we infer reactionTop from messageTop?

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -331,6 +418,20 @@ export const ChatScreen = () => {
       topbarRight={topBarRight}
     >
       <ScreenContent>
+        {/* Everything inside the portal displays on top of all other screen contents. */}
+        <Portal hostName='ChatReactionsPortal'>
+          {shouldShowPopup && popupChatIndex !== null ? (
+            <ReactionPopup
+              reactionInfo={reactionInfo.current}
```
rickyrombo, line 425: nit: could pass each of the ref's props in separately to avoid the generic sounding "reactionInfo" within the component

rickyrombo: In fact, the `hasTail` and `isAuthor` could be calculated here using `chatMessages` and `popupIndex`, and don't need to be part of the ref imo

rickyrombo: I think maybe the only thing that needs to be part of the ref is the `y` position of the message

packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
```diff
@@ -64,53 +76,198 @@ const useStyles = makeStyles(({ spacing, palette, typography }) => ({
...
+  )
+}
+
+const formatChatReactions = (
+  reactions: ChatMessageReaction[],
+  isAuthor: boolean,
+  reactionPosition: { x: number; y: number }
+) => {
+  // When there are multiple reactions, shift the earlier reactions closer
+  // towards the inside of of the message body.
```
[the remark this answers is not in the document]
dharit-tan: good call! unfortunately still need to offset the container by `reactions.length * offset` to shift reactions closer to screen edge when there are multiple reactions.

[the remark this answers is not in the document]
rickyrombo: like if the code was something like:
```
  reactionContainer: {
    position: absolute;
    left: -spacing(4);
  },
  reaction: {
    margin-inline-end: -spacing(7);
  }
// ...
<View style={styles.reactionContainer}>
  {reactions.map(reaction => <Reaction ... />)} 
</View>
```

similar to like it is on web?

packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
```diff
@@ -199,6 +159,21 @@ export const ChatMessageListItem = forwardRef<View, ChatMessageListItemProps>(
       y: number
     }>({ x: 0, y: 0 })
 
+    const reactionPositionStyle = [
```
rickyrombo, line 162: does this not get inferred if the bubble is marked as `position: relative`?

rickyrombo: On web what I did was basically set the `right` or `left` styles relative to the parent based on `isAuthor`, then changed text direction (in your case, could be just reversing the container list: `flexDirection: 'row-reverse'` or `flexDirection: 'row'`)

rickyrombo: Basically if we can avoid using state to get positions of the parent by inferring using position relative, that would be good, and then also if we can get rid of manually creating styles within the component and have two "classes" instead

dharit-tan: Amazing! I think I tried something similar but was having trouble with position the tail when the reaction container position was anything other than absolute. But I realized I can set the height of the container to 0 and it'll still be visible.

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
dharit-tan: Don't think it even needs to be a state variable once it's inside the popup!

packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
```diff
@@ -182,31 +161,46 @@ export const ChatMessageListItem = forwardRef<View, ChatMessageListItemProps>(
...
+                        {message.reactions.map((reaction, index) => {
+                          return (
+                            <ChatReaction key={index} reaction={reaction} />
+                          )
+                        })}
+                      </View>
+                    ) : null}
+                    {/* Add an 8px margin between messages when there are reactions. */}
+                    {!hasTail ? (
+                      <View style={styles.reactionMarginBottom} />
```
rickyrombo, line 197: I'd make this a style on the root view rather than having a proxy element

rickyrombo (approved):

packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
dharit-tan: ty for late & speedy reviews!

PR #3026, opened 2023-03-13
[C-583] Fix section/flat list generics

dylanjeffers:
### Description

Fixes our internal section and flat list components to properly handle generic typing. This uncovered two hidden bugs in the SearchResults section list and ChatMessage flatlist
- Had to move "Search" types to common, as will search bar reducers in a follow up pr.


audius-infra:
Preview this change https://demo.audius.co/dj-c-583-section-flat-types

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -17,15 +17,16 @@ import {
...
-import { View, Text, Image } from 'react-native'
 import type { FlatList as RNFlatList } from 'react-native'
+import { View, Text, Image, FlatList } from 'react-native'
 import { useDispatch, useSelector } from 'react-redux'
 
 import WavingHand from 'app/assets/images/emojis/waving-hand-sign.png'
 import IconKebabHorizontal from 'app/assets/images/iconKebabHorizontal.svg'
 import IconSend from 'app/assets/images/iconSend.svg'
-import { TextInput, Screen, FlatList, ScreenContent } from 'app/components/core'
+import { TextInput, Screen, ScreenContent } from 'app/components/core'
```
dharit-tan, line 29: Don't we actually want to be using our Flatlist and not RN's?

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -344,9 +345,9 @@ export const ChatScreen = () => {
                 <FlatList
                   contentContainerStyle={styles.flatListContainer}
                   data={chatMessages}
-                  keyExtractor={(message) => message.chat_id}
+                  keyExtractor={(message) => message.message_id}
                   renderItem={({ item, index }) => (
-                    <Fragment key={item.key}>
+                    <Fragment key={item.message_id}>
```
dharit-tan, line 350: Actually just double-checked RN docs and it seems we should use `item.key`. Think if we use `message.message_id` in `keyExtractor` it'll handle this for us. But will this mess up the typing? https://reactnative.dev/docs/flatlist#required-renderitem

dylanjeffers: oh yeah, okay so we should just remove key from fragment since this isn't a map, so

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
dylanjeffers: yeah def, i think i removed to test if our impl was messed up, will revert

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
dharit-tan: Think so!

packages/mobile/src/components/core/FlatList.tsx
```diff
@@ -15,13 +12,20 @@ import { CollapsibleTabNavigatorContext } from '../top-tab-bar'
 import { PlayBarChin } from './PlayBarChin'
 import { PullToRefresh, useOverflowHandlers } from './PullToRefresh'
 
-type FlatListProps = RNFlatListProps<any>
+declare module 'react' {
```
sliptype, line 15: Should we put this somewhere higher in the hierarchy?

packages/mobile/src/components/core/FlatList.tsx
```diff
@@ -15,13 +12,20 @@ import { CollapsibleTabNavigatorContext } from '../top-tab-bar'
...
+} // Allows forwardRef to properly type generic components
 
-type CollapsibleFlatListProps = {
+export type FlatListT<ItemT> = RNFlatList<ItemT>
+export type AnimatedFlatListT<ItemT> = Animated.FlatList<ItemT>
+
+type CollapsibleFlatListProps<ItemT> = {
   sceneName: string
-} & RNFlatListProps<any>
+} & Animated.AnimatedProps<FlatListProps<ItemT>>
```
sliptype, line 26: 🔥

sliptype (approved):
Great work dude

audius-infra:

packages/mobile/src/components/core/FlatList.tsx
dylanjeffers: When i did, it seemed to break normal react so i got scared, but can try again

sliptype: I thought declaring modules like this was global anyways, I could be wrong though. Might require the `global` keyword

dylanjeffers: updated, great call

PR #3029, opened 2023-03-14
Upgrade mobile app versions for release

dylanjeffers:
### Description

I thought this commit was pushed up with the notifications changes, my b

### DIFF
--- a/packages/mobile/ios/AudiusReactNative/Info.plist
+++ b/packages/mobile/ios/AudiusReactNative/Info.plist
@@ -17,9 +17,6 @@
 		<key>CFBundlePackageType</key>
 		<string>APPL</string>
 		<key>CFBundleShortVersionString</key>
-    <!-- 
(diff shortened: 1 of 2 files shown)


packages/mobile/ios/AudiusReactNative/Info.plist
```diff
@@ -17,7 +17,10 @@
 		<key>CFBundlePackageType</key>
 		<string>APPL</string>
 		<key>CFBundleShortVersionString</key>
-		<string>1.1.62</string>
+    <!--
```
dylanjeffers, line 20: I think this was a holdover from previous commits, do we want it still?

sliptype (approved):

packages/mobile/ios/AudiusReactNative/Info.plist
sliptype: Nah we don't need it

PR #3032, opened 2023-03-14
Fix collapsible list typo

dylanjeffers:
### Description

Fixes issue where a @ts-ignored line had a typo... :(

also fixes introduced type issue with overflow menus.


rickyrombo (approved):
Sneaky!

PR #3042, opened 2023-03-15
[PAY-1061] - Only show confetting when unlocking from how to unlock section

sddioulde:
### Description

Only show confetting when unlocking from how to unlock section on track page or modal

### Dragons

### How Has This Been Tested?

Tested on local dapp v stage

### Feature Flags ###


rickyrombo (approved):
I don't know if this is how I would have done it, but seems ok?

I would have had the UI decide to show the confetti when it sees the state change and taking the confetti out of the saga, eg using a ref + use effect. Would have saved a lot of plumbing I think

PR #3053, opened 2023-03-17
[C-2266] Disabled the PlayBar favorite button for the track owner

Kyle-Shanks:
### Description
Disabled the favorite button in the PlayBar if the user is the track's owner

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/mobile/src/components/now-playing-drawer/PlayBar.tsx
+++ b/packages/mobile/src/components/now-playing-drawer/PlayBar.tsx
@@ -4,10 +4,11 @@ import type { Nullable, Track, User } from '@audius/common'
 import {
   SquareSizes,
   FavoriteSource,
+  accountSelectors,
   tracksSocialActions
 } from '@audius/common'
 import { TouchableOpacity, Animated, View, Dimensions } from 'react-native'
-import { useDispatch } from 'react-redux'
+import { useDispatch, useSelector } from 'react-redux'
 
 import { FavoriteButton } from 'app/components/favorite-button'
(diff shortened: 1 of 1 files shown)


packages/mobile/src/components/now-playing-drawer/PlayBar.tsx
```diff
@@ -108,6 +111,7 @@ export const PlayBar = (props: PlayBarProps) => {
   const renderFavoriteButton = () => {
     return (
       <FavoriteButton
+        isDisabled={currentUser?.user_id === track?.owner_id}
```
dylanjeffers, line 114: nice add, im starting to wonder if we should add "id" + "type" to favorite buttons and just let it select and do all the logic, cause i think we have so much duplicate code for "onPress" for favorite button, and then also have to add "disabled" everywhere lol. do you think it would be worth a ticket? another addition to it would be moving from lottie to rive.

dylanjeffers (approved):
great catch

packages/mobile/src/components/now-playing-drawer/PlayBar.tsx
Kyle-Shanks: yea i agree. i've done this type of thing in other places too. and it also would make it easier if we want to update animations, add toasts, etc.

dylanjeffers: cool! ill go ahead and add a low pri backlog item

PR #3054, opened 2023-03-17
[C-2286] Update placement of the repeat single track check from play saga to next saga

Kyle-Shanks:
### Description
Move the check for repeating a single track from the play saga to the next saga to prevent restarting when we go from pause to play.
(This must have been like this for a long time 👀 )

### Dragons

This touches core player sagas on web so there is a bit of risk there

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A


dylanjeffers (approved):
great catch!

PR #3055, opened 2023-03-17
[C-429] Fix edit-profile screen scroll

dylanjeffers:
### Description

Fixes issue where users cant scroll on edit-profile screen, which is bad for users with a long description

### DIFF
--- a/packages/mobile/src/screens/edit-profile-screen/EditProfileScreen.tsx
+++ b/packages/mobile/src/screens/edit-profile-screen/EditProfileScreen.tsx
@@ -60,7 +60,7 @@ const useStyles = makeStyles(({ palette, spacing }) => ({
     width: 'auto'
   },
   textFields: {
-    paddingTop: spacing(12)
+    paddingTop: spacing(16)
   }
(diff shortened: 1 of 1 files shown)


packages/mobile/src/screens/edit-profile-screen/EditProfileScreen.tsx
```diff
@@ -59,6 +58,9 @@ const useStyles = makeStyles(({ palette }) => ({
   },
   profilePictureImage: {
     width: 'auto'
+  },
+  textFields: {
+    paddingTop: spacing(12)
```
Kyle-Shanks, line 63: padding changes from 64 to 48. making sure that this was on purpose

Kyle-Shanks (approved):

packages/mobile/src/screens/edit-profile-screen/EditProfileScreen.tsx
dylanjeffers: I can count clearly 😫

PR #3056, opened 2023-03-17
Fix heavy rotation playlist on client

sabrina-kiam:
### Description
This updates the heavy rotation playlist to hit a discovery endpoint for getting user track listen histories. Previously it was hitting an identity endpoint which was outdated when we migrated listening info from identity to discovery, so the rotation playlist page was either blank entirely for new users or contained really old listening data.

This gets a playlist of the top 20 listened to tracks, but can easily change the value to whatever seems best

Note: we've recently updated the GetUserTrackHistory endpoint on the discovery side to start collecting play counts of tracks for a user from now moving forward, and we did not backfill for previous play counts of tracks. So the playlist may reflect most recently played songs for a bit until we've been indexing play counts in listening history for long enough

### How Has This Been Tested?

Tested locally against stage. Logged in and go to explore -> heavy rotation. also confirmed the history page under `Library` on the left panel still works as it uses the same endpoint

### DIFF
--- a/packages/common/src/services/audius-api-client/AudiusAPIClient.ts
+++ b/packages/common/src/services/audius-api-client/AudiusAPIClient.ts
@@ -411,6 +411,7 @@ type GetUserTrackHistoryArgs = {
   currentUserId: Nullable<ID>
   limit?: number
   offset?: number
+  sortMethod?: string
 }
 
 type GetReactionArgs = {
@@ -1523,16 +1524,19 @@ export class AudiusAPIClient {
     currentUserId,
     userId,
     offset = 0,
-    limit = 100
+    limit = 100,
+    sortMethod = ''
   }: GetUserTrackHistoryArgs) {
     const encodedUserId = this._encodeOrThrow(userId)
     const encodedCurrentUserId = encodeHashId(currentUserId)
     this._assertInitialized()
-    const params = {
+    let params = {
       user_id: encodedCurrentUserId || undefined,
       limit,
-      offset
+      offset,
+      sort_method: sortMethod || undefined
     }
+    params = sortMethod ? { ...params, sort_method: sortMethod } : params
 
     const response: Nullable<APIResponse<APIActivity[]>> =
       await this._getResponse(
--- a/packages/common/src/services/audius-api-client/AudiusAPIClient.ts
+++ b/packages/common/src/services/audius-api-client/AudiusAPIClient.ts
@@ -1525,18 +1525,17 @@ export class AudiusAPIClient {
     userId,
     offset = 0,
     limit = 100,
-    sortMethod = ''
+    sortMethod
   }: GetUserTrackHistoryArgs) {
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://demo.audius.co/sk-fix-heavy-rot

packages/common/src/services/audius-api-client/AudiusAPIClient.ts
```diff
@@ -1523,16 +1524,19 @@ export class AudiusAPIClient {
     currentUserId,
     userId,
     offset = 0,
-    limit = 100
+    limit = 100,
+    sortMethod = ''
```
dylanjeffers, line 1528: no need for default arg, since '' will || undefined in the params arg

dylanjeffers: so can just be sort_method: sortMethod, and with that i think you can remove the bottom line too so in the end im thinking: 
```
  let params = {
      user_id: encodedCurrentUserId || undefined,
      limit,
      offset,
      sort_method: sortMethod
    }
```

packages/web/src/common/store/smart-collection/sagas.ts
```diff
@@ -34,22 +35,36 @@ const { getUserId } = accountSelectors
...
-  const explore = yield* getContext('explore')
-  const topListens = yield* call([explore, 'getTopUserListens'])
+  const currentUserId = yield* select(getUserId)
+  const apiClient = yield* getContext('apiClient')
+  const userListeningHistoryMostListenedByUser = yield* call(
+    [apiClient, apiClient.getUserTrackHistory],
+    {
+      userId: currentUserId!, // userId
+      currentUserId,
+      limit: 20, // limit
```
dylanjeffers, line 45: no need for comments here since its an object

packages/web/src/common/store/smart-collection/sagas.ts
```diff
@@ -34,22 +35,36 @@ const { getUserId } = accountSelectors
...
     .filter(
       (track) =>
-        users.entries[track.userId] &&
-        !users.entries[track.userId].is_deactivated
+        users.entries[track.owner_id] &&
+        !users.entries[track.owner_id].is_deactivated
     )
-    .map((listen) => ({
-      track: listen.trackId
+    .map((track) => ({
```
dylanjeffers, line 66: do we need to filter tracks by deleted/unavailable etc?

dylanjeffers (approved):
nice additions, looking forward to using this. is the api and everything ready in prod too?

packages/web/src/common/store/smart-collection/sagas.ts
sabrina-kiam: hmm it didn't filter them out before but thats probably a good call, will add

audius-infra:
Preview this change https://demo.audius.co/sk-fix-heavy-rot

PR #3058, opened 2023-03-17
[C-2263 C-2283] Add setCurrentUser profile action and call when we focus a profile screen

Kyle-Shanks:
### Description
* Add setCurrentUser action for profile slice
* Call setCurrentUser action when we focus the profile screen to make sure that the currentUser is always up to date

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A


dylanjeffers (approved):
lets goo

PR #3060, opened 2023-03-17
[C-2329] Fix notification navigation

dylanjeffers:
### Description

Updates notification navigation to check payload, payload.data and payload.data.data like we had originally. Once we start receiving notifs, ill be able to confirm which one it actually is too.

### DIFF
--- a/packages/mobile/src/notifications.ts
+++ b/packages/mobile/src/notifications.ts
@@ -48,7 +48,7 @@ class PushNotifications {
         body
       })
     )
-    this.navigation?.navigate(payload)
+    this.navigation?.navigate(payload?.data?.data ?? payload?.data ?? payload)
   }
 
   // Method used to open the push notification that the user pressed while the app was closed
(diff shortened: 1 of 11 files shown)


packages/mobile/src/notifications.ts
```diff
@@ -48,7 +48,7 @@ class PushNotifications {
         body
       })
     )
-    this.navigation?.navigate(payload)
+    this.navigation?.navigate(payload?.data?.data ?? payload?.data ?? payload)
```
raymondjacobson, line 51: is data.data because of axios?

raymondjacobson (approved):

packages/mobile/src/notifications.ts
dylanjeffers: honestly not sure, this was a check from previous impl, so it might just be wrong. im adding both just to make sure it actually works before then figuring out which one works :)

dylanjeffers: its really hard to test this locally unfortunately

audius-infra:
Preview this change https://demo.audius.co/dj-c-2329-fix-notification-deeplinks

PR #3075, opened 2023-03-21
[C-2344] Update the web playbar scrubber to respect the playback speed of podcasts

Kyle-Shanks:
### Description
Update the scrubber in stems to accept and use playbackRate
Update the web playbar to pass in the playbackRate when playing podcasts

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/stems/src/components/Scrubber/Slider.tsx
+++ b/packages/stems/src/components/Scrubber/Slider.tsx
@@ -24,6 +24,7 @@ export const Slider = ({
   isDisabled,
   elapsedSeconds,
   totalSeconds,
+  playbackRate,
   onScrub,
   onScrubRelease,
   includeExpandedTargets = true,
@@ -54,7 +55,8 @@ export const Slider = ({
     trackRef,
     handleRef,
     elapsedSeconds,
-    totalSeconds
+    totalSeconds,
+    playbackRate
   )
 
   /**
@@ -218,6 +220,11 @@ export const Slider = ({
       else pause()
     }
   }, [isPlaying, dragPercent, play, pause])
+
+  useEffect(() => {
+    setPercent(elapsedSeconds / totalSeconds)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [playbackRate])
 
   // When the key changes, reset the animation
   useEffect(() => {
--- a/packages/web/src/components/play-bar/desktop/PlayBar.js
+++ b/packages/web/src/components/play-bar/desktop/PlayBar.js
@@ -15,7 +15,8 @@ import {
   playerActions,
   playerSelectors,
   queueSelectors,
-  FeatureFlags
+  FeatureFlags,
+  playbackRateValueMap
 } from '@audius/common'
 import { Scrubber } from '@audius/stems'
 import { push as pushRoute } from 'connected-react-router'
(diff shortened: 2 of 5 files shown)


packages/stems/src/components/Scrubber/Slider.tsx
```diff
@@ -219,6 +221,15 @@ export const Slider = ({
     }
   }, [isPlaying, dragPercent, play, pause])
 
+  const playbackRateRef = useRef<number | null>(null)
```
dylanjeffers, line 224: I wonder if we should avoid refs as much as we can. What about usePrevious?

packages/stems/src/components/Scrubber/Slider.tsx
```diff
@@ -219,6 +221,15 @@ export const Slider = ({
...
+
+  useEffect(() => {
+    if (playbackRate !== playbackRateRef.current) {
```
dylanjeffers, line 227: Another option is to update the dependencies of this effect to only change when playback rate changes, simulating useEventEffect that's coming out soon?

packages/web/src/components/play-bar/desktop/PlayBar.js
```diff
@@ -393,6 +396,9 @@ class PlayBar extends Component {
                 isPlaying={isPlaying && !isBuffering}
                 isDisabled={!uid && !collectible}
                 includeTimestamps
+                playbackRate={
+                  isLongFormContent ? playbackRateValueMap[playbackRate] : 1
```
dylanjeffers, line 400: Just wondering if we should always have playback rate set correctly in state so we don't have to make checks on what type of content we are dealing with?

dylanjeffers (approved):
Nice adds. Curious your thoughts on the effect and the check for long form content

packages/stems/src/components/Scrubber/Slider.tsx
```diff
@@ -219,6 +221,15 @@ export const Slider = ({
...
+      setPercent(elapsedSeconds / totalSeconds)
+      playbackRateRef.current = playbackRate
+    }
+  }, [elapsedSeconds, totalSeconds, playbackRate, setPercent, play])
```
sliptype, line 231: This is where useEventEffect would be helpful!

packages/web/src/components/play-bar/desktop/PlayBar.js
dylanjeffers: Additionally do we need to use the value map all over the place or make that part of a selector?

sliptype (approved):
Looks good to me!

packages/web/src/components/play-bar/desktop/PlayBar.js
Kyle-Shanks: yea it gets a bit annoying so maybe it could be a separate selector, but i like having a selector for the string for rendering out svgs and other UI things

packages/stems/src/components/Scrubber/Slider.tsx
Kyle-Shanks: yea I run into this type of thing a lot, lots of ignore-exhaustive-deps. I'll go with this option though.

packages/web/src/components/play-bar/desktop/PlayBar.js
dylanjeffers: Can possibly have two selectors or would that be just as confusing?

PR #3094, opened 2023-03-23
[C-2248, C-2373] Use playlistUpdates, remove legacyNotifications

dylanjeffers:
### Description

- Connects new playlistUpdates reducer
- Removes legacyNotifications and adds notificationsUiSlice


packages/web/src/common/store/social/collections/sagas.ts
```diff
@@ -354,9 +355,7 @@ export function* saveCollectionAsync(
   )
 
   if (!collection.is_album) {
-    yield* put(
-      notificationActions.updatePlaylistLastViewedAt(action.collectionId)
-    )
+    yield* put(updatedPlaylistViewed({ playlistId: action.collectionId }))
```
raymondjacobson, line 358: nice

packages/web/src/components/notification/Notification/components/EntityLink.tsx
```diff
@@ -35,17 +24,14 @@ export const useGoToEntity = (
...
   const handleClick: MouseEventHandler = useCallback(
     (event) => {
       if (!entity) return
       event.stopPropagation()
       event.preventDefault()
       const link = getEntityLink(entity)
-      if (isNotificationPanelOpen) {
-        dispatch(toggleNotificationPanel())
-      }
+      dispatch(closeNotificationPanel())
```
raymondjacobson, line 34: do we know why this was guarded before?

raymondjacobson (approved):
lgtm! the small QOL changes are so _sick_

packages/web/src/components/notification/Notification/components/EntityLink.tsx
dylanjeffers: yes cause toggle could be called multiple times, kinda same issue we were seeing with saliou's flag drama.

packages/web/src/store/application/ui/notifications/notificationsUISlice.ts
```diff
@@ -0,0 +1,48 @@
+import { Nullable } from '@audius/common'
```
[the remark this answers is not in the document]
dylanjeffers: yeah the idea here is that instead of everything being "slice.ts", its the name of the slice, i see that's what's preferred in redux docs, and kinda something weve discussed with "saga.ts" being hard to work with as well. agreed it feels redundant given the directory naming though. im down to do whatever we think

PR #3095, opened 2023-03-24
[C-2275] Fix android drawer offsets

dylanjeffers:
### Description

Fixes issues with android drawer positioning. Simplifies approach by relying on safe-area-context insets for nav-bar height, removing a dependency and extra complication. Now ios and android have the exact same implementation.


audius-infra:

sliptype (approved):
Code looks good but I am a little worried about this one tbh. This is the pr that added the navbar-height dep https://github.com/AudiusProject/audius-client/pull/1277

Did you test the nowplaying bar on android with the os navigation bar enabled and the os navigation bar disabled (gestures only)?

dylanjeffers:
> Code looks good but I am a little worried about this one tbh. This is the pr that added the navbar-height dep #1277
> 
> Did you test the nowplaying bar on android with the os navigation bar enabled and the os navigation bar disabled (gestures only)?

Totally understood, maybe it would be best for us to review together to make sure? I tested both navigation styles, (bar and gestures), and also ios. After realizing that "safe-view" gives height of the nav-bar, that helped a lot cause we could just use that instead of the other library. I wasn't able to post videos here to confirm because for some reason the screen cast removes the nav-bar in final product.

PR #3101, opened 2023-03-24
[C-2365] Update play buttons on web and mobile to show resume when track is current

Kyle-Shanks:
### Description
Update the play buttons on web and mobile to display resume the track is completed, but is still the current track

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/mobile/src/components/details-tile/DetailsTile.tsx
+++ b/packages/mobile/src/components/details-tile/DetailsTile.tsx
@@ -1,12 +1,13 @@
 import { useCallback } from 'react'
 
-import type { Track } from '@audius/common'
+import type { CommonState, Track } from '@audius/common'
 import {
   FeatureFlags,
   Genre,
   usePremiumContentAccess,
   squashNewLines,
   accountSelectors,
+  playerSelectors,
   playbackPositionSelectors
 } from '@audius/common'
 import { TouchableOpacity, View } from 'react-native'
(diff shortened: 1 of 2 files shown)


packages/mobile/src/components/details-tile/DetailsTile.tsx
```diff
@@ -220,6 +222,8 @@ export const DetailsTile = ({
   const navigation = useNavigation()
 
   const currentUserId = useSelector(accountSelectors.getUserId)
+  const currentPlayerTrackId = useSelector(getTrackId)
+  const isCurrentTrack = track && track.track_id === currentPlayerTrackId
```
sliptype, line 226: Can you put this equality check into the selector, that will prevent extra rerenders of the details tile when the playing track changes:

```suggestion
  const isCurrentTrack = useSelector((state) => {
    return track && track.track_id === getTrackId(state)
 }
```

sliptype (approved):
Makes sense!

packages/mobile/src/components/details-tile/DetailsTile.tsx
sliptype: Also same deal on web

Kyle-Shanks: 🧠

PR #3104, opened 2023-03-27
[C-2382] Update dapp-store cli, short_description length

dylanjeffers:
### Description

There was an issue with dapp-store cli not validating short_description length which caused our release to pass with an invalid short_description length. Updating the cli, and updating short_description fixes the issue.

- Also added a step in readme to run validate to ensure our config is correct before trying to publish.


sliptype (approved):

audius-infra:

PR #3108, opened 2023-03-27
[PAY-1090] - Subscribe to special access gates to unlock new tracks

sddioulde:
### Description

There is an edge case where an artist can be followed or tipped before their tracks are loaded, and if their tracks happen to be loaded in the small window between the follow/tip and its confirmation, the UI will show a locked state instead of an unlocking/unlocked state for those tracks.
To handle this scenario, this PR stores followee and tipped user ids so that those the UI knows to be in an unlocking state as those users' special access tracks load.

### Dragons

### How Has This Been Tested?

local web app vs stage

### Feature Flags ###

### DIFF
--- a/packages/common/src/store/premium-content/sagas.ts
+++ b/packages/common/src/store/premium-content/sagas.ts
@@ -270,6 +270,57 @@ function* handleSpecialAccessTrackSubscriptions(tracks: Track[]) {
   )
 }
 
+// Request premium content signatures for the relevant nft-gated tracks
+// which the client believes the user should have access to.
+function* updateCollectibleGatedTracks(trackMap: { [id: ID]: string[] }) {
+  const account = yield* select(getAccountUser)
+  if (!account) return
+
+  const apiClient = yield* getContext('apiClient')
+
+  const premiumContentSignatureResponse = yield* call(
+    [apiClient, apiClient.getPremiumContentSignatures],
+    {
+      userId: account.user_id,
+      trackMap
+    }
+  )
+
+  if (premiumContentSignatureResponse) {
+    // Keep record of number of tracks that have a signature
+    // so that we can later track their metrics.
+    let numTrackIdsWithSignature = 0
+
+    const premiumContentSignatureMap: {
+      [id: ID]: Nullable<PremiumContentSignature>
+    } = { ...premiumContentSignatureResponse }
+    // Set null for tracks for which signatures did not get returned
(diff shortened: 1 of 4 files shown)


dharit-tan (approved):

packages/common/src/store/premium-content/sagas.ts
```diff
@@ -191,15 +206,132 @@ function* updateNewPremiumContentSignatures({
...
+  const remoteConfigInstance = yield* getContext('remoteConfigInstance')
+  yield* call(remoteConfigInstance.waitForRemoteConfig)
+  const gatedTrackPollIntervalMs = remoteConfigInstance.getRemoteVar(
+    IntKeys.GATED_TRACK_POLL_INTERVAL_MS
+  )
+
+  yield* all(
+    tracksThatNeedSignature.map((track) => {
+      const trackId = track.track_id
+      return call(pollPremiumTrack, {
```
rickyrombo, line 261: Users might have a lot of gated tracks so this might lead to a lot of requests. I fear we might hit the 6 requests to the same domain limit once profiles have a ton of gated tracks. It feels like we could be smart somehow with the confirmer on follow/tip and fetch _once_ for signatures per track? Or is the signature not guaranteed to be there after indexing?

EDIT: I think my idea might miss the point of this PR though. Would this work?
1. User loads a profile, immediately clicks follow
2. Tracks start loading in from that first request, without signatures since the follow isn't indexed
3. Follow gets indexed, confirmer dispatches an action, triggering a refetch for tracks with follower premium conditions _once_ (this could be the same for the normal flow of unlocking too)

It just _feels_ like there shouldn't be a need to poll separately here but I don't know enough to know for sure

sddioulde: good point. i think your confirmer logic here would work for follow-gated tracks but not tip-gated tracks as tipping does not use the confirmer to determine that a tip has been sent.

probably still a good idea for follow-gated tracks to use the confirmer i.e. instead of adding the followeeId right after clicking follow, wait until confirmer confirms the follow. but that means the UI wouldn't show 'UNLOCKING' in this case until confirmer confirms the follow... which is not exactly the UI spec.

also, this whole PR is to handle an edge case. there is only a relatively small window of time when this can happen. maybe we should just ignore the 6 requests limit scenario when users have very many follow-gated tracks, and say that we're okay with some of those requests getting blocked for a second or so? thoughts?

PR #3117, opened 2023-03-28
[C-2387] Improve permalink and user handle caching

dylanjeffers:
### Description

Improves how we save permalinks and handles, by adding it as part of the "ADD_SUCCESSFUL" cache action, rather than watching that action and posting another action. This should reduce the number of redux actions by a lot, and also speed up our store's state.
- Updates our `asCache` function to allow "postprocessing" on an action by the Entity specific reducers.


rickyrombo (approved):
This looks amazing! We should be extra careful since the cache layer is so integral to everything, but honestly it seems pretty straightforward, I don't se how it could go wrong.

PR #3121, opened 2023-03-28
Fix dapp store versionCode

dylanjeffers:
We needed to redeploy dapp-store audius version with a manually updated versionCode, (this is normally done in ci for our android build)


audius-infra:

sliptype (approved):

PR #3127, opened 2023-03-29
Allow Optimizely targeting w CodePush on mobile C-2394

nicoback2:
### Description
Separated all the info into different attributes so we can define more specific target audiences
(note - couldn't combine mobile app version and CodePush version into one version string in a way that would work the way we want, so they are two separate attributes)

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/common/src/services/remote-config/remote-config.ts
+++ b/packages/common/src/services/remote-config/remote-config.ts
@@ -37,6 +37,12 @@ type State = {
   id: Nullable<number>
 }
 
+type MobileClientInfo = {
+  mobilePlatform: 'android' | 'ios'
+  mobileAppVersion: string
+  codePushUpdateNumber: number | undefined
+}
+
 export type RemoteConfigOptions<Client> = {
   createOptimizelyClient: () => Promise<Client>
   getFeatureFlagSessionId: () => Promise<Nullable<number>>
@@ -44,7 +50,8 @@ export type RemoteConfigOptions<Client> = {
   setLogLevel: () => void
   environment: Environment
   appVersion: string
-  getMobileClientVersion?: () => Promise<string> | string
+  platform: 'web' | 'mobile'
+  getMobileClientInfo?: () => Promise<MobileClientInfo> | MobileClientInfo
 }
 
 export const remoteConfig = <
@@ -64,7 +71,8 @@ export const remoteConfig = <
   setLogLevel,
   environment,
   appVersion,
-  getMobileClientVersion
+  platform,
+  getMobileClientInfo
 }: RemoteConfigOptions<Client>) => {
   const state: State = {
     didInitialize: false,
@@ -76,8 +84,9 @@ export const remoteConfig = <
 
   // Optimizely client
   let client: Client | undefined
-  /** Mobile app version that is being run */
-  let mobileClientVersion: string | undefined
+
+  /** Mobile app info if platform is mobile */
+  let mobileClientInfo: MobileClientInfo | undefined
 
   const emitter = new EventEmitter()
   emitter.setMaxListeners(1000)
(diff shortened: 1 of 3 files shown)


packages/mobile/src/services/remote-config/remote-config-instance.ts
```diff
@@ -19,31 +19,43 @@ const { version: appVersion } = packageInfo
...
-/** Returns version in the form of `a1.1.33+codepush:v5` if Android, `i1.1.33+codepush:v5` if iOS. The `+codepush:*` portion is omitted if there is no CodePush update installed. */
-const getMobileClientVersion = async () => {
-  const baseVersion = `${Platform.OS === 'android' ? 'a' : 'i'}${
-    VersionNumber.appVersion
-  }`
-  let res = baseVersion // This is our default value if getting the CodePush update metadata fails for some reason, or if there is no CodePush update installed.
-  let codePushUpdateMetadata: LocalPackage | null
+/** Returns mobile platform (ios or android), mobile app version, and code push update number (if any) */
+const getMobileClientInfo = async () => {
+  const mobilePlatform = (Platform.OS === 'android' ? 'android' : 'ios') as
```
dylanjeffers, line 24: can do "as const" i think here?

packages/common/src/services/remote-config/remote-config.ts
```diff
@@ -37,14 +37,21 @@ type State = {
   id: Nullable<number>
 }
 
+type MobileClientInfo = {
+  mobilePlatform: 'android' | 'ios'
```
sliptype, line 41: Yessssss

packages/mobile/src/services/remote-config/remote-config-instance.ts
dylanjeffers: just wondering if it could maybe even just be set to "Platform.OS"?

packages/mobile/src/services/remote-config/remote-config-instance.ts
```diff
@@ -19,31 +19,43 @@ const { version: appVersion } = packageInfo
...
-    res = `${baseVersion}+codepush:${codePushUpdateMetadata.label}`
+
+  let codePushUpdateNumber: number | undefined
+  if (
+    codePushUpdateMetadata &&
+    codePushUpdateMetadata.label &&
+    codePushUpdateMetadata.label.length > 1
+  ) {
+    // Codepush version nunbers are formatted as e.g."v10" - remove the leading "v".
+    codePushUpdateNumber = Number(codePushUpdateMetadata.label.slice(1))
```
sliptype, line 46: Nice

packages/web/src/services/remote-config/remote-config-instance.ts
```diff
@@ -18,6 +18,7 @@ export const FEATURE_FLAG_LOCAL_STORAGE_SESSION_KEY = 'featureFlagSessionId-2'
 
 export const remoteConfigInstance = remoteConfig({
   appVersion,
+  platform: 'web',
```
sliptype, line 21: One extra thing we could add is differentiating between web and desktop here. Doesn't necessarily have to be in this pr, but that would help when we want to set a min version

sliptype (approved):
This looks great! I thinking splitting them up makes a lot of sense

packages/mobile/src/services/remote-config/remote-config-instance.ts
nicoback2: yeah i think that'd be better, i orginally thought this would make it clear that we only expect android or ios

dylanjeffers: yeah Os is android, i think the "saga" part is like Platform.constants.Model

PR #3134, opened 2023-03-30
Optimizely attributes - use null if empty instead of undefined

nicoback2:
### Description
https://docs.developers.optimizely.com/experimentation/v3.1.0-full-stack/docs/pass-in-audience-attributes-swift
Looks like undefined isn't a valid value for an attribute, but null is fine

Issue: https://audius-internal.slack.com/archives/C0322K9M048/p1680195739448289

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/common/src/services/remote-config/remote-config.ts
+++ b/packages/common/src/services/remote-config/remote-config.ts
@@ -37,6 +37,13 @@ type State = {
   id: Nullable<number>
 }
 
+type MobileClientInfo = {
+  /** This is the type of Platform.OS, but we only expect ios or android here */
+  mobilePlatform: 'ios' | 'android' | 'web' | 'windows' | 'macos'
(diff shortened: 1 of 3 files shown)


packages/common/src/services/remote-config/remote-config.ts
```diff
@@ -234,7 +244,10 @@ export const remoteConfig = <
       return client.isFeatureEnabled(f, id.toString(), {
         userId: id,
         appVersion,
-        mobileClientVersion: mobileClientVersion ?? 'N/A'
+        platform,
+        mobilePlatform: mobileClientInfo?.mobilePlatform ?? null,
```
nicoback2, line 248: These three lines have changes, rest is the same as before revert

sliptype (approved):

PR #3136, opened 2023-03-30
[PAY-1107][PAY-1112] Center mobile chat text input, enlarge icon

dharit-tan:
### Description
Centers the text in the textinput. This was a bit tricky because using `multiline` in the textinput disallows centering. This feels a bit janky but unfortunately it's the only way I got it to work!

@sliptype including you because you helped me find the first fix, which was to use `lineHeight`, but we actually need to use lineHeight for its intended purpose (line spacing) unfortunately.

### Dragons

on iOS the cursor is larger than the text.. punting for now.

### How Has This Been Tested?

Local ios/android on stage
Screenshot 2023-03-30 at 7 03 19 PM
Simulator Screen Shot - iPhone 14 Pro - 2023-03-30 at 19 03 08

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/screens/chat-screen/ChatScreen.tsx
+++ b/packages/mobile/src/screens/chat-screen/ChatScreen.tsx
@@ -76,15 +76,16 @@ const useStyles = makeStyles(({ spacing, palette, typography }) => ({
     alignItems: 'center',
     backgroundColor: palette.neutralLight10,
     paddingLeft: spacing(4),
-    paddingRight: spacing(3),
     borderRadius: spacing(1)
   },
   composeTextInput: {
-    fontSize: typography.fontSize.medium
+    fontSize: typography.fontSize.medium,
+    lineHeight: spacing(6),
+    paddingTop: 0
   },
   icon: {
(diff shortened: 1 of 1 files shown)


packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -416,23 +417,25 @@ export const ChatScreen = () => {
...
+                    fill={styles.icon.fill}
                     onPress={() => handleSubmit(inputMessage)}
                   />
                 )}
                 styles={{
                   root: styles.composeTextContainer,
-                  input: styles.composeTextInput
+                  input: [
+                    styles.composeTextInput,
+                    Platform.OS === 'ios' ? { paddingBottom: spacing(1) } : null
```
rickyrombo, line 431: that's super interesting 👀 

for reference, on Web, we "fake" the input and have it transparent and set the lineheight exactly for one line of text, and then have the containing element have padding. Not suggesting that would work here, but calling out in case it's useful

rickyrombo (approved):

PR #3137, opened 2023-03-30
[C-2295] Only show now-playing-artwork-tile when song enqueued

dylanjeffers:
### Description

- Improves default ui state by removing the empty now-playing artwork tile until a track is selected.
- Adds fade and slide in animation.
- Cleans up prop drilling and component isolation
- Simplified `NowPlayingArtworkTile` component hierarchy and styles
- Used stems/react-router components and added proper a11y

https://user-images.githubusercontent.com/8230000/228989710-db9cb719-95f0-443a-ae72-cf1d5d49e463.mov


sliptype (approved):
Great changes! In the video the animation looks a little choppy, is there something rendering preventing it from being smooth?

Kyle-Shanks (approved):
wow, this is much cleaner. nice work here

dylanjeffers:
> Great changes! In the video the animation looks a little choppy, is there something rendering preventing it from being smooth?

yeah agreed :/ my guess is that it's a non-production version so lets see what happens, but im guessing it might have to do with playback and actions going as a track is starting up

PR #3138, opened 2023-03-31
Fix track upload cover art name

theoilie:
### Description
Fixes cover art image uploads to storage v2 having the name "blob." This was caused by client-size resizing (why do we resize on the client when we also do it on the backend?)

### Dragons
Gated by `storage_v2` feature flag so unlikely to be much risk

### How Has This Been Tested?
Ran local client with code from https://github.com/AudiusProject/audius-protocol/pull/5032 and confirmed that files had the right name both in metadata and on the backend. UI here shows the first image was named "blob" and then after this change it had the correct name ("beignet.png")
Screen Shot 2023-03-30 at 5 59 40 PM

Ensure there's no weirdness around current (non storage v2) image uploads on staging after merging

### Feature Flags ###
`storage_v2` - existing feature flag


packages/common/src/services/audius-backend/AudiusBackend.ts
```diff
@@ -1213,7 +1213,16 @@ export const audiusBackend = ({
   ) {
     const storageV2Enabled = await getFeatureEnabled(FeatureFlags.STORAGE_V2)
     if (storageV2Enabled) {
-      // do storage v2 upload
+      try {
```
theoilie, line 1216: I added this try-catch here because I removed it from the libs uploadTrackV2 function. I think it was a code smell to have the uploadTrackV2 function wrap its whole body in try-catch

sliptype (approved):

PR #3141, opened 2023-03-31
[C-2391 PLAT-833] Handle get-discovery-notification error, notification processing

dylanjeffers:
### Description

Fixes a few error cases related to new discovery notifications.
- Handles case where api returns null
- Handles case where `mapNotifications` wasn't truly transforming `DiscoveryNotification` to `Notification`
- Fixes processNotification function to include id data from new notifs.


packages/web/src/common/store/notifications/parseAndProcessNotifications.ts
```diff
@@ -85,58 +85,65 @@ export function* parseAndProcessNotifications(
...
       )
     }
-    if (notification.type === NotificationType.RemixCreate) {
+    if (type === NotificationType.RemixCreate) {
       trackIdsToFetch.push(
         notification.parentTrackId,
         notification.childTrackId
       )
       notification.entityType = Entity.Track
-      notification.entityIds = [
```
sabrina-kiam, line 104: just confirming this is no longer needed?

sabrina-kiam (approved):
thanks for cleaning up as well as fixing! so should this fix clicking share on tastemaker and it disappearing as well as the feed issue?

packages/web/src/common/store/notifications/parseAndProcessNotifications.ts
dylanjeffers: yes! it was a mistyped notif

dylanjeffers:
> thanks for cleaning up as well as fixing! so should this fix clicking share on tastemaker and it disappearing as well as the feed issue?

Yes! since the issue was due to the user not being there basically

PR #3143, opened 2023-03-31
[PLAT-834]: Fix notifications last seen new tag

sabrina-kiam:
### Description
Fixes the isViewed attribute on a notification to be whether a notification has a `seen_at` value. If a notification has seen_at timestamp, isViewed is true because it's been viewed.

Also marks notifications as viewed once we close the notifs menu as opposed to once we open the notif menu, to allow users to see the tag when they first open a new notif

### DIFF
--- a/packages/web/src/components/nav/desktop/NavColumn.tsx
+++ b/packages/web/src/components/nav/desktop/NavColumn.tsx
@@ -166,9 +166,9 @@ const NavColumn = ({
       record(make(Name.NOTIFICATIONS_OPEN, { source: 'button' }))
     } else {
       closeNotificationPanel()
-    }
-    if (notificationCount > 0) {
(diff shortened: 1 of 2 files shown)


packages/web/src/components/nav/desktop/NavColumn.tsx
```diff
@@ -166,9 +166,9 @@ const NavColumn = ({
       record(make(Name.NOTIFICATIONS_OPEN, { source: 'button' }))
     } else {
       closeNotificationPanel()
-    }
-    if (notificationCount > 0) {
-      markAllNotificationsAsViewed()
+      if (notificationCount > 0) {
```
dylanjeffers, line 169: nice fix

dylanjeffers (approved):
awesomeeee

PR #3145, opened 2023-04-03
[C-2398] Clean up playlist-update feature flags

dylanjeffers:
### Description

Cleans up playlist-update feature flags now that this is live in prod

### Dragons

Do we still need to save to identity?


raymondjacobson (approved):
probably not! where is that done?

dylanjeffers:
> probably not! where is that done?

its done in the function that removes the feature flag, it hits identity, then hits libs

PR #3147, opened 2023-04-03
[C-2227] Clear wallet state on sign-out

dylanjeffers:
### Description

Fixes issue where signing out and logging in with new user gives the wallet address as previous user. This is dangerous and also has a lot of negative effects like showing incorrect balance and connected wallets


packages/common/src/store/pages/token-dashboard/slice.ts
```diff
@@ -293,6 +293,12 @@ const slice = createSlice({
     },
     resetRemovedStatus: (state) => {
       state.associatedWallets.removeWallet.status = null
+    },
+    resetState: (state) => {
+      state.associatedWallets = initialState.associatedWallets
```
sliptype, line 298: Is there are reason for only resetting parts of this slice? Why not set back to the initialState

sliptype (approved):

packages/common/src/store/pages/token-dashboard/slice.ts
dylanjeffers: for some weird reason doing `state = initialState` doesn't seem to work...

sliptype: I've seen this weirdness with immer before, can you try returning initial state from this reducer?

dylanjeffers: great idea, will give that a go

PR #3148, opened 2023-04-03
[C-2406] Prevent non finite set of currentTime

sliptype:
### Description

Fixes an app ending error when seeking a track: https://audius.sentry.io/discover/audius-client:150a6610d90e49369a416a7402b9f7d2/?field=title&field=event.type&field=project&field=user.display&field=timestamp&homepage=true&id=17670&query=%22Caught+saga+error%3A%22&sort=-timestamp&statsPeriod=7d&topEvents=5&widths=1236&yAxis=count%28%29

I was unable to repro, but this is a new bug maybe related to the podcast feature. I couldn't figure out why the time was potentially undefined, but decided to prevent this at the level where HTMLAudioElement.currentTime is being set. This should prevent this type of bug in the future too

### Dragons

### How Has This Been Tested?

Tested seeking locally against prod

### Feature Flags ###


dylanjeffers (approved):
wow very interesting catch, nice!

PR #3151, opened 2023-04-03
[PLAT-819]: Fix copy for repost and save of albums notifications

sabrina-kiam:
### Description
This updates the copy for repost and save notifications. It was printing `X reposted your playlist my_album` for albums which is wrong. this is because in the backend reposts are saved as either playlist or track so factoring in is_album when we generate the copy

### Dragons

### How Has This Been Tested?

Tested locally by re enacting a repost notif for an album 

### Feature Flags ###


dylanjeffers (approved):
excellent find!

dylanjeffers:
@sabrina-kiam we dont need to worry about this with repostofrepost?

sabrina-kiam:
oops yes added those!

PR #3152, opened 2023-04-03
[C-2348] Change offline download location

amendelsohn:
### Description

Change offline download location from Audius Documents directory to the Cache directory. This hides the files from the iOS file browser.

Also adds a best-effort sync process that will copy files from the old to new locations before loading the offline data from disk. On failure, the sync will abort, delete the old data, and the sync process will resume downloads in the new location.

### Dragons

Putting the wrong start/end directories can wipe users's existing downloads on first load.

### How Has This Been Tested?

iOS simulator & android simulator
1. Download tracks on old build
2. Switch to new build
3. Check for migration debug logs
4. Assert tracks are no longer visible in file browser and are visible in app offline


packages/mobile/src/store/offline-downloads/sagas/migrateOfflineDataPathSaga.ts
```diff
@@ -0,0 +1,63 @@
...
+export function* migrateOfflineDataPathSaga() {
+  const isOfflineModeEnabled = yield* call(getIsOfflineEnabled)
+  if (!isOfflineModeEnabled) return
+
+  const legacyFilesExist = yield* call(exists, legacyDownloadsRoot)
+  if (!legacyFilesExist) return // TODO: fire action?
+
+  try {
+    yield* call(copyRecursive, legacyDownloadsRoot, downloadsRoot)
+  } finally {
```
dylanjeffers, line 29: awesome, so what do we do with the redux-persist data? if we fail, maybe we want to update status for all items?

dylanjeffers (changes requested):
this looks great, just a q regarding redux-persist data

packages/mobile/src/store/offline-downloads/sagas/migrateOfflineDataPathSaga.ts
amendelsohn: Yeah I was thinking about that too. Probably a good idea to put them all back in queue again?

dylanjeffers: yeah that makes sense... do we even want to remove, or just keep those around indefinitely? just note that we tried and those will remain indefinitely? basically trying to prevent user impact. but if we think failure is unlikely, then maybe we dont need to worry about the whole thing? just main thing is to make sure the states stay in-sync, whether that's full deleted, or migrated?

amendelsohn: Two rules I'd like to keep:
1. I think it's best if we remove all trace of downloaded files visible in the browser.
2. I'd like to keep it such that nobody has files in both places unless the transfer is currently in progress.

With that in mind, when we fail, I want to treat the destination path as the source of truth and update state to match. In theory, since you should only need the migration on first load, the destination should be empty, so if we fail, we could:
1. Wipe both dirs and throw everything marked for download back into the queue
2. Hope that some of it got there and make sure sync is up to the task of figuring out what files are missing. Notably, I don't think it will do that correctly at the moment for tracks until the staleness window has passed.

I think the cleanest approach will be to wipe everything. This is a fail case anyway, so I don't necessarily mind incurring the additional cost.

dylanjeffers: awesome that makes a ton of sense, lets go with wiping everything in a failure case

packages/mobile/src/store/offline-downloads/sagas/migrateOfflineDataPathSaga.ts
```diff
@@ -0,0 +1,95 @@
...
+// dirs.DocumentDir -> dirs.CacheDir
+const legacyDownloadsRoot = path.join(dirs.DocumentDir, 'downloads')
+
+// Move downloads from legacy storage location to the updated path (currently: /Documents -> /Caches)
+export function* migrateOfflineDataPathSaga() {
+  const isOfflineModeEnabled = yield* call(getIsOfflineEnabled)
+  if (!isOfflineModeEnabled) return
+
+  const legacyFilesExist = yield* call(exists, legacyDownloadsRoot)
+  if (!legacyFilesExist) return // TODO: fire action?
```
dylanjeffers, line 34: just wondering if this should be resolved?

packages/mobile/src/store/offline-downloads/sagas/migrateOfflineDataPathSaga.ts
```diff
@@ -0,0 +1,95 @@
...
+  } catch (e) {
+    // If we fail, nuke the legacy directory to ensure we don't retry the process on every startup
+    // also requeue everything for download
+    yield* call(migrationRecovery)
+  } finally {
+    yield* call(unlink, legacyDownloadsRoot)
+  }
+}
+
+// Util to recursively copy a directory since neither RNFS or rn-fetch-blob come with one
```
dylanjeffers, line 47: epick

packages/mobile/src/store/offline-downloads/sagas/migrateOfflineDataPathSaga.ts
```diff
@@ -0,0 +1,95 @@
...
+      id: parseInt(collectionId),
+      type: 'collection'
+    }))
+  const trackJobs: OfflineJob[] = Object.keys(tracks).map((trackId) => ({
+    id: parseInt(trackId),
+    type: 'track'
+  }))
+
+  const newJobs = difference([...collectionJobs, ...trackJobs], queueJobs)
+  yield* put(redownloadOfflineItems({ items: newJobs }))
```
dylanjeffers, line 94: potentially check if newJobs exists, cause most times it wont right, if the migration is successful?

packages/mobile/src/store/offline-downloads/sagas/rehydrateOfflineDataSaga.ts
```diff
@@ -36,6 +37,9 @@ export function* rehydrateOfflineDataSaga() {
   const isOfflineModeEnabled = yield* call(getIsOfflineEnabled)
   if (!isOfflineModeEnabled) return
 
+  // Can remove this after all clients are likely updated
+  yield* migrateOfflineDataPathSaga()
```
dylanjeffers, line 41: do we want some analytics around this to make a best guess? no worries but a thought.

packages/mobile/src/store/offline-downloads/sagas/migrateOfflineDataPathSaga.ts
```diff
@@ -0,0 +1,95 @@
...
+  // creates destination directory
+  if (!(await RNFS.exists(destination))) {
+    await RNFS.mkdir(destination)
+  }
+
+  await Promise.all(
+    items.map(async (item) => {
+      //  item is a file
+      if (item.isFile() === true) {
+        const destinationFile = destination + '/' + item.name
```
dylanjeffers, line 61: string interp or we cool like this?

dylanjeffers (approved):
hell yeah so clean

packages/mobile/src/store/offline-downloads/sagas/migrateOfflineDataPathSaga.ts
amendelsohn: oh yeah I think it's unnecessary. ty!

packages/mobile/src/store/offline-downloads/sagas/migrateOfflineDataPathSaga.ts
dylanjeffers: ah actually no worries, cause this whole function is called once.

amendelsohn: yeah this is only called on caught error

packages/mobile/src/store/offline-downloads/sagas/rehydrateOfflineDataSaga.ts
amendelsohn: yeah I was thinking the same thing. Maybe one for success and one for error? When we stop seeing both for a while we're likely good to rip it out.

PR #3154, opened 2023-04-04
Mobile chats use push instead of navigate

dharit-tan:
### Description
Use `navigation.push` instead of `navigate` to navigate to chat screens. Also pass in a `getId` function so that the chats are unique. Otherwise, if a user
- navs to a chat
- clicks on the username header to nav to the profile
- then navs somewhere else and lands on a different user
- then hits the chat entry from that new user's profile page
the navigator will go back (left) instead of forwards.

### Dragons

Had to ts-ignore the type warning in `getId`. I think it's a known issue that typing is pretty wonky with react navigation, have run into this before and confirmed with client team. cc @sliptype in case i'm wrong.

### How Has This Been Tested?

local ios stage

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/screens/app-screen/AppTabScreen.tsx
+++ b/packages/mobile/src/screens/app-screen/AppTabScreen.tsx
@@ -286,7 +286,14 @@ export const AppTabScreen = ({ baseScreen, Stack }: AppTabScreenProps) => {
       <Stack.Group>
         <Stack.Screen name='ChatList' component={ChatListScreen} />
         <Stack.Screen name='ChatUserList' component={ChatUserListScreen} />
-        <Stack.Screen name='Chat' component={ChatScreen} />
+        <Stack.Screen
+          name='Chat'
+          component={ChatScreen}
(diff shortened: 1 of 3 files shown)


packages/mobile/src/screens/app-screen/AppTabScreen.tsx
```diff
@@ -286,7 +286,14 @@ export const AppTabScreen = ({ baseScreen, Stack }: AppTabScreenProps) => {
       <Stack.Group>
         <Stack.Screen name='ChatList' component={ChatListScreen} />
         <Stack.Screen name='ChatUserList' component={ChatUserListScreen} />
-        <Stack.Screen name='Chat' component={ChatScreen} />
+        <Stack.Screen
+          name='Chat'
+          component={ChatScreen}
+          getId={({ params }) => {
+            // @ts-ignore
```
rickyrombo, line 293: nit: would be good to leave a little explainer in the comment

rickyrombo (approved):

PR #3162, opened 2023-04-04
[PLAT-847]: Fix user subscription create album notification copy

sabrina-kiam:
### Description
Was testing uploads and the create album notification for subscribed users was rending a weird copy. Fixed the id parsing in audius backend discovery notifications

before:
IMG_1690

after:
Screenshot 2023-04-04 at 4 24 56 PM

### How Has This Been Tested?

tested locally against stage


dylanjeffers (approved):
nice find!

PR #3167, opened 2023-04-05
[QA-443] Fix cache case matching for handles and permalinks

amendelsohn:
### Description

Broken since https://github.com/AudiusProject/audius-client/pull/3117

toLower the keys so cache hits properly


dylanjeffers (approved):
thank you!

PR #3170, opened 2023-04-05
[PAY-1127] Improve mobile chats performance: remove index dep

dharit-tan:
### Description
Improves performance by removing the dependency on index so we can memoize `ChatMessageListItem`. We calculate `hasTail` in the slice now, and use message ids instead of indices for the `earliestUnreadIndex` comparison. Also includes a few other optimizations.

Todo: use reselect for 2 new selectors. Will follow up with another commit tmr.

Notes:
- @dylanjeffers may still be right that reversing the index order makes more sense, but it's good to remove the dependency on index either way. One thing we do get from this ordering is new messages coming in at the bottom instead of top (when starting from fresh chat).
- Seems like we should be able to get the `CellRenderer` to not re-render as well, but this is probably good enough for now.

### Dragons

### How Has This Been Tested?

Local ios stage, confirmed `ChatMessageListItem` is successfully memoized!

Screenshot 2023-04-04 at 10 46 14 PM

### Feature Flags ###


packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
```diff
@@ -128,101 +128,105 @@ const ChatReaction = ({ reaction }: ChatReactionProps) => {
...
+                    <View
+                      style={[
+                        styles.reactionContainer,
+                        isAuthor
+                          ? styles.reactionContainerIsAuthor
+                          : styles.reactionContainerOtherUser
+                      ]}
+                    >
+                      {message.reactions.map((reaction, index) => {
+                        return <ChatReaction key={index} reaction={reaction} />
```
dylanjeffers, line 199: any way we can key not on index?

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -306,6 +307,48 @@ export const ChatScreen = () => {
...
+      earliestUnreadMessageId,
+      handleMessagePress,
+      styles.unreadSeparator,
+      styles.unreadTag,
+      styles.unreadTagContainer,
+      unreadCount
+    ]
+  )
+
+  const mvcpProp = useMemo(
```
dylanjeffers, line 343: prob best to name same as the actual prop name?

packages/common/src/models/Chat.ts
```diff
@@ -0,0 +1,8 @@
+import type { ChatMessage } from '@audius/sdk'
+
+import { Status } from 'models'
+
+export type ChatMessageWithExtras = ChatMessage & {
+  status?: Status
+  hasTail: boolean
```
dylanjeffers, line 7: this is also a good way to compute, nice

dylanjeffers (approved):
awesome job

packages/common/src/store/pages/chat/slice.ts
```diff
@@ -284,8 +296,22 @@ const slice = createSlice({
     ) => {
       // triggers saga to get chat if not exists
       const { chatId, message, status } = action.payload
+
+      // Recalculate hasTail of previous message
+      const prevLatestMessageId = state.messages[chatId].ids[0]
```
sddioulde, line 301: assuming ids will always have length > 0 here?

packages/common/src/store/pages/chat/slice.ts
```diff
@@ -284,8 +296,22 @@ const slice = createSlice({
...
+      const prevLatestMessage =
+        state.messages[chatId].entities[prevLatestMessageId]
+      if (prevLatestMessage) {
+        const prevMsgHasTail = hasTail(prevLatestMessage, message)
+        chatMessagesAdapter.updateOne(state.messages[chatId], {
```
sddioulde, line 306: nit: ignore if idk what im talking about, but maybe we should only update if `prevMsgHasTail` is diff than current value

packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
```diff
@@ -128,101 +128,105 @@ const ChatReaction = ({ reaction }: ChatReactionProps) => {
...
-                    ) : null}
-                  </>
-                ) : null}
-              </View>
-            </TouchableWithoutFeedback>
-          </View>
-          {hasTail ? (
-            <>
+  const handleLongPress = useCallback(() => {
+    return onLongPress ? onLongPress(message.message_id) : null
```
sddioulde, line 156: nit: don't need `return` here. could just do `onLongPress?.(message.message_id)`

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -229,30 +235,28 @@ export const ChatScreen = () => {
...
-    if (
-      chatId &&
-      chat?.messagesStatus !== Status.LOADING &&
-      chat?.messagesSummary &&
-      chat?.messagesSummary.prev_count > 0
-    ) {
+  const shouldFetchMoreMessages =
+    chatId &&
+    chat?.messagesStatus !== Status.LOADING &&
+    chat?.messagesSummary &&
```
sddioulde, line 251: nit: can probably combine the 2 lines into `chat?.messagesSummary?.prev_count > 0`

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -295,9 +296,9 @@ export const ChatScreen = () => {
     })
     // Need to subtract spacing(2) to account for padding in message View.
     messageTop.current = messageY - spacing(2)
-    setPopupChatIndex(index)
+    setPopupMessageId(id)
     setShouldShowPopup(true)
-  }
+  }, [])
```
sddioulde, line 301: are we voluntarily using an empty deps array here? what if `itemsRef` updates?

sddioulde (approved):
looks good in general. left a few comments

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
dharit-tan: Interesting q.. So adding a dependency to `itemsRef.current` is definitely wrong. Adding dep of `itemsRef` is maybe correct, it would only trigger when we set the pointer to something other than the ref it was initialized with at the top? But since it's a const I don't think this is possible.. Let me know if this makes sense @sddioulde.

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
dharit-tan: Would look like `(chat?.messagesSummary?.prev_count ?? 0) > 0`. I think what we have here is cleaner but lmk @sddioulde .

packages/common/src/store/pages/chat/slice.ts
dharit-tan: So I think it'll just come out to `undefined` if not, which is fine because then the following if statement will fail.

packages/common/src/store/pages/chat/slice.ts
dharit-tan: Great callout. I think it doesn't matter much if we update with the same value as before. There's a potential argument for slightly better performance if we do the check for if `prevMsgHasTail` is diff from current value, but I'd personally opt for one less if statement in the code. Lmk what you think @sddioulde

packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
dharit-tan: Will use `created_at`

packages/mobile/src/screens/chat-screen/ChatMessageListItem.tsx
dharit-tan: Didn't know you could do that!

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
sddioulde: hmm interesting. that makes sense to me thx

packages/common/src/store/pages/chat/slice.ts
sddioulde: the reason why I called it out is because there may be selectors that use this value that would have to update in the UI. but yeah, this is fine

dharit-tan: Ah good call. I think since we're in `addMessage` the UI is gonna update anyway so in this case doesn't matter, but I will remember to think of this.

PR #3179, opened 2023-04-05
[C-2427] Fix preinstall checks for ruby and python versions

amendelsohn:
### Description

bad logic in the preinstall script caused any version to pass the precheck regardless of actual installed version


sliptype (approved):
oof

PR #3180, opened 2023-04-06
[C-1750] Refactor top tags

dylanjeffers:
### Description

Refactors top tags using correct endpoint, stems, and selectors
- Fixes issue where currentUser tags overrode artist profile's tags
- Moves "Tag" component to stems
- Adds "topTags" to profile state, replacing "mostUsedTags"
- Adds "topTags" saga, replacing "fetchMostUsedTags" saga
- Uses /users/:id/tags endpoint instead of fetching entire users tracks, resulting in massive performance improvement
- Uses `react-router` "Link" component to improve DX with tags

TODO: fix margin so profile-left-nav spacing is correct


packages/common/src/store/pages/profile/types.ts
```diff
@@ -31,7 +31,8 @@ export type ProfileState = {
   updateError: boolean
   collectionIds: number[]
   collectionStatus: Status.IDLE
-  mustUsedTags: string[]
```
sliptype, line 34: "must" used tags 🫣

packages/web/src/common/store/profile/fetchTopTagsSaga.ts
```diff
@@ -0,0 +1,27 @@
...
+  fetchTopTags,
+  fetchTopTagsSucceeded,
+  fetchTopTagsFailed
+} = profilePageActions
+
+export function* watchFetchTopTags() {
+  yield* takeLatest(FETCH_TOP_TAGS, fetchTopTagsWorker)
+}
+
+function* fetchTopTagsWorker(action: ReturnType<typeof fetchTopTags>) {
```
sliptype, line 14: I like the pattern of appending `Worker` to the name here, pretty annoying when name collisions happen with the action itself

packages/web/src/pages/profile-page/useProfileRoute.ts
```diff
@@ -0,0 +1,21 @@
...
+import { useSelector } from 'react-redux'
+import { useLocation } from 'react-router-dom'
+
+import { AppState } from 'store/types'
+import { getPathname } from 'utils/route'
+import { parseUserRoute } from 'utils/route/userRouteParser'
+
+const { getUser } = cacheUsersSelectors
+
+export const useProfileRoute = () => {
```
sliptype, line 11: Kinda a weird name for this, maybe `useProfileParams`?

sliptype (approved):
Wow the way these were fetched before was insanely gross, nice work with the refactor!

packages/web/src/common/store/profile/fetchTopTagsSaga.ts
dylanjeffers: thanks! yeah i noticed the param name for second arg to "takeEvery" is "worker", so thought we could go with it

packages/web/src/pages/profile-page/useProfileRoute.ts
dylanjeffers: yeah totally agreed... yeah will do

PR #3181, opened 2023-04-06
Read CID metadata from discovery

michellebrier:
### Description
Try reading CID metadata from discovery before going to content nodes. Still need to backfill discovery so a lot of data is still missing in DNs.

### Dragons

### How Has This Been Tested?

### Feature Flags ###


packages/web/src/services/audio-player/AudioPlayer.ts
```diff
@@ -70,6 +70,7 @@ class fLoader extends Hls.DefaultConfig.loader {
             this.getFallbacks(),
```
michellebrier, line 70: this looks like its for segments -- is codepath shouldn't be in use anymore right?

dylanjeffers: yeah good callout, @raymondjacobson do you think it's time for us to remove hls + track-segment references?

dylanjeffers (approved):
all makes sense! i wonder if we could add the `getFeatureEnabled` call in `fetchCID` so the callers dont have to change?

packages/web/src/services/audio-player/AudioPlayer.ts
raymondjacobson: yep, we should drop it all! 🪓

dylanjeffers: sweet added https://linear.app/audius/issue/C-2429/remove-segments-hls-in-client

michellebrier:
> all makes sense! i wonder if we could add the `getFeatureEnabled` call in `fetchCID` so the callers dont have to change?

yeah will add that in! initially added the flag at all callsites specifically because of the AudioPlayer callsite (discovery metadata doesn't support that) but sounds like its definitely not in use

audius-infra:
Preview this change https://demo.audius.co/mbrier/plat-856/read-cid-metadata-from-discovery

audius-infra:

audius-infra:
Preview this change https://demo.audius.co/mbrier/plat-856/read-cid-metadata-from-discovery

audius-infra:

michellebrier:
> It looks like there may be some changes to native mobile code, which requires triggering a full app release. Please follow the instructions here: https://www.notion.so/audiusproject/When-to-bump-app-version-2644a8f772364a4d91f44abcba44ce0b?pvs=4. cc @nicoback2 @sliptype

@nicoback2 @sliptype does these changes require a version bump? I bumped the sdk dependency and ran npm i so theres changes to `package.json`s and `package-lock.json`s

PR #3182, opened 2023-04-06
Fix missing track on Sammie's page C-2342

nicoback2:
### Description
@amendelsohn turns out this was indeed related to the hidden tracks issue. Have a quick fix here for mobile, but we can close this if your PR is going to address it already!

Sammie's first track wasn't showing up on her profile on mobile, but only when she was logged in. This is because her total_count for # of tracks was 9, but she also had one hidden track. The code was setting the lineup limit to total_count, so it would only load the first 9 tracks (including her hidden track) but not that oldest track.
Previously all 10 tracks were loading for her because her oldest track was set to her artist pick, which is always loaded first/separately.

https://linear.app/audius/issue/C-2342/[qa]-track-missing-on-artist-pagez

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/screens/profile-screen/TracksTab.tsx
+++ b/packages/mobile/src/screens/profile-screen/TracksTab.tsx
@@ -16,13 +16,11 @@ const { getProfileTracksLineup } = profilePageSelectors
 export const TracksTab = () => {
   const dispatch = useDispatch()
 
-  const { handle, user_id, track_count, artist_pick_track_id } =
-    useSelectProfile([
-      'handle',
-      'user_id',
-      'track_count',
-      'artist_pick_track_id'
-    ])
+  const { handle, user_id, artist_pick_track_id } = useSelectProfile([
+    'handle',
+    'user_id',
+    'artist_pick_track_id'
+  ])
 
   const handleLower = handle.toLowerCase()
 
@@ -52,7 +50,6 @@ export const TracksTab = () => {
       leadingElementId={artist_pick_track_id}
       actions={tracksActions}
       lineup={lineup}
-      limit={track_count}
       loadMore={loadMore}
       disableTopTabScroll
       LineupEmptyComponent={<EmptyProfileTile tab='tracks' />}


packages/mobile/src/screens/profile-screen/TracksTab.tsx
```diff
@@ -52,7 +50,6 @@ export const TracksTab = () => {
       leadingElementId={artist_pick_track_id}
       actions={tracksActions}
       lineup={lineup}
-      limit={track_count}
```
nicoback2, line 55: if we don't include this, it still works properly because the lineup component can calculate that there are no more tracks to load once we reach the end.

Kyle-Shanks, line 55: oh interesting. is this limit needed for anything style-wise or something like that?

Kyle-Shanks (approved):
nice

packages/mobile/src/screens/profile-screen/TracksTab.tsx
Kyle-Shanks: nvm saw your comment

sliptype (commented):
Doesn't this mean that any user with hidden tracks is missing tracks on their profile screen? I have hidden tracks on mine but am not missing any tracks 🤔 My track_count is 38 but I'm seeing 40 tracks on my profile screen

edit: Looks like `limit` is the max number of items to fetch per request. `count` is the max number of items in the lineup https://github.com/AudiusProject/audius-client/blob/main/packages/mobile/src/components/lineup/types.ts#L40

packages/mobile/src/screens/profile-screen/TracksTab.tsx
sliptype: We set the max items in the lineup on Track screen to 6, but idk why we would need it here. Removing this should be fine!

sliptype (approved):

nicoback2:
> Doesn't this mean that any user with hidden tracks is missing tracks on their profile screen? I have hidden tracks on mine but am not missing any tracks 🤔 My track_count is 38 but I'm seeing 40 tracks on my profile screen
> 
> edit: Looks like `limit` is the max number of items to fetch per request. `count` is the max number of items in the lineup https://github.com/AudiusProject/audius-client/blob/main/packages/mobile/src/components/lineup/types.ts#L40

@sliptype i think it's because she had exactly 9 tracks + 1 hidden track. seems like we load 9 tracks at a time and then decide whether to load more using this logic: https://github.com/AudiusProject/audius-client/blob/main/packages/mobile/src/components/lineup/Lineup.tsx#L288-L289. But in this case, since limit was set to 9 and we had already loaded 9 tracks, we didn't load any more.

PR #3186, opened 2023-04-06
[C-2306] Fix sign-in sign-up invalid email bug

dylanjeffers:
### Description

Fixes issue where we see "invalid email" when switching from sign-in to sign-up.  Essentially removing all cases where we clear the email between form switches


sliptype (approved):
Dope

PR #3188, opened 2023-04-07
[C-2412] Opaque id deeplinks

sliptype:
### Description

* Support opaque id deeplinks (for example coming from linktree). Web supports these routes but mobile deeplinking did not
* Fix issue with embed player deeplinks

Coming in next PR:
* Jest setup (it's way more frustrating than anticipated)
* Refactor of this deep linking logic to return state objects instead of trying to mangle multiple urls into one mapping
* Unit test

### Dragons

Pretty low risk because I've only added new cases that match these specific routes

### How Has This Been Tested?

Tested locally on simulator

### Feature Flags ###


dylanjeffers (approved):
Niceee

PR #3191, opened 2023-04-07
[C-2431] Add tastmaker notification to mobile

dylanjeffers:
### Description
Adds tastemaker notification to mobile
- Updates tastemaker svg name
- Updates a notification type that was unnecessarily complex

image


sabrina-kiam (approved):
looks great thank u so much!

PR #3195, opened 2023-04-08
[C-2303] Ensure consistent play/active state for entity tiles

dylanjeffers:
### Description

Ensures we have consistent play/active states for track/collection tiles in mobile, web, and mobile-web
- Improves component naming around memo and removes a few duplicate memos. This caused a lot of indentation changes, but overall better cause we drop a layer of nesting.

### DIFF
--- a/packages/common/src/services/audius-backend/AudiusBackend.ts
+++ b/packages/common/src/services/audius-backend/AudiusBackend.ts
@@ -2277,10 +2277,9 @@ export const audiusBackend = ({
           const data = action.data
           if ('playlist_id' in data) {
             entityType = data.is_album ? Entity.Album : Entity.Playlist
-            const playlist_ids = Array.isArray(data.playlist_id)
-              ? data.playlist_id
-              : [data.playlist_id]
-            return playlist_ids.map((playlist_id) => decodeHashId(playlist_id))
+            return data.playlist_id.map((playlist_id) =>
+              decodeHashId(playlist_id)
+            )
           }
           entityType = Entity.Track
           return decodeHashId(data.track_id)
--- a/packages/web/src/components/track/desktop/ConnectedPlaylistTile.tsx
+++ b/packages/web/src/components/track/desktop/ConnectedPlaylistTile.tsx
@@ -103,448 +103,445 @@ type ConnectedPlaylistTileProps = OwnProps &
   ReturnType<typeof mapStateToProps> &
   ReturnType<typeof mapDispatchToProps>
 
-const ConnectedPlaylistTile = memo(
-  ({
-    ordered,
-    index,
-    size,
-    collection,
-    userHandle,
-    containerClassName,
-    user,
-    tracks,
-    togglePlay,
-    playTrack,
-    pauseTrack,
-    playingUid,
-    isBuffering,
-    isPlaying,
-    goToRoute,
-    record,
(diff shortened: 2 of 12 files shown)


packages/common/src/services/audius-backend/AudiusBackend.ts
```diff
@@ -2277,10 +2277,9 @@ export const audiusBackend = ({
           const data = action.data
           if ('playlist_id' in data) {
             entityType = data.is_album ? Entity.Album : Entity.Playlist
-            const playlist_ids = Array.isArray(data.playlist_id)
-              ? data.playlist_id
-              : [data.playlist_id]
-            return playlist_ids.map((playlist_id) => decodeHashId(playlist_id))
+            return data.playlist_id.map((playlist_id) =>
```
raymondjacobson, line 2280: possible for this to be non array still?

packages/web/src/components/track/desktop/ConnectedPlaylistTile.tsx
```diff
@@ -103,448 +103,445 @@ type ConnectedPlaylistTileProps = OwnProps &
...
+  }, [setFavoriteUsers, id, setModalVisibility])
+
+  const onClickStatRepost = useCallback(() => {
+    setRepostUsers(id!)
+    setModalVisibility()
+  }, [setRepostUsers, id, setModalVisibility])
+
+  const renderStats = () => {
+    const contentTitle = 'track' // undefined,  playlist or album -  undefined is track
+    const sz = 'large'
```
raymondjacobson, line 359: nit `size`

unless u meant to write cz binance

packages/web/src/components/track/mobile/ConnectedPlaylistTile.tsx
```diff
@@ -62,242 +62,234 @@ type ConnectedPlaylistTileProps = PlaylistTileProps &
...
-              id: `${trackId}`,
-              source: PlaybackSource.PLAYLIST_TILE_TRACK
-            })
-          )
-        }
-      } else {
-        pauseTrack()
+  const togglePlay = useCallback(() => {
+    if (uploading) return
+    if (!isPlaying || !isActive) {
```
raymondjacobson, line 178: this ternary stuff is a little hard to grok, could be

```
if (uploading) return
if (isActive && !isPlaying) {
...
} else if (!isActive && isPlaying) {
...
}

```

packages/web/src/components/track/types.ts
```diff
@@ -232,9 +235,12 @@ export type DesktopPlaylistTileProps = {
   /** Prefix order number displayed on the left side of the track tile */
   order?: number
 
-  /** If the track is currently playing */
+  /** If the playlist is currently active */
```
raymondjacobson, line 238: 🤌

raymondjacobson (approved):
lemme just rewrite the track and playlist tiles on all the app surfaces we support the last day of bugbash. np.

packages/common/src/services/audius-backend/AudiusBackend.ts
dylanjeffers: lol! yes you think i did more than i really did, its all indentation basically, since we are memozing after rather than as part of the def itself, happy to handle these callouts anyway though :)

PR #3198, opened 2023-04-08
[C-2332] Fix follow artist stat

dylanjeffers:
### Description

Fixes interesting edge-case where feed api returns follower_count as a string, which means following a user on feed triggers a `follower_count + 1` change to entity which means "1" + 1 is "11" giving the appearance your follow increased follow by a multiple. Fix for now is to marshall all user object's follower_count to number, and then giving protocol headsup to fix this value.


amendelsohn (approved):
Wow. Ty javascript

PR #3199, opened 2023-04-10
[PAY-1152] Chats selector to check if is equal to popupMessageId

dharit-tan:
### Description
There was one more dependency on index in `ChatMessageListItem` that was used to determine whether to show/hide the reactions. This was because without this, we would render 2 reactions on top of each other which looked jank. Moved all the logic to a selector so we can still memoize `ChatMessageListItem`.

### Dragons

### How Has This Been Tested?

Local ios

### Feature Flags ###

### DIFF
--- a/packages/common/src/store/pages/chat/selectors.ts
+++ b/packages/common/src/store/pages/chat/selectors.ts
@@ -131,13 +131,16 @@ export const getChatMessageById = (
   return selectById(chatMessagesState, messageId)
 }
 
-export const getPopupMessageId = (state: CommonState) => {
-  return state.pages.chat.popupMessageId
+export const getReactionsPopupMessageId = (state: CommonState) => {
+  return state.pages.chat.reactionsPopupMessageId
 }
 
-export const isEqualToPopupMessageId = createSelector(
-  [getPopupMessageId, (_state: CommonState, messageId: string) => messageId],
-  (popupMessageId: string | null, messageId: string) => {
-    return messageId === popupMessageId
+export const isIdEqualToReactionsPopupMessageId = createSelector(
+  [
+    getReactionsPopupMessageId,
+    (_state: CommonState, messageId: string) => messageId
+  ],
+  (reactionsPopupMessageId: string | null, messageId: string) => {
+    return messageId === reactionsPopupMessageId
   }
 )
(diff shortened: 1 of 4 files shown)


packages/common/src/store/pages/chat/selectors.ts
```diff
@@ -130,3 +130,14 @@ export const getChatMessageById = (
   const chatMessagesState = state.pages.chat.messages[chatId]
   return selectById(chatMessagesState, messageId)
 }
+
+export const getPopupMessageId = (state: CommonState) => {
+  return state.pages.chat.popupMessageId
+}
+
+export const isEqualToPopupMessageId = createSelector(
+  [getPopupMessageId, (_state: CommonState, messageId: string) => messageId],
```
sddioulde, line 139: not sure, but wondering if this in fact does memoize, given the second arg is an unnamed lambda function. I don't know though.
did your test confirm memoization?
curious cuz i havent used this specific pattern before

piazzatron (approved):
Few nits on names, looks good – nice pulling out the selector so we don't re-render everything every time the popup index changes

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -379,12 +383,12 @@ export const ChatScreen = () => {
...
         {/* Everything inside the portal displays on top of all other screen contents. */}
         <Portal hostName='ChatReactionsPortal'>
-          {shouldShowPopup && popupMessageId && popupMessage ? (
+          {shouldShowPopup && popupMessage ? (
             <ReactionPopup
               chatId={chatId}
               messageTop={messageTop.current}
               containerBottom={chatContainerBottom.current}
-              isAuthor={decodeHashId(popupMessage.sender_user_id) === userId}
+              isAuthor={decodeHashId(popupMessage?.sender_user_id) === userId}
```
sddioulde, line 391: seems to me that decoHashId takes in a string, but here the arg may be undefined.

sddioulde (approved):
left comments. but generally looks good

packages/common/src/store/pages/chat/selectors.ts
dharit-tan: Tbh my first time using reselect. I needed a selector that would accept an arg, docs indicated this was the way to do it.

I confirmed the components are memoized, but not sure how to confirm the selector was memoized. Not strictly necessary to use reselect here but thought it was a good chance to try it out.

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
dharit-tan: Yep but I think that's ok, confirmed `decodeHashId(undefined)` comes out to `null` which will still give us desired behavior.

PR #3200, opened 2023-04-10
[C-2435] Fix issue with collectibles cache update

dylanjeffers:
### Description

Fixes weird issue where `fetchProfileCustomizedCollectibles` updates the entire user object, which as a lot of incorrect data from the `fetchCID` function, resulting in a profile screen going from "artist" to "user", which turns into runtime error as the profile tabs don't know how to transition focus (i think basically tabs are not expecting to drop by 3). not sure what is best here, but most likely we just want to update the collectibles key. Are there any other keys we want to include from this function? maybe @raymondjacobson would know the most here?

### DIFF
--- a/packages/web/src/common/store/profile/sagas.js
+++ b/packages/web/src/common/store/profile/sagas.js
@@ -96,7 +96,7 @@ function* fetchProfileCustomizedCollectibles(user) {
           {
             id: user.user_id,
             metadata: {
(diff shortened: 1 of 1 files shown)


amendelsohn (approved):
Looks good, but I don't have context on the other keys we might need.

raymondjacobson (approved):
lgtm

PR #3202, opened 2023-04-10
[PAY-827][PAY-1154] Use DN stream endpoint for track download and offline mode

sddioulde:
### Description

Update track downline and offline mode to both use DN /stream endpoint. Related to https://github.com/AudiusProject/audius-protocol/pull/5075 and helps us towards deperecating all non-cid DN + CN stream paths.

### Dragons

### How Has This Been Tested?

local web and mobile app vs stage

### Feature Flags ###


packages/mobile/src/store/offline-downloads/sagas/offlineQueueSagas/workers/downloadTrackWorker.ts
```diff
@@ -157,13 +160,36 @@ function* downloadTrackAudio(track: UserTrackMetadata) {
...
 
-  for (const creatorNodeEndpoint of creatorNodeEndpoints) {
-    const trackAudioUri = `${creatorNodeEndpoint}/tracks/stream/${encodedTrackId}`
-    const response = yield* call(downloadFile, trackAudioUri, trackFilePath)
-    const { status } = response.info()
-    if (status === 200) return
+  const audiusBackendInstance = yield* getContext('audiusBackendInstance')
+  const apiClient = yield* getContext('apiClient')
+  const getFeatureEnabled = yield* getContext('getFeatureEnabled')
+  const isGatedContentEnabled = yield* call(() => {
```
dharit-tan, line 166: Haven't seen this pattern before. The one I'm familiar with is:
```
  const isNewPodcastControlsEnabled = yield* call(
    getFeatureEnabled,
    FeatureFlags.PODCAST_CONTROL_UPDATES_ENABLED,
    FeatureFlags.PODCAST_CONTROL_UPDATES_ENABLED_FALLBACK
  )
```

packages/mobile/src/store/offline-downloads/sagas/offlineQueueSagas/workers/downloadTrackWorker.ts
```diff
@@ -157,13 +160,36 @@ function* downloadTrackAudio(track: UserTrackMetadata) {
...
+    }
   }
 
+  const trackAudioUri = apiClient.makeUrl(
+    `/tracks/${encodedTrackId}/stream`,
+    queryParams
+  )
+  const response = yield* call(downloadFile, trackAudioUri, trackFilePath)
+  const { status } = response.info()
+  if (status === 200) return
```
dharit-tan, line 191: super nit but "don't special-case the happy path" is something I've tried to adhere to. In this case maybe `if (status !== 200) throw error else return`, but up to you.

packages/web/src/common/store/social/tracks/sagas.ts
```diff
@@ -634,11 +638,59 @@ export function* watchUnsetArtistPick() {
...
+}: {
+  track: Track
+  filename: string
+}) {
+  try {
+    const audiusBackendInstance = yield* getContext('audiusBackendInstance')
+    const apiClient = yield* getContext('apiClient')
+    const trackDownload = yield* getContext('trackDownload')
+    const getFeatureEnabled = yield* getContext('getFeatureEnabled')
+    const isGatedContentEnabled = yield* call(() => {
```
dharit-tan, line 653: ditto above

packages/web/src/services/track-download.ts
```diff
@@ -1,15 +1,25 @@
...
-  ) {
-    const audiusLibs = await this.audiusBackend.getAudiusLibs()
-    return audiusLibs.File.downloadCID(cid, creatorNodeEndpoints, filename)
+  async downloadTrack({ url, filename }: { url: string; filename: string }) {
+    const response = await window.fetch(url)
+    if (!response.ok) {
+      throw new Error('Download unsuccessful')
+    }
+
+    const downloadURL = (url: string, filename: string) => {
```
dharit-tan, line 12: I'm not familiar with this code path at all, but is this really how we download tracks? make a link and calling click on it seems strange, but if it's how we do things otherwise then ok!

dharit-tan (approved):

packages/mobile/src/store/offline-downloads/sagas/offlineQueueSagas/workers/downloadTrackWorker.ts
sddioulde: yeah I've seen both. it's that typescript was complaining because of `getFeatureEnabled` returning a Promise. Weird that the other pattern works too. Can dig or chat with a client folk in case im missing sthg

packages/mobile/src/store/offline-downloads/sagas/offlineQueueSagas/workers/downloadTrackWorker.ts
sddioulde: fair. i did some copy/pasta :P

packages/web/src/services/track-download.ts
sddioulde: yeah ive seen this a bunch in my years writing code. including in libs where i got this from, which is how current track download works. check [this](https://github.com/AudiusProject/audius-protocol/blob/main/libs/src/api/File.ts#L14) out.

piazzatron:
Would be good to include someone from @AudiusProject/client on this PR since it touches client-team code

packages/mobile/src/store/offline-downloads/sagas/offlineQueueSagas/workers/downloadTrackWorker.ts
piazzatron: Yeah this is odd, worth digging in to figure out why our usual pattern doesn't work. @sddioulde looking around I found a few places in the codebase where I think we're not awaiting this promise:

```
  const getFeatureEnabled = yield* getContext('getFeatureEnabled')
  const isGatedContentEnabled = getFeatureEnabled(
    FeatureFlags.GATED_CONTENT_ENABLED
  )
```

Worth a search for `getFeatureEnabled` in `**/src/**/sagas**`

piazzatron (changes requested):
Generally looking good, I'd like for us to consolidate the signature generation / signature map code before we merge this. Let's also check in with client team since they have familiarity with the track download flow. 

Also, I'm not sure we even need to feature flag this anymore! Worth having a discussion offline but since CN expects signatures now, I don't see us going back.

packages/mobile/src/store/offline-downloads/sagas/offlineQueueSagas/workers/downloadTrackWorker.ts
amendelsohn: Yeah I think the usual pattern should be possible

sddioulde: fixed

packages/web/src/common/store/social/tracks/sagas.ts
```diff
@@ -634,11 +639,53 @@ export function* watchUnsetArtistPick() {
...
+    const audiusBackendInstance = yield* getContext('audiusBackendInstance')
+    const apiClient = yield* getContext('apiClient')
+    const trackDownload = yield* getContext('trackDownload')
+    const getFeatureEnabled = yield* getContext('getFeatureEnabled')
+    const isGatedContentEnabled = yield* call(getFeatureEnabled, FeatureFlags.GATED_CONTENT_ENABLED)
+    let queryParams: QueryParams = {}
+    if (isGatedContentEnabled) {
+      const premiumTrackSignatureMap = yield* select(
+        getPremiumTrackSignatureMap
+      )
```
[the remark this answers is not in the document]
sddioulde: added this as part of the refactor card on linear. hope to get to soon, along with other stuff

PR #3203, opened 2023-04-10
[PAY-748] Fix supporter count alignment

schottra:
### Description
The code for positioning the "+NN" text relies on some negative margins and math related to the width of the underlying image. With the "dimming" element meant to stack beneath it, this can get tricky to keep in sync with the variable-sized image support. So I've updated the method to be a little closer to what I think the intention was:

* The text is now a child of the View with the dimming background and will be centered within it.
* The View containing the text will be absolute positioned to the parent above it, so that the parent will only be the width/height of the first child (the `ProfilePicture` we are drawing above)
* Said View has its width/height computed based on the incoming image size as before, but this is just to give it a fixed amount of padding from the edges to maintain the original effect. Otherwise, it occupies the entire container and centers its content.

The end result is that the text should always stay centered above the ProfilePicture, without the use of negative margins.

I also update the borderRadius to be calculated based on the image size. We want a circle, which should be a 50% border radius. So we take half of the width and round up based on the assumption that ProfilePictures are always square dimensions (which they are).

### Dragons
None

### How Has This Been Tested?
Manually on mobile emulators (both Android and iOS). View any account with a "Supporting" list > 19.

Android screenshot
image

iOS screenshot
image

No monitoring necessary other than to verify in production once shipped.

### Feature Flags ###
No new features.


packages/mobile/src/screens/notifications-screen/Notification/ProfilePictureList.tsx
```diff
@@ -16,6 +16,7 @@ const USER_LENGTH_LIMIT = 9
  * use the default of spacing(10) - 2 (which is equal to 38).
  * We use the dimensions to determine how to position the
  * extra profile picture +N text.
+ *
```
dharit-tan, line 19: nit: delete extra line

packages/mobile/src/screens/notifications-screen/Notification/ProfilePictureList.tsx
```diff
@@ -34,9 +35,13 @@ const useStyles = makeStyles(({ spacing, palette, typography }) => ({
...
-    borderRadius: 15,
-    width: 24,
-    height: 24
+    position: 'absolute',
+    alignItems: 'center',
+    justifyContent: 'center',
+    top: PROFILE_PICTURE_BORDER_WIDTH,
+    right: PROFILE_PICTURE_BORDER_WIDTH,
+    bottom: PROFILE_PICTURE_BORDER_WIDTH,
+    left: PROFILE_PICTURE_BORDER_WIDTH
```
piazzatron, line 44: nice

packages/mobile/ios/Podfile.lock
```diff
@@ -1202,4 +1202,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 6927948487d04eb4c86fe5056d0c6dc0bd531bf3
 
-COCOAPODS: 1.11.3
+COCOAPODS: 1.12.0
```
piazzatron, line 1205: Intentional?

piazzatron (approved):
🚀 Looks great! Appreciate the thorough documentation in PR doc as well as code comment :)

packages/mobile/ios/Podfile.lock
dharit-tan: Might wanna check with client team before checking this in.

packages/mobile/src/screens/notifications-screen/Notification/ProfilePictureList.tsx
```diff
@@ -114,17 +128,18 @@ export const ProfilePictureList = (props: ProfilePictureListProps) => {
...
+                width: dimWidth,
+                height: dimHeight,
+                // borderRadius of 50% or greater gives us a circle
+                borderRadius: Math.ceil(dimWidth / 2)
+              }
             ]}
           >
-            {`+${formatCount(remainingUsersCount)}`}
-          </Text>
+            <Text style={[styles.imageCount]}>
```
dharit-tan, line 139: nit: probably don't need `[]` around the style.

dharit-tan (approved):

packages/mobile/ios/Podfile.lock
schottra: cc @sliptype . I believe we discussed this a few days ago and your answer was along the lines of "should be fine as long as you can run the mobile app" 😁

sliptype: Yup I have this change locally too! Commit it 💪

audius-infra:

dylanjeffers (dismissed):
hey small thing! all this looks wonderful, just requesting changes just so we can talk about the web package.json-lock changes. we will be looking to move to node 16 soon, but in meantime we are noticing folks machines can diverge with package-lock creation. we have just agreed to basically not commit locks unless a pr is changing packages in package.json. when it does come time to add a package, lmk and i can show you how i do it to minimize changes. thanks :)

schottra:
@dylanjeffers Makes sense! Ideally, if we're all using the same npm version locally, the install shouldn't result in the lockfile changing?
I think I pulled a more recent version of node when I did a machine setup and/or a previous modification to packages.json didn't include a lockfile change.

But happy to abide by that policy going forward. Want me to revert the lockfile change in this PR?

dylanjeffers:
> @dylanjeffers Makes sense! Ideally, if we're all using the same npm version locally, the install shouldn't result in the lockfile changing? I think I pulled a more recent version of node when I did a machine setup and/or a previous modification to packages.json didn't include a lockfile change.
> 
> But happy to abide by that policy going forward. Want me to revert the lockfile change in this PR?

gotcha, yeah there is weirdness potentially due to different versions of node, but we havent tracked down a reason why, i think node 14 basically has this issue, hence the goal to update. reverting that change would be great for now, thanks. do you have nvm set up btw? if not can show you how we do that so ideally you have more consistent install experience

schottra:
> 

Yep! I'm using nvm and let it installed the Node version specified in the local config file.

audius-infra:
Preview this change https://demo.audius.co/fix-supporter-count-alignment

PR #3206, opened 2023-04-10
Configure Jest for mobile testing

sliptype:
### Description
image

* Configure Jest to run successfully for the mobile app
* Use babel to compile typescript instead of `ts-jest` which wasn't working due to Flow files with `.js` extensions in the react-native codebase
  > The drawback to this is that our tests won't be typechecked when run
* Add babel plugin to transform `react-use` imports so that browser env specific hooks aren't imported

Coming next:
* Test for the NavigationContainer linking config

### Dragons

Added `process.env` checks to determine if running via Jest. Will confirm that doesn't break anything in non-test envs

### How Has This Been Tested?

### Feature Flags ###


audius-infra:

packages/common/src/services/solana-client/SolanaClient.ts
```diff
@@ -22,7 +22,9 @@ export class SolanaClient {
     try {
       this.connection = new Connection(solanaClusterEndpoint!, 'confirmed')
     } catch (e) {
-      console.error('Could create Solana RPC connection', e)
+      if (process.env.JEST_WORKER_ID === undefined) {
```
dylanjeffers, line 25: fun! just wondering what the idea is here? is this to avoid extra logs when running tests? would this be the way we would handle logs like this going forward? Would we want to just mock out SolanaClient?

packages/common/src/store/pages/trending/reducer.ts
```diff
@@ -22,7 +22,9 @@ import {
   makeInitialState
 } from './lineup/reducer'
 
-const urlParams = new URLSearchParams(window.location.search)
+const urlParams = new URLSearchParams(
```
dylanjeffers, line 25: ah interesting, this is cool to see where we make assumptions. since this is in common... i wonder if we should update this at som point?

packages/mobile/babel.config.js
```diff
@@ -1,7 +1,23 @@
 module.exports = (api) => {
   const babelEnv = api.env()
   const plugins = [
-    ['@babel/plugin-transform-react-jsx', { runtime: 'automatic' }]
+    ['@babel/plugin-transform-react-jsx', { runtime: 'automatic' }],
```
dylanjeffers, line 4: wow amazing work making this work. is this based on an issue we could potentially reference? or maybe some comments? cause this is pretty wild

packages/mobile/jest.config.js
```diff
@@ -0,0 +1,55 @@
+// These are required to be transformed because
+// they are ESM and jest doesn't support ESM
```
dylanjeffers, line 2: ive heard decent things about https://vitest.dev/, i wonder if that would support esm, cause its pretty wild just doesn't. obvs not something for rn, just a long term thing

packages/mobile/jest.setup.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+}
+
+const Blob = require('blob-polyfill').Blob
+
+window.Blob = Blob
+
+global.navigator = {
+  userAgent: 'node.js'
+}
```
dylanjeffers, line 125: wow incredible mocks! just checking that we need to manually mock things like `@react-native-cookies/cookies`, cause i think if you simply say `jest.mock('@react-native-cookies/cookies')` it will try and fill in the blanks and mock every exported module, not sure though.

packages/mobile/package.json
```diff
@@ -170,28 +170,33 @@
...
+    "@babel/preset-typescript": "7.21.4",
     "@babel/runtime": "7.21.0",
     "@react-native-community/eslint-config": "3.2.0",
     "@svgr/plugin-svgo": "5.5.0",
     "@tsconfig/react-native": "2.0.3",
     "@types/array.prototype.flat": "1.2.1",
     "@types/bn.js": "5.1.0",
     "@types/hls.js": "0.13.0",
-    "@types/jest": "24.9.1",
+    "@types/jest": "29.5.0",
```
dylanjeffers, line 184: big changes, nice

packages/mobile/src/app/App.test.tsx
```diff
@@ -6,7 +6,7 @@ import 'react-native'
 // Note: test renderer must be required after react-native.
 import renderer from 'react-test-renderer'
 
-import { App } from 'app/app'
+import { App } from './App'
```
dylanjeffers, line 9: wow it actually renders!

dylanjeffers (approved):
amazing work getting this set up! excited to add more frameworks and see where we can take this.

packages/common/src/services/solana-client/SolanaClient.ts
sliptype: Yeah this was to avoid big errors when testing, would probably be better to mock solana client. Let me try that!

packages/common/src/store/pages/trending/reducer.ts
sliptype: Yeah most definitely, shouldn't be relying on window.location in the reducer

packages/mobile/babel.config.js
sliptype: Good call, let me add a comment. For reference here is the issue:
https://github.com/streamich/react-use/issues/483
https://github.com/streamich/react-use/blob/master/docs/Usage.md

packages/common/src/services/solana-client/SolanaClient.ts
schottra: Also curious about using JEST_WORKER_ID? I thought JEST sets `NODE_ENV='test'`. (https://jestjs.io/docs/environment-variables)

packages/common/src/store/pages/trending/reducer.ts
schottra: The pattern I've seen most often is to just throw any global access into its own module as a re-export such that you can do
`import {location} from './location';`

Then it's easy to automock it for testing.

packages/mobile/jest.setup.js
schottra: Auto-mocking depends heavily on _how_ the module exports things. I've rarely had it just work, unfortunately.

packages/mobile/src/app/App.test.tsx
schottra: Is our goal to render the whole app and test it in an integration or e2e style fashion?
In my experience, the smaller the scope for your tests, the less you need to do things like mocking all of the external libraries and browser API usage etc.

sliptype: I think some high level integration / smoke tests would be good to have, seeing as we have no tests for mobile at all right now. We at least have probers for web. But generally I agree that the scope of most tests should be small, the whole reason I went down this path was to unit test the deep-linking code lol

audius-infra:

packages/mobile/src/app/App.test.tsx
dylanjeffers: Yeah agreed we can start running smaller tests now, but it's great you covered all the bad cases that we might experience along the way.

audius-infra:

PR #3208, opened 2023-04-11
[C-2082] Fix busted profile edits

raymondjacobson:
### Description

Couple things:
* Pull to refresh totally breaks profile updates. In commit 1, make it so that we never override something that's in the confirmer. The confirmer finishes by refetching and caching the new object, so we should avoid doing this while confirming. Mobile web (prior to reloaded) had a solve for this that I've deleted since it's pretty brittle
* Fix profile/cover photo edits by allowing the local updates to be used in fastimage, etc.

### Dragons

### How Has This Been Tested?

Local simulator & web browser for commit 1

### Feature Flags ###


audius-infra:

packages/mobile/src/components/image/UserCoverImage.tsx
```diff
@@ -31,12 +31,21 @@ type CoverImageUser = Nullable<
...
+  const contentNodeImage = useContentNodeImage({
     cid,
     user,
     size: WidthSizes.SIZE_640,
     fallbackImageSource: imageCoverPhotoBlank
   })
+
+  if (user.updatedCoverPhoto) {
+    return {
+      source: { uri: user.updatedCoverPhoto.url },
```
dylanjeffers, line 43: That's fun

packages/mobile/src/components/image/UserImage.tsx
```diff
@@ -19,12 +19,20 @@ type UseUserImageOptions = {
...
 
-  return useContentNodeImage({
+  const contentNodeImage = useContentNodeImage({
     cid,
     size,
     user,
     fallbackImageSource: profilePicEmpty
   })
+
+  if (user.updatedProfilePicture) {
```
dylanjeffers, line 29: Would we want to add this to the core image hook or just a mobile only thing for now?

packages/web/src/common/store/cache/sagas.js
```diff
@@ -98,15 +99,21 @@ export function* retrieve({
     call(getEntriesTimestamp, uniqueIds)
   ])
 
+  const confirmer = yield select(getConfirmCalls)
   // Figure out which IDs we need to retrive from source
   const idsToFetch = []
   uniqueIds.forEach((id) => {
-    if (
+    let isConfirming = false
+    if (cachedEntries?.[id]?.[idField]) {
```
dylanjeffers, line 107: Awesome adds. Isn't the cache add step check if confiming, and if so don't call add_successful?

packages/web/src/common/store/cache/users/sagas.js
```diff
@@ -213,17 +213,17 @@ function* watchFetchProfilePicture() {
...
-                [size]: url
-              }
-            }
             yield put(
               cacheActions.update(Kind.USERS, [
-                { id: userId, metadata: userWithProfilePicture }
+                {
+                  id: userId,
+                  metadata: {
+                    _profile_picture_sizes: {
```
dylanjeffers, line 221: Nice fix, so basically don't merge whole object?

dylanjeffers (approved):
Awesome stuff

packages/mobile/src/components/image/UserImage.tsx
raymondjacobson: It's handled separately on web for now, so this at least brings parity. Would be good to move deeper into common though

packages/web/src/common/store/cache/users/sagas.js
raymondjacobson: yeah same as your fix

packages/web/src/common/store/cache/sagas.js
raymondjacobson: You're right, we already had code to do this. It seems like pull to refresh explicitly passes through an arg to blow away the current entry in the cache. Maybe we can huddle abt this, not quite sure what the reasoning for that was

raymondjacobson: Ok I removed the code I added here and also dropped the arg in pull to refresh and I see the same behavior as with my fix. Mine might be better b/c we lose a network hop, but it's definitely a riskier change so will merge without.

audius-infra:

packages/web/src/common/store/cache/sagas.js
dylanjeffers: Whatever you think is best, this can be something we discuss during irl too

PR #3210, opened 2023-04-11
Attempt to update circle config to move off python2.7-jessie

schottra:
### Description
We're seeing build failures for CircleCI steps which still leverage the `circleci/python:2.7-jessie` image. This is due to `jessie` being long deprecated and removed from distribution sites. We were only using that image to support installing the aws-cli package via pip. However, there is now an orb that we can use to install this package.

This PR updates the web workflows using the outdated image to use the orb for installation of aws-cli, and to build from the `cimg/python` image used by the other web workflows.

### Dragons
It's a new image and way of installing the SDK, so there could be dependency or configuration issues we aren't aware of.

### How Has This Been Tested?
The plan is to run through the workflows in CircleCI to verify things are working. There are no app code changes as part of this, so the deploy steps should result in identical code.

Directly monitor output of all affected CI workflows. In the event the deployment workflows are unsuccessful, we can deploy an old branch to restore them and revert the PR.

### Feature Flags ###
No feature flags

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,6 +29,7 @@ parameters:
 setup: true
 
 orbs:
+  aws-cli: circleci/aws-cli@3.1.5
   continuation: circleci/continuation@0.2.0
   path-filtering: circleci/path-filtering@0.1.1
 
--- a/.circleci/src/commands/@web-commands.yml
+++ b/.circleci/src/commands/@web-commands.yml
@@ -109,11 +109,6 @@ web-distribute:
           cd packages/web
           npm run dist:<< parameters.build-type >>
 
-web-install-awscli:
-  steps:
-    - run:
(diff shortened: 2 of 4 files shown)


sliptype (approved):
Awesome! Thanks for doing this

schottra:
@dylanjeffers It probably makes sense to update CI. I was going for minimal change here to unblock other PRs. So I updated the images to match other commands/workflows in the config file which were all using the 3.7 image.
I can create a ticket to update them all together.

dylanjeffers:
> @dylanjeffers It probably makes sense to update CI. I was going for minimal change here to unblock other PRs. So I updated the images to match other commands/workflows in the config file which were all using the 3.7 image. I can create a ticket to update them all together.

sounds like a plan!

PR #3215, opened 2023-04-11
Fix safari refresh event not being captured

dharit-tan:
### Description
Fixes safari beforeunload events not being captured properly.
Note: cmd+r hotkey remapping still doesn't work, we get the generic safari alert instead of our nice "hang tight" one. But with this fix at least we prevent a user losing their changes unknowingly.

### Dragons

### How Has This Been Tested?

Local web on safari, brave, chrome. Confirmed behavior on brave/chrome stayed the same.

### Feature Flags ###


raymondjacobson (approved):
nice nice, i think this was the behavior once upon a time and safari must have changed some things. it's very hard to hijack their unbeforeunload

PR #3218, opened 2023-04-12
[PAY-1128] Update logic for showing "earn/claim $AUDIO" pill

schottra:
### Description
This updates the logic in `NavAudio` to wait on values used for conditional rendering to finish loading. The logic here uses information from 3 different calls. So depending on order, you could see the pill in multiple states in rapid succesion.

Also did some cleanup of the logic and styling:
* Updated balance calculations to prefer wallet balance if loaded, falling back to account balance and then a constant of `0`.
* Updated logic for displaying bubble to wait on loading of challenges and wallet for the two cases where we show "claim" and "earn" variants
* Removed the `useState` and `useEffect` pair for updating the bubble type. This should be unnecessary, since the hooks we use to select data will cause a re-render when the results change.
* Make the token counter and earn/claim pills separate click targets. They point to the same place now, but that may change. It also fixes a behavior where a big empty space to the right of the token count is clickable when there is no element there.
* Consolidate some styles that appear to have been originally applied conditionally, but are now always applied.
* Slightly odd: Due to the way `Transition` will render the component when it changes, we get two elements: one for the old text and one for the updated text. They were rendering side-by-side and animating together, which is what caused the temporary change in width. I updated the styling to stack them, so that we get a fade between the two in-place.

Here's what happens if the wallet loads and reflects a balance of zero (even if the balance from the account record originally showed positive), when there are no claimable rewards
Kapture 2023-04-14 at 10 47 59

Here's a case where there are claimable challenges and the wallet loads and updates the balance to 0.
Kapture 2023-04-14 at 11 02 11

And then here's what happens if the challenges take longer to load than the wallet and the balance in the wallet is zero (the original source of the flashing bug)
Kapture 2023-04-14 at 11 11 24

### Dragons
None

### How Has This Been Tested?
Tested manually in my local environment. I completed some challenges on a new account to try out the flow and also hard-coded some states and timers to mimic the behavior of the variables changing over time.

N/A

### Feature Flags ###
N/A

### DIFF
--- a/packages/web/src/components/nav/desktop/NavAudio.tsx
+++ b/packages/web/src/components/nav/desktop/NavAudio.tsx
@@ -6,6 +6,7 @@ import {
   StringKeys,
   formatWei,
   accountSelectors,
+  audioRewardsPageSelectors,
   walletSelectors,
   useSelectTierInfo,
   useAccountHasClaimableRewards
@@ -23,8 +24,9 @@ import { useSelector } from 'utils/reducer'
 import { AUDIO_PAGE } from 'utils/route'
 
 import styles from './NavAudio.module.css'
-const { getAccountTotalBalance } = walletSelectors
+const { getAccountTotalBalance, getAccountBalanceLoading } = walletSelectors
 const getAccountUser = accountSelectors.getAccountUser
+const { getUserChallengesLoading } = audioRewardsPageSelectors
 
 type BubbleType = 'none' | 'claim' | 'earn'
 
@@ -35,7 +37,10 @@ const messages = {
 
 const NavAudio = () => {
   const navigate = useNavigateToPage()
+  const userChallengesLoading = useSelector(getUserChallengesLoading)
   const account = useSelector(getAccountUser)
+  const balanceLoading = useSelector(getAccountBalanceLoading)
+
   let totalBalance = useSelector(getAccountTotalBalance)
   if (totalBalance.eq(new BN(0)) && account?.total_balance) {
     totalBalance = new BN(account?.total_balance) as BNWei
@@ -61,24 +66,23 @@ const NavAudio = () => {
     } else if (!positiveTotalBalance) {
       setBubbleType('earn')
     } else {
-      setBubbleType('none')
     }
   }, [setBubbleType, hasClaimableRewards, positiveTotalBalance])
 
-  if (!account) {
+  // Wait for all the states we care about to load to prevent flashing
+  // of conditional content
+  if (balanceLoading || userChallengesLoading || !account) {
     return null
   }
 
   return (
-    <div
-      className={cn(
-        styles.audio,
-        { [styles.hasBalance]: positiveTotalBalance },
-        { [styles.show]: true }
-      )}
-      onClick={goToAudioPage}
-    >
-      <div className={styles.amountContainer}>
+    <div className={styles.audio}>
+      <div
+        className={cn(styles.amountContainer, styles.interactive, {
+          [styles.hasBalance]: positiveTotalBalance
+        })}
+        onClick={goToAudioPage}
+      >
         {positiveTotalBalance && audioBadge ? (
           cloneElement(audioBadge, {
             height: 16,
@@ -103,9 +107,10 @@ const NavAudio = () => {
             item !== 'none' && (
               <animated.span
                 style={props}
-                className={cn(styles.actionBubble, {
+                className={cn(styles.actionBubble, styles.interactive, {
                   [styles.claimRewards]: item === 'claim'
                 })}
+                onClick={goToAudioPage}
               >
                 <span>
                   {item === 'claim'
--- a/packages/web/src/components/nav/desktop/NavAudio.tsx
+++ b/packages/web/src/components/nav/desktop/NavAudio.tsx
@@ -1,8 +1,7 @@
-import { cloneElement, useCallback, useEffect, useState } from 'react'
+import { cloneElement, useCallback } from 'react'
 
 import {
   BadgeTier,
-  BNWei,
   StringKeys,
   formatWei,
   accountSelectors,
@@ -35,16 +34,11 @@ const messages = {
   claimRewards: 'Claim Rewards'
 }
 
-const NavAudio = () => {
+const RenderNavAudio = () => {
   const navigate = useNavigateToPage()
-  const userChallengesLoading = useSelector(getUserChallengesLoading)
   const account = useSelector(getAccountUser)
-  const balanceLoading = useSelector(getAccountBalanceLoading)
 
-  let totalBalance = useSelector(getAccountTotalBalance)
-  if (totalBalance.eq(new BN(0)) && account?.total_balance) {
-    totalBalance = new BN(account?.total_balance) as BNWei
-  }
+  const totalBalance = useSelector(getAccountTotalBalance)
   const positiveTotalBalance = totalBalance.gt(new BN(0))
   // we only show the audio balance and respective badge when there is an account
   // so below null-coalescing is okay
@@ -54,25 +48,15 @@ const NavAudio = () => {
   const challengeRewardIds = useRemoteVar(StringKeys.CHALLENGE_REWARD_IDS)
   const hasClaimableRewards = useAccountHasClaimableRewards(challengeRewardIds)
 
-  const [bubbleType, setBubbleType] = useState<BubbleType>('none')
-
   const goToAudioPage = useCallback(() => {
     navigate(AUDIO_PAGE)
   }, [navigate])
 
-  useEffect(() => {
-    if (hasClaimableRewards) {
-      setBubbleType('claim')
-    } else if (!positiveTotalBalance) {
-      setBubbleType('earn')
-    } else {
-    }
-  }, [setBubbleType, hasClaimableRewards, positiveTotalBalance])
-
-  // Wait for all the states we care about to load to prevent flashing
-  // of conditional content
-  if (balanceLoading || userChallengesLoading || !account) {
-    return null
+  let bubbleType: BubbleType = 'none'
+  if (hasClaimableRewards) {
+    bubbleType = 'claim'
+  } else if (!positiveTotalBalance) {
+    bubbleType = 'earn'
   }
 
   return (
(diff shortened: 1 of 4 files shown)


audius-infra:
Preview this change https://demo.audius.co/fix-claim-button-flickering

audius-infra:
Preview this change https://demo.audius.co/fix-claim-button-flickering

packages/web/src/components/nav/desktop/NavAudio.tsx
```diff
@@ -103,9 +91,10 @@ const NavAudio = () => {
             item !== 'none' && (
```
dharit-tan, line 91: not your code and not your problem but figured i'd point out for learning purposes: we're trying stick to our styleguide and avoid using `&&` in favor of ternaries.

schottra: That's what I get for not reading the style guide yet.
I've typically been in favor of the `condition ? (component) : null` pattern. But I also did recently read something about that being an anti-pattern now due to the way React handles `null` vs `false`

dharit-tan: Oh wut. Please do share if you find that.

packages/web/src/components/nav/desktop/NavAudio.tsx
```diff
@@ -121,4 +110,18 @@ const NavAudio = () => {
   )
 }
 
+const NavAudio = () => {
+  const userChallengesLoading = useSelector(getUserChallengesLoading)
```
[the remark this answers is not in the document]
piazzatron: have some thoughts, will do a thorough pass

packages/web/src/components/nav/desktop/NavAudio.tsx
schottra: Alright I did an overkill modification for you and just moved the logic to another component. It was kind of driving me nuts anyway.
https://github.com/AudiusProject/audius-client/pull/3218/commits/cfbb6283d8da8e4b709e2f3174a0b0662fed2403

packages/web/src/components/nav/desktop/NavAudio.tsx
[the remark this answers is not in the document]
schottra: Yeah. So there are quite a few places where we have selector logic normalizing things like this to empty values. I presume it was to avoid errors when attempting to render values which haven't loaded yet.
I was waiting on the cache/fetch refactoring to define a pattern before trying to address this more directly.

Ideally, I'd like to have _all_ fetchable items share the same interface where you have `data` and then metadata indicating the state of the fetch. So you can safely ignore the value of `data` until the fetch is in a terminal state.

The pattern I have here with the fetching and presentational components is IMO a clean way to deal with the issue. Normally, I'd do something like `return isLoading ? <NavAudioSkeleton /> : <NavAudio balance={balance} />`

packages/web/src/components/nav/desktop/NavAudio.tsx
```diff
@@ -87,22 +117,13 @@ const RenderNavAudio = () => {
...
           config={{ duration: 100 }}
         >
-          {(item) => (props) =>
-            item !== 'none' && (
-              <animated.span
-                style={props}
-                className={cn(styles.actionBubble, styles.interactive, {
-                  [styles.claimRewards]: item === 'claim'
-                })}
+          {(bubbleType) => (bubbleStyle) =>
```
dharit-tan, line 120: Hm not familiar with this pattern, does omitting line 120 and just doing `style={styles.bubbleStyle}` not work?

schottra: `style` is passed in by `<Transition />`. It's how react-spring does the CSS to animate the element, I believe.

audius-infra:
Preview this change https://demo.audius.co/fix-claim-button-flickering

piazzatron (changes requested):
Code wise, this looks great! Love the cleanups, explicit loading states, separation of component, and removing the superfluous state + useEffect.

I have some thoughts about delaying the load for the audio amount + earn/claim badge:

- As it stands in this PR, when the nav row renders in the entire nav column gets pushed down, which I find quite visually jarring and maybe more attention grabbing than the strobing we have today. 
- In terms of an 80/20 happy path, most users probably don't have challenges that are waiting to be disbursed, though they may have a few AUDIO sitting around. So for that class of users, this change is delaying showing them information they used to see immediately (their AUDIO balance - I tested this against my prod account). I added some logs and at least in my case, I see this in terms of response ordering:
- The DN user account comes back first (with total_balance); this is often cached in local storage so we're potentially reading stale values.
- The undisbursed endpoint comes second
- The wallet slice comes third, which comes from chain & is the most 'correct' AUDIO balance (though it agrees with DN most of the time!)

I probably misled you on slack convos and should have thought more deeply about this beforehand, but I think this is what we should do:

- Always have some space for that row, so the nav bar never gets pushed down
- Show the AUDIO amount as soon as we can; keeping the prior logic preferring the chain value, but showing the value from DN beforehand if we have it.
- Only show the earn/claim pill when we know what state it will be in initially (i.e. wait until the undisbursed challenges endpoint loads). After that first load, anything is game (you complete a challenge, it switches from earn -> claim, etc).

packages/web/src/components/nav/desktop/NavAudio.tsx
```diff
@@ -33,13 +34,41 @@ const messages = {
   claimRewards: 'Claim Rewards'
 }
 
-const NavAudio = () => {
+type RewardsActionBubbleProps = {
+  bubbleType: BubbleType
+  onClick(): void
```
sddioulde, line 39: nit: interesting, was checking where else we use this pattern and I see one other occurrence in the client code. no biggie ofc but we've been doing `onClick: () => void`. maybe this is another client styleguide thing we should have...

sddioulde:
> Code wise, this looks great! Love the cleanups, explicit loading states, separation of component, and removing the superfluous state + useEffect.
> 
> I have some thoughts about delaying the load for the audio amount + earn/claim badge:
> 
> * As it stands in this PR, when the nav row renders in the entire nav column gets pushed down, which I find quite visually jarring and maybe more attention grabbing than the strobing we have today.
> * In terms of an 80/20 happy path, most users probably don't have challenges that are waiting to be disbursed, though they may have a few AUDIO sitting around. So for that class of users, this change is delaying showing them information they used to see immediately (their AUDIO balance - I tested this against my prod account). I added some logs and at least in my case, I see this in terms of response ordering:
> * The DN user account comes back first (with total_balance); this is often cached in local storage so we're potentially reading stale values.
> * The undisbursed endpoint comes second
> * The wallet slice comes third, which comes from chain & is the most 'correct' AUDIO balance (though it agrees with DN most of the time!)
> 
> I probably misled you on slack convos and should have thought more deeply about this beforehand, but I think this is what we should do:
> 
> * Always have some space for that row, so the nav bar never gets pushed down
> * Show the AUDIO amount as soon as we can; keeping the prior logic preferring the chain value, but showing the value from DN beforehand if we have it.
> * Only show the earn/claim pill when we know what state it will be in initially (i.e. wait until the undisbursed challenges endpoint loads). After that first load, anything is game (you complete a challenge, it switches from earn -> claim, etc).

Can confirm that the very first implementation attempt of this also had similar logic that pushed down the rest of the nav items and I remember we wanted to avoid that. So yes, we should always have the space there.

sddioulde (commented):
looking v nice overall! waiting on some of the comments being addressed to approve :)

schottra:
> Code wise, this looks great! Love the cleanups, explicit loading states, separation of component, and removing the superfluous state + useEffect.
> 
> I have some thoughts about delaying the load for the audio amount + earn/claim badge:
> 
> * As it stands in this PR, when the nav row renders in the entire nav column gets pushed down, which I find quite visually jarring and maybe more attention grabbing than the strobing we have today.
> * In terms of an 80/20 happy path, most users probably don't have challenges that are waiting to be disbursed, though they may have a few AUDIO sitting around. So for that class of users, this change is delaying showing them information they used to see immediately (their AUDIO balance - I tested this against my prod account). I added some logs and at least in my case, I see this in terms of response ordering:
> * The DN user account comes back first (with total_balance); this is often cached in local storage so we're potentially reading stale values.
> * The undisbursed endpoint comes second
> * The wallet slice comes third, which comes from chain & is the most 'correct' AUDIO balance (though it agrees with DN most of the time!)
> 
> I probably misled you on slack convos and should have thought more deeply about this beforehand, but I think this is what we should do:
> 
> * Always have some space for that row, so the nav bar never gets pushed down
> * Show the AUDIO amount as soon as we can; keeping the prior logic preferring the chain value, but showing the value from DN beforehand if we have it.
> * Only show the earn/claim pill when we know what state it will be in initially (i.e. wait until the undisbursed challenges endpoint loads). After that first load, anything is game (you complete a challenge, it switches from earn -> claim, etc).

Comments make sense! I also don't love the jumping. It's easy enough to solve. And if you're okay with the audio balance quickly changing after the wallet loads, I can put that logic back in. There is one slight issue to get around due to the fact that the existing logic will prefer the DN balance when the wallet balance is falsey. This includes a balance of zero. The state itself is initialized to `null`, but the selectors above it normalize it to `0` before it's loaded. And those are used in multiple places, so I don't feel comfortable updating it back to a `null` value. I can get around that by relying on the new loading state variable and always preferring the wallet value after the initial load succeeds. I just wanted to point out that difference in logic. If I _don't_ fix that, there are a few edge cases where we show the wrong balance and/or don't show the "Earn $AUDIO" pill

packages/web/src/components/nav/desktop/NavAudio.tsx
schottra: I believe they both define the same type (a function returning nothing), so it's just aesthetic preference. But I can see the benefit of consistency across the codebase. Another way I've seen it done is to create an explicit type for the callback:
```
type ClickEventHandler = () => void

type RewardsActionBubbleProps = {
  onClick: ClickEventHandler
}
```

That has the benefit of being able to define the type in one place and use it in several, with the linter/transpiler complaining at you when you change the type and forget to update all the places that use it.

schottra:
@piazzatron @sddioulde I've updated the logic to work as follows:
* audio balance will prefer wallet once loaded, fall back to account if available, and return 0 otherwise
* Bubble will show claimable rewards as soon as challenges have loaded. If there are no claimable rewards, bubble will show the "earn" version once the wallet has loaded and we see a non-positive balance.

Here's what happens if the wallet loads and reflects a balance of zero (even if the balance from the account record originally showed positive), when there are no claimable rewards
Kapture 2023-04-14 at 10 47 59

Here's a case where there are claimable challenges and the wallet loads and updates the balance to 0.
Kapture 2023-04-14 at 11 02 11

And then here's what happens if the challenges take longer to load than the wallet and the balance in the wallet is zero (the original source of the flashing bug)
Kapture 2023-04-14 at 11 11 24

Please give it another look!

audius-infra:
Preview this change https://demo.audius.co/fix-claim-button-flickering

piazzatron (approved):
Looks great!! Such an improvement :D

dylanjeffers (approved):
looks great!

audius-infra:
Preview this change https://demo.audius.co/fix-claim-button-flickering

PR #3225, opened 2023-04-14
[C-2429] Remove hls

dylanjeffers:
### Description

Removes hls now that we exclusively use mp3
- Tested on tracks/albums/offline/collectible-playlists
- Tested on web and mobile


audius-infra:

audius-infra:
Preview this change https://demo.audius.co/dj-c-2429-remove-hls

packages/common/src/utils/hls.test.js
```diff
@@ -1,112 +0,0 @@
-import { generateM3U8, generateM3U8Variants } from './hls'
```
raymondjacobson, line 1: rip. we should have more tests

packages/web/scripts/publishScripts.sh
```diff
@@ -9,13 +9,11 @@ make -C "./node_modules/exif-parser" build-browser-bundle
 declare -a modules=(
   "./node_modules/jimp/browser/lib/jimp.min.js"
   "./node_modules/exif-parser/dist/exif-parser-0.1.12-min.js"
-  "./node_modules/hls.js/dist/hls.min.js"
```
raymondjacobson, line 12: i guess this was no longer used

raymondjacobson (approved):

PR #3234, opened 2023-04-18
[PAY-937] Fix empty state for transactions table

schottra:
### Description
The `isLoading` logic for the AudioTransactionsTable was previous relying on an `every` check against the loaded data, instead of tracking success of the fetch. Due to the way `Array.prototype.every()` works, an empty array will always cause the loading condition to be true.

To fix this, I added loading state tracking to the slice for both transactions and count, and then wait on them explicitly to determine table loading state.
Additionally, I moved the reset logic mostly into the slice to simplify the component. The slice will now reset the items array when a fetch begins and the offset is zero (fetching the first page). This means the component doesn't have to manually clear items when paginating, and gets us closer to a pagination functionality that is separate from the display logic.

I also moved most of the helper functions in `AudioTransactionsTable` out to constants. They weren't referencing any props, so they don't need to be defined in the component with `useCallback`.

Previous behavior:
Kapture 2023-04-18 at 00 00 05

New behavior:
Kapture 2023-04-18 at 00 02 11

### Dragons
Nope

### How Has This Been Tested?
Tested manually.

Verify functionality in staging and RC

### Feature Flags ###
N/A

### DIFF
--- a/packages/common/src/models/Status.ts
+++ b/packages/common/src/models/Status.ts
@@ -4,3 +4,8 @@ export enum Status {
   SUCCESS = 'SUCCESS',
   ERROR = 'ERROR'
 }
+
+/** Detects if a status is in a non-terminal state */
+export function statusIsUnloaded(status: Status) {
+  return [Status.IDLE, Status.LOADING].includes(status)
+}
--- a/packages/common/src/models/Status.ts
+++ b/packages/common/src/models/Status.ts
@@ -6,6 +6,6 @@ export enum Status {
 }
 
 /** Detects if a status is in a non-terminal state */
-export function statusIsUnloaded(status: Status) {
+export function statusIsNotTerminal(status: Status) {
   return [Status.IDLE, Status.LOADING].includes(status)
 }
--- a/packages/common/src/models/Status.ts
+++ b/packages/common/src/models/Status.ts
@@ -6,6 +6,6 @@ export enum Status {
 }
 
 /** Detects if a status is in a non-terminal state */
-export function statusIsNotTerminal(status: Status) {
+export function statusIsNotFinalized(status: Status) {
   return [Status.IDLE, Status.LOADING].includes(status)
 }
--- a/packages/web/src/components/audio-transactions-table/AudioTransactionsTable.tsx
+++ b/packages/web/src/components/audio-transactions-table/AudioTransactionsTable.tsx
@@ -74,6 +74,114 @@ export const isChangePositive = (tx: TransactionDetails) => {
   )
 }
 
+// Cell Render Functions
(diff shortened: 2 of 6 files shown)


audius-infra:
Preview this change https://demo.audius.co/pay-937-transactions-table-empty-state

packages/web/src/components/audio-transactions-table/AudioTransactionsTable.tsx
```diff
@@ -74,6 +74,114 @@ export const isChangePositive = (tx: TransactionDetails) => {
   )
 }
 
+// Cell Render Functions
```
dharit-tan, line 77: Love it!

dharit-tan (approved):
Awesome, thanks for cleaning this up!!

packages/common/src/models/Status.ts
```diff
@@ -4,3 +4,8 @@ export enum Status {
   SUCCESS = 'SUCCESS',
   ERROR = 'ERROR'
 }
+
+/** Detects if a status is in a non-terminal state */
```
dharit-tan, line 8: I like this, should circulate this in #eng-client. V minor nit with the name, is "unloaded" a conventional term? To me it sounds more like something was loaded and then removed. Maybe "statusIsNotLoaded" but not a strong opinion.

schottra: I sat here for a good five minutes trying to name this because:
* I don't like things that have to be inverted to be useful (ex. `const isLoading = !isLoadedStatus()`)
* I was trying not to be to obscure with the terminology (`statusIsNonTerminal`)
Your suggestion is a good one!

schottra: Oh wait, I also wanted it to indicate that `failed` is considered a loaded status. So I might go with NonTerminal in the end

piazzatron: nit: `statusIsNotFinalized` or `statusIsPending`? pending sounds a bit like loading though...

piazzatron (approved):
Looks great!! Nice refactors. 

Let's check in with Sammie and see whether we want the linked-wallets explainer tile to stay visible when we have no transactions (I suspect we probably do).

audius-infra:
Preview this change https://demo.audius.co/pay-937-transactions-table-empty-state

schottra:
@piazzatron Checked with design and updated it to keep the tile when the table is empty

Kapture 2023-04-18 at 13 57 43

audius-infra:
Preview this change https://demo.audius.co/pay-937-transactions-table-empty-state

audius-infra:
Preview this change https://demo.audius.co/pay-937-transactions-table-empty-state

PR #3236, opened 2023-04-18
Fix dev-mode hot-key modals

dylanjeffers:
### Description

Fixes issue where dev-mode hot-key modals didn't register their hotkey hook since they were not rendered. Fixed by reserving a space for them in the Modals component


sliptype (approved):
Nice! Thank you

PR #3242, opened 2023-04-19
PLAT-725: browser push misconfigured

alecsavvy:
### Description
It looks like the subscribe function was misspelled as `notifcation` instead of `notification`. My hunch is other files referencing this name correctly were just returning undefined.

### Dragons
Spellcheckers?

### How Has This Been Tested?
Fixed the typo and did a repository search for both the old and correct name. Corrected all references to match the right one.

Web push should correctly subscribe now because the func ref now matches up.

### Feature Flags ###
None


dylanjeffers (approved):
Wonderful

audius-infra:
Preview this change https://demo.audius.co/as/plat-725-browser-push-register

PR #3244, opened 2023-04-19
Fix React errors about `key` property when rendering tables

schottra:
### Description
When rendering the table headers, we get errors about children not having keys.
image

I looked into this, and the rows in our table headers were not being assigned keys. This is due to a difference in behavior between react-table v7 and v8+. The examples demonstrate using `headerGroup.id` as the key property. However, this property is `undefined` in the version we are using. The examples for `v7` pull the key from the `headerGroupProps` instead.

The linter isn't able to determine that the `headerGroupProps` always has a `key` value and the spread should provide it. So I updated the logic to pull the key out individually and assign it explicitly.

### Dragons
None

### How Has This Been Tested?
Manual verification locally

Will verify no table regressions in staging

### Feature Flags ###
N/A

### DIFF
--- a/packages/web/src/components/table/Table.tsx
+++ b/packages/web/src/components/table/Table.tsx
@@ -277,11 +277,13 @@ export const Table = ({
         (header) => header.id === 'trackActions' || header.id === 'overflowMenu'
       )
 
+      const { key: headerGroupKey, ...headerGroupProps } =
+        headerGroup.getHeaderGroupProps()
       return (
         <tr
           className={styles.tableHeadRow}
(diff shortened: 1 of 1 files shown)


dylanjeffers (approved):
Hell yeah great find

sliptype (approved):
Awesome! I would love for us to do a pass of all the console errors that get thrown in dev, there are a lot 😅

Kyle-Shanks (approved):
Nice 🔥

schottra:
CI test seems to be flaky and unrelated, going to merge and we will figure out the test failure separately

PR #3245, opened 2023-04-19
[C-574] Add migration path to @react-spring/web

dylanjeffers:
### Description

- Adds @react-spring/web (v9 of react-spring), as a step towards migrating off react-spring (v8)
- Implements change on one component: NowPlayingArtworkTile
- Adds eslint-rule to ensure no new components are implemented with v8
- Adds eslint-ignores to all current instances of v8, with comment to migrate

### DIFF
--- a/packages/web/.eslintrc.js
+++ b/packages/web/.eslintrc.js
@@ -38,5 +38,18 @@ module.exports = {
         extensions: ['.ts', '.tsx', '.js', '.jsx', '.json']
       }
     }
+  },
+  rules: {
+    'no-restricted-imports': [
+      'error',
+      {
+        paths: [
+          {
+            name: 'react-spring',
+            message: 'Please use @react-spring/web instead'
+          }
+        ]
+      }
+    ]
   }
 }
--- a/packages/web/.eslintrc.js
+++ b/packages/web/.eslintrc.js
@@ -43,9 +43,9 @@ module.exports = {
     'no-restricted-imports': [
       'error',
       {
-        paths: [
+        patterns: [
           {
-            name: 'react-spring',
+            group: ['react-spring*'],
             message: 'Please use @react-spring/web instead'
           }
         ]
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -75,7 +75,7 @@
     "@opentelemetry/semantic-conventions": "1.6.0",
     "@optimizely/optimizely-sdk": "4.9.2",
     "@project-serum/sol-wallet-adapter": "0.2.5",
-    "@react-spring/web": "^9.7.2",
+    "@react-spring/web": "9.7.2",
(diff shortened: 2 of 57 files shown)


packages/web/package.json
```diff
@@ -75,6 +75,7 @@
     "@opentelemetry/semantic-conventions": "1.6.0",
     "@optimizely/optimizely-sdk": "4.9.2",
     "@project-serum/sol-wallet-adapter": "0.2.5",
+    "@react-spring/web": "^9.7.2",
```
sliptype, line 78: 📍

sliptype (approved):
😎

packages/web/package.json
dylanjeffers: strikes again

audius-infra:
Preview this change https://demo.audius.co/d-c-574-upgrade-react-spring

packages/web/src/components/nav/desktop/NowPlayingArtworkTile.tsx
```diff
@@ -9,9 +9,9 @@ import {
   averageColorSelectors
 } from '@audius/common'
 import { IconButton } from '@audius/stems'
+import { animated, useSpring } from '@react-spring/web'
```
schottra, line 12: I know the APIs changed considerably between versions. I'm assuming you audited the usage of `useSpring` here to make sure we are configuring things correctly based on the current docs with respect to what the old docs say?
(I think you have to use internet archive to get a reliable version of the old docs 😬 )

schottra (approved):
This is awesome. Thanks for updating it! And also the lint rule is a great idea.
I'm surprised there were no issues running these two versions side-by-side?

packages/web/.eslintrc.js
```diff
@@ -38,5 +38,18 @@ module.exports = {
...
       }
     }
+  },
+  rules: {
+    'no-restricted-imports': [
+      'error',
+      {
+        paths: [
+          {
+            name: 'react-spring',
```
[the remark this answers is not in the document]
dylanjeffers: ohh great call, yeah let me do an audit of all the usages, there is a "patterns" option for this rule we can maybe use

packages/web/src/components/nav/desktop/NowPlayingArtworkTile.tsx
dylanjeffers: yeah exactly, it was the one component i knew before/after, and referencing the docs and its the same api for this one, since its config object.

packages/web/.eslintrc.js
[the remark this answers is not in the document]
dylanjeffers: @schottra updated to pattern if you're curious, also added to stems

PR #3247, opened 2023-04-19
[C-2493] separate state for suggested follows vs related artists

amendelsohn:
### Description

Artists you already follow were being filtered out from the related artists preview. This stores a separate state for each type so when suggestedFollows falls back to showing topArtists instead of related artists, we will still show the related artists.

### DIFF
--- a/packages/common/src/store/ui/related-artists/sagas.ts
+++ b/packages/common/src/store/ui/related-artists/sagas.ts
@@ -32,10 +32,13 @@ export function* fetchRelatedArtists(action: PayloadAction<{ artistId: ID }>) {
     )
 
     let showingTopArtists = false
-    let filteredArtists = relatedArtists
-      .filter((user) => !user.does_current_user_follow && !user.is_deactivated)
+    const filteredArtists = relatedArtists.filter(
+      (user) => !user.is_deactivated
+    )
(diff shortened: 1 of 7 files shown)


audius-infra:
Preview this change https://demo.audius.co/amendel-related-artists-follow-filter

packages/common/src/store/ui/related-artists/sagas.ts
```diff
@@ -32,27 +32,32 @@ export function* fetchRelatedArtists(action: PayloadAction<{ artistId: ID }>) {
...
+    if (suggestedFollows.length === 0) {
       const showTopArtistRecommendationsPercent =
         remoteConfigInstance.getRemoteVar(
           DoubleKeys.SHOW_ARTIST_RECOMMENDATIONS_FALLBACK_PERCENT
         ) || 0
       const showTopArtists = Math.random() < showTopArtistRecommendationsPercent
 
       if (showTopArtists) {
-        filteredArtists = yield fetchTopArtists()
+        suggestedFollows = yield fetchTopArtists()
```
dylanjeffers, line 49: do we want yield* call ?

dylanjeffers (approved):
word, nice fix with few changes

Kyle-Shanks (approved):

PR #3250, opened 2023-04-19
[C-2501] Use proxy-wormhole libs

dylanjeffers:
### Description

Updates web client to use proxy-wormhole libs, shoutout @nicoback for enabling this. This results in 30% smaller overall bundle, and a 20% higher (49 to 60) lighthouse score on trending!

- Re-enables our analyzeBundle command
- Adds "proxy-wormhole" feature flag that allows us to switch back between performant an non-performant libs depending on any errors we figure out
- Also merges tons of duplicate libraries, moment, lodash, @solana/web3.js, and bn.js, resulting in many more savings across the board.

### DIFF
--- a/packages/web/src/services/audius-backend/getLibs.ts
+++ b/packages/web/src/services/audius-backend/getLibs.ts
@@ -1,0 +1,11 @@
+import { FeatureFlags, RemoteConfigInstance } from '@audius/common'
+
+export const getLibs = async (remoteConfig: RemoteConfigInstance) => {
+  await remoteConfig.waitForRemoteConfig()
+  const isProxyWormholeEnabled = remoteConfig.getFeatureEnabled(
+    FeatureFlags.PROXY_WORMHOLE
+  )
+  return isProxyWormholeEnabled
+    ? import('@audius/sdk/dist/web-libs')
+    : import('@audius/sdk/dist/legacy')
+}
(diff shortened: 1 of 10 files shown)


audius-infra:

packages/web/src/services/audius-backend/getLibs.ts
```diff
@@ -0,0 +1,11 @@
+import { FeatureFlags, RemoteConfigInstance } from '@audius/common'
+
+export const getLibs = async (remoteConfig: RemoteConfigInstance) => {
+  await remoteConfig.waitForRemoteConfig()
+  const isProxyWormholeEnabled = remoteConfig.getFeatureEnabled(
+    FeatureFlags.PROXY_WORMHOLE
+  )
+  return isProxyWormholeEnabled
```
sliptype, line 8: Could this slow down page load because we are waiting for a feature flag to start initializing libs?

sliptype (approved):
Huge find!!

packages/web/src/services/audius-backend/getLibs.ts
dylanjeffers: since our read flow uses eager-load i think that is okay, just write flow will take slightly longer to initialize. our sdk currently waits for remote-config too, btw

PR #3252, opened 2023-04-20
PLAT-725: fix browser push

alecsavvy:
### Description
The saga for subscribing to web push was never registered in the root saga. The toggle for the settings was also not reversing the flag correctly and before was eternally on deactivated.

### Dragons
Peruvian Vipertooths are pretty rad

### How Has This Been Tested?
Built against stage and was able to successfully register my web app for notifs. Confirmed in the identity db that my records were there.

The notifications plugin will be attempting to push to these endpoints. 

### Feature Flags ###
None


packages/web/src/components/switch/Switch.tsx
```diff
@@ -23,6 +23,7 @@ const Switch = ({
         checked={allowCheckedWhileDisabled ? isOn : isDisabled ? false : isOn}
         className={styles.inputCheckbox}
         type='checkbox'
+        onChange={_ => {}}
```
alecsavvy, line 26: this fixes a react error

packages/web/src/store/sagas.ts
```diff
@@ -86,6 +86,7 @@ import reachabilitySagas from 'store/reachability/sagas'
 import routingSagas from 'store/routing/sagas'
 import signOutSagas from 'store/sign-out/sagas'
 import tokenDashboardSagas from 'store/token-dashboard/sagas'
+import uncommonAccountSagas from 'store/account/sagas'
```
dylanjeffers, line 89: ah, so looks like store/account/sagas merges commonSagas with it, so can just rename this to `accountSagas` and remove the `accountSagas` from common in this file?

packages/web/src/components/switch/Switch.tsx
dylanjeffers: ah thank you for uncovering this bug! can we actually undo this change, and instead play a story to properly untangle the input/label onchange vs onclick issue? i just added it to my todo for tomorrow so ill add you to that pr?

dylanjeffers (approved):
Blues eyes white dragon are also pretty rad!

michellebrier (approved):

PR #3253, opened 2023-04-20
C-2502: extend store history

alecsavvy:
### Description
extends the redux history to 1000 actions before they start to fall off

### Dragons
None, dev tool

### How Has This Been Tested?
My redux dev tools has tons of actions now!

N/A

### Feature Flags ###
N/A


dylanjeffers (approved):
This is will make web dev so much nicer, thank you

PR #3254, opened 2023-04-20
Fix react-router issues

dylanjeffers:
### Description

Fixes issue where multiple instances of react-router were present, this also improves perf!


schottra:
Oops 😅

schottra (approved):

PR #3256, opened 2023-04-20
[PAY-1162] Update rendering overflow for some "Supporting" lists

schottra:
### Description
The overflow count that renders on the last icon of a `ProfilePictureList` will overflow the container when it's more than one digit. We are planning a more in-depth overhaul of the way we render lists of profile pictures with the "view all" link. But for now, this will address the two cases in the native mobile Profile page.
* The last of the Supporting tiles will now render a fixed set of up to six, with a "View All X" link underneath
* The Top Supporters picture list will render with a smaller font size to make it _less_ likely to overflow (though this is still possible with larger counts)

Also did some cleanup:
* Moved some repeated code for computing the ranked supporting list into a shared hook.
* Updated a constant for fetching the top X supporting to ensure we get enough to render 6 in the SupportingTiles section. Since we show the top 3 individually, we need 9 total.
* Updated the text for the Supporting section on a web profile page to use the same formatting (View All X)
* Adjusted dimming layer behind remaining user count to be slightly darker, for more contrast with the text.
* Change z-indexes for ProfilePictures when we aren't showing the overflow count above the last image

image

image

### Dragons
None

### How Has This Been Tested?
Manual testing on web/mobile

Manual verification before release

### Feature Flags ###
N/A


packages/common/src/hooks/useRankedSupporters.ts
```diff
@@ -0,0 +1,36 @@
+import { ID } from '../models/Identifiers'
+import { tippingSelectors } from '../store/tipping'
+import { stringWeiToBN } from '../utils'
+
+import { useProxySelector } from './useProxySelector'
+
+const { getOptimisticSupportingForUser } = tippingSelectors
+
+/**
```
dylanjeffers, line 9: boss

packages/mobile/src/screens/profile-screen/ProfileHeader/ViewAllSupportingTile.tsx
```diff
@@ -51,31 +53,29 @@ export const ViewAllSupportingTile = () => {
...
-      return amount1BN.gte(amount2BN) ? -1 : 1
-    })
-    .map((k) => supportingForProfile[k as unknown as ID])
-    .map((s) => s.receiver_id)
-  const rankedSupporting = useSelector((state) => {
-    const usersMap = getUsers(state, { ids: rankedSupportingIds })
-    return rankedSupportingIds.map((id) => usersMap[id]).filter(Boolean)
-  })
+
+  const rankedSupportingList = useRankedSupportingForUser(user_id)
```
dylanjeffers, line 57: would it make sense to merge useRankedSupportingForUser with the other proxySelector, so its just one?

packages/mobile/src/screens/profile-screen/ProfileHeader/ViewAllSupportingTile.tsx
```diff
@@ -51,31 +53,29 @@ export const ViewAllSupportingTile = () => {
...
+      return rankedIds.map((id) => usersMap[id]).filter(Boolean)
+    },
+    [rankedSupportingList]
+  )
 
   const handlePress = useCallback(() => {
     navigation.push('SupportingUsers', { userId: user_id })
   }, [navigation, user_id])
 
+  const viewAllString = useMemo(
```
dylanjeffers, line 74: i think this can just be a string since string equality is sufficient

dylanjeffers (approved):
really awesome changes!

packages/mobile/src/screens/profile-screen/ProfileHeader/ViewAllSupportingTile.tsx
sddioulde: was wondering the same

sddioulde (approved):
tight!

packages/mobile/src/screens/profile-screen/ProfileHeader/ViewAllSupportingTile.tsx
schottra: The reason I didn't merge them is that the components consuming the `Supporting[]` lists handle them differently. Calculating the ranked list of `Supporting[]` objects is where the common code path ends. Though we could probably update them all to just accept a list of `User` objects as calculated here.

packages/mobile/src/screens/profile-screen/ProfileHeader/ViewAllSupportingTile.tsx
schottra: Oh yeah, you're right. I think it bothered me to do the formatting logic on every render since the number doesn't change often. But this component shouldn't re-render unless the supporing list changes anyway, so that's probably not a useful optimization.

audius-infra:
Preview this change https://demo.audius.co/pay-1162-update-supporters-list

packages/mobile/src/screens/profile-screen/ProfileHeader/ViewAllSupportingTile.tsx
dylanjeffers: Gotcha, makes sense that there isn't common code, im wondering if having a single selector would just be more performant for this one? Ultimately I think this is minor but though

schottra: If it helps, we limit the list that's being sorted to 9 items...something I didn't learn until right before I opened this PR 😆

PR #3258, opened 2023-04-21
[C-2514] Fix empty related artists when logged out (web)

amendelsohn:
### Description

image

### DIFF
--- a/packages/common/src/services/audius-api-client/AudiusAPIClient.ts
+++ b/packages/common/src/services/audius-api-client/AudiusAPIClient.ts
@@ -250,10 +250,9 @@ type GetUserTracksByHandleArgs = {
   getUnlisted: boolean
 }
 
-type GetRelatedArtistsArgs = CurrentUserIdArg &
-  PaginationArgs & {
-    userId: ID
-  }
+type GetRelatedArtistsArgs = PaginationArgs & {
+  userId: ID
+}
 
 type GetFavoritesArgs = {
   currentUserId: ID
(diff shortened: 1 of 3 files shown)


packages/common/src/services/audius-api-client/AudiusAPIClient.ts
```diff
@@ -1160,18 +1159,12 @@ export class AudiusAPIClient {
...
-    offset,
-    limit
-  }: GetRelatedArtistsArgs) {
+  async getRelatedArtists({ userId, offset, limit }: GetRelatedArtistsArgs) {
     this._assertInitialized()
-    const encodedCurrentUserId = encodeHashId(currentUserId)
     const encodedUserId = this._encodeOrThrow(userId)
     const response = await this._getResponse<APIResponse<APIUser[]>>(
       FULL_ENDPOINT_MAP.getRelatedArtists(encodedUserId),
-      { user_id: encodedCurrentUserId || undefined, offset, limit }
```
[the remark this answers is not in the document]
amendelsohn: There's a user_id arg in the request path (here we pass `encodedUserId`) but this one is not used. I think I just copied it from somewhere by accident.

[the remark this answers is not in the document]
Kyle-Shanks: interesting that sending bad info would make this not work if it's not used, but good that this is fixed 👌

Kyle-Shanks (approved):

packages/web/src/common/store/user-list/related-artists/sagas.ts
```diff
@@ -58,12 +53,10 @@ function* errorDispatcher(error: Error) {
 
 function* getRelatedArtists(currentPage: number, pageSize: number) {
   const id = yield* select(getId)
-  const currentUserId = yield* select(getUserId)
-  if (!id || !currentUserId) return { userIds: [], hasMore: false }
+  if (!id) return { userIds: [], hasMore: false }
```
amendelsohn, line 56: @Kyle-Shanks the functional fix was actually here, I just removed the user_id arg there because it was unused

amendelsohn: we were checking for `currentUserId` here where we shouldn't since you don't need to be logged in to get related artists

Kyle-Shanks: ah I see. nice 👍

audius-infra:
Preview this change https://demo.audius.co/amendel-related-artists-empty-fix

PR #3259, opened 2023-04-21
[C-2516] Always filter followed users from suggested

amendelsohn:
### Description

Recently followed users were still showing up in suggested follows. This ensures they won't make it to the suggestion box.


dylanjeffers (approved):
lgtm

PR #3260, opened 2023-04-21
[C-2511] Don't delay mutuals fetch

amendelsohn:
### Description

- mutuals were loading in well after top supporters and related artists, pushing the others down the page
- fix removes the delay loading mutual follows

### Dragons

There must have been a reason we did it this way in the past.
Maybe perf concerns with this happening during initial loading.

We could defer all of the fetches if that's more correct.

### DIFF
--- a/packages/web/src/common/store/profile/sagas.js
+++ b/packages/web/src/common/store/profile/sagas.js
@@ -353,9 +353,6 @@ function* fetchProfileAsync(action) {
     }
 
     if (!isNativeMobile) {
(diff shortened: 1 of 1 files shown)


amendelsohn:
I'm wondering if there's a better way to do this where we select all the content for the boxes one level up and don't show any until they're all done loading. This has the downside of allowing one of them stuck in loading to break them all.

dylanjeffers (approved):
awesome find

PR #3261, opened 2023-04-21
[C-2498] Update mobile verification flow to use the account user handle for verification

Kyle-Shanks:
### Description
An update to the mobile verification flow to pass along the signed in user's handle as opposed to the handle that we received from the authed site. Prevents errors where the handles don't match and we verify the wrong account on our side

### Dragons

An update to the mobile verification system that has not be tested thoroughly. Potential for bugs here

### How Has This Been Tested?

Not tested. Will try to figure out a good testing method for this

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/mobile/src/screens/settings-screen/AccountVerificationScreen.tsx
+++ b/packages/mobile/src/screens/settings-screen/AccountVerificationScreen.tsx
@@ -136,8 +136,8 @@ export const AccountVerificationScreen = () => {
 
   const handleField: EditableField = useSelector(getHandleField)
 
-  const name = accountUser?.name
-  const handle = accountUser?.handle
+  const accountName = accountUser?.name
+  const accountHandle = accountUser?.handle
 
   const onVerifyButtonPress = useCallback(() => {
     setStatus(Status.LOADING)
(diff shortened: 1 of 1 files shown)


packages/mobile/src/screens/settings-screen/AccountVerificationScreen.tsx
```diff
@@ -157,15 +157,16 @@ export const AccountVerificationScreen = () => {
       const handle = type === 'twitter' ? profile.screen_name : profile.username
       const verified =
         type === 'twitter' ? profile.verified : profile.is_verified
-      dispatch(signOnActions.validateHandle(handle, verified))
+
+      dispatch(signOnActions.validateHandle(accountHandle ?? handle, verified))
```
amendelsohn, line 161: is it safe to fall back to handle?

amendelsohn (approved):

packages/mobile/src/screens/settings-screen/AccountVerificationScreen.tsx
Kyle-Shanks: I don't think that we should be able to get here without the user being logged it so I don't think we would ever fallback, but it seemed safe enough bc that's what we have been using until now

PR #3262, opened 2023-04-21
[PAY-1122] Mobile chats keyboard improvements

dharit-tan:
### Description
Custom implementation of `KeyboardAvoidingView`.

- Fixes issue where keyboard would render over textinput at bottom of screen (plz lmk if you still see this issue).
- Add white background behind/under textinput so it looks less jank during the animation.
- Allows customization of in/out animation durations.

Side note: I tried react reanimated animations and the react reanimated `useAnimatedKeyboard` but the results all felt the same as this one.

### Dragons

Not tested on android yet, I'm sure this will be buggy but we can always fall back to the RN `KeyboardAvoidingView` which worked fine (no animations, just snapping in and out but that is the case for other messaging apps on android).

Hardcoded `heightOffsetRatio` to default to 0.75, just anecdotally this works but need to test on other screen sizes. On all the web examples I found, others also had to hardcode some value here because the keyboard height returned from the event never actually lined up with reality.

### How Has This Been Tested?

Local ios simulator

https://user-images.githubusercontent.com/3893871/233731340-f5ea76d6-a148-4102-a132-fe09dcbd6811.mp4

### Feature Flags ###


dylanjeffers (approved):
Really clean implementation!!! So surprised the normal keyboard-avoiding-view doesn't do this, so great that we have options. interested to learn more about this part of react-native at some point too, and stoked you were able to figure out a better way!

piazzatron:
As @dylanjeffers said, nice and clean! Very curious to understand why the normal RN keyboard avoiding view doesn't work properly, but this looks like a nice fix :)

PR #3264, opened 2023-04-24
Notification settings improvements

michellebrier:
### Description
- Only request permissions and overwrite settings with `initialState` on sign in if the user doesn't have permissions enabled.
- Write to identity `UserNotificationMobileSettings` to set all relevant subfields to false when turning off all notifications. I know we deregister their device if they toggle notifs off which effectively disables notification but would prefer the data in the settings table be accurate. Did the same for browser. Can change back if you guys strongly dislike this
- Misc stuff like add messages to the enable notifications reminder

Let me know your thoughts!

### How Has This Been Tested?

- Make new account. Sign in on mobile. Accept push notification settings. Verify in identity that all types of push notifs are enabled
- Toggle mobile push notif settings all on / off / one at a time. Make sure all changes persisted correctly. Leave some toggled off and sign out
- Sign in again. Make sure settings are the same and have not been overridden
- Same for browser


packages/mobile/src/notifications.ts
```diff
@@ -89,17 +89,15 @@ class PushNotifications {
...
+  async hasPermission(): Promise<boolean> {
+    return await Notifications.isRegisteredForRemoteNotifications()
+  }
+
   requestPermission() {
     isRegistering = true
     Notifications.registerRemoteNotifications()
   }
 
-  async checkPermission(
```
michellebrier, line 97: this doesnt seem to be called anywhere and does the same thing as isRegisteredForRemoteNotifications but ios-specific

packages/mobile/src/store/settings/sagas.ts
```diff
@@ -18,13 +18,51 @@ import PushNotifications from 'app/notifications'
...
+
+  yield* call(audiusBackendInstance.registerDeviceToken, token, os)
+}
+
+function* disablePushNotifications() {
+  const audiusBackendInstance = yield* getContext('audiusBackendInstance')
+  const newSettings = { ...initialState.pushNotifications }
+  for (const key in newSettings) {
+    newSettings[key] = false
+  }
```
sliptype, line 51: We could use lodash `mapValues` for this!

```suggestion
  const newSettings = mapValues(initialState.pushNotifications, false)
```

packages/web/src/common/store/pages/signon/sagas.js
```diff
@@ -531,12 +532,9 @@ function* signIn(action) {
...
       if (isNativeMobile) {
-        yield put(
-          togglePushNotificationSetting(
-            PushNotificationSetting.MobilePush,
-            true
-          )
-        )
+        // If permissions not already enabled, request permission to send push notifications
+        // and enable all if accepted
+        yield put(requestPushNotificationPermissions())
```
sliptype, line 537: Nice!

packages/mobile/src/notifications.ts
sliptype: Good catch! Looks fine to delete for sure

sliptype (approved):
Love it, very clean! Just one callout about a small potential improvement with the mapping of settings values

packages/mobile/src/store/settings/sagas.ts
michellebrier: amazing, will add

PR #3270, opened 2023-04-24
[PLAT-903] Fix missing remix notifications

dylanjeffers:
### Description

Fixes issue where remix notifications relied on legacy selectors for old notification data shape. Now it's simply selecting tracks based on parent/child id, rather that picking tracks out of a list.

### DIFF
--- a/packages/mobile/src/screens/notifications-screen/Notification/EntityLink.tsx
+++ b/packages/mobile/src/screens/notifications-screen/Notification/EntityLink.tsx
@@ -1,12 +1,12 @@
 import { useCallback } from 'react'
 
-import type { EntityType } from '@audius/common'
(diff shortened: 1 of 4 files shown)


packages/mobile/src/screens/notifications-screen/Notification/EntityLink.tsx
```diff
@@ -19,7 +19,7 @@ export const EntityLink = (props: EntityLinkProps) => {
         id: entity.track_id,
         fromNotifications: true
       })
-    } else if (entity.user) {
```
raymondjacobson, line 22: wow nice

raymondjacobson (approved):
Awesome!! ty

PR #3271, opened 2023-04-25
[PAY-1106] Unmount reaction popup animations

dharit-tan:
### Description
Unmount animations for reactions popup (from convos with @dylanjeffers). Using a callback passed to `Animated.timing.start()` that sets the `shouldShowPopup` state to false when the animation completes.

### Dragons

### How Has This Been Tested?

Local ios stage

### Feature Flags ###


rickyrombo (approved):

packages/mobile/src/screens/chat-screen/ReactionPopup.tsx
```diff
@@ -95,24 +99,47 @@ export const ReactionPopup = ({
     (r) => r.user_id === userIdEncoded
   )?.reaction
 
-  const beginAnimation = useCallback(() => {
-    Animated.spring(backgroundOpacityAnim.current, {
+  const beginMountAnimation = useCallback(() => {
+    Animated.timing(backgroundOpacityAnim.current, {
```
dylanjeffers, line 103: Now that I see this, can you use the Animated.parallel composer function to just have one .start()?

packages/mobile/src/screens/chat-screen/ReactionPopup.tsx
```diff
@@ -95,24 +99,47 @@ export const ReactionPopup = ({
...
+    }).start()
+    Animated.timing(otherOpacity.current, {
+      toValue: 0,
+      useNativeDriver: true,
+      duration: UNMOUNT_ANIMATION_DURATION_MS
+    }).start(onClose)
+    Animated.spring(translationAnim.current, {
+      toValue: REACTION_CONTAINER_HEIGHT,
+      useNativeDriver: true
+    }).start()
```
dylanjeffers, line 133: And same thing here, can have the callback on the start for the .parallel part

dylanjeffers (approved):
Awesome changes, lmk if the parallel changes make sense

packages/mobile/src/screens/chat-screen/ReactionPopup.tsx
piazzatron, line 103: Support using parallel, I think we could also stand to pull these out (plus the refs) into their own standalone hook to keep the components simpler

piazzatron (approved):
Looks great!

packages/mobile/src/screens/chat-screen/ReactionPopup.tsx
dharit-tan: Had to keep the translation animation separate bc it takes long to complete, and while it's incomplete the user wouldn't be able to open a new reaction popup. Couldn't find a way to cut off the animation time or change the duration (not sure we would want to do that anyway) so this seemed like the best solution.

PR #3273, opened 2023-04-25
[C-1998] Reset notification state on sign-out

dylanjeffers:
### Description

Resets notification state when sign-out action is dispatched

### DIFF
--- a/packages/common/src/store/notifications/notificationsSlice.ts
+++ b/packages/common/src/store/notifications/notificationsSlice.ts
@@ -1,6 +1,7 @@
 import { createEntityAdapter, createSlice } from '@reduxjs/toolkit'
 
 import { Status } from 'models/Status'
+import { signOut } from 'store/sign-out/slice'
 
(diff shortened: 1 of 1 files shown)


sabrina-kiam:
wow thank you!! quick question do you know why this only affects mobile and not web?? it looks like the fix is in common so i would think both versions would have the bug?

dylanjeffers:
> wow thank you!! quick question do you know why this only affects mobile and not web?? it looks like the fix is in common so i would think both versions would have the bug?

yes so on web we refresh the entire page, causing the react app to reset entirely, wheras mobile it still runs. so this general class of problem, stale data across sessions, only affects mobile. However we should ideally get to a world where we dont need to reset the react app on web, and also have a correct state on sign-out. the pattern added here, where the notification slice listens to the sign-out action, is a really powerful one i think, and should be applied across the board.

sabrina-kiam (approved):

PR #3277, opened 2023-04-26
[CON-660] Prefer duration from track metadata over track_segments calculation

theoilie:
### Description
- Displays and passes around `duration` from track metadata when it's truthy in all places that used `track_segments` to calculate duration. Keeps the calculation as a fallback if `duration` is not truthy
- See https://github.com/AudiusProject/audius-protocol/pull/5165

### Dragons
Everything still falls back to track_segments, so it should be safe

### How Has This Been Tested?
Uploaded (v1 and v2 flow) multiple times locally, checked the duration on the track page, and played around with moving the scrubber

Manual testing of track playback on staging

### Feature Flags ###
None

### DIFF
--- a/packages/common/src/services/audius-backend/AudiusBackend.ts
+++ b/packages/common/src/services/audius-backend/AudiusBackend.ts
@@ -577,10 +577,12 @@ export const audiusBackend = ({
     return {
       ...track,
       // TODO: This method should be renamed as it does more than images.
-      duration: track.track_segments.reduce(
-        (duration, segment) => duration + parseFloat(segment.duration),
-        0
-      ),
+      duration:
+        track.duration ||
+        track.track_segments.reduce(
+          (duration, segment) => duration + parseFloat(segment.duration),
+          0
+        ),
       _cover_art_sizes: coverArtSizes
     }
   }
--- a/packages/web/src/common/store/player/sagas.ts
+++ b/packages/web/src/common/store/player/sagas.ts
@@ -132,7 +132,11 @@ export function* watchPlay() {
 
       const endChannel = eventChannel((emitter) => {
         audioPlayer.load(
-          track.track_segments,
+          track.duration ||
+            track.track_segments.reduce(
+              (duration, segment) => duration + parseFloat(segment.duration),
+              0
+            ),
           () => {
             if (onEnd) {
               emitter(onEnd({}))
@@ -212,7 +216,7 @@ export function* watchCollectiblePlay() {
       const audioPlayer = yield* getContext('audioPlayer')
       const endChannel = eventChannel((emitter) => {
         audioPlayer.load(
-          [],
+          0,
           () => {
             if (onEnd) {
               emitter(onEnd({}))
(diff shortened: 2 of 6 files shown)


audius-infra:
Preview this change https://demo.audius.co/theo-use-track-duration-metadata

packages/common/src/services/audius-backend/AudiusBackend.ts
```diff
@@ -577,10 +577,12 @@ export const audiusBackend = ({
     return {
       ...track,
       // TODO: This method should be renamed as it does more than images.
-      duration: track.track_segments.reduce(
-        (duration, segment) => duration + parseFloat(segment.duration),
-        0
-      ),
+      duration:
+        track.duration ||
```
dylanjeffers, line 581: Use ?? instead of || cause 0 is technically falsy

packages/web/src/common/store/player/sagas.ts
```diff
@@ -132,7 +132,11 @@ export function* watchPlay() {
 
       const endChannel = eventChannel((emitter) => {
         audioPlayer.load(
-          track.track_segments,
+          track.duration ||
```
dylanjeffers, line 135: Are there cases where duration won't be there now that we have added the duration field in the AudiusBackend side? Or are there cases where we haven't covered on api adapter side that this handles?

dylanjeffers (approved):
Looks great, just some minor qs around when to do the track segment redude

packages/common/src/services/audius-backend/AudiusBackend.ts
theoilie: I think we want to do this when falsey because the discovery migration sets `duration` to 0 for every track. I'm not sure if we should even support a track actually being 0 seconds, but that would break this (could fix by optional chaining the track_segments and fall back to 0)

packages/web/src/common/store/player/sagas.ts
theoilie: duration should always be there in discovery (assuming the node upgraded / ran the new migration). it's possible I missed a spot in client. it would be 0 for all tracks up until now, though

packages/common/src/services/audius-backend/AudiusBackend.ts
dylanjeffers: Oh yeah great call! Let's do this

packages/web/src/common/store/player/sagas.ts
dylanjeffers: Okay cool prob fine to have the dupe

packages/common/src/services/audius-backend/AudiusBackend.ts
theoilie: sweet 🚀 just realized I misspoke re: edge case where a track is actually 0 seconds - `track_segments` will never be null (it'll be at worst an empty array), so no need for optional chaining. this will just return 0 if someone for some reason uploads an empty track (lol)

PR #3279, opened 2023-04-26
[C-2524] Fix mobile notification reactions

dharit-tan:
### Description
Changes to the mobile `ReactionList` for chats broke the other usage of `ReactionList` in notifications (tip sent). The fix:
- Originally measurements were happening inside a `useEffect` that fired whenever `isVisible` changed. Re-add this back so notifications reactions work.
- Using `isVisible` properly on the chats `ReactionPopup` side, but this did not fix so also need to...
- Still perform measurements during `onLayout`.

### Dragons

### How Has This Been Tested?

Local web stage.
Test by confirming both chats `ReactionPopup` and notifications reactions work. Also check that notifications still works when navigating away/back to the notifications screen.

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/screens/notifications-screen/Reaction/ReactionList.tsx
+++ b/packages/mobile/src/screens/notifications-screen/Reaction/ReactionList.tsx
@@ -120,7 +120,11 @@ export const ReactionList = (props: ReactionListProps) => {
 
   const handleMeasure: OnMeasure = useCallback((config) => {
     const { x, width, reactionType } = config
-    positions.current = { ...positions.current, [reactionType]: { x, width } }
+    // Sometimes this function is called before the View being measured is laid out,
+    // so check that x and width are valid before updating the positions.
+    if (x > 0 && width > 0) {
+      positions.current = { ...positions.current, [reactionType]: { x, width } }
+    }
   }, [])
 
   return (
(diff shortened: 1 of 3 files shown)


packages/mobile/src/screens/notifications-screen/Reaction/ReactionList.tsx
```diff
@@ -120,7 +120,11 @@ export const ReactionList = (props: ReactionListProps) => {
 
   const handleMeasure: OnMeasure = useCallback((config) => {
     const { x, width, reactionType } = config
-    positions.current = { ...positions.current, [reactionType]: { x, width } }
+    // Sometimes this function is called before the View being measured is laid out,
+    // so check that x and width are valid before updating the positions.
+    if (x > 0 && width > 0) {
```
schottra, line 125: I wonder if `useLayoutEffect` would be useful here? It's supposed to run synchrnously _after_ the DOM has been set up for the component. So it should help ensure the layout is valid? Not sure if it behaves the same in ReactNative, though.

dharit-tan: Yeah I had the same thought, tried it and behavior was the same as `useEffect` so I bailed on it. Tbh I don't fully understand why we're getting 0 values here and also sometimes when we call it during `onLayout` bc the component should be laid out afaik, maybe we could discuss sometime but for now this is a bit of a hacky solution.

piazzatron (approved):

schottra (approved):

PR #3281, opened 2023-04-26
[CON-664] Split storage v2 feature flag into signup+upload

theoilie:
### Description
Splits `storage_v2` feature flag into `storage_v2_track_upload` and `storage_v2_signup` so we can rollout uploads separately first.

@endline I removed the ability to prefix an email with "storage_v2_test_". We'll want to use real emails in prod/staging anyway so just run `localStorage.setItem('FeatureFlagOverride:storage_v2_signup', 'enabled')` in Console and then use a real email with the "+" like theo+42623@audius.co

### How Has This Been Tested?
Will be testing on staging today. Flags are set in optimizely as well

### Feature Flags ###
removed: `storage_v2`
added: `storage_v2_track_upload` and `storage_v2_signup`

### DIFF
--- a/packages/common/src/services/audius-backend/AudiusBackend.ts
+++ b/packages/common/src/services/audius-backend/AudiusBackend.ts
@@ -1266,8 +1266,13 @@ export const audiusBackend = ({
     metadata: TrackMetadata,
     onProgress: (loaded: number, total: number) => void
   ) {
-    const storageV2Enabled = await getFeatureEnabled(FeatureFlags.STORAGE_V2)
-    if (storageV2Enabled) {
+    const storageV2SignupEnabled = await getFeatureEnabled(
+      FeatureFlags.STORAGE_V2_SIGNUP
+    )
+    const storageV2UploadEnabled = await getFeatureEnabled(
+      FeatureFlags.STORAGE_V2_TRACK_UPLOAD
+    )
+    if (storageV2SignupEnabled || storageV2UploadEnabled) {
       try {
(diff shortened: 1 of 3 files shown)


audius-infra:
Preview this change https://demo.audius.co/theo-split-storage-v2-flags

packages/common/src/services/audius-backend/AudiusBackend.ts
```diff
@@ -1266,8 +1266,13 @@ export const audiusBackend = ({
...
   ) {
-    const storageV2Enabled = await getFeatureEnabled(FeatureFlags.STORAGE_V2)
-    if (storageV2Enabled) {
+    const storageV2SignupEnabled = await getFeatureEnabled(
+      FeatureFlags.STORAGE_V2_SIGNUP
+    )
+    const storageV2UploadEnabled = await getFeatureEnabled(
+      FeatureFlags.STORAGE_V2_TRACK_UPLOAD
+    )
+    if (storageV2SignupEnabled || storageV2UploadEnabled) {
```
sliptype, line 1275: Why the `or` here?

sliptype (approved):

packages/common/src/services/audius-backend/AudiusBackend.ts
theoilie: if you signed up with the v2 flow then you won't have a replica set, so you have to upload through the v2 flow. we're gonna rollout upload before signup since once you signup there's no going back (without extra bandaid work, that is)

PR #3283, opened 2023-04-27
Fix parsing create notification track owner id

sabrina-kiam:
### Description
Create notifications for tracks weren't showing up in discovery UI 🤦 not sure how we missed this. a while ago joe changed the specifier of track create notifications which audiusbackend was expecting for tracks, so had to update audiusbackend to parse for the information elsewhere in the notif

### DIFF
--- a/packages/common/src/services/audius-backend/AudiusBackend.ts
+++ b/packages/common/src/services/audius-backend/AudiusBackend.ts
@@ -2341,12 +2341,12 @@ export const audiusBackend = ({
         .filter(removeNullable)
       // playlist owner ids are the specifier of the playlist create notif
       const userId =
-        entityType !== Entity.Track
-          ? // album/playlist create notifications store album owner
+        entityType === Entity.Track
+          ? // track create notifs store track owner id in the group id
+            parseInt(notification.group_id.split(':')[3])
+          : // album/playlist create notifications store album owner
             // id as the specifier
             (decodeHashId(notification.actions[0].specifier) as number)
-          : // track create notifs store track owner id in the group id
-            parseInt(notification.group_id.split(':')[3])
       return {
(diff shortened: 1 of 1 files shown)


sabrina-kiam:
Screenshot 2023-04-26 at 9 20 00 PM
Screenshot 2023-04-26 at 9 19 25 PM

packages/common/src/services/audius-backend/AudiusBackend.ts
```diff
@@ -2339,7 +2339,14 @@ export const audiusBackend = ({
         })
         .flat()
         .filter(removeNullable)
-      const userId = decodeHashId(notification.actions[0].specifier) as number
+      // playlist owner ids are the specifier of the playlist create notif
+      const userId =
+        entityType !== Entity.Track
```
raymondjacobson, line 2344: can entity type be something besides track & collection? will this break if so?

raymondjacobson (approved):

packages/common/src/services/audius-backend/AudiusBackend.ts
isaacsolo: 'create' types only happen on track and playlist handler

maybe use === and reverse the logic for clarity

sabrina-kiam: also `entityType` var is created on line 2327 and only set in 3 places, once on initialization, and in lines 2332 and 2337 where its all either set to track, playlist, or album

sabrina-kiam: the code could definitely be more safeguarded

dylanjeffers (approved):
Nice catch

PR #3284, opened 2023-04-27
[PAY-1148] Mobile chat user search row disabled when not permitted

dharit-tan:
### Description
On search users screen, disable row when not permitted to message user.

Per @sliptype's suggestion, selecting relevant data in flatlist item component rather than passing it in from parent.

Also fixed a bug where `searchUsers` would dispatch infinitely when result count was low.

### Dragons

### How Has This Been Tested?

Local ios simulator stage tested with blocker, blockee, account with followers only inbox settings. Tested that `searchUsers` changes didn't break web.
Simulator Screen Shot - iPhone 14 Pro - 2023-04-27 at 00 27 53

### Feature Flags ###

### DIFF
--- a/packages/common/src/store/pages/chat/selectors.ts
+++ b/packages/common/src/store/pages/chat/selectors.ts
@@ -1,4 +1,4 @@
-import type { UserChat } from '@audius/sdk'
+import type { UserChat, ValidatedChatPermissions } from '@audius/sdk'
 import { createSelector } from 'reselect'
 
 import type { ID } from 'models/Identifiers'
@@ -152,15 +152,26 @@ export const getUnfurlMetadata = (
   return message?.unfurlMetadata
 }
 
-export const isPermittedForUser = (state: CommonState, userId: ID) => {
-  const currentUserId = getUserId(state) ?? -1
-  const blockerList = getBlockers(state)
-  const hasBlocker = blockerList.includes(currentUserId)
-  const blockeeList = getBlockees(state)
-  const hasBlockee = blockeeList.includes(userId)
-  const permissionsMap = getUserChatPermissions(state)
-  const isPermitted =
-    permissionsMap[userId]?.current_user_has_permission ?? true
+export const isPermittedForUser = createSelector(
+  [
+    getUserId,
+    getBlockers,
+    getBlockees,
+    getUserChatPermissions,
+    (_: CommonState, userId: ID) => userId
+  ],
+  (
+    currentUserId,
+    blockerList,
+    blockeeList,
+    permissionsMap: Record<ID, ValidatedChatPermissions>,
+    userId
+  ) => {
+    const hasBlocker = blockerList.includes(currentUserId ?? -1)
+    const hasBlockee = blockeeList.includes(userId)
+    const isPermitted =
+      permissionsMap[userId]?.current_user_has_permission ?? true
 
-  return !hasBlocker && !hasBlockee && isPermitted
-}
+    return !hasBlocker && !hasBlockee && isPermitted
+  }
+)
--- a/packages/common/src/store/pages/chat/selectors.ts
+++ b/packages/common/src/store/pages/chat/selectors.ts
@@ -152,26 +152,24 @@ export const getUnfurlMetadata = (
   return message?.unfurlMetadata
 }
 
-export const isPermittedForUser = createSelector(
+export const getCanMessageUser = createSelector(
   [
-    getUserId,
     getBlockers,
     getBlockees,
     getUserChatPermissions,
     (_: CommonState, userId: ID) => userId
   ],
   (
-    currentUserId,
     blockerList,
     blockeeList,
     permissionsMap: Record<ID, ValidatedChatPermissions>,
     userId
   ) => {
-    const hasBlocker = blockerList.includes(currentUserId ?? -1)
-    const hasBlockee = blockeeList.includes(userId)
+    const isBlocked = blockerList.includes(userId)
+    const doesBlock = blockeeList.includes(userId)
     const isPermitted =
       permissionsMap[userId]?.current_user_has_permission ?? true
 
-    return !hasBlocker && !hasBlockee && isPermitted
+    return !isBlocked && !doesBlock && isPermitted
   }
 )
--- a/packages/common/src/store/pages/chat/selectors.ts
+++ b/packages/common/src/store/pages/chat/selectors.ts
@@ -1,6 +1,7 @@
 import type { UserChat } from '@audius/sdk'
 import { createSelector } from 'reselect'
 
+import type { ID } from 'models/Identifiers'
 import { accountSelectors } from 'store/account'
 import { cacheUsersSelectors } from 'store/cache'
 import { CommonState } from 'store/reducers'
@@ -149,4 +150,17 @@ export const getUnfurlMetadata = (
 ) => {
   const message = getChatMessageById(state, chatId, messageId)
   return message?.unfurlMetadata
+}
+
+export const isPermittedForUser = (state: CommonState, userId: ID) => {
+  const currentUserId = getUserId(state) ?? -1
+  const blockerList = getBlockers(state)
+  const hasBlocker = blockerList.includes(currentUserId)
+  const blockeeList = getBlockees(state)
+  const hasBlockee = blockeeList.includes(userId)
+  const permissionsMap = getUserChatPermissions(state)
+  const isPermitted =
+    permissionsMap[userId]?.current_user_has_permission ?? true
+
+  return !hasBlocker && !hasBlockee && isPermitted
 }
--- a/packages/common/src/store/ui/search-users-modal/sagas.ts
+++ b/packages/common/src/store/ui/search-users-modal/sagas.ts
@@ -26,7 +26,9 @@ function* doSearchUsers(action: ReturnType<typeof searchUsers>) {
     })
     const users = Object.values(res.users) as User[]
     yield* call(processAndCacheUsers, users)
-    yield* put(searchUsersSucceeded({ userIds: users.map((u) => u.user_id) }))
+    yield* put(
+      searchUsersSucceeded({ userIds: users.map((u) => u.user_id), limit })
+    )
   } catch (e) {
     console.error(e)
   }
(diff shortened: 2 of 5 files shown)


audius-infra:
Preview this change https://demo.audius.co/rt-perms-mobile

audius-infra:
Preview this change https://demo.audius.co/rt-perms-mobile

packages/common/src/store/pages/chat/selectors.ts
```diff
@@ -150,3 +151,25 @@ export const getUnfurlMetadata = (
   const message = getChatMessageById(state, chatId, messageId)
   return message?.unfurlMetadata
 }
+
+export const getCanMessageUser = createSelector(
+  [
+    getBlockers,
+    getBlockees,
+    getUserChatPermissions,
+    (_: CommonState, userId: ID) => userId
```
schottra, line 160: Might be missing something, but wondering why this selector takes state and then doesn't use it? Can it be rewritten to not accept the state? Or does it always have to accept it as the first argument?

packages/common/src/store/ui/search-users-modal/sagas.ts
```diff
@@ -26,7 +26,9 @@ function* doSearchUsers(action: ReturnType<typeof searchUsers>) {
     })
     const users = Object.values(res.users) as User[]
     yield* call(processAndCacheUsers, users)
-    yield* put(searchUsersSucceeded({ userIds: users.map((u) => u.user_id) }))
+    yield* put(
+      searchUsersSucceeded({ userIds: users.map((u) => u.user_id), limit })
```
schottra, line 30: I'm guessing the control paginates results as you scroll?

packages/common/src/store/ui/search-users-modal/slice.ts
```diff
@@ -38,13 +38,12 @@ const slice = createSlice({
...
+      action: PayloadAction<{ userIds: number[]; limit: number }>
     ) => {
-      state.userList.userIds = state.userList.userIds.concat(
-        action.payload.userIds
-      )
+      const { userIds, limit } = action.payload
+      state.userList.userIds = state.userList.userIds.concat(userIds)
       state.userList.status = Status.SUCCESS
-      state.userList.hasMore = action.payload.userIds.length > 0
+      state.userList.hasMore = userIds.length === limit
```
schottra, line 46: Just checking that things are alright if another request happens after this and returns no results, in the case where the search results in exactly `limit` items and there aren't any more to return after that.

If you wanted, you could move the `hasMore` calculation into the saga and let it compare the limit it requested against the results that were returned. Then we don't have to pipe the limit through across two action payloads.

packages/mobile/src/screens/chat-screen/ChatUserListItem.tsx
```diff
@@ -0,0 +1,165 @@
...
+    fontSize: typography.fontSize.small,
+    color: palette.neutral
+  },
+  followersContainer: {
+    display: 'flex',
+    flexDirection: 'row',
+    alignItems: 'center'
+  },
+  followersCount: {
+    fontWeight: '800',
```
schottra, line 72: Do we have typography constants for weights?

packages/mobile/src/screens/chat-screen/ChatUserListItem.tsx
```diff
@@ -0,0 +1,165 @@
...
+type ChatUserListItemProps = {
+  user: User
+}
+
+export const ChatUserListItem = ({ user }: ChatUserListItemProps) => {
+  const styles = useStyles()
+  const palette = useThemeColors()
+  const dispatch = useDispatch()
+  const canMessageUser = useSelector((state) =>
+    getCanMessageUser(state, user.user_id)
```
schottra, line 111: How often do we refresh the info that drives this selector? Just wondering if it would cause an issue if I open this screen and keep it open while the other user blocks me.

packages/mobile/src/screens/chat-screen/ChatUserListItem.tsx
```diff
@@ -0,0 +1,165 @@
...
+  )
+
+  const handlePress = useCallback(
+    (user) => {
+      dispatch(createChat({ userIds: [user.user_id] }))
+    },
+    [dispatch]
+  )
+
+  if (!user) {
```
schottra, line 121: * Would love to define things like this as explicit loading states since `props.user` could be technically be falsey for other reasons. 
* Is there anything else we need to wait for to ensure this component works correctly
* You could move this logic up a level so that this component is guaranteed to be rendered with a `User`. If I understand the code correctly, we're using this to render a spinner while we load results? That logic probably belongs in the parent.

packages/mobile/src/screens/chat-screen/ChatUserListScreen.tsx
```diff
@@ -177,6 +113,17 @@ export const ChatUserListScreen = (props: ChatUserListScreenProps) => {
...
+  useEffect(() => {
+    dispatch(fetchBlockees())
+    dispatch(fetchBlockers())
+  }, [dispatch])
+
+  useEffect(() => {
+    if (userIds.length > 0) {
+      dispatch(fetchPermissions({ userIds }))
+    }
+  }, [dispatch, userIds])
```
schottra, line 125: Do you think it would make sense to push this logic into the sagas? i.e.:
* The loading state for this component is based on waiting for the user list, list of blockees, and list of blockers. We could fetch them all in parallel in the same loading saga and only switch the state once we've finished all three
* Can/should we move fetching permissions to the saga that populates user ids? If we fork it there, it's not waiting for a re-render and effect hook loop before fetching the permissions.

packages/common/src/store/pages/chat/selectors.ts
dharit-tan: Will double check but I think the state has to be the first arg.

packages/common/src/store/ui/search-users-modal/sagas.ts
dharit-tan: Yup `onEndReached={handleLoadMore}` prop of flatlist and the action concats the results to the previous results.

packages/mobile/src/screens/chat-screen/ChatUserListItem.tsx
dharit-tan: Oo good call.. should we poll this?
Personally I think it's ok to leave as is since if a user is blocked, we restrict the creation of new chats as well as disabling the textinput for existing chats, on both sides. So even if a user is able to send messages in the case you mentioned, by the next page load they will be restricted. cc @piazzatron @rickyrombo on this and i'll bring it up during the DMs sync tomorrow.

rickyrombo: good catch @schottra! def don't think we should poll this. let's handle chat creation failures gracefully instead - maybe navigating to that chat screen with the proper "you cannot message this user" text

packages/mobile/src/screens/chat-screen/ChatUserListScreen.tsx
dharit-tan: Hm.. I don't think we want to couple the user list fetching with permissions fetching. Right now this is the only place we use this `UserList` sort of component but I'd like to make it generic.

piazzatron: Another possibility is to pull this out into a shared hook if this logic applies in web as well - cc @rickyrombo

packages/mobile/src/screens/chat-screen/ChatUserListItem.tsx
piazzatron: agreed with @rickyrombo

schottra: Definitely agree we shouldn't poll. I more imagined that we will discover upon attempting to send a message that you aren't allowed to do it and then add them into the block list at that point.
The real question is do we remove the message from the screen? Or mark it as undelivered? I think the experience could be confusing.

packages/mobile/src/screens/chat-screen/ChatUserListScreen.tsx
schottra: Oh okay. I thought the saga for fetching the user list was specifically for chats. So I would have thought it could fetch the users and decorate them with their status. Or fetch both users and block information and combine.
Or, fetch users and proactively start fetching blocks.
Anyway a hook composing the two also makes sense.

packages/mobile/src/screens/chat-screen/ChatUserListItem.tsx
dharit-tan: Ah man, just remembered why I did this - we do have constants but don't have one for 800. There is something wonky with our font weight constants not matching up with the specs in figma as well. Something that's on the to-do list.

dharit-tan: EDIT just gonna use `bold` here, confirmed with design that this is ok.

packages/mobile/src/screens/chat-screen/ChatUserListScreen.tsx
rickyrombo: I still need to adopt this actually in web! I'll work on extracting a hook for this

packages/mobile/src/screens/chat-screen/ChatUserListItem.tsx
dharit-tan: Good call. Actually we already ensure the user exists in the parent so didn't need this check anyway. Yay!

dharit-tan: Great call. Actually the data we pass in from the parent is from a selector, so we don't need this check anyway. Yay!

packages/common/src/store/ui/search-users-modal/slice.ts
dharit-tan: Good call but gonna punt on this refactor.

I think we're ok bc if the previous request has exactly `limit` items, the next request will have 0 items and it'll just be a no-op and stop refreshing. We send 1 extra request but I think that's ok for now. Just checked with `limit = 1` and it works.

schottra (approved):

PR #3287, opened 2023-04-27
[C-2543, C-2544] Initial updates to mobile playlist management

Kyle-Shanks:
### Description
* Add drawer for playlist publish flow
* Add playlist updates feature flags
* Update the action buttons on playlist details tiles for playlists that the user owns

[Will add some screenshots]

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

All visible changes should be behind the PLAYLIST_UPDATES_PRE_QA flag


packages/common/src/store/ui/publish-playlist-confirmation-modal/selectors.ts
```diff
@@ -0,0 +1,4 @@
+import { CommonState } from 'store/commonStore'
+
+export const getPlaylistId = (state: CommonState) =>
```
dylanjeffers, line 3: ah so i think this is definitely fine to do, having a separate state+slice+reducer for this, but one potential option is to use the "data" object in our modals, so you can maybe do `setVisible({name: "PublishPlaylist", data: {playlistId }})` then you can just leverage the existing modal strat and not have to have all the extra code around it. what you think?

dylanjeffers: also this is not something we have established well yet, but i think another option could be to render the "drawer" right next to the component that needs it, and we can portal the drawer to the root. this would be a larger change overal, but a headsup as far as where we might want to go with this long-term

packages/common/src/store/ui/toast/slice.ts
```diff
@@ -20,6 +20,14 @@ const slice = createSlice({
       const toast = action.payload
       state.toasts.push(toast)
     },
+    manualClearToast: (state, action: DissmissToastAction) => {
+      const toastIdx = state.toasts.findIndex(
+        (t) => t.key === action.payload.key
```
dylanjeffers, line 25: so cleanest imo would be to remove the toast element from the state, and then the toast component would see the element is no longer there, and animate out? this is a cool way to make it happen though.

dylanjeffers: if we go with this implementation, would something like this work?
```
const toastToClear = state.toasts.find(....)
toastToClear.timeout = 0
```

dylanjeffers: and just final callout, i think as much as we can, use full words for things, even known words that can be minimized, so instead of idx, use index. maybe would be a good thing to check with everyone on client, and possibly add to styleguide.

packages/mobile/src/components/details-tile/DetailsTile.tsx
```diff
@@ -195,9 +195,13 @@ export const DetailsTile = ({
   hideShare,
   isPlaying,
   isPlayable = true,
+  isPlaylist = false,
+  isPublished = true,
```
dylanjeffers, line 199: do we want to provide default args for published, or should that always come from the metadata?

packages/mobile/src/components/details-tile/DetailsTile.tsx
```diff
@@ -195,9 +195,13 @@ export const DetailsTile = ({
...
+  onPressEdit = () => {},
```
dylanjeffers, line 200: i personally dont provide default blank functions for things anymore, and just do `onPressEdit?.()` where appropriate.

packages/mobile/src/components/details-tile/DetailsTileActionButtons.tsx
```diff
@@ -54,19 +61,25 @@ const useStyles = makeStyles(({ palette }) => ({
 export const DetailsTileActionButtons = ({
   hasReposted,
   hasSaved,
+  isPlaylist = false,
```
dylanjeffers, line 64: since isPlaylist is set in the parent component, we dont need the default value here. typically we should avoid all default args all the way down, and just make sure it's set in one spot, like the highest parent?

packages/mobile/src/components/details-tile/DetailsTileActionButtons.tsx
```diff
@@ -54,19 +61,25 @@ const useStyles = makeStyles(({ palette }) => ({
...
   isOwner,
   isPublished = true,
   hideFavorite,
   hideOverflow,
   hideRepost,
   hideShare,
+  onPressEdit = () => {},
```
dylanjeffers, line 71: same personal feeling with edit here :)

packages/mobile/src/components/details-tile/DetailsTileActionButtons.tsx
```diff
@@ -105,11 +118,35 @@ export const DetailsTileActionButtons = ({
...
+      onPress={onPressEdit}
+      styles={{ icon: styles.actionButton }}
+    />
+  )
+
+  const publishButton = (
+    <IconButton
+      fill={neutralLight4}
+      icon={IconRocket}
+      // TODO: Add isDisabled based on if playlist is publishable logic
```
dylanjeffers, line 134: just checking this is a todo for not this pr?

packages/mobile/src/components/publish-playlist-drawer/PublishPlaylistDrawer.tsx
```diff
@@ -0,0 +1,144 @@
...
+    textTransform: 'uppercase'
+  },
+  container: {
+    marginHorizontal: spacing(4)
+  },
+  body: {
+    margin: spacing(4),
+    lineHeight: spacing(6),
+    textAlign: 'center',
+    color: palette.neutral,
```
dylanjeffers, line 54: this is the default color for text just btw

packages/mobile/src/components/publish-playlist-drawer/PublishPlaylistDrawer.tsx
```diff
@@ -0,0 +1,144 @@
...
+  const dispatch = useDispatch()
+  const playlistId = useSelector(getPlaylistId)
+  const neutral = useColor('neutral')
+  const { toast } = useToast()
+  const { toast: manualToast } = useManualToast()
+  const styles = useStyles()
+  const { isOpen, onClose } = useDrawerState('PublishPlaylistConfirmation')
+
+  const displayPublishToast = useCallback(() => {
+    const publishingToastContent = (
```
dylanjeffers, line 82: nit but maybe `publishingPlaylistToastContent`

packages/mobile/src/components/publish-playlist-drawer/PublishPlaylistDrawer.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+    return manualToast({
+      content: publishingToastContent
+    })
+  }, [manualToast, styles])
+
+  const handlePublish = useCallback(() => {
+    if (playlistId) {
+      const dismissPublishingToast = displayPublishToast()
+      const onConfirmPublish = () => {
```
dylanjeffers, line 97: ah cool, yeah i think this will be the main question mark for us to discuss cause it's not advisable to send a callback into an action :)

packages/mobile/src/components/publish-playlist-drawer/PublishPlaylistDrawer.tsx
```diff
@@ -0,0 +1,144 @@
...
+  },
+  container: {
+    marginHorizontal: spacing(4)
+  },
+  body: {
+    margin: spacing(4),
+    lineHeight: spacing(6),
+    textAlign: 'center',
+    color: palette.neutral,
+    fontSize: typography.fontSize.large,
```
dylanjeffers, line 55: these styles can also be added directly to text which can sometimes be faster, just an fyi: 

```
<Text fontSize='large' weight='medium' />
```

packages/mobile/src/components/toasts/Toast.tsx
```diff
@@ -74,9 +74,11 @@ export const Toast = (props: ToastProps) => {
       toValue: 1,
       useNativeDriver: true
     }).start(() => {
-      setTimeout(() => {
-        animateOut()
-      }, timeout)
+      if (timeout !== 'MANUAL') {
```
dylanjeffers, line 77: so im coming around to some thoughts for the props/config of toast, so instead of overloading the "timeout" prop with "MANUAL", maybe we have a new prop/config value like noTimeout, or manualDismiss, or something along those lines? this is a smaller thing, but just feeling like from a user perspective, it's a bit weird to have a value "timeout" be a number or a string === "MAUNAL" know what i mean?

packages/mobile/src/hooks/useManualToast.ts
```diff
@@ -0,0 +1,28 @@
...
+  const dispatch = useDispatch()
+  const handleToast = useCallback(
+    (toast: ManualToastAction) => {
+      const key = uuid()
+      dispatch(addToast({ ...toast, key, timeout: 'MANUAL' }))
+
+      // Return function to manually clear the toast
+      // Will be picked up by the Toast component and will dismiss the toast after the animation
+      return () => {
+        dispatch(manualClearToast({ key }))
```
dylanjeffers, line 21: so maybe the move is that we simply keep track of the "key" of the toast we want to manually dismiss, and pass that along instead of the callback?

packages/mobile/src/screens/collection-screen/CollectionScreen.tsx
```diff
@@ -48,6 +50,7 @@ import { makeStyles } from 'app/styles'
 import { CollectionScreenDetailsTile } from './CollectionScreenDetailsTile'
 import { CollectionScreenSkeleton } from './CollectionScreenSkeleton'
 
+const { open: openEditPlaylist } = createPlaylistModalUIActions
```
dylanjeffers, line 53: so how come this is edit, should be create?

packages/web/src/common/store/cache/collections/sagas.js
```diff
@@ -869,10 +869,21 @@ function* publishPlaylistAsync(action) {
     ])
   )
 
-  yield call(confirmPublishPlaylist, userId, action.playlistId, playlist)
+  yield call(
+    confirmPublishPlaylist,
+    userId,
+    action.playlistId,
+    playlist,
+    action.confirmCallback
```
dylanjeffers, line 877: so maybe here we have the toastKey or something?

dylanjeffers: another option is to allow us to pass a toast key to the toast dispatch, so we can simply reference it anywhere else in the app when we want to dismiss.

packages/web/src/common/store/cache/collections/sagas.js
```diff
@@ -905,6 +916,7 @@ function* confirmPublishPlaylist(userId, playlistId, playlist) {
             }
           ])
         )
+        if (confirmCallback) confirmCallback()
```
dylanjeffers, line 919: so here we can `yield* put(dismissToast({ key: publishPlaylistToastKey })`?

dylanjeffers (changes requested):
supremely clean and readable code, nice work! added lots of comments as i went on this journey :) high level i think the two things id like to see updated:

1. Avoid callback in action. like mentioned, maybe we pass around or have global context of the toast key we want to dismiss.
2. Consider using modal "data" field (i think this currently just exists for mobile drawers), instead of having to add yet another slice + saga for a single variable.
3. Is there a way to maybe just have "useToast" be the only hook, and the user can pass a new config option to it to specify that we want to manually dismiss it at a later time?

Going to "Request changes" just for request 1. above, just cause that's a strict rule :) but otherwise super close

packages/common/src/store/ui/toast/slice.ts
Kyle-Shanks: I tried to simply remove the toast from state, but the toast elements are rendered from the toasts and immediately disappears if it is removed. I went with this pattern to have the toast listen for the change and dismiss itself bc it looked nicer. Open to alternative solutions to this though bc it does feel a bit unintuitive

packages/common/src/store/ui/toast/types.ts
```diff
@@ -8,7 +8,7 @@ export type Toast = {
   content: ToastContent
   type?: ToastType
   key: string
-  timeout?: number
+  timeout?: number | 'MANUAL'
```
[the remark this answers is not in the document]
Kyle-Shanks: An empty timeout is set to default in a few places and I didn't want to change that in case there were places that were relying on that. I agree that having this be either a string or number isn't the best, but it felt ok bc the string is only used in the special case. Open to discussion here

packages/mobile/src/components/details-tile/DetailsTile.tsx
Kyle-Shanks: Good thought. will update

packages/mobile/src/components/details-tile/DetailsTile.tsx
Kyle-Shanks: I think it will always come from the metadata, but added this bc i wanted the default behavior to ignore the code that was added.

packages/mobile/src/components/details-tile/DetailsTileActionButtons.tsx
Kyle-Shanks: Yup, for later when the hidden/private logic is added

packages/mobile/src/components/overflow-menu-drawer/CollectionOverflowMenuDrawer.tsx
```diff
@@ -103,8 +111,12 @@ const CollectionOverflowMenuDrawer = ({ render }: Props) => {
     },
     [OverflowAction.DELETE_PLAYLIST]: () =>
       dispatch(openDeletePlaylist({ playlistId: id })),
-    [OverflowAction.PUBLISH_PLAYLIST]: () =>
-      is_album ? () => {} : dispatch(publishPlaylist(Number(id)))
+    [OverflowAction.PUBLISH_PLAYLIST]: () => {
+      if (is_album) return () => {}
```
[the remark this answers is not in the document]
Kyle-Shanks: i don't know what the check is for displaying this option. I was just keeping logic that was there from before

packages/mobile/src/components/toasts/Toast.tsx
Kyle-Shanks: Yea I can get behind this. I'll do some testing to see how this feels. Want to make sure that I maintain an easy method to have the toast dismiss itself based on changes in the slice

packages/mobile/src/screens/collection-screen/CollectionScreen.tsx
Kyle-Shanks: I think the same actions are used for creating and editing. this is for the edit button so called it openEditPlaylist to be clear

packages/common/src/store/ui/toast/slice.ts
dylanjeffers: yeah makes sense, i think "react-spring" can handle it but maybe not react-native animations

audius-infra:
Preview this change https://demo.audius.co/kj-Update-mobile-playlist-action-buttons-and-publish-flow

dylanjeffers (approved):
looks great

PR #3288, opened 2023-04-28
Mobile chat housekeeping

dharit-tan:
### Description
Cleaning up some stuff.
- Centering loading spinners
- Remove `ChatUser` - seemed unnecessary

### Dragons

### How Has This Been Tested?

Local ios stage

### Feature Flags ###


packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -178,7 +189,8 @@ export const ChatScreen = () => {
   )
   const unreadCount = chat?.unread_message_count ?? 0
   const isLoading =
-    chat?.messagesStatus === Status.LOADING && chatMessages?.length === 0
+    statusIsNotFinalized(chat?.messagesStatus ?? Status.IDLE) &&
```
dharit-tan, line 192: `chat.messagesStatus` starts as `undefined` since it's using `entityAdapter`, this seems best way to handle?

schottra: If I'm being really dogmatic, I would say that the bit of the component tree below that needs `isLoading` should be its own component that is only rendered if we have a valid `chatId`, and that it pulls the loading state in those cases (where it should be guaranteed to exist).
But, otherwise yeah I think this works. The only catch might be that for `undefined` cases, it's going to show a loading spinner? Is that what we want? As in, is there a "unselected" case where we shouldn't render anything at all?

packages/mobile/src/screens/chat-screen/ChatListItem.tsx
```diff
@@ -31,28 +65,33 @@ export const ChatListItem = ({ chat }: { chat: UserChat }) => {
   const navigation = useNavigation<AppTabScreenParamList>()
   const styles = useStyles()
 
-  const users = useProxySelector(
+  const user = useProxySelector(
     (state) => getOtherChatUsersFromChat(state, chat),
     [chat]
-  )
+  )[0]
```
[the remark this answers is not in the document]
dharit-tan: I think accessing empty array gives you `undefined` which works for code below. Not pretty tho, open to suggestions. Just thought adding a whole other if check was unnecessary.

packages/mobile/src/screens/chat-screen/ChatListItem.tsx
```diff
@@ -23,6 +22,41 @@ const useStyles = makeStyles(({ spacing, palette, typography }) => ({
...
+    marginBottom: spacing(2)
+  },
+  userNameContainer: {
+    display: 'flex',
+    flexDirection: 'row',
+    marginBottom: spacing(1)
+  },
+  userName: {
+    fontSize: typography.fontSize.large,
+    fontWeight: 'bold',
```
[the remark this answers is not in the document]
dharit-tan: Good callout

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
dharit-tan: Don't think there's a case where we shouldn't render anything at all. Think it should either be
- chats if they exist.
- `EmptyChatMessage` if we fetched already and there are no messages.
- loading spinner while fetch is underway.

packages/mobile/src/screens/chat-screen/ChatListItem.tsx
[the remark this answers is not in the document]
piazzatron: Should we have a version of this selector that does this internally?

packages/mobile/src/screens/chat-screen/ChatListItem.tsx
```diff
@@ -31,28 +65,33 @@ export const ChatListItem = ({ chat }: { chat: UserChat }) => {
   const navigation = useNavigation<AppTabScreenParamList>()
   const styles = useStyles()
 
-  const users = useProxySelector(
+  const user = useProxySelector(
```
piazzatron, line 68: nit: maybe `otherUser`?

piazzatron (approved):

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -190,7 +202,7 @@ export const ChatScreen = () => {
   const chatFrozenRef = useRef(chat)
 
   useEffect(() => {
-    if (chatId && (chat?.messagesStatus ?? Status.IDLE) === Status.IDLE) {
+    if (chatId && statusIsNotFinalized(chat?.messagesStatus ?? Status.IDLE)) {
```
rickyrombo, line 205: I don't think this is what we want. We only want to start fetching if we're not currently fetching - eg. iff `Status.IDLE` but not `Status.LOADING`

TIL we have this helper though! that's neat

dharit-tan: good looks

audius-infra:
Preview this change https://demo.audius.co/rt-spinners

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -182,10 +193,11 @@ export const ChatScreen = () => {
   )
   const unreadCount = chat?.unread_message_count ?? 0
   const isLoading =
-    chat?.messagesStatus === Status.LOADING && chatMessages?.length === 0
+    (chat?.messagesStatus ?? Status.LOADING) === Status.LOADING &&
```
dharit-tan, line 196: Probably should put this logic in its own selector but let's merge this for QA for now.

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -69,7 +69,8 @@ const useStyles = makeStyles(({ spacing, palette, typography }) => ({
     justifyContent: 'space-between'
   },
   listContainer: {
-    flex: 1
+    flexGrow: 1,
```
schottra, line 72: Maybe a comment based on what we discovered in the huddle?

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -422,30 +434,32 @@ export const ChatScreen = () => {
               hasCurrentlyPlayingTrack ? { bottom: PLAY_BAR_HEIGHT } : null
             ]}
           >
+            {chat?.messagesStatus === Status.SUCCESS &&
+            chatMessages?.length === 0 ? (
+              <EmptyChatMessages />
+            ) : null}
             {!isLoading ? (
```
schottra, line 441: Minor nit: Negated conditionals are a little harder to read. In the future, could arrange the ternary the other way, so that the loading component comes first and you can do `isLoading ? <LoadingComponent /> : <ContentComponent />`

schottra (approved):

audius-infra:
Preview this change https://demo.audius.co/rt-spinners

PR #3290, opened 2023-05-01
[PAY-1199] Mobile inbox settings functional

dharit-tan:
### Description
Wire up mobile inbox settings screen.

### Dragons

### How Has This Been Tested?

Local ios stage

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/rt-inbox-settings

packages/common/src/store/pages/chat/sagas.ts
```diff
@@ -355,6 +357,33 @@ function* doFetchPermissions(action: ReturnType<typeof fetchPermissions>) {
...
+          [currentUserId]: {
+            current_user_has_permission: true,
+            permits: action.payload.permissions,
+            user_id: encodeHashId(currentUserId)
+          }
+        }
+      })
+    )
+  } catch (e) {
+    console.error('setPermissionsFailed', e)
```
schottra, line 383: Not sure what the long term pattern is supposed to be here. Should we be firing a sentry event?

schottra: also what's the user experience for when we fail to set this? Should the controls revert state?

packages/mobile/src/screens/settings-screen/InboxSettingsScreen.tsx
```diff
@@ -64,8 +76,51 @@ const useStyles = makeStyles(({ spacing, palette, typography }) => ({
...
+    title: messages.tipperTitle,
+    description: messages.tipperDescription,
+    value: ChatPermission.TIPPERS
+  },
+  {
+    title: messages.noneTitle,
+    description: messages.noneDescription,
+    value: ChatPermission.NONE
+  }
+]
```
schottra, line 100: ❤️

packages/common/src/store/pages/chat/sagas.ts
schottra: Also love seeing the SDK usage!

dharit-tan: cc @rickyrombo I think the RPC will throw an error if it fails, which will prevent the UI from updating in the first place. Probably should fire a sentry event tho good call.

packages/mobile/src/screens/settings-screen/InboxSettingsScreen.tsx
```diff
@@ -64,8 +76,51 @@ const useStyles = makeStyles(({ spacing, palette, typography }) => ({
...
+
+  useEffect(() => {
+    if (userId) {
+      dispatch(fetchPermissions({ userIds: [userId] }))
+    }
+  }, [dispatch, userId])
+
+  const handlePress = useCallback(
+    (value) => {
+      dispatch(setPermissions({ permissions: value }))
```
rickyrombo, line 119: Thoughts on not using redux + sagas for the setPermissions stuff?

I skipped it in the web impl. could probably share a hook with ya if we're looking to keep things DRY:

https://github.com/AudiusProject/audius-client/blob/776596ff5f1b1dc09a148d834ab91e90884a6ebc/packages/web/src/components/inbox-settings-modal/InboxSettingsModal.tsx#L97-L116

packages/common/src/store/pages/chat/sagas.ts
rickyrombo: Great callout. I haven't been good about explicitly throwing sentry errors and have instead been relying on our catch-all sentry rule of console.errors but I shouldn't do that - I'll make a note to clean some of these up

packages/mobile/src/screens/settings-screen/InboxSettingsScreen.tsx
dharit-tan: What's the benefit of doing it this way?

rickyrombo: Mainly just colocation making things easier to reason about. I'm game either way, either I can adopt the new saga or abstract the component methods into a hook you can use

schottra: I would vote for not having so much async logic directly in the component. I know we don't have a story yet for how to do mutations without sagas. This could at least be moved to a shared state hook of some sort that exports a save callback?

I _am_ in favor of getting off of sagas and potentially skipping redux state management for local component state. Especially for new code.

dharit-tan: Got it, I think what was missing from my understanding was that this is basically local state. I'll convert to a hook.

rickyrombo: > This could at least be moved to a shared state hook of some sort that exports a save callback?

yep, precisely

dharit-tan:
Had discussions with @piazzatron and @rickyrombo about this. tl;dr current version is def overly complex and potentially would've been better to separate some of the code in the shared hook. Gonna leave it as it is for now bc it works and it's good fodder for philosophical conversations int its current state. But for now we've spent enough time on it. Here's a doc: https://www.notion.so/Hooks-vs-redux-case-study-57ef82c8dcfc436c80b36491ec870a42

rickyrombo (approved):

PR #3291, opened 2023-05-01
Fix devnet in wallet authorize for saga phone

raymondjacobson:
### Description

Dev-net was hard-coded. Our staging/test env uses mainnet-beta anyway, so this should be safe.


dylanjeffers (approved):
nice find

PR #3294, opened 2023-05-01
[C-2282] Show tracks tab on own profile if all tracks hidden

amendelsohn:
### Description

Fixes a bug where we wouldn't consider your own user an artist if all of your tracks are hidden. Other users will not see you displayed as an artist, but this change allows you to view tracks on your own profile.

image

TODO:
- [x] fix on mobile (follow-up PR)

### Dragons

The solution here involves adding a slice to fetch and store the count of the user's first page of uploaded tracks. Ideally, the API would return the correct count as part of getUser when you are fetching your own profile. In the meantime, this is the agreed upon workaround.

### How Has This Been Tested?

Local web:
1. upload hidden track for user1
2. view user1 profile as user1 (tracks tab)
3. view as guest/user2 (no tracks shown)


packages/common/src/store/ui/own-profile/slice.ts
```diff
@@ -0,0 +1,30 @@
+import type { PayloadAction } from '@reduxjs/toolkit'
+import { createSlice } from '@reduxjs/toolkit'
+
+import { Nullable } from 'utils/index'
+
+export type OwnProfileState = typeof initialState
+
+type State = {
+  trackCount: Nullable<number>
```
amendelsohn, line 9: it's somewhat misleading to call it count as we only count the first page of results, and since we're only using this as a zero vs non-zero value, perhaps we should just store the bool or null

packages/common/src/store/ui/own-profile/slice.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+const initialState: State = {
+  trackCount: null
+}
+
+const slice = createSlice({
+  name: 'ownProfile',
```
amendelsohn, line 17: wondering if this should be part of account?
perhaps `store.account.hasTracks`

dylanjeffers: yeah i could def see this being a part of account actually!

dylanjeffers (approved):
makes a ton of sense, and stoked this is actually unblocked now. agreed with maybe moving to account slice, and also having a boolean, like "isArtist" or "hasTracks" or something?

amendelsohn:
Now working on mobile

audius-infra:
Preview this change https://demo.audius.co/amendel-hidden-tracks-profile-fix

PR #3295, opened 2023-05-02
[PAY-1108] Chats Screen avoids play bar

dharit-tan:
### Description
Tried a couple of approaches but landed here. Flatlist only supports sticky headers, no sticky footers, and since we're using inverted we need a sticky header! There was a `stickyHeaderInverted` boolean prop but that didn't work. (Also no documentation for these.) Stackoverflow crawling indicates that a view outside the flatlist is the way to go.

Still need to figure out how to disable the shadow above the play bar.

### Dragons

### How Has This Been Tested?

Local ios stage

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/screens/chat-screen/ChatScreen.tsx
+++ b/packages/mobile/src/screens/chat-screen/ChatScreen.tsx
@@ -9,6 +9,7 @@ import {
   decodeHashId,
   encodeHashId,
   Status,
+  playerSelectors,
   isEarliestUnread
 } from '@audius/common'
 import { Portal } from '@gorhom/portal'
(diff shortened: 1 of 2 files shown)


packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -168,7 +172,9 @@ export const ChatScreen = () => {
   const messageTop = useRef(0)
   const chatContainerTop = useRef(0)
   const chatContainerBottom = useRef(0)
+  const isKeyboardOpen = useSelector(getIsKeyboardOpen)
 
+  const hasTrack = useSelector(getHasTrack)
```
schottra, line 177: Minor nit: At first read, this isn't clear to me that `hasTrack` means "there's a player control showing that we need to account for". Maybe either add a comment or rename the variable to make it more obvious?

schottra (approved):

packages/mobile/src/components/now-playing-drawer/NowPlayingDrawer.tsx
```diff
@@ -118,6 +119,13 @@ export const NowPlayingDrawer = memo(function NowPlayingDrawer(
 
   const { drawerNavigation } = useContext(AppDrawerContext)
 
+  // Don't show shadow when we're in chat screen, so that the drawer appears
+  // at same level as the chat screen textinput
+  const isInChatScreen = useNavigationState((state) => {
```
dylanjeffers, line 124: nice! yeah i think its def hacky, but at least consistent with the other hack we have for this type of introspection

dylanjeffers (approved):
nice, what can you do sometimes 😅

PR #3298, opened 2023-05-02
[PAY-1178] Fix new messages on new/unloaded chats

rickyrombo:
### Description

Sending a message in a new or unloaded chat for the other user would fail if they have the chats page open, as `addMessage` was getting called for a chat that didn't exist and the `hasTail` check expects there to be one.

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/common/src/store/pages/chat/slice.ts
+++ b/packages/common/src/store/pages/chat/slice.ts
@@ -164,11 +164,13 @@ const slice = createSlice({
         return { ...item, hasTail: hasTail(item, data[index - 1]) }
       })
       // Recalculate hasTail for latest message of new batch
-      if (state.messages[chatId].ids.length > 0) {
+      if (state.messages[chatId] && state.messages[chatId].ids.length > 0) {
         const prevEarliestMessageId =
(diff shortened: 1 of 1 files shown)


audius-infra:
Preview this change https://demo.audius.co/mjp-dms-check-tail

packages/common/src/store/pages/chat/slice.ts
```diff
@@ -164,11 +164,13 @@ const slice = createSlice({
         return { ...item, hasTail: hasTail(item, data[index - 1]) }
       })
       // Recalculate hasTail for latest message of new batch
-      if (state.messages[chatId].ids.length > 0) {
+      if (state.messages[chatId] && state.messages[chatId].ids.length > 0) {
```
dharit-tan, line 167: Think I forgot `chatsAdapter.getSelectors` existed when I wrote this, should we be using `selectById` here instead of `state.messages[chatId]`?

rickyrombo: No because those selectors are from the state root and the reducers are from the slice root

dharit-tan (approved):

PR #3308, opened 2023-05-03
Fix index issue with constructUrl

dylanjeffers:
### Description

Fixes issue where we are not indexing the param, just creating an array `[1]` 😢


audius-infra:
Preview this change https://demo.audius.co/dj-mp-fix-construct-url

raymondjacobson (approved):
wow

PR #3316, opened 2023-05-03
[C-2560] Improve checking for new notifs

dylanjeffers:
### Description

Improves `checkForNewNotifications` flow by making better checks for if notifications have changed. right now its really flaky because it compares notification order, which is incorrect since the api is not always accurate with this compared to the slice. By checking for "new" notifs or checking existing ones, instead of caring about index, this makes this much less flaky.

### DIFF
--- a/packages/web/src/common/store/notifications/checkForNewNotificationsSaga.ts
+++ b/packages/web/src/common/store/notifications/checkForNewNotificationsSaga.ts
@@ -6,6 +6,7 @@ import {
   NotificationType,
   walletActions
 } from '@audius/common'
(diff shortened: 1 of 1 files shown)


sabrina-kiam (approved):
lgtm! confirming this coveres if a notification's actions change?? i can't remember if that was an issue or not

dylanjeffers:
> lgtm! confirming this coveres if a notification's actions change?? i can't remember if that was an issue or not

yes it does, its captured by the userIds and entityIds check!

sabrina-kiam:
sweet thanks dylan!

PR #3317, opened 2023-05-03
[C-2393] Cache pods in CI

sliptype:
### Description
image

* Cache pods in mobile ci
* Ideally we would use the same version of `cocoapods` on CI as we do locally. Right now I've hardcoded the version to `1.12.0`, but in the future it may be better to include `cocoapods` in our Gemfile.lock
  Also added the `--deployment` flag which will cause the step to fail if any changes are generated in the Podfile.lock during installation

See: https://github.com/CocoaPods/CocoaPods/issues/8848

### Dragons

If we commit Podfile.lock with a new cocoapods version, ci will fail until we update the version in ci

### How Has This Been Tested?

Ran mobile ci pipeline and confirmed that cached pods improved run time

### Feature Flags ###

### DIFF
--- a/.circleci/src/commands/@mobile-commands.yml
+++ b/.circleci/src/commands/@mobile-commands.yml
@@ -34,8 +34,8 @@ mobile-prepare-ios:
     - attach_workspace:
         at: ./
     - run:
-        name: update cocoapods
-        command: sudo gem install cocoapods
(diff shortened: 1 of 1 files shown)


nicoback2 (approved):

Kyle-Shanks (approved):
Neat!

PR #3321, opened 2023-05-03
Fix search-results saga error due to yield*

dylanjeffers:
### Description

Fixes issue in search-results saga where we use yield* instead of yield since it's a .js file


audius-infra:
Preview this change https://demo.audius.co/dj-fix-search-results

sliptype (approved):

PR #3324, opened 2023-05-03
Update fetchPermission condition

dylanjeffers:
### Description

Small update to fetchProfile that gates `fetchPermission` around `CHAT_ENABLED` flag. this error is showing up in sentry, even if it's minor :))


rickyrombo (approved):
<3 thanks

PR #3325, opened 2023-05-03
Get call to action for chat permissions

rickyrombo:
### Description

Paves the way for the new UI describing why you can't message someone, by adding a new selector that infers the reason based on the permissions for a user.

### Dragons

### How Has This Been Tested?

### Feature Flags ###


packages/common/src/store/pages/chat/selectors.ts
```diff
@@ -150,3 +151,62 @@ export const getUnfurlMetadata = (
   const message = getChatMessageById(state, chatId, messageId)
   return message?.unfurlMetadata
 }
+
+export const getUserChatPermissions = createSelector(
```
rickyrombo, line 155: fyi: @dharit-tan breaking change for mobile potentially

packages/common/src/store/pages/chat/selectors.ts
```diff
@@ -150,3 +151,62 @@ export const getUnfurlMetadata = (
...
+    const currentUserPermissions = currentUserId
+      ? chatPermissions[currentUserId]
+      : null
+    const userPermissions = chatPermissions[userId]
+    const isBlockee = blockees.includes(userId)
+    const isBlocker = blockers.includes(userId)
+    const canChat =
+      !isBlockee &&
+      !isBlocker &&
+      (userPermissions?.current_user_has_permission ?? true)
```
michellebrier, line 188: this wont be false bidirectionally yet, so the logic below that calcs actions based on the `currentUserPermissions` gated with (`!canChat)` wont be evaluated if the non-current user is not a blocker/blockee and has permissions such that the current user is allowed to chat with them. 

tldr we need a bidirectional `current_user_has_permission` for this to work, and we can calc the action on the backend or the frontend but would prefer to just keep it all contained in the backend (and frontend reads a `to_enable` field in the resp)

rickyrombo: ahh yeah ok that checks out. Thanks!

packages/common/src/store/pages/chat/selectors.ts
dharit-tan: Love this

packages/common/src/store/pages/chat/selectors.ts
rickyrombo: @michellebrier / @dharit-tan does this work now that we have unidirectional permissions?

dharit-tan: Think so

dharit-tan (approved):

dharit-tan:
Lint + merge main!

audius-infra:
Preview this change https://demo.audius.co/mjp-dms-permissions-action

audius-infra:
Preview this change https://demo.audius.co/mjp-dms-permissions-action

PR #3327, opened 2023-05-04
[C-2566] Prevent player error on skip

dylanjeffers:
### Description

Prevents play error on skip.

### Dragons

Not sure if simply calling `.remove()` is enough, or if setting `src = ''` is also required. @raymondjacobson for thoughts 😄


raymondjacobson (approved):
Trial & error is how i've come to understand this stuff. Seems reasonable. when it's on stage, i can try to repro the other bug

PR #3329, opened 2023-05-04
DMs: Fix unread indicator not going away

rickyrombo:
### Description

Still some more bugs here to iron out I think, but this should help?

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/common/src/store/pages/chat/sagas.ts
+++ b/packages/common/src/store/pages/chat/sagas.ts
@@ -211,7 +211,7 @@ function* doMarkChatAsRead(action: ReturnType<typeof markChatAsRead>) {
     const audiusSdk = yield* getContext('audiusSdk')
     const sdk = yield* call(audiusSdk)
     const chat = yield* select((state) => getChat(state, chatId))
-    if (!chat || dayjs(chat?.last_read_at).isBefore(chat?.last_message_at)) {
+    if (!chat || !dayjs(chat?.last_read_at).isAfter(chat?.last_message_at)) {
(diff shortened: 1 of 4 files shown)


packages/common/src/store/pages/chat/sagas.ts
```diff
@@ -211,7 +211,7 @@ function* doMarkChatAsRead(action: ReturnType<typeof markChatAsRead>) {
     const audiusSdk = yield* getContext('audiusSdk')
     const sdk = yield* call(audiusSdk)
     const chat = yield* select((state) => getChat(state, chatId))
-    if (!chat || dayjs(chat?.last_read_at).isBefore(chat?.last_message_at)) {
+    if (!chat || !dayjs(chat?.last_read_at).isAfter(chat?.last_message_at)) {
```
dharit-tan, line 214: ??? plz explain

rickyrombo: wanted to allow it to mark as read if the timestamps were equal, in case it was reading the optimistically set value

dharit-tan (approved):

packages/common/src/store/pages/chat/sagas.ts
dharit-tan: maybe add comment for that

rickyrombo: I think it reads pretty straightforward in context: if the last read isn't after the last message, call chats.read

PR #3334, opened 2023-05-04
Remove canBeUnlisted and clean up fetchTrack

dylanjeffers:
### Description

Removes `canBeUnlisted` option for `fetchTrack` flow, instead favoring `retrieveByHandleAndSlug` for most requests, and `retrieveTracks` for the case where there is only a trackId.

This should essentially cover all our cases:
   - If you have a track permalink, you can view the track, whether or not it's hidden
   - If you only have the trackId, you can only view the track if its public.

### Dragons
  - Need to ensure stem uploads on a hidden track work with this new system. In particular @rickyrombo can i use `track.title` for slug, or should i use `permalink` and grab the slug part of it?

### DIFF
--- a/packages/common/src/store/pages/track/actions.ts
+++ b/packages/common/src/store/pages/track/actions.ts
@@ -38,7 +38,6 @@ export const fetchTrack = (
   trackId: Nullable<number>,
   slug?: string,
   handle?: string,
-  canBeUnlisted?: boolean,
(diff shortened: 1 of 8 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-remove-can-be-unlisted

raymondjacobson (approved):
lgtm, want to thoroughly test on stage

PR #3337, opened 2023-05-04
[PAY-1202] Refactor saved collections fetching

schottra:
### Description
This is the first round of changes to improve the saved albums/playlists experience. This PR focuses on updates to the fetching logic.
Previously, the fetching was done in account sagas (`fetchSavedAlbumsAsync` and `fetchSavedPlaylistsAsync`), and all records were fetched in one API call. This breaks for more than about 80 ids passed. We were also using a large provider to fetch the data and then prop-drilling it down to `SavedPage` and its children. 

Changes
* Added new slice, selectors, and saga specifically for fetching saved collections. The saga includes a helper that will break lists of collections to be fetched into smaller chunks if they exceed a conservative limit (50) that might cause an issue with the API
* Updated all places using `retrieveCollections` directly to to use the new saga helper.
* Added hooks for fetching saved albums and playlists(TODO). Implemented a saved albums hook which allows for `fetchMore` style pagination of full playlist metadata. It uses derived selectors to determine which playlists from the fully fetched album id list do _not_ have local metadata. `fetchMore` will collect the first `page_size` ids and initiate a fetch for them. This will be used in future work to implement an `InfiniteScroll` component. (_Note_: For now, we are using the new hook with a page size of 9999, which has the same effect as the previous "load everything at once" behavior)
* Removed albums-related logic from `SavedPageProvider`
* Refactored album rendering code in `web` to follow the mobile patterns
  * Added content components for displaying saved albums which use the new hook to fetch the data instead of consuming it through props
  * Moved logic for album cards into separate components
  * Removed a lot of prop drilling for utilities which could be consumed directly through hooks or constants
* Added `useGoToRoute` hook for components to consume if they need to push routes. We don't need to pass this down through props.
* Fixed some propType warnings with card components

There are some follow-up issues that I will address in future PRs, but nothing that wasn't already broken :-)

### Dragons
This does change fetching logic on both mobile and web, though it should be in a way that doesn't break anything for users which were not _already_ broken (i.e. users who had too many playlists to fetch safely)

### How Has This Been Tested?
Tested on desktop, mobile web, and local emulator to ensure playlists and albums still load correctly.

Will test as part of client release process. Will monitor sentry and rollback if any major issues occur.

### Feature Flags ###
N/A


audius-infra:
Preview this change https://demo.audius.co/pay-1202-refactor-saved-collections

audius-infra:
Preview this change https://demo.audius.co/pay-1202-refactor-saved-collections

sliptype (approved):
Awesome work! All looks good to me

packages/common/src/store/reducers.ts
```diff
@@ -263,6 +268,8 @@ export type CommonState = {
   tracks: TracksCacheState
   users: UsersCacheState
 
+  savedCollections: SavedCollectionsState
```
piazzatron, line 271: nit: IMO a bit cleaner to use the `ReturnType` of the reducer here, and then we don't have to export the state

packages/common/src/store/saved-collections/types.ts
piazzatron, line 1: We've tended to colocate action types in their corresponding slice in the past – this is nice too but would opt for consistency one way or the other (cc @sliptype)

packages/web/src/common/store/saved-collections/sagas.ts
```diff
@@ -0,0 +1,51 @@
...
+const COLLECTIONS_BATCH_LIMIT = 50
+
+/** Fetches an arbitrarily long list of collection ids in reasonably-sized chunks. Helpful both to
+ * reduce request load and prevent errors from extremely long URLs.
+ */
+export function* fetchCollectionChunks(ids: ID[]) {
+  // Request ids in chunks if we're asked for too many
+  const chunks = chunk(ids, COLLECTIONS_BATCH_LIMIT)
+  for (let i = 0; i < chunks.length; i += 1) {
+    yield call(retrieveCollections, null, chunks[i])
```
piazzatron, line 26: I wonder if we should rename/deprecate `retrieveCollections` or add a comment indicating that it doesn't paginate and that future callers should hit `fetchCollectionChunks` instead?

packages/web/src/components/actions-tab/ActionsTab.js
```diff
@@ -248,14 +248,14 @@ ActionsTab.propTypes = {
...
+  isPublic: PropTypes.bool,
   isDisabled: PropTypes.bool,
   includeEdit: PropTypes.bool,
   direction: PropTypes.oneOf(['vertical', 'horizontal']),
   variant: PropTypes.oneOf(['track', 'playlist', 'album']),
   containerStyles: PropTypes.string,
-  currentUserReposted: PropTypes.boolean,
-  currentUserSaved: PropTypes.boolean,
+  currentUserReposted: PropTypes.bool,
+  currentUserSaved: PropTypes.bool,
```
piazzatron, line 258: ty for this 😍

packages/web/src/hooks/useGoToRoute.ts
```diff
@@ -0,0 +1,9 @@
+import { useCallback } from 'react'
+
+import { push as pushRoute } from 'connected-react-router'
+import { useDispatch } from 'react-redux'
+
+export function useGoToRoute() {
```
piazzatron, line 6: 🔥

packages/web/src/pages/saved-page/components/desktop/SavedPage.tsx
```diff
@@ -39,6 +44,80 @@ const messages = {
...
+    )
+  })
+
+  return cards.length > 0 ? (
+    <CardLineup cards={cards} cardsClassName={styles.cardsContainer} />
+  ) : (
+    <EmptyTable
+      primaryText='You haven’t favorited any albums yet.'
+      secondaryText='Once you have, this is where you’ll find them!'
+      buttonLabel='Go to Trending'
```
piazzatron, line 115: nit: could pull these out into a `messages` object as campground cleaning (this file might have been around prior to that convention!)

packages/web/src/pages/saved-page/components/desktop/SavedPage.tsx
```diff
@@ -39,6 +44,80 @@ const messages = {
   filterPlaceholder: 'Filter Tracks'
 }
 
+type AlbumCardProps = Pick<CardProps, 'index' | 'isLoading' | 'setDidLoad'> & {
+  album: CollectionWithOwner
+}
+
+const AlbumCard = ({ album, index, isLoading, setDidLoad }: AlbumCardProps) => {
```
piazzatron, line 51: Good looks pulling these out into standalone components

packages/common/src/hooks/useSavedCollections.ts
```diff
@@ -0,0 +1,84 @@
...
+  const { unfetched: unfetchedAlbums, fetched: albumsWithDetails } =
+    useSelector(getAlbumsWithDetails)
+  const { status } = useSelector(getSavedAlbumsState)
+
+  const fetchMore = useCallback(() => {
+    if (status === Status.LOADING || unfetchedAlbums.length === 0) {
+      return
+    }
+    const ids = unfetchedAlbums
+      .slice(0, Math.min(pageSize, unfetchedAlbums.length))
```
piazzatron, line 51: Doesn't the saga layer already chunk this via your `fetchCollectionChunks` fn?

packages/web/src/pages/saved-page/SavedPageProvider.tsx
piazzatron, line 1: Could we lose `getFilteredAlbums`?

piazzatron (approved):
Very clean and easy to review! Only nits :)

packages/common/src/hooks/useSavedCollections.ts
schottra: Yes, but there are two layers. This is intended for future use where we are calling `fetchMore` here and it's fetching the next X album records using the underlying saga. This has a configurable page size based on how many you want to fetch at once in a particular component.
The saga has no notion of page sizes. It's just a bulk fetch that guards against 400s by chunking up the list if you ask for too many at once.
For most purposes, the `pageSize` here will be much less than the chunk size in the saga. (i.e. we shoudn't ever trigger the saga's chunking.

packages/common/src/store/saved-collections/types.ts
schottra: I was using `notifications` as the most recent example, [which puts types in a separate file](https://github.com/AudiusProject/audius-client/blob/main/packages/common/src/store/notifications/types.ts).

audius-infra:
Preview this change https://demo.audius.co/pay-1202-refactor-saved-collections

packages/web/src/pages/saved-page/SavedPageProvider.tsx
schottra: Oh yep. Good catch. I had left it until I decided what to do about filtering. But it won't be used either way.

packages/web/src/pages/saved-page/components/desktop/SavedPage.tsx
schottra: Oh yep. Definitely just moved code around without looking at it here. Will put into a messages object.

packages/common/src/hooks/useSavedCollections.ts
```diff
@@ -0,0 +1,84 @@
...
+type UseSavedAlbumDetailsConfig = {
+  pageSize?: number
+}
+export function useSavedAlbumsDetails({
+  pageSize = DEFAULT_PAGE_SIZE
+}: UseSavedAlbumDetailsConfig) {
+  const dispatch = useDispatch()
+  const [hasFetched, setHasFetched] = useState(false)
+  const { unfetched: unfetchedAlbums, fetched: albumsWithDetails } =
+    useSelector(getAlbumsWithDetails)
```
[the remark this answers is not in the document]
schottra: I can update the selector name and add a comment

packages/common/src/hooks/useSavedCollections.ts
piazzatron: makes sense!

packages/common/src/store/saved-collections/types.ts
schottra: Got it. The other complicating factor here is that there are types shared between the `slice` and `selectors` module and I don't really want to import `selectors` just to get at the types.
Historically, in TS, as soon as a type is used in multiple places I pull it out into a types file to remove the future possibility of circular dependencies.

So I can move just the action types if that's the pattern we are going for. But we will still have the `types` file.

packages/web/src/common/store/saved-collections/sagas.ts
schottra: Updated to do this in `retrieveCollections`. Turns out we were already doing chunking for track fetching! So I updated both patterns to match.

schottra:
@piazzatron @sliptype Mind looking at these two changes as well?
https://github.com/AudiusProject/audius-client/pull/3337/commits/5a483562ff68eaedb8472e4e29c85b81ed2bcd98
https://github.com/AudiusProject/audius-client/pull/3337/commits/2863089a6e0ec248ea28f564180dfb547739e59b

audius-infra:
Preview this change https://demo.audius.co/pay-1202-refactor-saved-collections

sliptype (approved):

audius-infra:
Preview this change https://demo.audius.co/pay-1202-refactor-saved-collections

PR #3340, opened 2023-05-04
[PAY-1221] Fix message hasTail recalculations when server resends messages

dharit-tan:
### Description
Now that the backend is resending recent messages, hasTail calculations were getting re-triggered for old messages and they would randomly show tails when websocket reconnects happened. Unfortunately that means we do need to check if the incoming message already exists and don't add if it does.

Extracted hasTail recalculation to a helper function to be used in both `addMessages` and `fetchMoreMessagesSucceeded`.

### Dragons

### How Has This Been Tested?

local ios stage

### Feature Flags ###

### DIFF
--- a/packages/common/src/store/pages/chat/slice.ts
+++ b/packages/common/src/store/pages/chat/slice.ts
@@ -79,7 +79,7 @@ const {
   selectIds: getAllMessageIds
 } = chatMessagesAdapter.getSelectors()
 
-// Recalculate hasTail for the message at index - 1 (the previous message of
+// Recalculate hasTail for the message at index + 1 (the previous message of
 // the message at index).
 const recalculatePreviousMessageHasTail = (
   chatState: EntityState<ChatMessageWithExtras>,
--- a/packages/common/src/store/pages/chat/slice.ts
+++ b/packages/common/src/store/pages/chat/slice.ts
@@ -73,7 +73,30 @@ export const chatMessagesAdapter = createEntityAdapter<ChatMessageWithExtras>({
   sortComparer: messageSortComparator
 })
 
-const { selectById: getMessage } = chatMessagesAdapter.getSelectors()
+const {
+  selectById: getMessage,
+  selectEntities: getAllMessages,
+  selectIds: getAllMessageIds
+} = chatMessagesAdapter.getSelectors()
+
+// Index of message for which to recalculate the hasTail property
+// of the _previous_ message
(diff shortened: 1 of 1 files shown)


audius-infra:
Preview this change https://demo.audius.co/rt-recalculate-hastail

packages/common/src/store/pages/chat/slice.ts
```diff
@@ -73,7 +73,30 @@ export const chatMessagesAdapter = createEntityAdapter<ChatMessageWithExtras>({
...
 })
 
-const { selectById: getMessage } = chatMessagesAdapter.getSelectors()
+const {
+  selectById: getMessage,
+  selectEntities: getAllMessages,
+  selectIds: getAllMessageIds
+} = chatMessagesAdapter.getSelectors()
+
+// Recalculate hasTail for the message at index - 1 (the previous message of
```
sddioulde, line 82: nit: maybe update comment and say index + 1 and explain that message oder is backwards?

packages/common/src/store/pages/chat/slice.ts
```diff
@@ -73,7 +73,30 @@ export const chatMessagesAdapter = createEntityAdapter<ChatMessageWithExtras>({
...
+  const messageIds = getAllMessageIds(chatState)
+  const chatMessages = getAllMessages(chatState)
+  const prevMessageId = messageIds[index + 1]
+  const prevMessage = chatMessages[prevMessageId]
+  if (!prevMessage) return
+  const newMessageId = messageIds[index]
+  const newMessage = chatMessages[newMessageId]
+  chatMessagesAdapter.updateOne(chatState, {
+    id: prevMessageId,
+    changes: { hasTail: hasTail(prevMessage, newMessage) }
```
schottra, line 97: Is it worth determining if `hasTail` has actually changed before calling `updateOne`? If `updateOne` results in a new state that would cause a re-render, we may not want to do that unless the value has actually changed?

sddioulde (approved):
looks good to me, but maybe someone else with more context should also check

packages/common/src/store/pages/chat/slice.ts
dharit-tan: So you were right, it actually is an object reassign so it would re-render. However it's only for 1 message at a time which I think is fine. It won't re-render the entire screen.

audius-infra:
Preview this change https://demo.audius.co/rt-recalculate-hastail

PR #3341, opened 2023-05-05
[C-2519] Ignore TOS deeplink

dylanjeffers:
### Description

Updates apple-site-association file to ignore /legal/terms-of-use url. Unfortunately we cannot do the same thing for android. Their intent-filter's dont allow exclusions, and since we have audius.co/user/... scheme there is no way to exclude this specific url.

This SO answer is super clear regarding android limitations: https://stackoverflow.com/a/45247815

### DIFF
--- a/packages/web/public/.well-known/apple-app-site-association
+++ b/packages/web/public/.well-known/apple-app-site-association
@@ -1,11 +1,15 @@
 {
-    "applinks": {
-        "apps": [],
-        "details": [
-            {
-                "appID": "LRFCG93S85.co.audius.audiusmusic",
-                "paths": ["NOT /oauth/auth*", "*"]
-            }
+  "applinks": {
+    "apps": [],
+    "details": [
+      {
+        "appID": "LRFCG93S85.co.audius.audiusmusic",
+        "paths": [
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-c-2519-ignore-tos-deeplink

packages/web/public/.well-known/apple-app-site-association
```diff
@@ -1,11 +1,15 @@
...
-                "paths": ["NOT /oauth/auth*", "*"]
-            }
+  "applinks": {
+    "apps": [],
+    "details": [
+      {
+        "appID": "LRFCG93S85.co.audius.audiusmusic",
+        "paths": [
+          "NOT /oauth/auth*",
+          "NOT /legal/terms-of-use",
```
raymondjacobson, line 9: privacy-policy and service-terms as well. maybe we should just legal\*?

raymondjacobson (approved):

audius-infra:
Preview this change https://demo.audius.co/dj-c-2519-ignore-tos-deeplink

PR #3342, opened 2023-05-05
[PAY-1219] Fix mobile chat reactions popup message getting cut off

dharit-tan:
### Description
For some reason `onLayout` for the bottom TextInput's parent view was getting fired when the keyboard shows, but not when the keyboard hides? Weird.. This would cause the ref that stores the flatlist content container height to be stuck at the smaller height it has when the keyboard is shown. But luckily since we have our own handrolled KeyboardAvoidingView, could just add a callback to the show/hide handlers and trigger them from there manually.

### Dragons

### How Has This Been Tested?

Local ios stage

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/core/KeyboardAvoidingView.tsx
+++ b/packages/mobile/src/components/core/KeyboardAvoidingView.tsx
@@ -20,8 +20,8 @@ type KeyboardAvoidingViewProps = {
   keyboardHidingDuration?: number
   // Offset is subtracted from the desired height when the keyboard is showing.
   keyboardShowingOffset?: number
-  onKeyboardShowCallback?: () => void
-  onKeyboardHideCallback?: () => void
+  onKeyboardShow?: () => void
+  onKeyboardHide?: () => void
   children: React.ReactNode
 }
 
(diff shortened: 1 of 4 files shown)


packages/mobile/src/components/core/KeyboardAvoidingView.tsx
```diff
@@ -20,6 +20,8 @@ type KeyboardAvoidingViewProps = {
   keyboardHidingDuration?: number
   // Offset is subtracted from the desired height when the keyboard is showing.
   keyboardShowingOffset?: number
+  onKeyboardShowCallback?: () => void
```
schottra, line 23: nit: 'Callback' is redundant? I feel like the normal pattern is `onEvent` as a prop and `handleEvent` as the handler

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -438,6 +448,7 @@ export const ChatScreen = () => {
               styles.keyboardAvoiding,
               hasCurrentlyPlayingTrack ? { bottom: PLAY_BAR_HEIGHT } : null
             ]}
+            onKeyboardHide={() => measureView(composeRef, chatContainerBottom)}
```
schottra, line 451: This is going to change on every render. Might be worth pulling it out into a `useCallback` if `composeRef` and `chatContainerBottom` don't change on every render.

rickyrombo (approved):

PR #3343, opened 2023-05-05
[C-2577] Fix mobile web social buttons by removing pull-to-refresh

dylanjeffers:
### Description

Fixes runtime error when unauthenticated users try to like or repost content. This is weirdly due to lottie eventListener bug in pull-to-refresh. The fix for now is just removing pull-to-refresh entirely for mobile web, since all modern mobile browers implement their own pull to refresh anyway. chatted with ray about this and is down for removing


sliptype (approved):
Less code on mobile web?! Love it

PR #3344, opened 2023-05-05
[C-2565] Improve mobile app-redirect-popover experience

dylanjeffers:
### Description

Improves app-redirect-popover experience in a few key ways:
- Only show it once per session
- Allow users to dismiss by clicking background

- Also drastically improves code simplicity

> This was all vetted with product


Kyle-Shanks (approved):
neat

PR #3352, opened 2023-05-10
[PAY-1032][PAY-892] Mobile DMs unread indicator, prefetch chats

dharit-tan:
### Description
Purple unread circle to indicate unread chats. 
To enable this, also fetching chats + connecting to websockets at root screen load, and moved fetching chat messages for the first 10 messages on ChatListScreen load.
Changed `takeLatest` to `takeEvery` so multiple `fetchMoreMessages` calls in a row would all go through.

### Dragons

### How Has This Been Tested?

Local ios stage 

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/rt-unread-purple

packages/common/src/store/pages/chat/sagas.ts
```diff
@@ -392,7 +392,7 @@ function* watchFetchChats() {
 }
 
 function* watchFetchChatMessages() {
-  yield takeLatest(fetchMoreMessages, doFetchMoreMessages)
+  yield takeEvery(fetchMoreMessages, doFetchMoreMessages)
```
rickyrombo, line 395: would you mind also doing a pass of the `doFetchMoreMessages` saga and making sure we take the proper max/min of the pointers?

packages/common/src/store/pages/chat/selectors.ts
```diff
@@ -92,6 +92,11 @@ export const getChatMessages = createSelector(
   }
 )
 
+export const getHasUnreadMessages = (state: CommonState) => {
```
rickyrombo, line 95: fwiw I don't think this is sufficient - the user might have older unread chats from chats beyond the fetched chats page size. @michellebrier added a new endpoint for checking if there are any unreads

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -207,12 +207,16 @@ export const ChatScreen = () => {
...
   const chatFrozenRef = useRef(chat)
 
+  // Initial fetch, but only if messages weren't fetched on app load
   useEffect(() => {
-    if (chatId && (chat?.messagesStatus ?? Status.IDLE) === Status.IDLE) {
-      // Initial fetch
+    if (
+      chatId &&
+      (chat?.messagesStatus ?? Status.IDLE) === Status.IDLE &&
+      chatMessages.length === 0
```
rickyrombo, line 215: possible edge case: chats that are created in the same session with no message sent? Though I _think_ with the idle check it's fine

packages/common/src/store/pages/chat/sagas.ts
dharit-tan: sorry not sure what you mean by this? logic seems ok to me.

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
dharit-tan: Yeah we'd probably fetch once, which would fail, but then status will be `SUCCESS`. Let me verify.

packages/common/src/store/pages/chat/sagas.ts
rickyrombo: https://github.com/AudiusProject/audius-client/blob/72fa511ef95cba111f67976c8ab24b6dfb9c12cf/packages/common/src/store/pages/chat/slice.ts#L194

We take the summary from the latest fetch, because we assume that's going to have the latest pointer/cursor, but we should probably be getting the min and max so the next fetch starts at the appropriate place regardless of order

audius-infra:
Preview this change https://demo.audius.co/rt-unread-purple

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
dharit-tan: Yeah we fetch once on chat creation, which succeeds with 0 messages, and then on subsequent loads to that chat screen nothing happens due to idle check.

dharit-tan:
Still need to figure out how to decrement global unread count during `markChatAsRead`, probably just fetch unread count in the saga there.

audius-infra:
Preview this change https://demo.audius.co/rt-unread-purple

packages/mobile/src/screens/chat-screen/ChatListScreen.tsx
```diff
@@ -109,24 +111,28 @@ export const ChatListScreen = () => {
...
+  // Prefetch messages for initial loaded chats
+  useEffect(() => {
+    if (
+      chats.length > 0 &&
+      chats.every(
+        (chat) => !chat.messagesStatus || chat.messagesStatus === Status.IDLE
+      )
+    ) {
+      chats.slice(0, CHATS_MESSAGES_PREFETCH_LIMIT).map((chat) => {
+        dispatch(fetchMoreMessages({ chatId: chat.chat_id }))
```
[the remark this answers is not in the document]
rickyrombo: Yeah let's leave it as is, if only to focus on other issues - can revisit if there are perf issues

audius-infra:
Preview this change https://demo.audius.co/rt-unread-purple

packages/common/src/store/pages/chat/slice.ts
```diff
@@ -175,24 +188,27 @@ const slice = createSlice({
...
-      chatsAdapter.updateOne(state.chats, {
-        id: chatId,
-        changes: { messagesStatus: Status.SUCCESS, messagesSummary: summary }
-      })
+      if (!summary) {
+        console.error('fetchMoreMessagesSucceeded: no summary')
+        return
+      }
+      const existingSummary = state.chats.entities[chatId]?.messagesSummary
+      if (summary.next_count > (existingSummary?.next_count ?? -1)) {
```
rickyrombo, line 196: I think we really want to take the max of the `next_cursor` and the min of the `prev_cursor` (or vice versa? can't remember which way they go)?

rickyrombo (approved):

packages/common/src/store/pages/chat/selectors.ts
```diff
@@ -93,6 +93,25 @@ export const getChatMessages = createSelector(
...
+ * to coalesce count fetched from backend + optimistic reads, to be used in
+ * getHasUnreadMessages.
+ */
+export const getUnreadMessagesCount = createSelector(
+  [getChats, (state: CommonState) => state.pages.chat.unreadMessagesCount],
+  (chats, unreadCount) => {
+    const optimisticUnreadCount = chats.reduce((acc, currentChat) => {
+      return acc + (currentChat.unread_message_count ?? 0)
+    }, 0)
+    return Math.min(unreadCount, optimisticUnreadCount)
```
rickyrombo, line 108: On thinking further...

- `markChatAsRead` sets `unread_message_count` to 0
- `incrementUnreadCount` adds 1 to `unread_message_count`

but neither of those right now impact the `unreadCount` pulled from the new SDK endpoint. That means:

- if `unreadCount` is 0, but we get a new message in that session, optimisticUnreadCount will be 1 but this method returns zero
- if `unreadCount` is positive, but we optimistically mark every chat _loaded_ as read, optimistic read count will be 0 and the method returns zero (but `unreadCount` may still be counting unreads for chats we haven't paginated through yet, there may still be unreads)

So unfortunately I don't think we can take the min or max here, I think we do need to independently update the store for `unreadMessagesCount` in `incrementUnreadCount` and `markAsRead`

dharit-tan: I am incrementing global `unreadMessagesCount` in `incrementUnreadCount`, so that takes care of the first case. But yeah I think you're right in 2nd case.

packages/common/src/store/pages/chat/slice.ts
```diff
@@ -360,6 +376,7 @@ const slice = createSlice({
           changes: { unread_message_count: existingUnreadCount + 1 }
         })
       }
+      state.unreadMessagesCount += 1
```
rickyrombo, line 379: ah, sneaky! I missed this sorry

packages/common/src/store/pages/chat/slice.ts
dharit-tan: Chatted with @rickyrombo on slack, his way is the way to go. "the summary we get back is for that query but I want the store to have the summary for all the stuff itself."

dharit-tan: Turns out we want min of both i believe.

packages/common/src/store/pages/chat/slice.ts
```diff
@@ -140,6 +141,7 @@ const slice = createSlice({
       action: PayloadAction<{ unreadMessagesCount: number }>
     ) => {
       state.unreadMessagesCount = action.payload.unreadMessagesCount
+      state.optimisticUnreadMessagesCount = action.payload.unreadMessagesCount
```
rickyrombo, line 144: maybe don't do this actually... this will clear any pending optimistic changes. Perhaps instead of an "optimisticUnreadMessagesCount" we should keep track of "optimisticUnreadMessageCountChanges"?

audius-infra:
Preview this change https://demo.audius.co/rt-unread-purple

rickyrombo (approved):

packages/common/src/store/pages/chat/slice.ts
rickyrombo: - UI shows 10 unreads
- start fetching unread count
- start mark chat as read that has 4 unreads
  - optimistic unread count decremented 4 to 6
  - UI shows 6 unreads
- fetch unread count succeeds with 10 since the write hasn't completed yet
  - optimistic unread count set to unread count, 10
  - UI shows 10 unreads
- mark chat as read succeeds
  - unread count updated to optimistic unread count (10)
  - optimistic unread count deleted
  - UI shows 10 unreads

packages/common/src/store/pages/chat/slice.ts
```diff
@@ -288,6 +323,7 @@ const slice = createSlice({
       // Set the true state
       const { chatId } = action.payload
       delete state.optimisticChatRead[chatId]
+      delete state.optimisticUnreadMessagesCount
```
rickyrombo, line 326: this action needs to update the unreadMessageCount.

the failure case doesn't, so that it goes back to normal

packages/common/src/store/pages/chat/slice.ts
dharit-tan: When should we reset `optimisticUnreadMessageCountChanges`? We need to reconcile it with the actual unread count eventually, right? Eg:
- On first load we fetch unread count, it's 4.
- We mark a chat as read that had all 4 unread messages, optimisticUnreadCountChanges = -4.
- 2 new messages come in on a different chat, optimisticUnreadCountChanges = -2
- Open left nav bar which re-fetches unreadCount, it is now 2.
- We would incorrectly show no unread messages bc unreadCount + optimisticUnreadCountChanges = 0.

rickyrombo: Hmm yeah this is quite tricky... Before I say anything, I think we could `delete state.optimisticUnreadMessagesCount` here and merge this as v1 and revisit before launch

Wonder if we could get away with only fetching the unread message count once, maybe on initial app load. The websocket is always listening, right? It'll "reconcile" on app kill + reopen. It sounds a little risky but it might not be...

Might have to sleep on this and think more

audius-infra:
Preview this change https://demo.audius.co/rt-unread-purple

PR #3353, opened 2023-05-12
C-2483 Fix queue overshot empty track player bug

nicoback2:
### Description
Before:  If a user skips to the end of their favorites lineup, they run into a blank/empty track player. From this state, if the user either clicks the play button or the rewind button, the final track in the lineup will start playing without updating the track player to indicate what is playing.

After: 
Load the autoplay queue early in order to avoid overshooting the queue.
Fallback: If the queue is still overshot, replay the last track while we wait for the autoplay queue to load.

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/web/src/common/store/queue/sagas.ts
+++ b/packages/web/src/common/store/queue/sagas.ts
@@ -117,17 +117,12 @@ function* handleQueueAutoplay({
   const length = yield* select(getLength)
   const shuffle = yield* select(getShuffle)
   const repeatMode = yield* select(getRepeat)
-  const isCloseToEndOfQueue = index + 1 >= length
-  const isOnlySongInQueue = index === 0 && length === 1
+  const isCloseToEndOfQueue = index + 9 >= length
   const isNotRepeating =
     repeatMode === RepeatMode.OFF ||
     (repeatMode === RepeatMode.SINGLE && (skip || ignoreSkip))
 
-  if (
-    !shuffle &&
-    isNotRepeating &&
-    (isCloseToEndOfQueue || isOnlySongInQueue)
-  ) {
+  if (!shuffle && isNotRepeating && isCloseToEndOfQueue) {
     yield* waitForAccount()
     const userId = yield* select(getUserId)
     yield* put(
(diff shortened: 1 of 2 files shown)


packages/web/src/common/store/queue/sagas.ts
```diff
@@ -117,17 +117,12 @@ function* handleQueueAutoplay({
   const length = yield* select(getLength)
   const shuffle = yield* select(getShuffle)
   const repeatMode = yield* select(getRepeat)
-  const isCloseToEndOfQueue = index + 1 >= length
-  const isOnlySongInQueue = index === 0 && length === 1
+  const isCloseToEndOfQueue = index + 9 >= length
```
dylanjeffers, line 120: Is 9 the right value here?

packages/web/src/common/store/queue/sagas.ts
```diff
@@ -117,17 +117,12 @@ function* handleQueueAutoplay({
...
   const isNotRepeating =
     repeatMode === RepeatMode.OFF ||
     (repeatMode === RepeatMode.SINGLE && (skip || ignoreSkip))
 
-  if (
-    !shuffle &&
-    isNotRepeating &&
-    (isCloseToEndOfQueue || isOnlySongInQueue)
-  ) {
+  if (!shuffle && isNotRepeating && isCloseToEndOfQueue) {
```
dylanjeffers, line 125: What is the situation with shuffle actually? Id assume if we get to the end of the queue in that case we'd want same logic?

dylanjeffers (approved):
Clean fix, and doing this in the middle of travel, nicee

audius-infra:
Preview this change https://demo.audius.co/nkang--empty-track-player

packages/web/src/common/store/queue/sagas.ts
nicoback2: shuffle mode basically jumps around to different parts of the queue so this logic wouldn't apply correctly. but yeah i wonder how we deal with that, prob worth qa'ing to see if we get a similar issue

PR #3356, opened 2023-05-15
[PAY-1215] Fix create new message crash

dharit-tan:
### Description
On a new chat, `getChatMessageById` would fail (used for the reactions popup on mobile) bc the `chatMessagesState` would be undefined (line 141) and we'd try to access `entities` of it.

### Dragons

### How Has This Been Tested?

Local ios stage + web stage - tested with user Saliou_2 as well as others.

### Feature Flags ###

### DIFF
--- a/packages/common/src/store/pages/chat/selectors.ts
+++ b/packages/common/src/store/pages/chat/selectors.ts
@@ -128,6 +128,7 @@ export const getChatMessageByIndex = (
   messageIndex: number
 ) => {
   const chatMessagesState = state.pages.chat.messages[chatId]
+  if (!chatMessagesState) return
   const messageIds = getChatMessageIds(chatMessagesState)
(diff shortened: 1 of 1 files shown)


packages/common/src/store/pages/chat/selectors.ts
```diff
@@ -128,6 +128,7 @@ export const getChatMessageByIndex = (
   messageIndex: number
 ) => {
   const chatMessagesState = state.pages.chat.messages[chatId]
+  if (!chatMessagesState) return
```
rickyrombo, line 131: should we return undefined or null?

rickyrombo (approved):

packages/common/src/store/pages/chat/selectors.ts
rickyrombo: Makes sense - I think being explicit is good

PR #3361, opened 2023-05-16
Update edit playlist flow and components

Kyle-Shanks:
### Description
Updating the edit playlist flow to conform with the new designs. Keeping the legacy flow and adding the new stuff behind a feature flag

Screenshot 2023-05-16 at 12 41 15 PM

Screenshot 2023-05-16 at 12 41 27 PM

### Dragons

N/A

### How Has This Been Tested?

Manually Tested

N/A

### Feature Flags ###

PLAYLIST_UPDATES_PRE_QA


packages/mobile/src/screens/app-screen/AppTabScreen.tsx
```diff
@@ -180,11 +185,13 @@ export const AppTabScreen = ({ baseScreen, Stack }: AppTabScreenProps) => {
         component={CollectionScreen}
         options={screenOptions}
       />
-      <Stack.Screen
-        name='EditPlaylist'
-        component={EditPlaylistScreen}
-        options={screenOptions}
-      />
+      {!isPlaylistUpdatesEnabled ? (
```
dylanjeffers, line 188: do we want this in the app-tab-screen and the app-screen? kinda forget where we should put it, but id imagine one spot?

packages/mobile/src/screens/edit-playlist-screen/EditPlaylistForm.tsx
```diff
@@ -0,0 +1,215 @@
...
+          {values.tracks?.length ? (
+            <>
+              <View>
+                {values.tracks.length ? (
+                  <Text style={styles.reorderTitle}>
+                    {messages.reorderTitle}
+                  </Text>
+                ) : null}
+              </View>
+              <TrackList
```
dylanjeffers, line 198: is it possible to move the handleReorder/handleRemove logic to a "DraggableTrackList" component that can manage itself? that way this form can just focus on the core layout? small architecture thing, lmk what you think.

packages/mobile/src/screens/edit-playlist-screen/EditPlaylistFormScreen.tsx
```diff
@@ -0,0 +1,95 @@
...
+  bottomSection: {
+    padding: spacing(4),
+    paddingBottom: spacing(12),
+    backgroundColor: palette.white,
+    borderTopWidth: 1,
+    borderTopColor: palette.neutralLight7
+  }
+}))
+
+export const FormScreen = (props: FormScreenProps) => {
```
dylanjeffers, line 45: so iirc there already is a "FormScreen" component out there that we can maybe use here?

packages/mobile/src/screens/edit-playlist-screen/EditPlaylistNavigator.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+const screenOptionOverrides = { headerRight: () => null }
+
+export const EditPlaylistNavigator = (
+  props: FormikProps<EditPlaylistValues> & { playlistId: number }
+) => {
+  const screenOptions = useAppScreenOptions(screenOptionOverrides)
+
+  return (
+    <>
```
dylanjeffers, line 19: I think in the case of this screen, we dont need a navigator since it is all one screen, so maybe we can remove this part of things? lmk if you want to pair on the component hierarchy here!

packages/mobile/src/screens/page-form-screen/PageFormScreen.tsx
```diff
@@ -0,0 +1,133 @@
...
+  bottomSection: {
+    padding: spacing(4),
+    paddingBottom: spacing(12),
+    backgroundColor: palette.white,
+    borderTopWidth: 1,
+    borderTopColor: palette.neutralLight6
+  }
+}))
+
+export const FormScreen = (props: FormScreenProps) => {
```
dylanjeffers, line 56: ah, right, like using this component for edit-playlist-screen?

dylanjeffers (approved):
looks very clean, just a few callouts regarding FormScreen, and then the general EditPlaylistScreen component hierarchy

dylanjeffers (approved):
looks very clean, just a few callouts regarding FormScreen, and then the general EditPlaylistScreen component hierarchy

packages/mobile/src/components/core/IconButton.tsx
```diff
@@ -52,6 +55,13 @@ export const IconButton = ({
   const styles = useStyles()
   const { scale, handlePressIn, handlePressOut } = usePressScaleAnimation(0.9)
   const { neutral } = useThemeColors()
+  const { toast } = useToast()
+
+  const onDisabledPress = () => {
```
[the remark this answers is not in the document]
Kyle-Shanks: Yea this might work better for Button, I'll look to see how easy it would be to update

packages/mobile/src/screens/edit-playlist-screen/EditPlaylistFormScreen.tsx
Kyle-Shanks: oh this is actually not supposed to be here haha. I made this first and then scrapped it

packages/mobile/src/screens/edit-playlist-screen/EditPlaylistNavigator.tsx
Kyle-Shanks: this was the only way that the header and top nav things were showing up. I might be mistaken, but down to chat about this

packages/mobile/src/screens/app-screen/AppTabScreen.tsx
Kyle-Shanks: yea the two different types have different headers so they have to be in one vs the other unless im missing something

packages/mobile/src/screens/edit-playlist-screen/EditPlaylistForm.tsx
Kyle-Shanks: moved this code over from the previous implementation. it seems like the reordering has to update form values so it seems like that's why the callback was defined here, but down to discuss further if you have ideas about this

PR #3365, opened 2023-05-16
[C-2603] Fix readonly object error in audius-query reducer

amendelsohn:


dylanjeffers (approved):
great catch

audius-infra:
Preview this change https://demo.audius.co/audius-query-status-error-fix

PR #3367, opened 2023-05-17
[C-2596] Add disabled option to audius-query hooks

amendelsohn:
Disables the hook fetch while hookOptions.disabled is truthy


audius-infra:
Preview this change https://demo.audius.co/audius-query-disabled-option

dylanjeffers (approved):
Wow so easy to add and such a big time feature

PR #3371, opened 2023-05-17
[PAY-1136][PAY-1129] Add infinite loading to albums on desktop / mobile web

schottra:
#### Refactored fetching code
* Now generic to all collections
* Accepts a list of IDs to fetch, which enables filtering of the list _before_ fetching and preserves proper `hasMore` / `fetchMore` behavior based on the filtered list
* Simplified logic which only returns the list of ids which have been fetched, allowing `Card` components to fetch the full metadata themselves

#### Pushed data connection down into Cards
For Desktop and Mobile web, the `AlbumCard` component now accepts an ID and fetches the collection and user info directly. This follows our normalized data patterns and will also help prevent excessive re-renders for already loaded cards just because we've added more items to the albums list.

#### Added infinite scrolling
* Created a version of `CardLineup` which does infinite scrolling. I didn't modify the existing `CardLineup` to do this because we use it in other places where we probably don't want infinite scrolling behavior
* Cleaned up `CardLineup` a little bit to remove unnecessary legacy redux patterns

### Dragons
N/A

### How Has This Been Tested?
Tested locally on desktop and mobile web

Will test via normal RC flows and in staging

### Feature Flags ###
N/A

### DIFF
--- a/packages/common/src/hooks/useSavedCollections.ts
+++ b/packages/common/src/hooks/useSavedCollections.ts
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useState } from 'react'
+import { useCallback, useEffect } from 'react'
 
 import { useDispatch, useSelector } from 'react-redux'
 
@@ -38,7 +38,7 @@ export function useSavedAlbumsDetails({
   pageSize = DEFAULT_PAGE_SIZE
 }: UseSavedAlbumDetailsConfig) {
   const dispatch = useDispatch()
-  const [hasFetched, setHasFetched] = useState(false)
+
   const { unfetched: unfetchedAlbums, fetched: albumsWithDetails } =
     useSelector(getFetchedAlbumsWithDetails)
   const { status } = useSelector(getSavedAlbumsState)
@@ -51,24 +51,14 @@ export function useSavedAlbumsDetails({
       .slice(0, Math.min(pageSize, unfetchedAlbums.length))
       .map((c) => c.id)
     dispatch(fetchCollections({ type: 'albums', ids }))
-    setHasFetched(true)
-  }, [status, unfetchedAlbums, pageSize, dispatch, setHasFetched])
+  }, [status, unfetchedAlbums, pageSize, dispatch])
 
-  // Fetch first page if we don't have any items fetched yet
-  // Needs to wait for at least some albums to be fetchable
-  useEffect(() => {
-    if (
-      !hasFetched &&
-      // TODO: This check should change once InfiniteScroll is implemented
-      status !== Status.LOADING /* &&
-      unfetchedAlbums.length > 0 &&
-      albumsWithDetails.length === 0 */
-    ) {
-      fetchMore()
-    }
-  }, [albumsWithDetails, status, hasFetched, unfetchedAlbums, fetchMore])
-
-  return { data: albumsWithDetails, status, fetchMore }
+  return {
+    data: albumsWithDetails,
+    status,
+    hasMore: unfetchedAlbums.length > 0,
+    fetchMore
+  }
 }
 
 export function useSavedPlaylists() {
--- a/packages/common/src/hooks/useSavedCollections.ts
+++ b/packages/common/src/hooks/useSavedCollections.ts
@@ -1,24 +1,22 @@
-import { useCallback, useEffect } from 'react'
+import { useCallback, useMemo } from 'react'
 
 import { useDispatch, useSelector } from 'react-redux'
 
+import { ID } from 'models/Identifiers'
 import { Status } from 'models/Status'
+import { CommonState } from 'store/index'
+import { getFetchedCollectionIds } from 'store/saved-collections/selectors'
 
-import { accountActions } from '../store/account'
 import {
+  CollectionType,
   savedCollectionsActions,
   savedCollectionsSelectors
 } from '../store/saved-collections'
 
-const { fetchSavedPlaylists } = accountActions
 const { fetchCollections } = savedCollectionsActions
 
-const {
-  getAccountAlbums,
-  getSavedAlbumsState,
-  getFetchedAlbumsWithDetails,
-  getAccountPlaylists
-} = savedCollectionsSelectors
+const { getAccountAlbums, getSavedCollectionsState, getAccountPlaylists } =
+  savedCollectionsSelectors
 
 const DEFAULT_PAGE_SIZE = 50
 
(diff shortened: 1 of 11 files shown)


audius-infra:
Preview this change https://demo.audius.co/pay-1136-infinite-load-card-lineup

packages/web/src/pages/saved-page/components/desktop/SavedPage.tsx
```diff
@@ -89,36 +108,60 @@ const AlbumCard = ({ album, index, isLoading, setDidLoad }: AlbumCardProps) => {
...
 
 const AlbumsTabContent = () => {
   const goToRoute = useGoToRoute()
 
-  // Temporarily requesting large page size to ensure we get all albums
-  // until the list is updated to use `InfinteScroll`
-  const { data: albums } = useSavedAlbumsDetails({ pageSize: 9999 })
-  const { isLoading, setDidLoad } = useOrderedLoad(albums.length)
-  const cards = albums.map((album, i) => {
+  const { data: savedAlbums } = useSavedAlbums()
```
sliptype, line 117: I wonder if `useSavedAlbums` could accept a selector so we could avoid the extra `useMemo`?

```suggestion
  const savedAlbumIds = useSavedAlbums((savedAlbums) => savedAlbums.map((a) => a.id))
```

sliptype (approved):
Sooooo nice! Thanks for doing the refactors too

packages/common/src/hooks/useSavedCollections.ts
```diff
@@ -1,84 +1,92 @@
...
+  /** Whether any items remain unfetched */
+  hasMore: boolean
+  /** Triggers fetching of the next page of items */
+  fetchMore: () => void
+}
+/** Given a list of collectionIds and a type ('albums' or 'playlists'), returns state
+ * necessary to display a list of fully-fetched collections of that type, as well as
+ * load any remaining items which haven't been fetched.
+ */
+export function useFetchedSavedCollections({
```
piazzatron, line 51: nitty nit: curious for your thought process in naming this `useFetchedSavedCollections` rather than `useFetchSavedCollections`?

packages/web/src/pages/saved-page/components/desktop/SavedPage.tsx
piazzatron: We'd have to memo it on the mobile component that allows you to filter the name (why don't we let you do this on desktop??) anyways, so IMO fine as is

packages/common/src/hooks/useSavedCollections.ts
```diff
@@ -1,84 +1,92 @@
...
-/* TODO: Handle filtering
- * Option 1: This hook takes the list of album ids to fetch and computes the unfetched
- * based on that.
- * Option 2: Bake filter into selectors which drive this. Downside: Can't use this in multiple places...
- */
-type UseSavedAlbumDetailsConfig = {
+export function useSavedPlaylists() {
+  return useSelector(getAccountPlaylists)
+}
+
```
piazzatron, line 30: After spending a while staring at this, I think it would be nice to call these hooks something that indicates that they're fetching the abbreviated `AccountCollection` that essentially just has the IDs, which is then to be fetched later as a full `Collection`. If I just look at the callsites for these and the fetching hook below, it feels like we're selecting a `Collection` out of the store, and then re-fetching it for some reason. Maybe `useAccountAlbums` - eventually would be great to rename this type into something more descriptive (`CollectionMetadata`? `CollectionPreview`? hmm), but probably avoid for now :D

piazzatron (approved):
Looks great - just some ergonomics musings for me :). Really clean.

packages/common/src/hooks/useSavedCollections.ts
```diff
@@ -1,84 +1,92 @@
...
+        if (fetchedSet.has(id)) {
+          accum.fetched.push(id)
+        } else {
+          accum.unfetched.push(id)
+        }
+        return accum
+      },
+      { fetched: [], unfetched: [] }
+    )
+  }, [collectionIds, fetchedCollectionIDs])
```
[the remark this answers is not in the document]
schottra: Oh yeah good suggestion. There's an additional check I added in a follow-up PR. I will add `partition` there just to avoid rebase hassle.

packages/web/src/components/lineup/InfiniteCardLineup.tsx
```diff
@@ -0,0 +1,49 @@
...
+const InfiniteCardLineup = (props: InfiniteCardLineupProps) => {
+  const { hasMore, loadMore, ...lineupProps } = props
+  const scrollRef = useRef(null)
+
+  const getNearestScrollParent = useCallback(() => {
+    if (!scrollRef.current) {
+      return null
+    }
+    return (
+      (getScrollParent(scrollRef.current) as unknown as HTMLElement) ?? null
```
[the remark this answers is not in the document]
schottra: I looked into this. `getScrollParent` returns `document.scrollElement` in some cases, which is typed as `Element`. I'm not 100% sure why. But it's a built-in type, so I don't think we could change it, other than to move this cast into `getScrollParent` itself. Thoughts on that?

packages/web/src/pages/saved-page/components/desktop/SavedPage.tsx
schottra: My follow-up branch is changing some of this due to the filtering we have to do on mobile. Agreed that  it's weird that we don't allow filtering on desktop. I actually think it would be pretty easy to add and may do that if it doesn't take too long.

packages/common/src/hooks/useSavedCollections.ts
schottra: Naming is the hardest problem in software engineering :-)

I had originally just named this `useSavedCollections`, but changed it based on some good feedback to indicate the difference between the minimal saved collection records we read from the `account`, and the full records that we have to fetch from the `/playlists` endpoint. In my mind, the idea here is that it's retrieving the list of collections which have had their full records fetched. That could be through this hook, or elsewhere, they only have to be in the cache. The selector used under the hood will map the "minimal" records across all cached collections and return the intersection.

Also I think I'm just being picky about grammar? 😅  `useFetched...` reads cleaner to me than `useFetch` as the primary value being returned is the list of collections, not the `fetch` function?

packages/web/src/components/lineup/InfiniteCardLineup.tsx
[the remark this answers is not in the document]
sliptype: It looks like in most other places we are defining `getScrollParent` locally in the component file. I only see one other usage of `utils/scrollParent`. Maybe it would be better to define locally?

PR #3374, opened 2023-05-17
[C-2508] Update playback positions to be tracked per user id

Kyle-Shanks:
### Description
Update podcast playback positions to be tracked per user id so that users can log out and log into different accounts with predictable behaviour

### Dragons

N/A

### How Has This Been Tested?

Manually tested on web and mobile

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/common/src/store/playback-position/sagas.ts
+++ b/packages/common/src/store/playback-position/sagas.ts
@@ -2,6 +2,7 @@ import { call, delay, put, select, takeEvery } from 'typed-redux-saga'
 
 import { AudioPlayer } from 'services/audio-player'
 import { FeatureFlags } from 'services/remote-config'
+import { getUserId } from 'store/account/selectors'
 import { getTrack } from 'store/cache/tracks/selectors'
 import { getContext } from 'store/effects'
 import { getPlaying, getTrackId } from 'store/player/selectors'
@@ -13,7 +14,11 @@ import {
   initializePlaybackPositionState,
   setTrackPosition
 } from './slice'
-import { PlaybackPositionState, PLAYBACK_POSITION_LS_KEY } from './types'
+import {
+  LEGACY_PLAYBACK_POSITION_LS_KEY,
+  PlaybackPositionState,
+  PLAYBACK_POSITION_LS_KEY
+} from './types'
 
 const RECORD_PLAYBACK_POSITION_INTERVAL = 1000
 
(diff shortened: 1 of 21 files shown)


audius-infra:
Preview this change https://demo.audius.co/kj-Update-playback-positions-to-track-user-ids

packages/common/src/store/playback-position/sagas.ts
```diff
@@ -36,11 +42,35 @@ function* setInitialPlaybackPositionState() {
...
+    }
+    yield* put(
+      initializePlaybackPositionState({
+        playbackPositionState: convertedLegacyState
+      })
+    )
+    setLocalStorageItem(
+      PLAYBACK_POSITION_LS_KEY,
+      JSON.stringify(convertedLegacyState)
+    )
```
amendelsohn, line 72: Do we also want to clear out the old key?

packages/common/src/store/playback-position/slice.ts
```diff
@@ -10,16 +10,16 @@ type InitializePlaybackPositionStatePayload = {
 
 type SetTrackPositionPayload = {
   trackId: ID
+  userId?: ID | null
```
amendelsohn, line 13: should these accept empty? seems like it's a no-op if we don't have user id, so maybe just don't fire the action in the first place? idk which is better

packages/common/src/store/playback-position/slice.ts
```diff
@@ -31,22 +31,29 @@ const slice = createSlice({
       action: PayloadAction<InitializePlaybackPositionStatePayload>
     ) => {
       const { playbackPositionState } = action.payload
-      state.trackPositions =
-        playbackPositionState.trackPositions ?? state.trackPositions
+      const userIds = Object.keys(playbackPositionState)
+      userIds.forEach((userId) => {
+        state[userId] = playbackPositionState[userId]
+      })
```
amendelsohn, line 37: can we do `state = { ...playbackPositionState }` or does the proxy not like that?

packages/common/src/store/playback-position/types.ts
```diff
@@ -10,5 +11,7 @@ export type PlaybackPositionInfo = {
 }
 
 export type PlaybackPositionState = {
-  trackPositions: { [Key in ID]?: PlaybackPositionInfo }
+  [Key in ID]?: {
+    trackPositions: { [Key in ID]?: PlaybackPositionInfo }
+  }
```
amendelsohn, line 16: Don't need to fix here, but wondering if we should start naming these better to make the types more readable:

```
export type PlaybackPositionState = {
  [UserIdKey in ID]?: {
    trackPositions: { [TrackIdKey in ID]?: PlaybackPositionInfo }
  }
```

amendelsohn (approved):
Looks great! Nice job :)

packages/common/src/store/playback-position/slice.ts
Kyle-Shanks: yea i tried that first, it doesn't like that unfortunately

amendelsohn: dang unlukcy

packages/common/src/store/playback-position/slice.ts
Kyle-Shanks: This made it easier for type and not having to check it user existed everywhere this was called if i remember correctly, but I agree it's a bit weird

PR #3375, opened 2023-05-18
[C-2607] Pagination wrapper hooks for audius-query

amendelsohn:
### Description

Integrated usage example to come in separate PR

Adds `usePaginatedQuery` and `useAllPaginatedQuery` which wrap an existing audius-query hook and manage pagination args and response automatically using our standard `limit, offset, hasMore, loadMore` interface

```
  const {
    data,
    status,
	hasMore,
    loadMore
  } = usePaginatedQuery(
    useMyApiQuery,
    {
      myArg1,
	  myArg2
    },
    PAGE_SIZE
```

Also adds sdk to the AudiusQueryContext, making it available in endpiont fetch functions


audius-infra:
Preview this change https://demo.audius.co/audius-query-pagination

packages/common/src/api/hooks/usePaginatedQuery.ts
```diff
@@ -0,0 +1,73 @@
...
+      result.data?.length < pageSize
+  }
+}
+
+export const useAllPaginatedQuery = <
+  Data,
+  ArgsType extends { limit: number; offset: number },
+  HookType extends (args: ArgsType) => QueryHookResults<Data[]>
+>(
+  useQueryHook: HookType,
```
sliptype, line 37: Still not entirely sure this doesn't break the rules of hooks. If the caller were to change the hook being passed, then hooks would be getting called in a different order

```
useAllPaginatedQuery(something ? useGetTransactions : useGetTracks, {}, 50)
```

Not sure if it's actually a problem though

sliptype (approved):
Awesome stuff!

packages/web/src/app.tsx
```diff
@@ -24,7 +26,9 @@ import './index.css'
 const AudiusApp = () => {
   return (
     <Provider store={store}>
-      <AudiusQueryContext.Provider value={{ apiClient }}>
+      <AudiusQueryContext.Provider
```
dylanjeffers, line 29: love this, as it gets bigger, we can maybe add a provider in a separate file that does all the config

dylanjeffers (approved):
wonderful additions

dylanjeffers (approved):
wonderful additions

packages/common/src/api/hooks/usePaginatedQuery.ts
```diff
@@ -0,0 +1,73 @@
...
+  const [page, setPage] = useState(0)
+  const args = { ...baseArgs, limit: pageSize, offset: page * pageSize }
+  const result = useQueryHook(args)
+  return {
+    ...result,
+    loadMore: () => setPage(page + 1),
+    hasMore:
+      result.status === Status.IDLE ||
+      (!result.data && result.status === Status.LOADING) ||
+      result.data?.length < pageSize
```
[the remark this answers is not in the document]
amendelsohn: omg you're right. It's backwards

PR #3376, opened 2023-05-18
[PAY-1197] Mobile inbox unavailable modal from profile screen

dharit-tan:
### Description
Show message locked button on profile screen when current user is not permitted to message a user. Pressing it opens the InboxUnavailable drawer with various states depending on why the user is not permitted.

Also includes an update to UserBadges component that allows us to set an `inline` prop that changes the parent component to a `Text` instead of `View`, allowing it to be shown inline with other text.

### Dragons

There is an issue with UserBadges component where the name can be shown on a different line than the badges. I have notified client team and they've added it to their todo. See below screenshots.

### How Has This Been Tested?

Local ios stage
Simulator Screen Shot - iPhone 14 Pro - 2023-05-17 at 15 34 28
Simulator Screen Shot - iPhone 14 Pro - 2023-05-17 at 20 32 33

### Feature Flags ###


packages/mobile/src/components/inbox-unavailable-drawer/InboxUnavailableDrawer.tsx
```diff
@@ -0,0 +1,256 @@
...
+    }
+  }, [
+    handleLearnMorePress,
+    handleTipPress,
+    styles.buttonTextWhite,
+    handleUnblockPress,
+    closeDrawer
+  ])
+
+  const content = mapActionToContent(callToAction, user!, styles)
```
dharit-tan, line 219: I think the `user!` is safe because we only ever navigate to this drawer from that user's profile screen. Open to ideas tho.

rickyrombo: I'd say have the fn accept undefined and render null instead of user badges if undefined

packages/mobile/src/components/inbox-unavailable-drawer/InboxUnavailableDrawer.tsx
```diff
@@ -0,0 +1,256 @@
...
+    closeDrawer()
+  }, [closeDrawer])
+
+  const handleTipPress = useCallback(() => {
+    dispatch(beginTip({ user, source: 'profile' }))
+    navigation.navigate('TipArtist')
+    closeDrawer()
+  }, [closeDrawer, dispatch, navigation, user])
+
+  const actionToButtonsMap = useMemo(() => {
```
rickyrombo, line 179: Thoughts on this vs web impl?

rickyrombo (approved):

packages/mobile/src/components/inbox-unavailable-drawer/InboxUnavailableDrawer.tsx
dharit-tan: Was using web impl but now that we have 2 buttons for unblock case it didn't work as well so changed to this instead. I think you'll have to update web modal for that case too.

rickyrombo: thanks for the callout, missed that!

In that case, feels like we don't really save much from having these individual props vs just button components, but I'm down for keeping as is too

dharit-tan: Yeah I forgot I can return a component with `() =>` honestly, would be down for that.

dharit-tan: Chatted with Marcus on slack, gonna keep it as it is for now. Realized it's weird to pass in the onPress callbacks to `mapActionToContent`, would have to change it to a hook.

packages/mobile/src/screens/profile-screen/ProfileInfo.tsx
```diff
@@ -98,8 +107,25 @@ export const ProfileInfo = (props: ProfileInfoProps) => {
...
 
+  const profileUserId = useSelector((state) =>
+    getProfileUserId(state, params.handle)
+  )
+  const { canChat } = useSelector((state) => getCanChat(state, profileUserId))
+
+  useEffect(() => {
+    dispatch(fetchBlockees())
+    dispatch(fetchBlockers())
+  }, [dispatch, params, profileUserId])
```
piazzatron, line 121: nit: I know this is merged already but probably worth a comment about these extra dependencies

PR #3386, opened 2023-05-18
[C-2615] Fix favorite tracks error due to empty track entries on web mobile

Kyle-Shanks:
### Description
Made the decision to not add pagination to the web mobile favorites page at this time and simply filter out the empty track entries to get rid of the error

Also removed some class name assignments without definitions to get rid of eslint errors

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A


audius-infra:
Preview this change https://demo.audius.co/kj-Fix-favorited-tracks-error-on-web-mobile

sliptype (approved):
Seems reasonable to me!

So people can still see all their favorites on mobile web right? We just don't have a virtualized table

PR #3387, opened 2023-05-18
[PAY-1218] Mobile block dms drawer switches block/unblock

dharit-tan:
### Description
Switch between block/unblock state on ProfileActionsDrawer and BlockMessagesDrawer.

### Dragons

### How Has This Been Tested?

local ios stage

### Feature Flags ###

### DIFF
--- a/packages/common/src/store/pages/chat/selectors.ts
+++ b/packages/common/src/store/pages/chat/selectors.ts
@@ -182,6 +182,11 @@ export const getUserChatPermissions = createSelector(
   }
 )
 
+export const getDoesBlockUser = createSelector(
+  [getBlockees, (_state: CommonState, userId: ID) => userId],
+  (blockees, userId) => blockees.includes(userId)
+)
+
 export const getCanChat = createSelector(
   [
     getUserId,
--- a/packages/mobile/src/components/block-messages/BlockMessages.tsx
+++ b/packages/mobile/src/components/block-messages/BlockMessages.tsx
@@ -14,15 +14,27 @@ import { spacing } from 'app/styles/spacing'
 import { useColor } from 'app/utils/theme'
 
 const { getUser } = cacheUsersSelectors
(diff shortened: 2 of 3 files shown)


audius-infra:
Preview this change https://demo.audius.co/rt-block-drawer-fixes

packages/mobile/src/components/block-messages/BlockMessages.tsx
```diff
@@ -131,23 +142,25 @@ export const BlockMessagesDrawer = () => {
           <Text style={styles.title}>{messages.title}</Text>
         </View>
         <Text style={styles.confirm}>
-          {messages.confirmText1}
-          {user?.name}
-          {messages.confirmText2}
+          {doesBlockUser
```
sddioulde, line 145: nit: not sure if what im about to suggest is cleaner, so fine with what you decide, but seeing a lot of `doesBlockUser` relate conditions below. maybe it makes sense to have var e.g. message,title,variant in a colocated space here and then use below

sddioulde (approved):
lgtm

packages/common/src/store/pages/chat/selectors.ts
```diff
@@ -182,6 +182,11 @@ export const getUserChatPermissions = createSelector(
   }
 )
 
+export const getDoesBlockUser = createSelector(
```
rickyrombo, line 185: smarrrt

rickyrombo (approved):
I'll make a note to adopt that selector elsewhere in webland

packages/mobile/src/components/block-messages/BlockMessages.tsx
dharit-tan: Hm good call. I think if we had more than 2 possible states it would def make sense, but since it's only ever block/unblock here I think this is fine.

PR #3388, opened 2023-05-18
[PAY-1028] Fix mobile chat actions drawer touch feedback

dharit-tan:
### Description
Fix touch behavior for ChatActions drawer (from chat screen) so that it no longer has a black highlight.

### Dragons

I know I should use ActionDrawer, but that doesn't support passing data afaik, which we need here! So I'm stuck reimplementing it in a janky way.

### How Has This Been Tested?

local ios stage

### Feature Flags ###


dharit-tan:
> 

Chatted with @rickyrombo on slack, ActionDrawer doesn't support passing a `data` field like native drawers do right now. (which can be added to any caller of `setVisibility` to pass data to the drawer). Apparently this is easy to add but not gonna address that right now.

piazzatron:
> > 
> 
> Chatted with @rickyrombo on slack, ActionDrawer doesn't support passing a `data` field like native drawers do right now. (which can be added to any caller of `setVisibility` to pass data to the drawer). Apparently this is easy to add but not gonna address that right now.

Worth tracking this in Linear somewhere as tech debt

piazzatron (approved):

rickyrombo (approved):

PR #3392, opened 2023-05-19
Add useAuthenticatedCallback

dylanjeffers:
### Description

Adds authenticatedCallback hook, which can be used anytime we have an action that requires log-in first. Will be used in left-nav primarily


audius-infra:
Preview this change https://demo.audius.co/dj-use-authenticated-callback

amendelsohn (approved):

PR #3393, opened 2023-05-19
[PAY-1241] Skeleton for mobile chat list screen

dharit-tan:
### Description
Skeleton for empty items in ChatListScreen

### Dragons

### How Has This Been Tested?

Local ios stage

Simulator Screen Shot - iPhone 14 Pro - 2023-05-19 at 13 22 32

### Feature Flags ###

### DIFF
--- a/packages/common/src/store/pages/chat/selectors.ts
+++ b/packages/common/src/store/pages/chat/selectors.ts
@@ -133,16 +133,6 @@ export const getSingleOtherChatUser = (state: CommonState, chatId?: string) => {
   return getOtherChatUsers(state, chatId)[0]
 }
 
-export const getAllOtherChatUsers = createSelector(
-  [(state) => state, getChats],
-  (state, chats) => {
-    return chats.map((chat) => {
-      const otherUser = getOtherChatUsersFromChat(state, chat)
-      return otherUser[0]
-    })
-  }
-)
-
 export const getChatMessageByIndex = (
   state: CommonState,
   chatId: string,
--- a/packages/common/src/store/pages/chat/selectors.ts
+++ b/packages/common/src/store/pages/chat/selectors.ts
@@ -133,6 +133,16 @@ export const getSingleOtherChatUser = (state: CommonState, chatId?: string) => {
   return getOtherChatUsers(state, chatId)[0]
 }
 
+export const getAllOtherChatUsers = createSelector(
+  [(state) => state, getChats],
+  (state, chats) => {
+    return chats.map((chat) => {
+      const otherUser = getOtherChatUsersFromChat(state, chat)
+      return otherUser[0]
+    })
+  }
+)
+
 export const getChatMessageByIndex = (
   state: CommonState,
   chatId: string,
--- a/packages/common/src/store/pages/chat/selectors.ts
+++ b/packages/common/src/store/pages/chat/selectors.ts
@@ -303,3 +303,7 @@ export const getCanSendMessage = createSelector(
     }
   }
 )
+
+export const getNumberOfFetchChatRequests = (state: CommonState) => {
+  return state.pages.chat.numberOfFetchChatRequests
+}
--- a/packages/common/src/store/pages/chat/selectors.ts
+++ b/packages/common/src/store/pages/chat/selectors.ts
@@ -303,7 +303,3 @@ export const getCanSendMessage = createSelector(
     }
   }
 )
-
-export const getNumberOfFetchChatRequests = (state: CommonState) => {
-  return state.pages.chat.numberOfFetchChatRequests
-}
--- a/packages/mobile/src/screens/chat-screen/ChatListItem.tsx
+++ b/packages/mobile/src/screens/chat-screen/ChatListItem.tsx
@@ -105,40 +105,38 @@ export const ChatListItem = ({ chatId }: { chatId: string }) => {
 
   return (
     <TouchableHighlight onPress={handlePress}>
-      <View style={styles.root}>
-        {otherUser ? (
-          <>
-            <View style={styles.contentRoot}>
-              <View style={styles.userContainer}>
-                <ProfilePicture
-                  profile={otherUser}
-                  style={styles.profilePicture}
-                />
-                <View style={styles.userTextContainer}>
-                  <View style={styles.userNameContainer}>
-                    <Text style={styles.userName}>{otherUser.name}</Text>
-                    <UserBadges user={otherUser} hideName />
-                  </View>
-                  <Text style={styles.handle}>@{otherUser.handle}</Text>
+      {otherUser ? (
+        <View style={styles.root}>
+          <View style={styles.contentRoot}>
+            <View style={styles.userContainer}>
+              <ProfilePicture
+                profile={otherUser}
+                style={styles.profilePicture}
+              />
+              <View style={styles.userTextContainer}>
+                <View style={styles.userNameContainer}>
+                  <Text style={styles.userName}>{otherUser.name}</Text>
+                  <UserBadges user={otherUser} hideName />
+                </View>
+                <Text style={styles.handle}>@{otherUser.handle}</Text>
+              </View>
+            </View>
+            {chat?.unread_message_count && chat?.unread_message_count > 0 ? (
+              <View>
+                <View style={styles.unreadCountContainer}>
+                  <Text style={styles.unreadCount}>
+                    {clipMessageCount(chat?.unread_message_count ?? 0)}{' '}
+                    {messages.new}
+                  </Text>
                 </View>
               </View>
-              {chat?.unread_message_count && chat?.unread_message_count > 0 ? (
-                <View>
-                  <View style={styles.unreadCountContainer}>
-                    <Text style={styles.unreadCount}>
-                      {clipMessageCount(chat?.unread_message_count ?? 0)}{' '}
-                      {messages.new}
-                    </Text>
-                  </View>
-                </View>
-              ) : null}
-            </View>
-            <Text numberOfLines={1}>{lastMessage}</Text>
-          </>
-        ) : (
-          <ChatListItemSkeleton />
-        )}
-      </View>
+            ) : null}
+          </View>
+          <Text numberOfLines={1}>{lastMessage}</Text>
+        </View>
+      ) : (
+        <ChatListItemSkeleton />
+      )}
     </TouchableHighlight>
   )
 }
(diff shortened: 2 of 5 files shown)


packages/mobile/src/screens/chat-screen/ChatListItem.tsx
```diff
@@ -103,38 +111,38 @@ export const ChatListItem = ({ chatId }: { chatId: string }) => {
...
-                  profile={otherUser}
-                  style={styles.profilePicture}
-                />
-                <View style={styles.userTextContainer}>
-                  <View style={styles.userNameContainer}>
-                    <Text style={styles.userName}>{otherUser.name}</Text>
-                    <UserBadges user={otherUser} hideName />
-                  </View>
-                  <Text style={styles.handle}>@{otherUser.handle}</Text>
+      {otherUser ? (
```
rickyrombo, line 114: feels like this should be checking for `!!chat`, but I guess it's helpful to assert otherUser is defined

rickyrombo (approved):

packages/mobile/src/screens/chat-screen/ChatListItem.tsx
dharit-tan: Even if `chat` is defined, we still need to wait to load in the user details.
However I need to rewrite this a bit anyway, stay tuned.

piazzatron: What does this look like if `otherUser` exists but `chat` doesn't?

packages/mobile/src/screens/chat-screen/ChatListItemSkeleton.tsx
```diff
@@ -0,0 +1,77 @@
...
+    borderRadius: spacing(12)
+  }
+}))
+
+export const ChatListItemSkeleton = ({ index }: { index: number }) => {
+  const styles = useStyles()
+
+  return (
+    <View style={[styles.root, { opacity: (4 - index) * 0.25 }]}>
+      <View style={[styles.contentRoot]}>
```
piazzatron, line 63: Since it looks like this is using largely the same styles as the non-skeleton list item, could we could just export those and import them here to ensure consistency?

piazzatron (approved):
Looks great!!

packages/mobile/src/screens/chat-screen/ChatListItem.tsx
rickyrombo: this won't get rendered - @dharit-tan and I chatted in slack a bit about this, it's a bit wonky

- Right now in fetching chats we trigger fetching the appropriate users, but don't wait for that to complete
- The way this PR is now is only waiting on that second _users_ fetching, not the original chats loading (which if we have no chats rn it shows empty state I believe?)

This presents a problem because if the chats aren't loaded we want the fading skeletons, but what do we do if the users aren't loaded? We could check to make sure all users are loaded, but then if you paginate, the whole chat list is going be replaced with skeletons until the new users appear

We have a couple options:
- Show the fading skeletons, and make the saga wait for all the relevant users to be fetched before returning so we can switch straight from the fading skeletons to complete chat list items
- Show the fading skeletons, then when we have chats, show non-fading skeletons (one for each chat), then when the users are loaded, show the final chat list items

I think the former is better, @dharit-tan is looking into it

dharit-tan: Agree option 1 is better, but the fns for retrieving users is in web common and I couldn't figure out how to get them into actual common.

Implemented option 2 here, don't love how it looks in the code but I believe it works.

audius-infra:
Preview this change https://demo.audius.co/rt-skeleton

audius-infra:
Preview this change https://demo.audius.co/rt-skeleton

packages/mobile/src/screens/chat-screen/ChatListItem.tsx
rickyrombo: That checks out - I think that's why I didn't do option 1 initially 🙁 

will take a look at the option 2 impl

packages/common/src/store/pages/chat/selectors.ts
```diff
@@ -133,6 +133,16 @@ export const getSingleOtherChatUser = (state: CommonState, chatId?: string) => {
   return getOtherChatUsers(state, chatId)[0]
 }
 
+export const getAllOtherChatUsers = createSelector(
+  [(state) => state, getChats],
```
schottra, line 137: I could be wrong, but I think `state=>state` will negate the benefit of using `createSelector`, since the inputs to your compute function would change every time anything in the entire state changes?
I would say either use a proxy selector, or update `getOtherUsersFromChat` to take only the subset of the state it needs.

packages/mobile/src/screens/chat-screen/ChatListItemSkeleton.tsx
```diff
@@ -0,0 +1,68 @@
+import { View } from 'react-native'
+
+import Skeleton from 'app/components/skeleton'
+import { makeStyles } from 'app/styles'
+
+import { useStyles as useChatListItemStyles } from './ChatListItem'
```
schottra, line 6: nit: Might make sense to move both components into a subfolder and put the styles in a `styles` file.

packages/mobile/src/screens/chat-screen/ChatListItemSkeleton.tsx
```diff
@@ -0,0 +1,68 @@
...
+    marginBottom: spacing(3)
+  },
+  userName: {
+    height: spacing(4),
+    width: spacing(25),
+    borderRadius: spacing(12)
+  },
+  handle: {
+    height: spacing(3),
+    width: spacing(22),
```
schottra, line 26: Just curious why these widths are hard-coded to specific things? Could the values exceed that width?

packages/mobile/src/screens/chat-screen/ChatListItemSkeleton.tsx
```diff
@@ -0,0 +1,68 @@
...
+  index?: number
+}) => {
+  const styles = useStyles()
+  const ChatListItemStyles = useChatListItemStyles()
+
+  return (
+    <View
+      style={[
+        ChatListItemStyles.root,
+        shouldFade ? { opacity: (4 - index) * 0.25 } : null
```
schottra, line 51: Might need a comment on why you are doing this or a helper function to isolate this special case logic

packages/mobile/src/screens/chat-screen/ChatListScreen.tsx
```diff
@@ -132,6 +145,10 @@ export const ChatListScreen = () => {
     }
   }, [chats, dispatch])
 
+  const handleLoadMore = useCallback(() => {
```
schottra, line 148: Safe to be called many times? I usually see guards on these to prevent duplicate calls.

packages/mobile/src/screens/chat-screen/ChatListScreen.tsx
```diff
@@ -143,7 +160,17 @@ export const ChatListScreen = () => {
       <ScreenContent>
         <View style={styles.shadow} />
         <View style={styles.rootContainer}>
-          {chatsStatus === Status.SUCCESS ? (
+          {isLoading ? (
+            Array(4)
```
schottra, line 164: More nits:
You could do this more succinctly with the second arg to `Array.from`: 
```
Array.from(
  {length: 4},
  (_, index) => <ChatListItemSkeleton />
)
```
Also surprised that linting doesn't yell at you for using an index as key here. 
You could always just pull this out into a separate component that doesn't use a map

dharit-tan: Reason I didn't pull out to separate component is bc I wanted to re-use the same component for when we've already loaded some chats, and then scroll down and load more. So we use it in 2 places - on initial load (with fade-out), and on subsequent loads (no fade-out).
And yeah linting didn't complain for some reason. In this case tho I think it's fine bc we only ever use it once and the index will always be 0-4.

packages/mobile/src/screens/chat-screen/ChatListScreen.tsx
dharit-tan: Good callout.

packages/mobile/src/screens/chat-screen/ChatListItemSkeleton.tsx
dharit-tan: Since this is a skeleton we're just showing a generic glowy section that's always gonna be the same size.

schottra: Oh I missed that this was a skeleton, carry on!

packages/mobile/src/screens/chat-screen/ChatListScreen.tsx
```diff
@@ -110,7 +115,15 @@ export const ChatListScreen = () => {
   const navigation = useNavigation<AppTabScreenParamList>()
   const chats = useSelector(getChats)
   const chatsStatus = useSelector(getChatsStatus)
+  const otherUsers = useSelector(getAllOtherChatUsers)
 
+  // If this is the first fetch, we want to show the loading skeleton that
+  // fades out. Otherwise, we want to show a skeleton in each incoming chat row.
+  const numberOfFetchChatRequests = useSelector(getNumberOfFetchChatRequests)
```
rickyrombo, line 122: I don't think we need this - I think this is inferred by `chats.length > 0`

dharit-tan: I thought this too but the complicated bit is `chats.length` can be `> 0` but the users haven't loaded yet, and we still want to show the skeleton in that case. But *only* for that case and not subsequent loads.

dharit-tan: Spoke offline, we're gonna nix this fetch counter and just live with the fading skeleton transitioning into non-fading skeleton with rows popping in. We fetch chats from the root screen so ideally users will never see the fading skeleton anyway. I'll leave the fading skeleton code in bc once we finish the common migration, we can have the `doFetchMoreChats` saga wait for users to load in as well, which will make it more likely to see the fading skeleton.

audius-infra:
Preview this change https://demo.audius.co/rt-skeleton

packages/common/src/store/pages/chat/selectors.ts
dharit-tan: Decided to just remove this altogether and live with a _slightly_ less polished loading UI.

rickyrombo (approved):

PR #3397, opened 2023-05-20
Update locks

dylanjeffers:
### Description

Brings locks back to as close of a diff as possible from mainline, just to avoid too many further lock + module changes for folks. also brings back podfile


rickyrombo (approved):
ty

PR #3400, opened 2023-05-22
[PAY-1220] Fix chat receive bug

piazzatron:
### Description

**Bug:** receiving a new chat in an active session causes a crash because we try to use the ChatID of a chat that doesn't yet exist to index into the `messages` object.

**Fix**:
- First, in the `addMessage` flow exit early if we don't know about the chat yet.
- Fix additional bug in the `fetchChatIfNeeded` flow to fetch chat users so that the `ChatListItem` renders

Also adds a misc type cleanup

### Dragons

None

### How Has This Been Tested?

Web/stage

### Feature Flags ###

### DIFF
--- a/packages/common/src/store/pages/chat/slice.ts
+++ b/packages/common/src/store/pages/chat/slice.ts
@@ -25,13 +25,15 @@ type UserChatWithMessagesStatus = UserChat & {
   messagesSummary?: TypedCommsResponse<ChatMessage>['summary']
 }
 
+type ChatID = string
+
 type ChatState = {
   chats: EntityState<UserChatWithMessagesStatus> & {
     status: Status
     summary?: TypedCommsResponse<UserChat>['summary']
   }
   messages: Record<
-    string,
+    ChatID,
     EntityState<ChatMessageWithExtras> & {
       status?: Status
       summary?: TypedCommsResponse<ChatMessage>['summary']
@@ -362,6 +364,14 @@ const slice = createSlice({
     ) => {
       // triggers saga to get chat if not exists
       const { chatId, message, status } = action.payload
+
+      // If no chatId, don't add the message
+      // and abort early, relying on the saga
+      // to fetch the chat
+      if (!(chatId in state.messages)) {
+        console.info('addMessage: no chatId')
+        return
+      }
 
       const existingMessage = getMessage(
         state.messages[chatId],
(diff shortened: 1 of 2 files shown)


packages/common/src/store/pages/chat/slice.ts
```diff
@@ -363,6 +365,14 @@ const slice = createSlice({
       // triggers saga to get chat if not exists
       const { chatId, message, status } = action.payload
 
+      // If no chatId, don't add the message
+      // and abort early, relying on the saga
+      // to fetch the chat
+      if (!(chatId in state.messages)) {
+        console.info('addMessage: no chatId')
```
rickyrombo, line 372: Nit: don't think we need log here

audius-infra:
Preview this change https://demo.audius.co/piazz-fix-receiving-chats-bug

packages/common/src/store/pages/chat/slice.ts
```diff
@@ -25,13 +25,15 @@ type UserChatWithMessagesStatus = UserChat & {
...
+type ChatID = string
+
 type ChatState = {
   chats: EntityState<UserChatWithMessagesStatus> & {
     status: Status
     summary?: TypedCommsResponse<UserChat>['summary']
   }
   messages: Record<
-    string,
+    ChatID,
```
rickyrombo, line 36: I went through and did this once but then realized vscode shows string anyway. Not sure if it adds any real value?

rickyrombo (approved):

dharit-tan (approved):

packages/common/src/store/pages/chat/slice.ts
piazzatron: I think it's helpful when looking at the slice type here to see that the record key is a ChatId vs say a message UUID or something

rickyrombo: Yeah I thought so too, but when you hover over the type it shows `string` not `ChatID` in intellisense anyway.

audius-infra:
Preview this change https://demo.audius.co/piazz-fix-receiving-chats-bug

packages/common/src/store/pages/chat/slice.ts
piazzatron: Yeah I know, still useful to know in this file IMO. Otherwise could use `Brand`

PR #3402, opened 2023-05-22
[PAY-1195] Mobile chat screen unavailable states

dharit-tan:
### Description
Adds a message at the bottom of the chat screen when current user is not allowed to send messages to another user. Replaces the textinput. Also disables reactions.

Also added a saga to re-fetch a chat and update it if the recheck permissions flag was changed. This is helpful for if the user loses permissions mid-session. Not sure if strictly necessary, maybe it's ok to just let the user continue to message and only show the `ChatUnavailable` state on next app load.

Either way we should probably handle error states for sent messages.

### Dragons

### How Has This Been Tested?

local ios stage
simulator_screenshot_5628657F-0756-47E8-A26E-65ACA7DD14E5

### Feature Flags ###

### DIFF
--- a/packages/common/src/store/pages/chat/sagas.ts
+++ b/packages/common/src/store/pages/chat/sagas.ts
@@ -33,7 +33,7 @@ const {
   fetchUnreadMessagesCountSucceeded,
   fetchUnreadMessagesCountFailed,
   goToChat,
-  fetchChatRecheckPermissions,
+  fetchChatIfNecessary,
   fetchMoreChats,
   fetchMoreChatsSucceeded,
   fetchMoreChatsFailed,
@@ -114,7 +114,7 @@ function* doFetchMoreMessages(action: ReturnType<typeof fetchMoreMessages>) {
     const sdk = yield* call(audiusSdk)
 
     // Ensure we get a chat so we can check the unread count
-    yield* call(fetchChatIfNecessary, { chatId })
+    yield* call(doFetchChatIfNecessary, { chatId })
     const chat = yield* select((state) => getChat(state, chatId))
 
     // Paginate through messages until we get to the unread indicator
@@ -204,7 +204,7 @@ function* doCreateChat(action: ReturnType<typeof createChat>) {
       .sort()
       .join(':')
     try {
-      yield* call(fetchChatIfNecessary, { chatId })
+      yield* call(doFetchChatIfNecessary, { chatId })
     } catch {}
     const existingChat = yield* select((state) => getChat(state, chatId))
     if (existingChat) {
@@ -286,7 +286,10 @@ function* doSendMessage(action: ReturnType<typeof sendMessage>) {
   }
 }
 
-function* fetchChatIfNecessary(args: { chatId: string; bustCache?: boolean }) {
+function* doFetchChatIfNecessary(args: {
+  chatId: string
+  bustCache?: boolean
+}) {
   const { chatId, bustCache = false } = args
   const existingChat = yield* select((state) => getChat(state, chatId))
   if (!existingChat || bustCache) {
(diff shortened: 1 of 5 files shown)


audius-infra:
Preview this change https://demo.audius.co/rt-chat-perms

packages/mobile/src/screens/chat-screen/ChatUnavailable.tsx
```diff
@@ -0,0 +1,132 @@
...
+    color: palette.secondary
+  }
+}))
+
+type ChatUnavailableProps = {
+  user: User
+  chatId: string
+}
+
+export const ChatUnavailable = ({ user, chatId }: ChatUnavailableProps) => {
```
piazzatron, line 47: Could this component just take in the userId and select a user?

packages/common/src/store/pages/chat/sagas.ts
```diff
@@ -432,6 +447,12 @@ function* watchAddMessage() {
   yield takeEvery(addMessage, ({ payload }) => fetchChatIfNecessary(payload))
 }
 
+function* watchFetchChat() {
```
piazzatron, line 450: Could we reuse fetchChatIfNecessary instead of adding a new saga here? Pass a bustCache flag or something?

packages/mobile/src/screens/chat-screen/ChatUnavailable.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+import { UserBadges } from 'app/components/user-badges'
+import { useNavigation } from 'app/hooks/useNavigation'
+import { setVisibility } from 'app/store/drawers/slice'
+import { makeStyles } from 'app/styles'
+
+const { getCanSendMessage } = chatSelectors
+
+const messages = {
+  noAction: 'You can no longer send messages to asdf asdf asdf adsf asdf ',
```
piazzatron, line 16: asdf :)

packages/mobile/src/screens/chat-screen/ChatUnavailable.tsx
```diff
@@ -0,0 +1,132 @@
...
+    dispatch(
+      setVisibility({
+        drawer: 'BlockMessages',
+        visible: true,
+        data: { userId: user.user_id }
+      })
+    )
+  }, [dispatch, user])
+
+  const mapActionToProps = useMemo(() => {
```
piazzatron, line 68: nit: "action" is pretty overloaded in this fn name, and it's returning a JSX component rather than props. any ideas for a better name here?

piazzatron (changes requested):
Looking good - curious if we can see how to do this without adding a saga. Let me know if you think i'm barking up the wrong tree here!

audius-infra:
Preview this change https://demo.audius.co/rt-chat-perms

piazzatron (approved):
Looks good! Just a few nits

packages/common/src/store/pages/chat/sagas.ts
```diff
@@ -432,6 +433,12 @@ function* watchAddMessage() {
   yield takeEvery(addMessage, ({ payload }) => fetchChatIfNecessary(payload))
 }
 
+function* watchFetchChatRecheckPermissions() {
```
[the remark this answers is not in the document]
dharit-tan: I can just name it `fetchChat`, but I was worried that if we didn't make it super clear that one version was checking for recheck permissions and busting cache vs `fetchChatIfNecessary`, one might call this function thinking it would bust cache, but instead it would just return the existing chat.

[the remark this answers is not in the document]
piazzatron: Ahh, I see what you're saying. Maybe cleanest would just be to have a `watchFetchChatIfNecessary` action which itself has a `bustCache` parameter (default false)? Nice to keep it 1:1 with action + saga if possible

audius-infra:
Preview this change https://demo.audius.co/rt-chat-perms

PR #3405, opened 2023-05-23
[C-2299] Fix NowPlayingDrawer navigation deduping bug

sliptype:
### Description

Realized that our `transitionEnd` handlers weren't always firing when attached via `navigation.addListener` inside a screen, but will always fire when attached via `screenListeners` at the Navigator level. So decided on a simpler approach than what was previously discussed (see below). 

* Track `lastNavAction` at `AppTabScreen` level. This lives outside of the component tree but I don't think should cause any problems because we are only checking its value in an event handler. Would love thoughts here though!

#### Previously discussed approach

Was worried a bit about making such large navigation changes, would require a lot of testing. Also this approach made it more difficult to implement deduping, which is a requirement

* keep active tab in redux
* use navigation ref to navigate from nowplaying
* go to correct path using a nested route argument for `push`, based on redux value
* anywhere nav ref is used, make sure to use the hook to navigate correctly

### Dragons

Touches navigation

### How Has This Been Tested?
Confirmed that deduping still works.
Confirmed that all NowPlaying navigation works as expected.
Tested navigation in general

### Feature Flags ###


packages/mobile/src/screens/app-screen/AppTabScreen.tsx
```diff
@@ -155,14 +160,27 @@ export const AppTabScreen = ({ baseScreen, Stack }: AppTabScreenProps) => {
...
+
   useEffect(() => {
     drawerNavigation?.setOptions({ swipeEnabled: !isNowPlayingDrawerOpen })
   }, [drawerNavigation, isNowPlayingDrawerOpen])
 
   return (
     <Stack.Navigator
       screenOptions={screenOptions}
-      screenListeners={{ state: handleChangeState }}
+      screenListeners={{
```
dylanjeffers, line 180: would memozing this object help at all for general perf you think?

dylanjeffers (approved):
agreed a bit of a dicey pattern, but lets try this out and see how it feels, much much simpler than the alternative approach

packages/mobile/src/screens/app-screen/AppTabScreen.tsx
sliptype: Hard to say, depends on how often this gets rendered. We could dig in and see but I would say it's ok to leave as is for now

PR #3406, opened 2023-05-23
[C-2598] Add types for audius-query hooks

amendelsohn:
### Description

Fetch hooks generated by audius-query will now have accurate typing based on the args and return type from the endpoint's fetch function.

Related changes:
- Endpoint options and schema key are now required
- Schema key is used on the CreateApi side to wrap response for normalizr


dylanjeffers (approved):
cleaner and better typed, win win :) nice job

PR #3407, opened 2023-05-23
[PAY-1222] Fix unread messages count

piazzatron:
### Description

Fixes multiple bugs here. 

**Bugs:**
- Retransmitted messages would get double counted. Fix this by losing the `incrementUnreadCount` action entirely and performing the increment in the `addMessage` reducer.
- If scrolled up in a chat, own messages would be counted because the chat would be considered inactive. Now explicitly check for own messages.
- Fixed issues with scroll handler: need to throttle on leading and trailing edge (otherwise it might not fire on scrolling down to bottom of the screen), which necessitated a change to event.target instead of event.currentTarget, bc the latter may no longer existed when the throttled scrollhandler runs
- Fixed issue where chats wouldn't be immediately marked as read when going to a new chat page because we relied on the scroll handler to fire at least once; fix is to mark as read in a useEffect on load. 
- Adjusted scroll bottom threshold to something more permissive.

### Dragons

Need to test more on mobile.

### How Has This Been Tested?

Web

### Feature Flags ###

### DIFF
--- a/packages/common/src/store/pages/chat/middleware.ts
+++ b/packages/common/src/store/pages/chat/middleware.ts
@@ -7,13 +7,8 @@ import { encodeHashId } from 'utils/hashIds'
 
 import { actions as chatActions } from './slice'
 
-const {
-  connect,
-  disconnect,
-  addMessage,
-  incrementUnreadCount,
-  setMessageReactionSucceeded
-} = chatActions
+const { connect, disconnect, addMessage, setMessageReactionSucceeded } =
+  chatActions
 
 export const chatMiddleware =
   (audiusSdk: () => Promise<AudiusSdk>): Middleware =>
@@ -34,13 +29,17 @@ export const chatMiddleware =
             console.debug('[chats] WebSocket opened. Listening for chats...')
           }
           messageListener = ({ chatId, message }) => {
-            store.dispatch(
-              addMessage({ chatId, message, status: Status.SUCCESS })
-            )
             const currentUserId = getUserId(store.getState())
-            if (message.sender_user_id !== encodeHashId(currentUserId)) {
-              store.dispatch(incrementUnreadCount({ chatId }))
-            }
+            const isSelfMessage =
+              message.sender_user_id === encodeHashId(currentUserId)
+            store.dispatch(
+              addMessage({
+                chatId,
+                message,
+                status: Status.SUCCESS,
+                isSelfMessage
+              })
+            )
           }
           reactionListener = ({ chatId, messageId, reaction }) => {
             store.dispatch(
--- a/packages/common/src/store/pages/chat/sagas.ts
+++ b/packages/common/src/store/pages/chat/sagas.ts
@@ -245,9 +245,11 @@ function* doMarkChatAsRead(action: ReturnType<typeof markChatAsRead>) {
     const audiusSdk = yield* getContext('audiusSdk')
     const sdk = yield* call(audiusSdk)
     const chat = yield* select((state) => getChat(state, chatId))
-    if (!chat || !dayjs(chat?.last_read_at).isAfter(chat?.last_message_at)) {
+    if (!chat || dayjs(chat?.last_read_at).isBefore(chat?.last_message_at)) {
       yield* call([sdk.chats, sdk.chats.read], { chatId })
       yield* put(markChatAsReadSucceeded({ chatId }))
+    } else {
+      yield* put(markChatAsReadFailed({ chatId }))
     }
   } catch (e) {
     console.error('markChatAsReadFailed', e)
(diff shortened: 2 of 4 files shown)


packages/common/src/store/pages/chat/slice.ts
```diff
@@ -372,50 +373,52 @@ const slice = createSlice({
...
+          // If a new message comes through, we don't need to recheck permissions anymore
+          recheck_permissions: false
+        }
+      })
+
+      // Recalculate tails
+      recalculatePreviousMessageHasTail(state.messages[chatId], 0)
+
+      // Handle unread counts
+
```
dharit-tan, line 403: delete extra line

packages/common/src/store/pages/chat/slice.ts
```diff
@@ -372,50 +373,52 @@ const slice = createSlice({
...
+      if (!isSelfMessage) {
+        // If we're actively reading, this will immediately get marked as read.
+        // Ignore the unread bump to prevent flicker
+        if (state.activeChatId !== chatId) {
+          const existingChat = getChat(state, chatId)
+          const optimisticRead = state.optimisticChatRead[chatId]
```
dharit-tan, line 409: Were we also gonna delete the optimistic unread count in some case? I think it was when a new non-self message comes in?

packages/common/src/store/pages/chat/middleware.ts
```diff
@@ -34,13 +29,17 @@ export const chatMiddleware =
             console.debug('[chats] WebSocket opened. Listening for chats...')
           }
           messageListener = ({ chatId, message }) => {
+            const currentUserId = getUserId(store.getState())
```
dharit-tan, line 32: Have a strange feeling that we should be able to get the currentUserId inside the slice and do the comparison there instead of plumbing it down, but maybe not..

audius-infra:
Preview this change https://demo.audius.co/pay-1222

dharit-tan (approved):

packages/web/src/pages/chat-page/components/ChatMessageList.tsx
```diff
@@ -76,6 +76,12 @@ export const ChatMessageList = forwardRef<HTMLDivElement, ChatMessageListProps>(
 
     const ref = useRef<HTMLDivElement>(null)
 
+    // On first load, mark chat as read
+    useEffect(() => {
+      if (!chatId) return
```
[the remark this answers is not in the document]
piazzatron: +1

packages/common/src/store/pages/chat/slice.ts
piazzatron: I think we don't need to do this

packages/common/src/store/pages/chat/middleware.ts
piazzatron: discussed IRL

packages/common/src/store/pages/chat/sagas.ts
```diff
@@ -245,9 +245,11 @@ function* doMarkChatAsRead(action: ReturnType<typeof markChatAsRead>) {
     const audiusSdk = yield* getContext('audiusSdk')
     const sdk = yield* call(audiusSdk)
     const chat = yield* select((state) => getChat(state, chatId))
-    if (!chat || !dayjs(chat?.last_read_at).isAfter(chat?.last_message_at)) {
+    if (!chat || dayjs(chat?.last_read_at).isBefore(chat?.last_message_at)) {
       yield* call([sdk.chats, sdk.chats.read], { chatId })
       yield* put(markChatAsReadSucceeded({ chatId }))
+    } else {
+      yield* put(markChatAsReadFailed({ chatId }))
```
dharit-tan, line 252: Nit: bit confusing bc markChatAsRead didn't actually fail, we just want to clear optimistic read count. Could define an explicit clearOptimsticReadCount fn. But honestly this is probably fine.

audius-infra:
Preview this change https://demo.audius.co/pay-1222

packages/common/src/store/pages/chat/slice.ts
```diff
@@ -372,50 +373,52 @@ const slice = createSlice({
...
+            : existingChat?.unread_message_count ?? 0
+          chatsAdapter.updateOne(state.chats, {
+            id: chatId,
+            changes: { unread_message_count: existingUnreadCount + 1 }
+          })
+        }
+
+        // Web or mobile: update optimistic unread count
+        state.optimisticUnreadMessagesCount =
+          (state.optimisticUnreadMessagesCount ?? state.unreadMessagesCount) + 1
```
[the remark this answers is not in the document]
rickyrombo: Does this work?

rickyrombo (approved):

PR #3418, opened 2023-05-24
[PAY-925] Adds unread message icons on web

schottra:
### Description
This implements unread message notifications at the navigation level for web.

* Added an `Icon` helper component which can be used to load any of our icons and optionally add a decorator component
* Added `NotificationDot` component with small / large size variants
* Updated nav popup component to render notification dots on the icon and messages menu item when user has unread chat messages

### Dragons
N/A

### How Has This Been Tested?
Manual testing on web

N/A

### Feature Flags ###
N/A

image


rickyrombo (approved):
_love_ ❤️ 

EDIT: check stylelint in CI

schottra:
> Looks great! Worth flagging this `Icon` component to the client team / design system migration effort at some point

cc @dylanjeffers @sliptype

packages/web/src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,27 @@
...
+/** Renders a stems Icon component with optional decorator
+ * Ex: `<Icon icon={IconKebabHorizontal} decorator={<NotificationDot variant='small' />} />`
+ */
+export const Icon = ({
+  decorator,
+  icon: IconComponent,
+  ...iconProps
+}: IconProps) => {
+  return decorator ? (
+    <div className={styles.iconContainer}>
```
dylanjeffers, line 20: spans maybe?

packages/web/src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,27 @@
...
+      <IconComponent />
+      <div className={styles.iconDecoration}>{decorator}</div>
+    </div>
+  ) : (
+    React.createElement(IconComponent, iconProps)
```
dylanjeffers, line 25: cant just do <IconComponent {...iconProps} /> ?

packages/web/src/components/notification-dot/NotificationDot.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import styles from './NotificationDot.module.css'
+
+type NotificationDotVariant = 'small' | 'large'
+type NotificationDotProps = {
+  variant: NotificationDotVariant
+} & ComponentProps<'span'>
+
+/** Renders a basic notification dot used to decorate controls and indicate new information */
+export const NotificationDot = (props: NotificationDotProps) => {
```
dylanjeffers, line 13: do we not have this dot already? just in mobile i guess?

dylanjeffers (commented):
just adding some comments :)

packages/web/src/components/notification-dot/NotificationDot.tsx
schottra: Just in mobile, correct. We had a sort of similar dot for when playlists are updated on desktop. But it was messy. This is formalizing it as a reusable component.

packages/web/src/components/Icon/Icon.tsx
schottra: Good callout I'll try it and see if it affects the positioning behavior.

audius-infra:
Preview this change https://demo.audius.co/pay-925-new-message-indicator

packages/web/src/components/Icon/Icon.tsx
dylanjeffers: interesting, i do believe it is the case that spreading doesn't lead to a unnecessary rerenders

schottra: I did some investigating:

PR #3421, opened 2023-05-24
[PAY-1258] Add toast when chat messages are copied

schottra:
### Description
Adds some logic to show a toast when a user copies a chat message. Also cleaned up a couple of callbacks which were redefined every render unnecessarily

Kapture 2023-05-24 at 15 23 11

### Dragons
N/A

### How Has This Been Tested?
Manual testing on physical ios device and simulator

N/A

### Feature Flags ###
N/A

### DIFF
--- a/packages/mobile/src/screens/chat-screen/ReactionPopup.tsx
+++ b/packages/mobile/src/screens/chat-screen/ReactionPopup.tsx
@@ -11,6 +11,7 @@ import { View, Dimensions, Pressable, Animated } from 'react-native'
 import { useDispatch, useSelector } from 'react-redux'
 
 import { usePopupAnimation } from 'app/hooks/usePopupAnimation'
+import { useToast } from 'app/hooks/useToast'
 import { makeStyles } from 'app/styles'
 import { spacing } from 'app/styles/spacing'
 import { zIndex } from 'app/utils/zIndex'
@@ -26,6 +27,10 @@ import {
 
 const { getUserId } = accountSelectors
 const { setMessageReaction } = chatActions
+
+const messages = {
+  messageCopied: 'Message copied to clipboard'
+}
 
 const useStyles = makeStyles(({ spacing, palette, typography }) => ({
   dimBackground: {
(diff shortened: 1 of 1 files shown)


packages/mobile/src/screens/chat-screen/ReactionPopup.tsx
```diff
@@ -151,15 +158,27 @@ export const ReactionPopup = ({
     [userId, handleClosePopup, dispatch, chatId, userIdEncoded]
   )
 
-  const handleCopyPress = useCallback(
-    (message: string) => {
-      Clipboard.setString(message)
-      handleClosePopup()
+  const { message: messageText } = message
```
rickyrombo, line 161: super nit and up to you, but I could almost prefer doing `message.message` rather than destructuring so we don't have `message`, `messages`, _and_ `messageText`

rickyrombo (approved):

packages/mobile/src/screens/chat-screen/ReactionPopup.tsx
```diff
@@ -151,15 +158,27 @@ export const ReactionPopup = ({
...
+      }
     },
-    [handleClosePopup]
+    [message, handleReactionSelected]
   )
 
-  return shouldShowPopup ? (
+  if (!shouldShowPopup) {
+    return null
+  }
```
piazzatron, line 179: <3 early return

piazzatron (approved):
Looks great and nice cleanups :D

packages/mobile/src/screens/chat-screen/ReactionPopup.tsx
schottra: I would tend to agree but `message.message` grinds on me more than the multiple variables 😆 
I wish we had named the outer object and the inner field different things (`message.text` or `chat.message`)

rickyrombo: yeeeah. could have called it chatMessage.message but that's not much better :joy: oh well, there's two hard problems in programming - cache invalidation, naming things, and off by one errors

PR #3422, opened 2023-05-24
[C-2644] Fix issue where new playlist drag-drop not persisting

dylanjeffers:
### Description

Fixes issue where drag-dropping new (unfavorited) playlists into the left nav wouldnt appear. Also fixes case of drag-dropping to specific sections in the left nav, including inside folders
- Refactors playlist-library saga to use `typed-redux-saga` which caught a few type bugs.
- Small styling fix with overflowing text. Now that we wrap everything in a span, we also need to text wrap it.

### DIFF
--- a/packages/web/src/components/nav/desktop/PlaylistLibrary/CollectionNavItem.module.css
+++ b/packages/web/src/components/nav/desktop/PlaylistLibrary/CollectionNavItem.module.css
@@ -9,7 +9,7 @@
   display: flex;
   align-items: center;
   line-height: normal;
-  overflow: hidden;
+  width: 100%;
 }
 
 .level1 {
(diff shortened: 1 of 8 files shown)


packages/web/src/components/nav/desktop/PlaylistLibrary/CollectionNavItem.module.css
```diff
@@ -9,6 +9,7 @@
   display: flex;
   align-items: center;
   line-height: normal;
+  overflow: hidden;
```
Kyle-Shanks, line 12: why was this needed? just curious

dylanjeffers: oh i see your comment now :) yes as discussed not needed, thanks!

PR #3431, opened 2023-05-25
[CON-678] v2 user profile edit

theoilie:
### Description
Makes editing profiles work for v2 users and/or with v2 uploads for profile pic and cover art.

### Dragons
None

### How Has This Been Tested?
Tested updating a v1 and v2 user locally with https://github.com/AudiusProject/audius-protocol/pull/5291. Set the storage v2 upload feature flag to true/false to test both code paths.

Make sure staging profile edits work.

### Feature Flags ###
No new flags

### DIFF
--- a/packages/common/src/services/audius-backend/AudiusBackend.ts
+++ b/packages/common/src/services/audius-backend/AudiusBackend.ts
@@ -1650,7 +1650,7 @@ export const audiusBackend = ({
       }
     } catch (err) {
       console.error(getErrorMessage(err))
-      return false
+      throw err
     }
   }
 
--- a/packages/common/src/services/audius-backend/AudiusBackend.ts
+++ b/packages/common/src/services/audius-backend/AudiusBackend.ts
@@ -1345,7 +1345,7 @@ export const audiusBackend = ({
         metadata: updatedMetadata,
 
         // We don't need these properties, but the confirmer expects them.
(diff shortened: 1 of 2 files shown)


packages/common/src/services/audius-backend/AudiusBackend.ts
```diff
@@ -1618,74 +1634,20 @@ export const audiusBackend = ({
...
+            newMetadata.user_id,
+            newMetadata,
+            writeMetadataThroughChainEnabled
+          )
+        return { blockHash, blockNumber, userId }
       }
-
-      newMetadata = schemas.newUserMetadata(newMetadata, true)
-
-      const { blockHash, blockNumber } = await audiusLibs.User.updateUser(
```
theoilie, line 1684: this was removed from libs in [here](https://github.com/AudiusProject/audius-protocol/pull/5012/files#diff-6cf87bbc23e9cdb74c8f51008bafae99d2d6f17a1193a4999f17ee0ed1842653L637), so this whole function would always error. I only saw 2 instances of "function not found" in Sentry, though, because this path is only called when a user doesn't have a replica set, which should never happen because sanity checks on login should ensure they have one

audius-infra:
Preview this change https://demo.audius.co/theo-con678-edit-user-v2

sliptype (approved):
lgtm!

PR #3442, opened 2023-05-26
[PAY-1138] Fix headers on Chats and fix old header prod bug

piazzatron:
### Description

Solved the header gutter problem by making the gutter logic from `Header.js` reusable and including it in the `ChatHeader`. 
- Also refactored this from a bunch of sloppy hooks manually manipulating the DOM to a regular React component that Portals to body.
- Handle the shadow differences on ChatsPage by allowing style overrides to the header, which are also passed to the header gutter
- Fix an old bug, shown below, where the gutter would fail to account for the height of the banner when switching pages, resulting in a "notch". Fixed by having the gutter be responsible for offsetting itself via shared context, instead of app.js doing it.

Old broken bug:

https://github.com/AudiusProject/audius-client/assets/6780028/c799798c-a68d-4d6c-841e-b426ffef03a1

New fixedness:

https://github.com/AudiusProject/audius-client/assets/6780028/c9c83f7c-09da-4258-8682-12ae64369099

### Dragons
- Nests whole app in a non-memoized context provider. But this thing shouldn't change often so not worried about perf 
- Touches headers. But shouldn't change logic. 

### How Has This Been Tested?

Web

### Feature Flags ###

### DIFF
--- a/packages/web/src/components/header/desktop/Header.module.css
+++ b/packages/web/src/components/header/desktop/Header.module.css
@@ -124,3 +124,14 @@
   left: 0px;
   z-index: 11;
 }
+
+.gutter {
+  right: 0px;
+  position: fixed;
+  box-shadow: rgba(133, 129, 153, 0.18) 1px 2px 5px -2px;
+  z-index: -1;
+}
+
+.bannerMargin {
+  margin-top: 40px;
+}
--- a/packages/web/src/components/header/desktop/Header.module.css
+++ b/packages/web/src/components/header/desktop/Header.module.css
@@ -124,14 +124,3 @@
   left: 0px;
   z-index: 11;
 }
-
-.gutter {
-  right: 0px;
-  position: fixed;
-  box-shadow: rgba(133, 129, 153, 0.18) 1px 2px 5px -2px;
-  z-index: -1;
-}
-
-.bannerMargin {
-  margin-top: 40px;
-}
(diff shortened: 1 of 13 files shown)


packages/web/src/components/header/desktop/HeaderGutter.tsx
```diff
@@ -0,0 +1,77 @@
+import { useContext, createContext } from 'react'
+
+import cn from 'classnames'
+import { head } from 'lodash'
+
+import { usePortal } from 'hooks/usePortal'
+
+import styles from './Header.module.css'
```
rickyrombo, line 8: This should use its own `HeaderGutter.module.css`

packages/web/src/components/header/desktop/Header.module.css
```diff
@@ -124,3 +124,14 @@
...
+
+.gutter {
+  right: 0px;
+  position: fixed;
+  box-shadow: rgba(133, 129, 153, 0.18) 1px 2px 5px -2px;
+  z-index: -1;
+}
+
+.bannerMargin {
+  margin-top: 40px;
```
rickyrombo, line 136: this stinks 😢 maybe someday we can make the banner push the container down so we don't have to do this

rickyrombo (approved):

packages/web/src/components/header/desktop/Header.module.css
schottra: At the very least, I would say make the height a CSS variable that's used both in the banner component and here?

piazzatron: Didn't want to do a deeper refactor than I already did here (since this is how it worked before), but yeah agreed. I'll pull it out into a CSS variable at least

audius-infra:
Preview this change https://demo.audius.co/piazz-fix-headers

PR #3445, opened 2023-05-26
[PAY-1150] Disable scrolling when chat reaction popup is active

dharit-tan:
### Description
Partially solves a bug where while pressing and holding on a chat message, after the reactions popup appears, the user can still scroll vertically on the underlying flatlist, as well as swipe left to navigate back on the navigator. This commit disables scrolling when the popup is active. Unfortunately the user can still swipe left, and if time a scroll gesture right at the end of the longpress there will still be an offset between the popup and original chat message.

### Dragons

### How Has This Been Tested?

Local ios stage

### Feature Flags ###


dylanjeffers (approved):
great 80/20 for now

PR #3449, opened 2023-05-26
[C-2647] Use entire cache to denormalize data

amendelsohn:
### Description

Select the full cache from store to rehydrate entities in `audius-query`

Fixes issue where we didn't know what entities to use to replace normalized id references data on cache hit.

### Dragons

Selecting the whole cache may cause extra renders, but we're doing this already. UseProxySelector will hopefully avoid the worst of it


dylanjeffers (approved):
wow also looks cleaner too. excellent work

audius-infra:
Preview this change https://demo.audius.co/audius-query-rehydrate-single-entity

PR #3450, opened 2023-05-26
[C-2653] Add shallow argument to audius-query hooks

amendelsohn:
### Description

if `shallow: true`, don't rehydrate the full entity structure.

In the case where kind is undefined, the top-level data is not an entity, so just returning the nonNormalizedData will work. However, if `kind: true`, nonNormalizedData just contains an id, so we need to do one level of denormalization. We provide only the matching kind so we don't repopulate the nested entities.

### DIFF
--- a/packages/common/src/audius-query/utils.ts
+++ b/packages/common/src/audius-query/utils.ts
@@ -1,11 +1,9 @@
-import { mapValues, zipObject } from 'lodash'
-
 import { Kind } from 'models/Kind'
 import { CommonState } from 'store/reducers'
 
 import * as cacheSelectors from '../store/cache/selectors'
 
-import { EntityMap, StrippedEntityMap } from './types'
+import { EntityMap } from './types'
 
 export function capitalize(str: string) {
   return str.replace(str[0], str[0].toUpperCase())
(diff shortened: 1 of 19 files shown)


packages/common/src/audius-query/utils.ts
```diff
@@ -15,8 +16,18 @@ export const getKeyFromFetchArgs = (fetchArgs: any) => {
 
 export const selectCommonEntityMap = (
   state: CommonState,
-  kind?: Kind
+  kind?: Kind,
+  shallow?: boolean
 ): EntityMap | null => {
+  if (kind && shallow) {
```
dylanjeffers, line 22: sick so basically we dont need to grab the whole state right?

dylanjeffers (approved):
love this addition! extra credit for v2 is to add docs in the form of js-doc, and then can generate readme args (or just have them as dupes :))

packages/common/src/audius-query/utils.ts
amendelsohn: right and we only need to do this if the top level data is itself an entity. Without this step, we'd just get `useGetTrackById({ id: 123 }, { shallow: true }) returns { data: 123 }`

audius-infra:
Preview this change https://demo.audius.co/audius-query-shallow

PR #3452, opened 2023-05-26
[C-2646] Add idListArgKey for lists of entities; add fetchArgs types

amendelsohn:
### Description

Adds `idListArgKey` which allows cache hit for lists of entities if all entities are already present in the cache.
Also added argument types to the existing api endpoint fetch functions so the hooks can typecheck the fetchArgs.

Future work:
- Selective fetching so we can hit some of the entities from cache and fetch the rest

### DIFF
--- a/packages/common/src/audius-query/createApi.ts
+++ b/packages/common/src/audius-query/createApi.ts
@@ -1,7 +1,7 @@
 import { useContext, useEffect } from 'react'
 
 import { CaseReducerActions, createSlice } from '@reduxjs/toolkit'
-import { isEqual } from 'lodash'
+import { isEqual, mapValues } from 'lodash'
 import { denormalize, normalize } from 'normalizr'
 import { useDispatch, useSelector } from 'react-redux'
 
@@ -12,6 +12,7 @@ import { getCollection } from 'store/cache/collections/selectors'
 import { getTrack } from 'store/cache/tracks/selectors'
 import { CommonState } from 'store/reducers'
 import { getErrorMessage } from 'utils/error'
+import { removeNullable } from 'utils/typeUtils'
 
 import * as cacheActions from '../store/cache/actions'
 import * as cacheSelectors from '../store/cache/selectors'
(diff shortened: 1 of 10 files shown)


packages/web/src/pages/chat-page/components/ChatMessagePlaylist.tsx
```diff
@@ -44,9 +44,9 @@ export const ChatMessagePlaylist = ({
   const { data: playlist, status } = useGetPlaylistById(
     {
       playlistId,
-      currentUserId
+      currentUserId: currentUserId!
     },
-    { disabled: !playlistId }
+    { disabled: !playlistId || !currentUserId }
```
amendelsohn, line 49: Not in love with this pattern. Idk if it's possible to make the type of args understand that it's only called after the disabled condition is false. It would be great to have the behavior like

```
const disabled = !playlistId || !currentUserId
if (!disabled) {
	fetch({playlistId, currentUserId})
}
```
in which it would know that by the time we call fetch, the args must exist

packages/common/src/audius-query/createApi.ts
```diff
@@ -145,45 +146,59 @@ const buildEndpointHooks = <
...
-
-        let permalinkCachedEntity = null
-        if (kind === Kind.TRACKS && permalinkArgKey) {
-          permalinkCachedEntity = getTrack(state, {
-            permalink: fetchArgs[permalinkArgKey]
-          })
-        }
-        if (kind === Kind.COLLECTIONS && permalinkArgKey) {
-          permalinkCachedEntity = getCollection(state, {
-            permalink: fetchArgs[permalinkArgKey]
```
amendelsohn, line 177: @sddioulde Just a callout that I refactored some of the `permalinkArgKey` logic. Pretty sure it's still working, but if you run into trouble after merging, lmk

audius-infra:
Preview this change https://demo.audius.co/audius-query-idsargkey

dylanjeffers (approved):
I dont think i totally understand all the changes, but trust you if you want to merge this now :)

dylanjeffers (approved):
I dont think i totally understand all the changes, but trust you if you want to merge this now :)

PR #3459, opened 2023-05-30
[C-2635] Fix offline playback; broken signature fetch

amendelsohn:
### Description

Fetching signatures for gated content was breaking offline playback. The queueing of tracks would fail because the audiusBackend fetch failed.

Shoutouts @dylanjeffers for pairing help


dylanjeffers (approved):
noice

dylanjeffers (approved):
noice

sliptype (approved):
Wow! Great find

PR #3461, opened 2023-05-30
[PAY-1278] Indicator for end of chat history on mobile

dharit-tan:
### Description
Add an indicator that tells the user they've reached the end of the chat history. Only show it when the chat list is scrollable, ie. the content height is larger than the container.

### How Has This Been Tested?

Local ios stage
- Confirmed it does not flicker upon entering chat screen.
- Confirmed it does not flicker at the top of the chat messages list before older messages have loaded in.
Simulator Screen Shot - iPhone 14 Pro - 2023-05-30 at 16 48 41

### Feature Flags ###


packages/mobile/src/screens/chat-screen/Indicator.tsx
```diff
@@ -0,0 +1,41 @@
...
+    textTransform: 'uppercase',
+    letterSpacing: 0.5,
+    backgroundColor: palette.neutralLight5,
+    paddingHorizontal: spacing(2),
+    paddingVertical: spacing(1),
+    borderRadius: spacing(0.5)
+  }
+}))
+
+export const Indicator = ({ content }: { content: string }) => {
```
rickyrombo, line 31: Love that this is split out - maybe could use a slightly more descriptive name though?

rickyrombo (approved):

packages/mobile/src/screens/chat-screen/Indicator.tsx
rickyrombo: Maybe `MessageSeparator`? Or `SeparatorWithText`?

dharit-tan: Called it `ChatMessageSeparator`

PR #3463, opened 2023-05-30
[PAY-1291] Adds new message toast when scrolled up on mobile

schottra:
### Description
This PR adds a behavior to `ChatScreen` which will track the most recent _received_ message and fire a toast if it changes while the user is scrolled far enough from the bottom.

* Added memo/hook to track most recently received message (since chats array is most recent first, it's the first index we find where the sender is not the current user)
* Added handler/ref to track current scroll position of the message list in a chat
* Consolidated logic for calculating auto-scroll threshold on iOS into a helper function
* Added helper to compute minimum scroll threshold needed to show the new toast. The idea here is that on iOS, we're using [`maintainVisibleContentPosition`](https://reactnative.dev/docs/0.66/scrollview#maintainvisiblecontentposition-ios) to automatically scroll new messages into view if we're within a small distance of the "top" of the list. We don't want to show the toast in those cases. On Android, there is no auto-scrolling, so we'll always use the threshold (100 units, roughly the size of a line of text plus timestamp)
* Added hook to track change of most recently received message id and fire the toast if applicable
* Added some helper types to make working with FlatList events a little easier

Kapture 2023-05-30 at 17 44 49

### Dragons
N/A

### How Has This Been Tested?
Tested on iOS/Android physical devices and iOS simulator

N/A

### Feature Flags ###
N/A


packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -1,30 +1,38 @@
...
+  Status,
   accountSelectors,
+  chatActions,
+  chatCanFetchMoreMessages,
   chatSelectors,
-  encodeUrlName,
   decodeHashId,
   encodeHashId,
-  Status,
+  encodeUrlName,
```
schottra, line 13: Most of the changes up here are due to an `Organize Imports` removing unused items and resorting the rest.

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -296,7 +335,35 @@ export const ChatScreen = () => {
...
+    const idx = chatMessages.findIndex(
+      (chat) => chat.sender_user_id !== userIdEncoded
+    )
+    return idx >= 0 ? chatMessages[idx].message_id : null
+  }, [chatMessages, userIdEncoded])
+
+  // If most recent message changes and we are scrolled up, fire a toast
+  useEffect(() => {
+    if (
+      newestReceivedMessageId != null &&
```
dharit-tan, line 348: would `newestReceivedMessageId &&...` be more readable?

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -46,6 +55,11 @@ import { ChatUnavailable } from './ChatUnavailable'
 import { EmptyChatMessages } from './EmptyChatMessages'
 import { ReactionPopup } from './ReactionPopup'
 
+type ChatFlatListProps = FlatListProps<ChatMessageWithExtras>
```
dharit-tan, line 58: nicee

rickyrombo (approved):

dharit-tan (approved):
I have it in my to-do to use this package to support the same behavior on android: https://github.com/GetStream/flat-list-mvcp
tho I don't think this changes the "no auto-scrolling on android"

PR #3481, opened 2023-06-01
[C-2702] Remove initialSelectedNode from disc-node selector

dylanjeffers:
### Description

Removes the initialSelectedNode from discovery-node-selector to prevent issues where users get stuck on a node that may be unhealthy.


rickyrombo (approved):

raymondjacobson (approved):
nice!!

PR #3484, opened 2023-06-01
[PAY-1308] Use all messages for tracking latest received

schottra:
### Description
The logic I had before was well-intentioned (find the first message that isn't from the current user). But it has a flaw when the first page of messages are all from the current user and you have to scroll up to see the first "received" message.
I changed the logic to just run the effect whenever the most recent message changes, and only do the toast logic when the most recent message is from the other user.

### Dragons
N/A

### How Has This Been Tested?
Tested locally on physical devices

N/A

### Feature Flags ###
N/A

### DIFF
--- a/packages/mobile/src/screens/chat-screen/ChatScreen.tsx
+++ b/packages/mobile/src/screens/chat-screen/ChatScreen.tsx
@@ -325,31 +325,31 @@ export const ChatScreen = () => {
     }
   }, [earliestUnreadIndex, chatMessages])
 
-  const newestReceivedMessageId = useMemo(() => {
-    const idx = chatMessages.findIndex(
-      (chat) => chat.sender_user_id !== userIdEncoded
-    )
-    return idx >= 0 ? chatMessages[idx].message_id : null
-  }, [chatMessages, userIdEncoded])
+  const newestMessage = chatMessages.length > 0 ? chatMessages[0] : null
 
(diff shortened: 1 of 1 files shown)


packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -325,31 +325,31 @@ export const ChatScreen = () => {
     }
   }, [earliestUnreadIndex, chatMessages])
 
-  const newestReceivedMessageId = useMemo(() => {
-    const idx = chatMessages.findIndex(
-      (chat) => chat.sender_user_id !== userIdEncoded
-    )
-    return idx >= 0 ? chatMessages[idx].message_id : null
-  }, [chatMessages, userIdEncoded])
+  const newestMessage = chatMessages.length > 0 ? chatMessages[0] : null
```
dharit-tan, line 328: tiny nit: i've been using `latestMessage` vs `earliestMessage`. tho tbh newer vs older maybe is easier to understand.

dharit-tan (approved):
nice - much simpler logic :)

rickyrombo (approved):

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
schottra: `latest` is fine! I'd rather be consistent.

PR #3489, opened 2023-06-01
[PAY-1336] Updates to mobile chats user search screen

dharit-tan:
### Description
Changes to search text input on create new chat screen. Did not implement the moving label - would require a deeper dive to get it to work well with a right icon + clear icon.

### Dragons

- Updates to core `TextInput`. This is the first usage of `clearable` so I updated the icon and its size.
- Removed `height` and `width` in `styles.icon` - correct me if i'm wrong but don't think these do anything, and we should set via component props instead?
- Updated `iconSearch` to new version, will appear in other places in the app but it looks fine on homescreen.

### How Has This Been Tested?

Local ios stage
Simulator Screen Shot - iPhone 14 Pro - 2023-06-01 at 17 13 58
Simulator Screen Shot - iPhone 14 Pro - 2023-06-01 at 17 13 54

### Feature Flags ###


packages/mobile/src/components/core/TextInput.tsx
```diff
@@ -67,13 +68,12 @@ const useStyles = makeStyles(({ typography, palette, spacing }) => ({
     color: palette.neutral,
     fontFamily: typography.fontByWeight.medium,
     minWidth: 40,
+    flexGrow: 1,
     // Needed for android
     padding: 0
   },
   icon: {
-    fill: palette.neutralLight5,
-    height: spacing(4),
```
rickyrombo, line 75: worth noting the default size of these icons is 24x24 and we're removing this which would resize them to 16x16 - might want to check other usages?

rickyrombo (approved):

packages/mobile/src/components/core/TextInput.tsx
dharit-tan: nice good call. there are other uses which rely on 16. gonna add icon style as a prop.

packages/mobile/src/components/core/TextInput.tsx
```diff
@@ -331,11 +335,19 @@ export const TextInput = forwardRef<RNTextInput, TextInputProps>(
                   right: spacing(2)
                 }}
               >
-                <IconCloseAlt fill={styles.icon.fill} />
+                <IconCloseAlt
+                  fill={iconProp?.fill ?? styles.icon.fill}
```
rickyrombo, line 339: might be easier to create an object above like:
```
const iconProps = {
   ...styles.icon
   ...iconProp
 }
 ```
 to minimize the `??` checks

rickyrombo (approved):

PR #3499, opened 2023-06-02
[pay-1311] Reset chat state on sign out

schottra:
### Description
Since our chat state reducers and sagas are all based on the current signed-in user and don't support multiple user chat states, we should reset it on sign out. This isn't an issue for web, as the page is refreshed after sign out and we lose all redux state.

### Dragons
N/A

### How Has This Been Tested?
Tested on physical device against staging

N/A

### Feature Flags ###
N/A


dharit-tan (approved):

audius-infra:
Preview this change https://demo.audius.co/pay-1311-messages-persist-across-session

PR #3515, opened 2023-06-06
[PAY-1333] Mobile dms end of history only shows with >10 messages

dharit-tan:
### Description
Require 10 messages to show the "End of Message History" indicator.

Also removed the `minHeight: 100%` css on flatlist container bc it was causing the flatlist to be scrollable even when there were 1-2 messages. Thought it was needed to make messages appear at the bottom, but apparently not anymore (??).

This solution is not perfect - sometimes we will get smaller values than expected for `flatListInnerHeight` and not show the end of history indicator. I think this is preferable to showing it erroneously when we don't want to. Calling this good enough for now.

### Dragons

`onLayout` is finnicky:
- Sometimes `chatContainerBottom` measurements don't happen at the right time and it comes out to 0. Strangely it only happens on _some_ chats (Rufus the Roo consistently repros this but has no discernible reason as to why).
- Values returned from `measureInWindow` and `onContentsSizeChanged` are not the same. `onContentSizeChanged` seems to be correct, `measureInWindow` is consistently undersized.

### How Has This Been Tested?

Local ios stage

### Feature Flags ###


packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -231,15 +230,12 @@ export const ChatScreen = () => {
...
-  // starts at screen height due to height: 100% css property.
+  // Only show the end reached indicator if there are no previous messages, more than 10 messages,
+  // and the flatlist is scrollable.
   const shouldShowEndReachedIndicator =
     !(chat?.messagesSummary?.prev_count ?? true) &&
-    flatListInnerHeight.current - chatContainerTop.current >
-      chatContainerBottom.current -
-        chatContainerTop.current +
-        END_REACHED_OFFSET
+    chatMessages.length > END_REACHED_MIN_MESSAGES &&
```
dharit-tan, line 237: No longer need `END_REACHED_OFFSET` bc we don't use `minHeight: 100%` anymore.
Also removed `chatContainerTop` bc it's subtracted from both sides.

rickyrombo (approved):

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
schottra: Just curious why we need to do the math? Could we not just append an item to the end of the flatlist that renders the "end reached" message and let Flatlist handle the scrolling and positioning?

I've seen us use this pattern in tables where we give things special IDs (ex: `id: '__endReachedIndicator'`) that we can check for in `renderItem` to conditionally render something other than a ChatListMessageItem

dharit-tan: Good thought, I think we decided not to append things to flatlist for the unread indicator due to other issues but those might not apply here. Will create ticket.

PR #3518, opened 2023-06-06
[C-2474] Generate playlist artwork from track artwork

dylanjeffers:
### Description

Adds feature where playlist artwork is generated from track artwork. Specifically, when a user adds a fourth track to their playlist, all four track artworks are combined into a single mosaic. Implemented image manipulation logic on web and mobile.

- Refactors addTrackToPlaylist saga to own file and convert to typescript.
- Uses web-worker for image generation in web
- Uses main-thread in mobile
### Dragons
Performance is decently slow on mobile, so the image doesnt change right away.

### DIFF
--- a/packages/web/src/common/store/cache/collections/addTrackToPlaylistSaga.ts
+++ b/packages/web/src/common/store/cache/collections/addTrackToPlaylistSaga.ts
@@ -1,0 +1,288 @@
+import {
+  Name,
+  Kind,
+  makeKindId,
+  makeUid,
+  cacheCollectionsSelectors,
+  cacheCollectionsActions,
+  PlaylistOperations,
+  cacheActions,
+  getContext,
+  audioRewardsPageActions,
+  Collection,
+  Nullable,
+  SquareSizes,
+  UserTrackMetadata,
+  AudiusBackend,
+  UserCollection,
+  ID
+} from '@audius/common'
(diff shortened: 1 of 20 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-c-2474-playlist-image-creation

packages/web/src/common/store/cache/collections/addTrackToPlaylistSaga.ts
```diff
@@ -0,0 +1,288 @@
...
+  const playlist: Nullable<PlaylistWithArtwork> = collections[playlistId]
+  if (!playlist) return
+
+  // Retrieve track-to-add so we confirm with the
+  // most up-to-date information.
+  const [track] = yield* call(retrieveTracks, { trackIds: [trackId] })
+  const trackOwnerEntities = yield* call(fetchUsers, [track.owner_id])
+  const trackOwner = trackOwnerEntities.entries[track.owner_id]
+
+  if (track && trackOwner && playlist.track_count === 3) {
```
Kyle-Shanks, line 103: do we want the track count be exactly 3 here?

Kyle-Shanks: ah i see, i guess we would only need to generate the image again if the >4 tracks were reordered

PR #3521, opened 2023-06-06
[PAY-1391] ensure all collections are fetched before attempting to download

schottra:
### Description
The original logic in `requestDownloadAllFavorites` assumed that all collections were fetched before running. This was true because we would load everything when `FavoritesScreen` mounted. However, now that we are fetching full collection details in pages, the selector used for the "download all" flow will only include items that have already been fetched as part of viewing the Albums or Playlists tabs.

To fix this, I added a helper saga that will fetch all account collections and we will wait on that before starting the download process.

Note: I chose to implement it as parallel fetches of albums/playlists so that we reuse the same sagas/actions that power the favorites screen. The result is that if you open the favorites screen and toggle download all, the albums and playlists tabs will subsequently already have all of their data loaded.

### Dragons
N/A

### How Has This Been Tested?
Tested locally on physical device

N/A

### Feature Flags ###
N/A

### DIFF
--- a/packages/common/src/store/saved-collections/selectors.ts
+++ b/packages/common/src/store/saved-collections/selectors.ts
@@ -10,6 +10,7 @@ import { CommonState } from '../commonStore'
 import { CollectionType } from './types'
 
 const getAccountCollections = (state: CommonState) => state.account.collections
+
 export const getSavedCollectionsState = (
   state: CommonState,
   type: CollectionType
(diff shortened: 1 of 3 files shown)


amendelsohn (approved):
This is great, thank you! Good find and a clean fix 👍

audius-infra:
Preview this change https://demo.audius.co/PAY-1391-fix-download-all

PR #3524, opened 2023-06-07
[PAY-1361] Fix reactions button hit area

piazzatron:
### Description

Fixes issue where the edges of the reactions button wouldn't be clickable. This was caused by my `:active` scaling animation scaling the button instead of the content inside the button, causing the `onClick` to fail to fire. The fix is to animate/scale content inside the button instead of the button itself. 

Would attach video but airplane wifi; play around on demo to see!

### Dragons

CSS diff looks scary but it works

### How Has This Been Tested?

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/piazz-fix-reaction-onclick

dharit-tan (approved):

PR #3526, opened 2023-06-07
[PAY-1365] Disable fullscreen gestures on mobile chat screen

dharit-tan:
### Description
After much investigation with @dylanjeffers and @schottra we were unable to come up with the perfect solution. Current thinking is that if we were able to get the stack navigator gesture handler to cede handling the pan gesture via `onResponderTerminateRequest` then this might work. Would have to patch rngh to do this.

For now the best we could come up with is disabling full screen swipe left gestures on the ChatScreen entirely. User will still be able to swipe left from the leftmost edge of the screen.

### Dragons

### How Has This Been Tested?

Local ios stage

### Feature Flags ###


rickyrombo (approved):

dylanjeffers (changes requested):
One liner, piece of cake right?

PR #3538, opened 2023-06-07
[PAY-1348] clean up ChatUserListScreen and use followers as default list

schottra:
### Description
* Reuses the logic from the web user list screen
* Made two local hooks for the two data sources (default user list and search user list)
* Cleans up `ChatUserListScreen` to be a little less complicated and to use the same logic, just switching the data source based on presence of query text

### Dragons
N/A

### How Has This Been Tested?
N/A

N/A

### Feature Flags ###
N/A

### DIFF
--- a/packages/mobile/src/screens/chat-screen/ChatUserListScreen.tsx
+++ b/packages/mobile/src/screens/chat-screen/ChatUserListScreen.tsx
@@ -1,15 +1,21 @@
-import { useState, useCallback, useEffect } from 'react'
+import { useCallback, useEffect, useState } from 'react'
 
-import type { User, ID } from '@audius/common'
+import type { User } from '@audius/common'
 import {
-  searchUsersModalSelectors,
-  searchUsersModalActions,
-  useProxySelector,
-  chatActions,
+  FOLLOWERS_USER_LIST_TAG,
+  Status,
+  accountSelectors,
   cacheUsersSelectors,
-  Status
+  chatActions,
+  followersUserListActions,
+  followersUserListSelectors,
+  searchUsersModalActions,
+  searchUsersModalSelectors,
+  statusIsNotFinalized,
+  useProxySelector,
+  userListActions
 } from '@audius/common'
-import { View, Image, Keyboard } from 'react-native'
+import { Image, Keyboard, View } from 'react-native'
 import { useDispatch, useSelector } from 'react-redux'
 import { useDebounce } from 'react-use'
 
@@ -17,10 +23,10 @@ import IconCompose from 'app/assets/images/iconCompose.svg'
 import IconSearch from 'app/assets/images/iconSearch.svg'
 import MagnifyingGlass from 'app/assets/images/leftPointingMagnifyingGlass.png'
 import {
-  Screen,
-  Text,
   FlatList,
+  Screen,
   ScreenContent,
+  Text,
   TextInput
 } from 'app/components/core'
 import LoadingSpinner from 'app/components/loading-spinner'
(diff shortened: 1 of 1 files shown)


packages/mobile/src/screens/chat-screen/ChatUserListScreen.tsx
```diff
@@ -134,40 +141,71 @@ const ListEmpty = () => {
...
+  }, [loadMore])
+
+  return { hasMore, loadMore, status, userIds }
+}
+
+export const ChatUserListScreen = () => {
   const styles = useStyles()
   const [query, setQuery] = useState('')
-  const [hasQuery, setHasQuery] = useState(false)
+  const [inputValue, setInputValue] = useState('')
```
dharit-tan, line 189: gah ty for cleaning

rickyrombo (approved):

packages/mobile/src/screens/chat-screen/ChatUserListScreen.tsx
```diff
@@ -182,34 +220,24 @@ export const ChatUserListScreen = (props: ChatUserListScreenProps) => {
...
-      setHasQuery(!!query)
-    },
-    debounceMs,
-    [query, setHasQuery, dispatch]
-  )
-
-  const handleChange = useCallback(
-    (text: string) => {
-      setQuery(text)
+      setQuery(inputValue)
```
dharit-tan, line 223: interesting.. so you're debouncing setting the state var instead of the dispatch call. i think i see why this is better but if you have a sec to explain would appreciate :)

dharit-tan (approved):

packages/mobile/src/screens/chat-screen/ChatUserListScreen.tsx
schottra: It's functionally not much different than what was here already(one state value is directly tied to the input and updates immediately on change, and that value is synchronized to the rest of the state on a debounce). But this pattern feels easier to follow because all of the code depending on `query` can respond when its value changes instead of some things tracking `hasQuery` and some things tracking the value of the query itself.

So I think it was mostly that I didn't like two different state variables when one would do fine?

packages/mobile/src/screens/chat-screen/ChatUserListScreen.tsx
```diff
@@ -134,40 +141,71 @@ const ListEmpty = () => {
...
-export const ChatUserListScreen = (props: ChatUserListScreenProps) => {
-  const {
-    debounceMs = DEBOUNCE_MS,
-    defaultUserList = {
-      userIds: [],
-      loading: false,
-      loadMore: () => {}
-    }
-  } = props
+function useQueryUserList(query: string) {
```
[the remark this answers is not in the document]
schottra: Good question. This is something I cemented as personal preference a while ago. And I had to look up the differences again to remind myself why.

Short answer: It probably doesn't matter much in this case.

Longer answer:
* Arrow functions are appropriate when you need to capture the surrounding closure and use values in the function later on (callbacks in various places being a common use case). They incur a _slight_ performance hit due to the fact that they aren't created until the statement is encountered, and they will have space allocated for the closure that they capture each time. Long term that might result in more Garbage Collection pressure, but really only when you have hundreds/thousands of them and you're sensitive to performance hiccups. Arrow functions also have weird `this` semantics, can't be used with `bind`, and don't support `arguments`. Those things don't really matter often in our codebase.
* 'Normal' functions are hoisted to the top of the module and allocated earlier at runtime (more chance for optimization by the engine, but again may not matter much to us). They don't capture closure in the same way, but _do_ support binding and `arguments`. They are also, in my personal opinion, easier decorate with type information (TS didn't really even support full typing for arrow functions for a long time).

So, based on things that don't matter much but could add up in large codebases I generally:
* Use `function` when I'm at module level and don't need to capture any closure other than constants also defined at the module level. And also when I'm defining methods for classes.
* Use arrow functions when I'm inside of another function and need what I would consider a "temporary" function that is being passed as a callback and/or will use values through closure of the surrounding code.

PR #3543, opened 2023-06-08
[PAY-1408][PAY-1351][PAY-1383][PAY-1417][PAY-1240] DMs: Web UI polishes and bugfixes

rickyrombo:
### Description

See commit messages, review commit by commit

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/common/src/store/pages/chat/middleware.ts
+++ b/packages/common/src/store/pages/chat/middleware.ts
@@ -5,6 +5,7 @@ import { Status } from 'models/Status'
 import { getUserId } from 'store/account/selectors'
 import { encodeHashId } from 'utils/hashIds'
 
+import { getChatMessageById } from './selectors'
 import { actions as chatActions } from './slice'
 
 const { connect, disconnect, addMessage, setMessageReactionSucceeded } =
@@ -42,13 +43,20 @@ export const chatMiddleware =
             )
           }
           reactionListener = ({ chatId, messageId, reaction }) => {
-            store.dispatch(
-              setMessageReactionSucceeded({
-                chatId,
-                messageId,
-                reaction
-              })
+            const message = getChatMessageById(
+              store.getState(),
+              chatId,
+              messageId
             )
+            if (message) {
+              store.dispatch(
+                setMessageReactionSucceeded({
+                  chatId,
+                  messageId,
+                  reaction
+                })
+              )
+            }
           }
           closeListener = async () => {
             console.debug('[chats] WebSocket closed. Reconnecting...')
--- a/packages/web/src/components/data-entry/InputV2.tsx
+++ b/packages/web/src/components/data-entry/InputV2.tsx
@@ -1,4 +1,10 @@
-import { ComponentPropsWithoutRef, MutableRefObject, RefCallback } from 'react'
+import {
+  ComponentPropsWithoutRef,
+  MutableRefObject,
+  RefCallback,
+  useState,
+  useCallback
+} from 'react'
 
 import cn from 'classnames'
 
@@ -42,6 +48,7 @@ export const InputV2 = (props: InputV2Props) => {
     warning: warningProp,
     error,
     inputClassName,
+    disabled,
     ...other
   } = props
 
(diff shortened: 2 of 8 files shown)


packages/web/src/components/data-entry/TextAreaV2.tsx
rickyrombo, line 1: This file's diff is misleading. The changes:
- Added a `forwardRef` wrapper
- Renamed the existing ref to `rootRef`
- Merged the `forwardedRef` and the `textareaRef` on the `<textarea>`

audius-infra:
Preview this change https://demo.audius.co/mjp-dms-ui-polish

packages/common/src/store/pages/chat/middleware.ts
```diff
@@ -42,13 +43,20 @@ export const chatMiddleware =
             )
           }
           reactionListener = ({ chatId, messageId, reaction }) => {
-            store.dispatch(
-              setMessageReactionSucceeded({
-                chatId,
-                messageId,
-                reaction
-              })
+            const message = getChatMessageById(
```
schottra, line 46: I don't know why, but it feels strange to put this logic in the middleware. Could it not live in the slice?
Seems like it's not for _writing_ a reaction, but for adding a reaction that we received over the socket to a chat that we have locally?

packages/web/src/components/data-entry/InputV2.tsx
```diff
@@ -59,20 +66,44 @@ export const InputV2 = (props: InputV2Props) => {
     [styles.error]: error
   }
 
+  /**
+   * Since Firefox doesn't support the :has() pseudo selector,
+   * manually track the focused state and use classes for required and disabled
+   */
```
schottra, line 72: Oof.
* Change looks safe at surface level, but wondering if there are any other places using this `InputV2` component that should be tested?
* We should probably chat with client and follow up on any other code using `:has` selectors

packages/web/src/pages/chat-page/components/ChatComposer.tsx
```diff
@@ -81,10 +84,17 @@ export const ChatComposer = (props: ChatComposerProps) => {
     [handleSubmit]
   )
 
+  useEffect(() => {
+    if (chatId) {
+      ref.current?.focus()
+    }
+  }, [ref, chatId])
```
schottra, line 91: Not a big deal but just curious:

With effect deps, I tend to list them in order of likelihood to change, most likely first. The idea is that in evaluating the array each time we render, it has the probability of doing the least amount of work (assuming it short-circuits on finding the first value that has changed).

Not that there's any real performance basis for that preference, but I was curious why you do it the opposite way... 😁

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -510,6 +510,14 @@ export const ChatScreen = () => {
     flatListInnerHeight.current = event.nativeEvent.layout.height
   }, [])
 
+  // Scroll to the bottom if the user sends a message
+  const handleMessageSent = useCallback(() => {
+    // Set a timeout to ensure the full render happened and we got the final height
+    setTimeout(() => {
+      flatListRef.current?.scrollToOffset({ offset: 0, animated: true })
+    }, 0)
```
schottra, line 518: nit: According to the [docs](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#:~:text=a%20security%20risk.-,delay,-Optional), omitting delay causes it to default to `0` and has 'execute immediately' semantics.

schottra: Just confirming you tested it both within and outside of the range we pass to `maintainVisibleContent`. That changes the default behavior and I wonder if there is any strange behavior as a result.

schottra (approved):
Looks good! Just some thoughts and nits.

packages/common/src/store/pages/chat/middleware.ts
dharit-tan: feel similarly, could we not just return early if `existingMessage` doesn't exist here?
https://github.com/AudiusProject/audius-client/blob/e25bb63df325e1b21249c249ec9ce7ffd2d21baa/packages/common/src/store/pages/chat/slice.ts#L281

packages/web/src/pages/chat-page/components/ChatComposer.tsx
dharit-tan: whoa first i'm hearing about this

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
rickyrombo: Execute next event cycle is what I want I think - it just needs to happen after a single render cycle. Didn't test too heavily with the `maintainVisibleContent` stuff, but I think since this is guaranteed to execute last it should have no oddities

packages/common/src/store/pages/chat/middleware.ts
rickyrombo: Yea y'all are right, will change!

packages/web/src/pages/chat-page/components/ChatComposer.tsx
schottra: These are things I spend too much time thinking about 😆

audius-infra:
Preview this change https://demo.audius.co/mjp-dms-ui-polish

PR #3546, opened 2023-06-08
[C-2747] Fix track title layout with offline + ai attribution

amendelsohn:
### Description

Some AI attribution changes broke the flex layout for track title and offline indicator

Before:

After:
Screenshot 2023-06-08 at 1 13 44 PM
Screenshot 2023-06-08 at 1 13 28 PM

### DIFF
--- a/packages/mobile/src/screens/track-screen/TrackScreenDetailsTile.tsx
+++ b/packages/mobile/src/screens/track-screen/TrackScreenDetailsTile.tsx
@@ -139,6 +139,12 @@ const useStyles = makeStyles(({ palette, spacing, typography }) => ({
     justifyContent: 'center',
     paddingHorizontal: spacing(4)
   },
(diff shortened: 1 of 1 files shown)


packages/mobile/src/screens/track-screen/TrackScreenDetailsTile.tsx
```diff
@@ -139,6 +139,12 @@ const useStyles = makeStyles(({ palette, spacing, typography }) => ({
     justifyContent: 'center',
     paddingHorizontal: spacing(4)
   },
+  headerRow: {
+    display: 'flex',
```
dylanjeffers, line 143: Technically not needed cause the default is flex

dylanjeffers (approved):
Wow so basically been broken for a while due to ai attr? Thanks for jumping on it regardless

PR #3553, opened 2023-06-09
[PAY-1423] Fix empty chat message box with currently playing track

dharit-tan:
### Description
Change the empty chat message box to be an actual `ListEmptyComponent` of the flatlist which necessitated mirror-flipping it, and also add extra padding when currently playing track drawer is visible.

Keep in mind everything is flipped because the parent flatlist is inverted.

### Dragons

Feel like there's a clever css way to do this but couldn't figure it out.

### How Has This Been Tested?

Local ios stage
Simulator Screen Shot - iPhone 14 Pro - 2023-06-09 at 00 31 26

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/screens/chat-screen/EmptyChatMessages.tsx
+++ b/packages/mobile/src/screens/chat-screen/EmptyChatMessages.tsx
@@ -1,11 +1,7 @@
-import { playerSelectors } from '@audius/common'
 import { View, Text, Image } from 'react-native'
-import { useSelector } from 'react-redux'
 
 import WavingHand from 'app/assets/images/emojis/waving-hand-sign.png'
 import { makeStyles } from 'app/styles'
-import { spacing } from 'app/styles/spacing'
-const { getHasTrack } = playerSelectors
 
 const messages = {
   newMessage: 'New Message',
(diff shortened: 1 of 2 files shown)


rickyrombo:
Added a commit @dharit-tan :
- The inversion of the empty component is a known bug in react native, though it looks like it might be fixed in upcoming version? https://github.com/facebook/react-native/issues/21196 To defensively prevent against our workaround accidentally flipping the empty chat state _back_ in the future when this bug is fixed, I found a different workaround - wrapping our component in a view _inside the ListEmptyComponent prop_ doesn't have the inversion issue, at least on simulator
- I moved the play bar height logic to the ChatScreen, making the view that the list sits in the size of the actual visible contents. When setting the `bottom` to the offset, it shifted the whole thing up, so I used ~~`marginTop`~~ `paddingTop` to push its children back down appropriately

packages/mobile/src/screens/chat-screen/EmptyChatMessages.tsx
```diff
@@ -10,9 +10,8 @@ const messages = {
 }
 
 const useStyles = makeStyles(({ spacing, palette, typography }) => ({
-  emptyContainer: {
+  root: {
     marginTop: spacing(8),
-    marginHorizontal: spacing(6),
```
rickyrombo, line 15: removed this because the container already has `marginHorizontal: spacing(6)`

dharit-tan:
Can't approve my own PR but this looks great, thx @rickyrombo

rickyrombo (approved):

PR #3568, opened 2023-06-12
[C-2716] Improve notifications on fresh account

dylanjeffers:
### Description

Fixes issue where opening the notification panel doesn't trigger `fetchNotifications` call, which can lead to cases where the loading bar is spinning for up to 10 seconds waiting for the `checkForNewNotifications` call to occur.

Further cleans up the notification panel's styles and ui layout, improving hierarchy, a11y, and closer to notification spec.


packages/web/src/components/notification/NotificationPanel.tsx
```diff
@@ -124,6 +118,10 @@ export const NotificationPanel = ({ anchorRef }: NotificationPanelProps) => {
     }
   }, [panelIsOpen, unviewedNotificationCount, dispatch])
 
+  const userHasNoNotifications =
```
sliptype, line 121: Can we invert this? Seems like it would be clearer as `userHasNotifications`

sliptype (approved):
🙌

good stuff!

packages/web/src/components/notification/NotificationPanel.tsx
dylanjeffers: totes

PR #3574, opened 2023-06-12
[C-2713] Add track metadata fields to new upload form

amendelsohn:
### Description

Fields:
Screenshot 2023-06-09 at 3 29 49 PM

Submitted values:
Screenshot 2023-06-09 at 3 30 04 PM

### Feature Flags ###

`UPLOAD_REDESIGN_ENABLED`


audius-infra:
Preview this change https://demo.audius.co/amendel-upload-track-metadata-2

packages/web/src/pages/upload-page/fields/TrackMetadataFields.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const [genreField, genreMeta, genreHelpers] = useField('genre')
+  const [, moodMeta, moodHelpers] = useField('mood')
+  const [, tagsMeta, tagsHelpers] = useField('tags')
+  const [, descriptionMeta, descriptionHelpers] = useField('description')
+
+  const onDropArtwork = async (selectedFiles: File[], source: string) => {
+    try {
+      let file = selectedFiles[0]
+      file = await resizeImage(file)
+      const storageV2SignupEnabled = getFeatureEnabled(
```
nicoback2, line 47: should we use `useFlag` instead? seems to be more used in components.

i think the diff is that the hook can get updates for a given feature flag https://github.com/AudiusProject/audius-client/blob/main/packages/web/src/components/menu/TrackMenu.tsx#L98-L101

packages/web/src/pages/upload-page/fields/TrackMetadataFields.tsx
```diff
@@ -0,0 +1,135 @@
...
+        FeatureFlags.STORAGE_V2_SIGNUP
+      )
+      const storageV2UploadEnabled = getFeatureEnabled(
+        FeatureFlags.STORAGE_V2_TRACK_UPLOAD
+      )
+      if (storageV2SignupEnabled || storageV2UploadEnabled) {
+        // @ts-ignore
+        file.name = selectedFiles[0].name
```
nicoback2, line 55: does casting to string not work? ts ignore is a bit scary

packages/web/src/pages/upload-page/fields/TrackMetadataFields.tsx
```diff
@@ -0,0 +1,135 @@
...
+      }
+      const url = URL.createObjectURL(file)
+      artworkHelpers.setValue({ url, file, source })
+      setImageProcessingError(false)
+    } catch (err) {
+      setImageProcessingError(true)
```
nicoback2, line 61: prob want to log the error here (unless you were planning to add all the logging at the end of the project which is what i do lol)

nicoback2 (approved):

packages/web/src/pages/upload-page/fields/TrackMetadataFields.tsx
amendelsohn: Unfortunately, no. The type error is that `file.name` is a readonly field. This was copied from an untyped js file https://github.com/AudiusProject/audius-client/blob/main/packages/web/src/components/data-entry/FormTile.js#L842
I didn't want to break the logic, but I'll follow up with Theo to see if it's actually working as intended.

audius-infra:
Preview this change https://demo.audius.co/amendel-upload-track-metadata-2

PR #3579, opened 2023-06-13
[C-2760, C-2779] Fix header for playlist and artists tabs in explore screen

Kyle-Shanks:
### Description
Updated the component structure slightly to fix the headers
Screenshot 2023-06-12 at 11 20 15 PM

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/mobile/src/components/core/CardList.tsx
+++ b/packages/mobile/src/components/core/CardList.tsx
@@ -35,8 +35,10 @@ const DefaultLoadingCard = () => null
 
 const useStyles = makeStyles(({ spacing }) => ({
   cardList: {
-    padding: spacing(3),
     paddingRight: 0
+  },
+  columnWrapper: {
+    paddingLeft: spacing(3)
   },
   card: {
     width: '50%',
@@ -90,6 +92,7 @@ export function CardList<ItemT extends {}>(props: CardListProps<ItemT>) {
   return (
     <FlatListComponent
       style={styles.cardList}
+      columnWrapperStyle={styles.columnWrapper}
       ref={ref}
       data={data}
       renderItem={handleRenderItem}
--- a/packages/mobile/src/screens/search-results-screen/tabs/PlaylistsTab.tsx
+++ b/packages/mobile/src/screens/search-results-screen/tabs/PlaylistsTab.tsx
@@ -7,6 +7,7 @@ import {
 } from '@audius/common'
 
 import { CollectionList } from 'app/components/collection-list/CollectionList'
+import { spacing } from 'app/styles/spacing'
 
 import { EmptyResults } from '../EmptyResults'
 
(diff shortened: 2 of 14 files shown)


dylanjeffers (changes requested):
Hmm i think we should avoid nesting flat lists in scrollviews, does ListHeaderComponent not do what we expect?

Kyle-Shanks:
> Hmm i think we should avoid nesting flat lists in scrollviews, does ListHeaderComponent not do what we expect?

It does, but getting the styles to look nice across all of our card lists is a bit of a pain. I'll fix it though 👍

packages/mobile/src/components/core/CardList.tsx
```diff
@@ -35,9 +35,11 @@ const DefaultLoadingCard = () => null
 
 const useStyles = makeStyles(({ spacing }) => ({
   cardList: {
-    padding: spacing(3),
     paddingRight: 0
   },
+  contentWrapper: {
```
dylanjeffers, line 40: Can this be columnWrapper?

packages/mobile/src/screens/explore-screen/tabs/PlaylistsTab.tsx
```diff
@@ -32,12 +33,14 @@ export const PlaylistsTab = () => {
...
   return (
-    <CollectionList
-      isLoading={
-        exploreStatus === Status.LOADING || playlistsStatus !== Status.SUCCESS
-      }
-      ListHeaderComponent={<TabInfo header={messages.infoHeader} />}
-      collection={playlists}
-    />
+    <ScrollView>
+      <TabInfo header={messages.infoHeader} />
```
dylanjeffers, line 37: I feel pretty strongly this should be in the flat list header, what's the main reason this doesn't work?

packages/mobile/src/screens/search-results-screen/tabs/PlaylistsTab.tsx
```diff
@@ -34,6 +35,7 @@ export const PlaylistsTab = () => {
 
   return (
     <CollectionList
+      style={{ paddingTop: spacing(3) }}
```
dylanjeffers, line 38: I also thing it'd a bit weird we have to apply custom styles to a lot of collection lists now too. Was there something fundamentally wrong with the original styles for the list?

dylanjeffers (commented):
Sorry to be nit picky on this one, just have some more qs on this one!

packages/mobile/src/screens/explore-screen/tabs/PlaylistsTab.tsx
Kyle-Shanks: oh thats weird. I updated it, but it reverted somehow. I'll fix 👍

packages/mobile/src/screens/search-results-screen/tabs/PlaylistsTab.tsx
dylanjeffers: Word okay makes sense thanks for clarifying the long term game plan

PR #3582, opened 2023-06-13
[CON-742] Make users and track actions v2 for v2 users/tracks even when feature flags are disabled

theoilie:
### Description

Makes it safe to flip v2 tracks back to disabled after flipping to enabled. Otherwise updating v2 tracks would fail because it would try the old flow (which is what I saw for my prod v2 track on mobile where I didn't have the feature flag enabled - the app crashed and sentry confirmed it was trying to use the v1 edit-track flow).

This is also important for users who signup as v2 but don't get the feature flag for v2 uploads (due to feature flag flakiness we've run into before).

### How Has This Been Tested?
- Uploaded a v2 track to staging and confirmed that editing it works even with setting the feature flag to false (previously broken - caused heavy load screen)
- Tested the inverse (upload a v1 track with the track upload feature flag set to true and then to false)

### Feature Flags ###
Same upload flag: `storage_v2_track_upload`

### DIFF
--- a/packages/common/src/services/audius-backend/AudiusBackend.ts
+++ b/packages/common/src/services/audius-backend/AudiusBackend.ts
@@ -1047,13 +1047,11 @@ export const audiusBackend = ({
     metadata: TrackMetadata,
     onProgress: (loaded: number, total: number) => void
   ) {
-    const storageV2SignupEnabled = await getFeatureEnabled(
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://demo.audius.co/con742-check-v2-without-feature-flag

PR #3592, opened 2023-06-14
Fix mobile textinput borders + labels

dharit-tan:
### Description
Bug was that `map` passes index as 2nd arg, so the first color in the list was getting opacity 0. We didn't hit this before because `opacity` in `convertHexToRGBA` was actually not being used, we always set it to `1` regardless of the `opacity` arg.

### Dragons

### How Has This Been Tested?

Local ios stage:
Simulator Screen Shot - iPhone 14 Pro - 2023-06-14 at 17 13 44

Confirmed send button opacity still works on android:
Screenshot_1686777153
a41b2df0-1dce-4028-9def-5aa73af550fc)

### Feature Flags ###


Kyle-Shanks (approved):
What a wild bug haha, thanks for the fix

PR #3595, opened 2023-06-14
[PAY-1437] Fix drawers sharing state causing crash

dharit-tan:
### Description
Bug: if two different native drawers were opened/closed rapidly, app would crash. This is because the first drawer would set the shared `data`, then when it was closed it would be marked as 'closing' but wouldn't actually unset `data` until its close animation completed. If the 2nd drawer was opened before the animation completed, it would open and read `data`, but then `data` would get set to `null` by the 1st drawer.
Solution: make each drawer have its own `data` in a map.

### Dragons

Touches a bunch of drawers. I've tested all and have asked AJ to test WalletConnect for next release.

### How Has This Been Tested?

Local ios stage

### Feature Flags ###


packages/mobile/src/store/drawers/selectors.ts
```diff
@@ -2,8 +2,10 @@ import type { AppState } from '../'
 
 import type { Drawer, DrawerData } from './slice'
 
-export const getData = <TDrawer extends keyof DrawerData>(state: AppState) =>
-  state.drawers.data as DrawerData[TDrawer]
+export const getData = <TDrawer extends keyof DrawerData>(
```
rickyrombo, line 5: nit: I think typically selectors do `state` as the first argument always?

Although I see below we wrap the selector in a new fn so 🤷 idk what's the best strategy here

rickyrombo (approved):
Nice find and fix!

packages/mobile/src/store/drawers/selectors.ts
dharit-tan: yeah was also a bit confused by the next selector below.

dylanjeffers (approved):
wonderful improvement reed

PR #3596, opened 2023-06-14
[C-2770] Fix issue where playlist library can disappear

dylanjeffers:
### Description

Fixes issue where user's playlist library can disappear from left-nav. This has to do with the api not always returning a user's playlist_library, which actually makes sense for all users other that the active user. The fix for now is to just always keep a user's playlist_library on their user object if we ever get it back.

### DIFF
--- a/packages/common/src/store/cache/reducer.ts
+++ b/packages/common/src/store/cache/reducer.ts
@@ -85,8 +85,7 @@ const unwrapEntry = (entry: { metadata: any }) => {
 const forceUpdateKeys = new Set([
   'field_visibility',
   'followee_reposts',
-  'followee_saves',
-  'playlist_library'
+  'followee_saves'
 ])
 
 // Customize lodash recursive merge to never merge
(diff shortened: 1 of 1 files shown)


packages/common/src/store/cache/reducer.ts
```diff
@@ -100,6 +99,12 @@ export const mergeCustomizer = (objValue: any, srcValue: any, key: string) => {
     return srcValue || objValue
   }
 
+  // Not every user request provides playlist_library,
+  // so always prefer it's existence, starting with latest
+  if (key === 'playlist_library') {
+    return objValue ?? srcValue
```
raymondjacobson, line 105: nit we do || above. maybe we should be consistent?

raymondjacobson (approved):
amazing!!!

packages/common/src/store/cache/reducer.ts
dylanjeffers: yeee sure thing

PR #3598, opened 2023-06-14
[PAY-1465] Auto-nav to chat only when permitted

dharit-tan:
### Description
Previously we added a feature where if the user unblocks someone from the create chat list screen, they would automatically be taken to the chat. This is a problem if the user is actually not permitted for a different reason.
Say if A is tip-gated, then B blocks A. If B unblocks A, the drawer will immediately attempt to create a chat and nav to it. Instead it should just close the drawer and stay on the same screen, at which point A will be marked as tip-gated instead.
Solution is to re-check permissions before creating the chat.

### Dragons

### How Has This Been Tested?

Local ios stage

https://github.com/AudiusProject/audius-client/assets/3893871/e86eb0f1-f3c1-4be1-b599-7340b69087d1

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/block-messages-drawer/BlockMessagesDrawer.tsx
+++ b/packages/mobile/src/components/block-messages-drawer/BlockMessagesDrawer.tsx
@@ -16,7 +16,7 @@ import { spacing } from 'app/styles/spacing'
 import { useColor } from 'app/utils/theme'
 
 const { getUser } = cacheUsersSelectors
-const { getDoesBlockUser } = chatSelectors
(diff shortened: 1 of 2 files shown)


rickyrombo (approved):
I could actually see an argument for doing this in the `createChat` saga perhaps. In any case, I'll keep this in mind when doing this for the tip modal

dharit-tan:
> I could actually see an argument for doing this in the `createChat` saga perhaps. In any case, I'll keep this in mind when doing this for the tip modal

Yeah makes sense

PR #3621, opened 2023-06-16
Add remove from playlist overflow action

Kyle-Shanks:
### Description
Update overflow menu state and add new option for tracks to remove from the contextual playlist

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/common/src/store/ui/mobile-overflow-menu/types.ts
+++ b/packages/common/src/store/ui/mobile-overflow-menu/types.ts
@@ -8,6 +8,7 @@ export enum OverflowAction {
   UNFAVORITE = 'UNFAVORITE',
   SHARE = 'SHARE',
   ADD_TO_PLAYLIST = 'ADD_TO_PLAYLIST',
+  REMOVE_FROM_PLAYLIST = 'REMOVE_FROM_PLAYLIST',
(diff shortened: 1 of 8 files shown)


packages/common/src/store/ui/mobile-overflow-menu/types.ts
```diff
@@ -36,6 +37,7 @@ export enum OverflowSource {
 export type OpenOverflowMenuPayload = {
   source: OverflowSource
   id: ID | string
+  contextPlaylistId?: ID
```
dylanjeffers, line 40: would it make sense to just have a "context" field that any overflow action can add to? feels a little awkward that we are adding "playlist" word into a somewhat generic function?

dylanjeffers (approved):
great stuff

packages/common/src/store/ui/mobile-overflow-menu/types.ts
Kyle-Shanks: True yea I was thinking about too, but ultimately feel like this is a bit safer so that we can check for a specific key instead of having to check everywhere this is used what the value is and if its applicable

PR #3629, opened 2023-06-21
[PAY-1319] Fix emoji keyboard overlapping chat text input

dharit-tan:
### Description
The bug: when switching to the emoji keyboard, the keyboard would slightly overlap the textinput ([see linear issue](https://linear.app/audius/issue/PAY-1319/emoji-bar-overlaps-input)). 
Root cause: I discovered that the content inside the `KeyboardAvoidingView` was offset by 84 pixels, which matches the height of the `BottomTabBar`. For some reason it thought that it had to avoid that too. I'd gotten lucky by using `keyboardHeightOffset` and landed on a value that perfectly matched up with 84 pixels, but this was a house of cards. The real solution is to measure the height of `BottomTabBar` and subtract that from the amount to translate the content of `KeyboardAvoidingView` by.

EDIT: gonna stick with hardcoding. Also needed to handle ios vs android bottom padding separately for some reason.

This solution feels like a bit much for such a simple thing, open to suggestions. I did start by just hardcoding the height of `BottomTabBar` (see first commit) but I figured there might be cases where there are extra insets at the bottom of the screen or something, so actually measuring it felt more robust.

I tried a variety of approaches:
- Tried to animate away `BottomTabBar` the same way we do when opening the `NowPlayingDrawer`, but `KeyboardAvoidingView` didn't know about the animations and still thought they were visible.
- Set `AppTabBar` to return `null` when the keyboard was visible, but this caused the rive animations in `BottomTabBar` to run again when they re-entered the screen.
- Tried to figure out how to prevent the animations from running via Rive API but couldn't find a way.

### Dragons

### How Has This Been Tested?

https://github.com/AudiusProject/audius-client/assets/3893871/b7001f4a-01fa-4895-9b90-4168cfd3975e
Screenshot_1687397964
Screenshot_1687397970
Screenshot_1687397986
Screenshot_1687397992

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/rt-bottom

rickyrombo (approved):
Is the bottom tab bar height dynamic? I think if we explicitly set its height to the same constant you use in the chats screen (like in your first commit) we should be good - if anyone wanted to change the height they'd see the constant's usage and would have to adjust it there anyway.

I think what you landed on is probably good too but I'm not sure if it needs to be this dynamic

dharit-tan:
> Is the bottom tab bar height dynamic? I think if we explicitly set its height to the same constant you use in the chats screen (like in your first commit) we should be good - if anyone wanted to change the height they'd see the constant's usage and would have to adjust it there anyway.
> 
> I think what you landed on is probably good too but I'm not sure if it needs to be this dynamic

Yep fair enough, it's not dynamic just thought this would be safer. I was worried bottom insets would mess things up on some devices, but looked again and we're probably good.

packages/mobile/src/screens/app-screen/AppTabBar.tsx
```diff
@@ -1,29 +1,60 @@
...
   const { navigation, state } = props
+  const styles = useStyles()
+  const dispatch = useDispatch()
   // Set handlers for the NowPlayingDrawer and BottomTabBar
   // When the drawer is open, the bottom bar should hide (animated away).
   // When the drawer is closed, the bottom bar should reappear (animated in).
   const translationAnim = useRef(new Animated.Value(FULL_DRAWER_HEIGHT)).current
 
+  const viewRef = useRef<View | null>(null)
+  const handleOnLayout = useCallback(() => {
```
dylanjeffers, line 36: handleLayout

packages/mobile/src/screens/app-screen/AppTabBar.tsx
```diff
@@ -1,29 +1,60 @@
...
+
   return (
     <>
       <NowPlayingDrawer translationAnim={translationAnim} />
-      <BottomTabBar
-        translationAnim={translationAnim}
-        navigation={navigation}
-        state={state}
-      />
+      <View
```
dylanjeffers, line 47: Should this all be defined in the bottom bar itself?

dylanjeffers (approved):
Great work!

piazzatron (approved):
Great find!!

Re: keeping it dynamic or not - if we can show that just setting it as a fixed height works on multiple devices, IMO we should rip out the redux stuff and just use a constant. Every little bit of complexity one can avoid adding is a win :D 

(As an aside, for this sort of simple pattern of setting a global var and being able to read from it elsewhere, React context is very nice - you can avoid the whole reducer/action/dispatch song and dance) and just set things directly.

dharit-tan:
Going to merge the hardcoded version, have the measuring version backed up on `rt-bottom-bar-measure` if we ever want to go back to it.

rickyrombo (approved):
Looks good. I'd add the constant as the height of the bottom bar to make it explicitly the same as the constant - that way anyone looking to change the bottom bar height knows to modify the constant

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -516,6 +520,19 @@ export const ChatScreen = () => {
...
+  // For some reason the bottom padding behavior is different between the platforms.
+  const getPlaybarAvoidingBottomPadding = useCallback(() => {
+    if (Platform.OS === 'ios') {
+      return hasCurrentlyPlayingTrack ? PLAY_BAR_HEIGHT : 0
+    } else if (Platform.OS === 'android') {
+      return hasCurrentlyPlayingTrack && !keyboardShown ? PLAY_BAR_HEIGHT : 0
+    }
+  }, [hasCurrentlyPlayingTrack, keyboardShown])
+
+  console.log(
```
rickyrombo, line 532: log

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
```diff
@@ -563,13 +580,15 @@ export const ChatScreen = () => {
...
+                ? PLAY_BAR_HEIGHT + BOTTOM_BAR_HEIGHT
+                : BOTTOM_BAR_HEIGHT
             }
             style={[
               styles.keyboardAvoiding,
-              hasCurrentlyPlayingTrack
-                ? { bottom: PLAY_BAR_HEIGHT, paddingTop: PLAY_BAR_HEIGHT }
-                : null
+              {
+                bottom: getPlaybarAvoidingBottomPadding()
```
rickyrombo, line 590: we lost the "paddingTop" style, is it ok without it?

rickyrombo (approved):

packages/mobile/src/screens/chat-screen/ChatScreen.tsx
dharit-tan: yep! idk what it was doing

rickyrombo: It was to fix an alignment issue - setting the bottom of the container doesn't affect the height but it does move it up, so things at the top inside the container would appear underneath the header or nudged up the height of the playbar. Try checking Android on an empty chat

dharit-tan: damn thanks for the catch!

PR #3631, opened 2023-06-21
Refactor: extract reusable modal field component

amendelsohn:
### Description

Factor out the modal field so it can be reused for other fields in the track upload form.

### Feature Flags ###

`UPLOAD_REDESIGN_ENABLED`


audius-infra:
Preview this change https://demo.audius.co/amendel-upload-modal-field

dylanjeffers (approved):
excellent work here

PR #3632, opened 2023-06-21
[PAY-1412] Reset player state on app load

dharit-tan:
### Description
Hard to repro this bug but I think this should fix it. The theory is that after a crash, the player state from before the crash was persisted, which caused `getHasTrack` player selector to erroneously report `true`.

Would like to get this in for the QA and see if we get any repros of the bug.

### Dragons

Adding stuff to root screen!

### How Has This Been Tested?

Hard to prove this makes the bug go away, but I did confirm this action only runs once on startup and that normal app behavior is unaffected.

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/screens/root-screen/RootScreen.tsx
+++ b/packages/mobile/src/screens/root-screen/RootScreen.tsx
@@ -1,6 +1,11 @@
 import { useCallback, useEffect, useState } from 'react'
 
-import { accountSelectors, chatActions, Status } from '@audius/common'
+import {
+  accountSelectors,
+  chatActions,
+  playerActions,
+  Status
+} from '@audius/common'
 import type { NavigatorScreenParams } from '@react-navigation/native'
 import { createNativeStackNavigator } from '@react-navigation/native-stack'
 import { getHasCompletedAccount } from 'common/store/pages/signon/selectors'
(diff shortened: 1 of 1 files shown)


packages/mobile/src/screens/root-screen/RootScreen.tsx
```diff
@@ -77,6 +83,14 @@ export const RootScreen = ({
...
   }, [dispatch, isLoaded, accountStatus])
 
+  // Reset the player when the app is loaded for the first time. Fixes an issue
+  // where after a crash, the player would persist the previous state. PAY-1412.
+  useEffect(() => {
+    if (isLoaded) {
+      dispatch(playerActions.reset({ shouldAutoplay: false }))
+    }
+  }, [isLoaded, dispatch])
+
```
piazzatron, line 93: Could we just put this dispatch on L68 instead of having a new useEffect?

piazzatron (approved):
Nice find

packages/mobile/src/screens/root-screen/RootScreen.tsx
dharit-tan: I think we want to keep them separate bc we don't want to clear this state if a user was not signed in, then signs in? L68 runs when the user has successfully signed in only. We want this to run when the app loads for the first time regardless of sign-in state.

dharit-tan: I guess could add `if (isLoaded)`-only check to the above useEffect. I could go either way lmk.

dharit-tan: chatted with @piazzatron think he's right!

PR #3634, opened 2023-06-22
[C-2763] Fix desktop playlist qa issues

dylanjeffers:
### Description

- [x]  When page width is small, the icons for the action buttons are off center
- [x]  Newly created folders should be added to the top of the playlist library
- [x]  playlist artwork lingers when going to a playlist without an image
    - Displays both artwork from other playlist and the artwork placeholder icon
- [x]  Add tooltip for the share action button when disabled bc of empty playlist
- [x]  Text color for the playlist name doesn’t animate, but the icon does
    - text color should animate
- [x]  bounding box for the playlist name edit should not span across the entire card
- [x]  make public modal stays open after clicking confirm
- [x]  make public action button text is the wrong size (too big)
- [x]  edit page cancel button font size is the wrong size (too small)
- [x]  add to playlist empty artwork
- [x]  overflow menus in sidebar can overlap
    - they don’t close until you click away or open some other type of overflow menu
- [x]  action buttons can get clipped on certain screen sizes
    - The share and make public buttons should shrink at the same break point as the edit button
- [x]  Placeholder artwork background color on the playlist screen should match the grey from the edit modal (Web)
- [x]  Make sure that the choose artwork popup aligns properly
    - Pops up over the image and in the middle of the modal


audius-infra:
Preview this change https://demo.audius.co/dj-c-2763-desktop-qa-fixes

packages/stems/src/assets/styles/animations.css
```diff
@@ -8,5 +8,6 @@
 :root,
 root {
   --quick: 0.07s ease-in-out;
-  --calm: 0.2s ease-in-out;
+  --expressive: 0.2s ease-in-out;
```
Kyle-Shanks, line 11: does anything else use the calm var? change makes sense, but want to make sure that nothing else is affected

packages/stems/src/components/Button/Button.module.css
```diff
@@ -59,15 +59,15 @@
 }
 
 .button .icon.left.noText {
-  margin: 0;
+  margin: 0 !important;
```
Kyle-Shanks, line 62: Can't wait for the design updates to avoid things like this in the future

packages/web/src/components/collection/desktop/Artwork.tsx
```diff
@@ -54,6 +58,13 @@ export const Artwork = (props: ArtworkProps) => {
     )
   }, [dispatch, collectionId])
 
+  console.log({
+    gradient,
+    imageOverride,
+    image,
+    placeholderArt
+  })
+
```
Kyle-Shanks, line 67: log

packages/web/src/components/collection/desktop/Artwork.tsx
```diff
@@ -1,6 +1,10 @@
 import { ComponentType, SVGProps, useCallback, useEffect } from 'react'
 
-import { CoverArtSizes, SquareSizes } from '@audius/common'
+import {
+  imageBlank as placeholderArt,
+  CoverArtSizes,
+  SquareSizes
+} from '@audius/common'
```
Kyle-Shanks, line 7: were these used for anything?

packages/web/src/components/collection/desktop/CollectionHeader.module.css
```diff
@@ -1,3 +1,12 @@
+.hide {
+  opacity: 0;
+}
+
+.show {
+  opacity: 1;
+  transition: opacity ease-in-out 0.5s;
```
Kyle-Shanks, line 7: can use the animation var here right?

dylanjeffers: yes def!

packages/web/src/components/collection/desktop/Artwork.tsx
dylanjeffers: no great call

packages/stems/src/components/Button/Button.module.css
dylanjeffers: right! it's so frustrating

packages/stems/src/assets/styles/animations.css
dylanjeffers: yes i went ahead and updated all usages :)

packages/web/src/components/image-selection/ImageSelectionButton.js
```diff
@@ -65,7 +66,7 @@ const ImageSelectionButton = ({
       {includePopup ? (
         <>
           <Button
-            ref={anchorRef}
+            ref={anchorRefProp ? undefined : anchorRefInner}
```
Kyle-Shanks, line 69: curious what this ref stuff is for?

Kyle-Shanks (approved):

packages/web/src/components/image-selection/ImageSelectionButton.js
dylanjeffers: so remember how the artist popover was weirldy positioned relative to the artwork? the reason was is because the popover was anchored to the button that opens it, not the artwork itself, so i updated things to accept an anchorRef, and kept the inner one around for compatibility with other usages where we actually want to anchor to the button.

Kyle-Shanks: ah interesting. sounds good

PR #3638, opened 2023-06-23
[C-2697] Fix mobile playlist qa issues

dylanjeffers:
### Description
- Android Issues
    - [ ]  Clicking edit button on android multiple times doesn’t re-open edit button
    - [x]  Clicking artwork on android shows “Profile Photo”
        - What should this be changed to?
- [x]  Clicking “delete playlist” closes the edit screen incorrectly
    - Issues getting the drawer to render in front of the form drawer
- [x]  Artwork is not saving
    - Seems to be saving, but no optimistic update so it shows up late
    - Need to update store so that it shows up immediately
- [x]  Find artwork for me text and icon seem to be off
    - Make the icon slightly smaller
- [x]  Disabled action buttons should be neutral-light-7
    - They currently have a 50% opacity thing happening which, if switched to neutral7 might mess up other things that rely on the 50% opacity for other colors
    - need to check
- [x]  Remove track click event sometimes doesn’t work
- [x]  Making playlist public results in runtime error
    - successfully published playlist though
    - Seems to have resolved itself
- [x]  drag-drop vibration should be “hard” for initialized, and “soft” for reorder
    - No hard option, but set to medium on start and soft on reorder
- [x]  Vertical centering issue with playlist ‘making public…’ toast
- [x]  Action button sizes are weird
    - Edit icon was slightly too big
- [x]  Update ‘Submit’ test to ‘Save’ on edit screen
- [x]  Back arrow on the top of the edit screen should be an X
- [x]  re-arranging for drag-drop delay should be shorter
    - Changed to 100ms
- [x]  “Drag to reorder” text seems a bit different
- [x]  Confirmation button says “Submit” instead of “Save”
- [x]  Remove track icons seems a bit small
- [x]  32 pixels from bottom of edit playlist tile to “save” “cancel” bottom area
- [x]  Playlist artwork has incorrect margins, seems to be 16 pixels all around
- [x]  Playlist artwork border radius incorrect
- [x]  collection-screen action buttons should be neutral-light-2 instead of neutral-light-4
- [x]  Fix on drag opacity issue (Told to wait on this)
- [x]  Input updates (Might do these as a later issue)
    - [x]  Description field’s placeholder text same color as label, and has somewhat awkward spacing
    - [x]  The cursor is behind placeholder text too which is a bit strange.


test.txt
Kyle-Shanks, line 1: 👀

dylanjeffers: you passed the test

packages/mobile/src/components/core/IconButton.tsx
```diff
@@ -77,15 +77,9 @@ export const IconButton = ({
...
-        <View
-          style={[
-            styles.icon,
-            isDisabled && { opacity: 0.5 },
-            stylesProp?.icon
-          ]}
-        >
+        <View style={[styles.icon, stylesProp?.icon]}>
           <Icon
-            fill={fill}
```
Kyle-Shanks, line 88: are there any buttons that rely on the 50% opacity for colors other than a lighter grey?

packages/mobile/src/components/core/VirtualizedKeyboardAwareScrollView.tsx
Kyle-Shanks, line 1: neat

packages/mobile/src/components/core/IconButton.tsx
dylanjeffers: Ah that is a good point, to be safe I'll add disabled fill as an option

packages/mobile/src/components/details-tile/DetailsTileActionButtons.tsx
```diff
@@ -125,18 +112,18 @@ export const DetailsTileActionButtons = ({
 
   const shareButton = (
     <IconButton
-      fill={neutralLight2}
+      fill={neutralLight4}
```
Kyle-Shanks, line 115: are these supposed to be 4 or 2? was told in one of the syncs that these are all supposed to be 2

dylanjeffers: So I was double checking with Sammie and she said to stick with 4, which is the "diminished" color, and later when we do design system change the diminished color to higher contrast color. So for time being keep with the default color for icons

packages/mobile/src/screens/edit-playlist-screen/LegacyEditPlaylistScreen.tsx
```diff
@@ -32,7 +31,7 @@ const useStyles = makeStyles(({ spacing }) => ({
   }
 }))
 
-const EditPlaylistForm = (props: FormikProps<EditPlaylistValues>) => {
+const EditPlaylistForm = (props: FormikProps<any>) => {
```
Kyle-Shanks, line 34: do the edit values not work here anymore?

packages/mobile/src/screens/wallet-connect/components/ConfirmRemoveWalletDrawer.tsx
```diff
@@ -64,37 +43,11 @@ export const ConfirmRemoveWalletDrawer = () => {
   if (visibleState === false) return null
 
   return (
-    <NativeDrawer
+    <ConfirmationDrawer
```
Kyle-Shanks, line 46: nice

packages/mobile/src/screens/edit-playlist-screen/LegacyEditPlaylistScreen.tsx
dylanjeffers: They do just rather than keep the new and legacy ones around I just went with removing the ones for legacy since it'll be removed soon. Kinda silly though, I can just make a legacy one as well and add it back

Kyle-Shanks (approved):
awesome stuff here

packages/mobile/src/components/core/VirtualizedKeyboardAwareScrollView.tsx
dylanjeffers: yeah for your context it was complaining about flatlist in scrollview, this miraculously worked!

packages/mobile/src/screens/wallet-connect/components/ConfirmRemoveWalletDrawer.tsx
dylanjeffers: yeah im happy i was able to kill a few birds here

PR #3645, opened 2023-06-23
[PAY-1530] adds usdc feature flag and hooks for fetching it

schottra:
### Description
Skeleton code for fetching the usdc feature flag in web and mobile

### Dragons
N/A

### How Has This Been Tested?
N/A

N/A

### Feature Flags ###
usdc flag has been created and defaults to false. No code depends on it yet.

### DIFF
--- a/packages/mobile/src/hooks/useIsUSDCEnabled.ts
+++ b/packages/mobile/src/hooks/useIsUSDCEnabled.ts
@@ -1,0 +1,6 @@
+import { FeatureFlags } from '@audius/common'
+
+import { useFeatureFlag } from './useRemoteConfig'
+
+export const useIsUSDCEnabled = () =>
+  useFeatureFlag(FeatureFlags.USDC_PURCHASES).isEnabled
--- a/packages/common/src/services/remote-config/feature-flags.ts
+++ b/packages/common/src/services/remote-config/feature-flags.ts
@@ -47,7 +47,8 @@ export enum FeatureFlags {
   PLAYLIST_UPDATES_POST_QA = 'playlist_updates_post_qa',
   AI_ATTRIBUTION = 'ai_attribution',
   DEVELOPER_APPS_PAGE = 'developer_apps_page',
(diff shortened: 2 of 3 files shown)


packages/mobile/src/hooks/useIsUSDCEnabled.ts
```diff
@@ -0,0 +1,6 @@
+import { FeatureFlags } from '@audius/common'
+
+import { useFeatureFlag } from './useRemoteConfig'
+
+export const useIsUSDCEnabled = () =>
```
rickyrombo, line 5: nit: personally don't see the need of this hook wrapper, calling `useFlag`/`useFeatureFlag` seems less indirect

rickyrombo (approved):

audius-infra:
Preview this change https://demo.audius.co/pay-1530-add-usdc-feature-flag

dylanjeffers (approved):
and so it begins!

packages/mobile/src/hooks/useIsUSDCEnabled.ts
piazzatron: personally like this abstraction - fewer imports, no need to think about the feature flag API or enums when you just want to enable/disable some code

piazzatron (approved):
A journey of a thousand miles (hopefully not) begins with a single step 🔥

packages/mobile/src/hooks/useIsUSDCEnabled.ts
schottra: I agree. I was following our feature flag guide, which requests that if you're going to use a flag in more than a couple of places (which we definitely will), that you create a function for reading it.

But maybe that was written before the hooks existed?

rickyrombo: Guess I'm the odd one out 😅 it just seems like a lot of code to turn a one liner into another one liner

PR #3646, opened 2023-06-23
[plat-1085] persist track and collection is_hidden when saving or rep…

sabrina-kiam:
### Description
passes whether a track or collection is private/unlisted when saving or reposting. as of right now, users can favorite and repost unlisted tracks via the play bar, which should not be allowed. we want to stop transactions where users are favoriting or reposting unlisted items so passing this metadata lets entity manager decide whether to continue the transaction.

follows a very similar pattern to is repost of repost

### Dragons

### How Has This Been Tested?

going to test locally and then e2e against stage 

### Feature Flags ###

### DIFF
--- a/packages/common/src/services/audius-backend/AudiusBackend.ts
+++ b/packages/common/src/services/audius-backend/AudiusBackend.ts
@@ -994,7 +994,7 @@ export const audiusBackend = ({
 
   async function repostTrack(
     trackId: ID,
-    metadata?: { is_repost_of_repost: boolean }
+    metadata?: { is_repost_of_repost: boolean; is_hidden: boolean }
   ) {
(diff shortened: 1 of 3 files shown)


audius-infra:
Preview this change https://demo.audius.co/sk/plat-1085

isaacsolo:
https://github.com/AudiusProject/audius-protocol/pull/5478#discussion_r1240459341

similar to this comment, i think we should avoid using metadata for social features when it isn't necessary. maybe the client side change could just block playing if the track is hidden?

sabrina-kiam:
gonna close this since i went a different approach that doesn't require client changes, might make a different diff disabling the buttons if easy enough

PR #3647, opened 2023-06-23
[C-2677] Remix Settings Modal layout complete

amendelsohn:
### Description

Layout for remix settings modal

Future work:
- nested form for submission logic
- fix track preview image loading

image
image

### Feature Flags ###

`UPLOAD_REDESIGN_ENABLED`


audius-infra:
Preview this change https://demo.audius.co/amendel-upload-remix-field

packages/web/src/assets/img/iconRemixGray.svg
dylanjeffers, line 1: so can we not use the iconRemix and set fill?

packages/web/src/pages/upload-page/fields/RemixModalField.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+export type RemixOfField = Nullable<{ tracks: { parent_track_id: ID }[] }>
+
+const HIDE_REMIX_FIELD_NAME = 'hide_remixes'
+const REMIX_OF_FIELD_NAME = 'remix_of'
+const REMIX_LINK_FIELD_NAME = 'remix_of_link'
+
+export const RemixModalField = () => {
+  const [linkField] = useField(REMIX_LINK_FIELD_NAME)
+  const debugLink = linkField.value?.replace(
```
dylanjeffers, line 55: we want dis?

packages/web/src/pages/upload-page/fields/RemixModalField.tsx
```diff
@@ -0,0 +1,144 @@
...
+    </ModalField>
+  )
+}
+
+type TrackInfoProps = {
+  track: Track | null
+  user: UserMetadata | null
+}
+
+const g = withNullGuard(
```
dylanjeffers, line 118: i personally dont totally love the nullGuard pattern, was this what we had before or do you like it? very down to iterate on it :)

packages/web/src/pages/upload-page/fields/ToggleRowField.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+type ToggleFieldProps = PropsWithChildren & {
+  name: string
+  header: string
+  description: string
+}
+
+export const ToggleRowField = (props: ToggleFieldProps) => {
+  const { name, header, description, children } = props
+  const [{ value }, , { setValue }] = useField<boolean>(name)
```
dylanjeffers, line 17: can we use the onChange from the useField? personally i like using that over setValue whenever possible

dylanjeffers (approved):
swell, swell indeed

packages/web/src/pages/upload-page/fields/RemixModalField.tsx
amendelsohn: oh whoops yea doesn't work with localhost but shoudn't check this in :)

packages/web/src/pages/upload-page/fields/RemixModalField.tsx
amendelsohn: this is what was there. I'm down to change it

packages/web/src/pages/upload-page/fields/ToggleRowField.tsx
amendelsohn: yea I think I got into a habit of using setValue cause it wasn't working with the old fields, but this one should work with onChange

packages/web/src/assets/img/iconRemixGray.svg
amendelsohn: I tried that way. I think it's an issue because the fill in the existing version is a gradient and doesn't get overridden properly by the fill property. I see the correct color and fill values in the inspector, but it's still purple. The new icon works, but we'd have to apply the gradient styling to every other usage of it.

Tempted to punt until design audits for Harmony, but also happy to try again to apply the previous style to the new svg

PR #3649, opened 2023-06-24
[PAY-1522][PAY-1451] Fix issues with DMs notification dot

piazzatron:
### Description

Fixes two issues with the existing implementation:
- The `NotificationDot` decorator would scale with the main button hover animation, which differs from existing behavior on the notification button
- When the `NotificationDot` decorator was present, some CSS would break that was looking for a direct SVG child, causing incorrect SVG path color. Additionally, the outer wrapper in `Icon` needed to have `display: flex` set for the SVG to be positioned properly.

Ultimately I opted to delete the `Icon` component - since the scaling animation needs to be defined outside the component anyways, this new abstraction wasn't useful here since we can't pass the dot as a decorator. Willing to be persuaded otherwise by @schottra if we think this is useful to keep around, however!

### Dragons

### How Has This Been Tested?

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/piazz-fix-nav-menu-issues

schottra (approved):
Oof, really wanted that Icon abstraction. Does it still work if we just remove the `decorator` concept and render the dot outside of the Icon?
I think long-term the Icon is still a useful abstraction as it will keep us from having to duplicate logic around sizing, color, etc.

piazzatron:
> Oof, really wanted that Icon abstraction. Does it still work if we just remove the `decorator` concept and render the dot outside of the Icon? I think long-term the Icon is still a useful abstraction as it will keep us from having to duplicate logic around sizing, color, etc.

I'll bring it back, sans decorator :)

PR #3651, opened 2023-06-26
[PAY-1529] Add initial types for usdc purchases

schottra:
### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/common/src/models/Track.ts
+++ b/packages/common/src/models/Track.ts
@@ -10,6 +10,7 @@ import { Repost } from './Repost'
 import { StemCategory } from './Stems'
 import { Timestamped } from './Timestamped'
 import { User, UserMetadata } from './User'
+import { StringWei } from './Wallet'
 
 type EpochTimeStamp = number
 
@@ -74,12 +75,18 @@ export type PremiumConditionsSolNFTCollection = {
   externalLink: Nullable<string>
 }
 
+export type PremiumConditionsUSDCPurchase = {
+  price: StringWei
+  slot: number
+}
+
 export type PremiumConditions = {
   nft_collection?:
     | PremiumConditionsEthNFTCollection
     | PremiumConditionsSolNFTCollection
   follow_user_id?: number
   tip_user_id?: number
+  usdc_purchase?: PremiumConditionsUSDCPurchase
 }
 
 export type PremiumContentSignature = {
--- a/packages/common/src/models/Track.ts
+++ b/packages/common/src/models/Track.ts
@@ -10,7 +10,7 @@ import { Repost } from './Repost'
 import { StemCategory } from './Stems'
 import { Timestamped } from './Timestamped'
 import { User, UserMetadata } from './User'
-import { StringWei } from './Wallet'
+import { StringUSDC } from './Wallet'
 
 type EpochTimeStamp = number
 
@@ -76,7 +76,7 @@ export type PremiumConditionsSolNFTCollection = {
 }
 
 export type PremiumConditionsUSDCPurchase = {
-  price: StringWei
+  price: StringUSDC
   slot: number
 }
 
--- a/packages/common/src/models/USDCTransactions.ts
+++ b/packages/common/src/models/USDCTransactions.ts
@@ -1,0 +1,82 @@
+import { Nullable } from '../utils/typeUtils'
+
+import { StringUSDC } from './Wallet'
+
+export enum USDCTransactionType {
+  PURCHASE = 'PURCHASE',
+  PURCHASE_CONTENT = 'PURCHASE_CONTENT',
+  TRANSFER = 'TRANSFER'
+}
+
+export enum USDCTransactionMethod {
+  // Transfer methods
+  SEND = 'SENT',
+  // TODO: Is this needed? Do we write a transaction for both the sender and recipient?
+  RECEIVE = 'RECEIVED',
+
+  // Purchase Methods
+  STRIPE = 'STRIPE'
+}
+
+export enum USDCTransactionMetadataType {
+  PURCHASE_USDC = 'PURCHASE_USDC',
+  PURCHASE_CONTENT = 'PURCHASE_CONTENT',
+  TRANSER = 'TRANSFER'
+}
+
+export type InAppUSDCPurchaseMetadata = {
+  discriminator: USDCTransactionMetadataType.PURCHASE_USDC
+  usd: string
+  usdc: StringUSDC
+  purchaseTransactionId: string
+}
+
+export enum USDCContentPurchaseType {
+  TRACK = 'TRACK'
+}
+
+export type USDCPurchaseContentMetadata = {
+  discriminator: USDCTransactionMetadataType.PURCHASE_CONTENT
+  amount: StringUSDC
+  senderUserId: string
+  receiverUserId: string
+  purchaseType: USDCContentPurchaseType
+  contentId: string
+}
+
+export type USDCTransferMetadata = {
+  discriminator: USDCTransactionMetadataType.TRANSER
+  amount: StringUSDC
+  destination: string
+}
+
+export type USDCTransactionDetails =
+  | {
+      signature: string
+      transactionType: USDCTransactionType.PURCHASE
+      method: USDCTransactionMethod.STRIPE
+      date: string
+      change: StringUSDC
+      balance: StringUSDC
+      metadata?: Nullable<InAppUSDCPurchaseMetadata>
+    }
+  | {
+      signature: string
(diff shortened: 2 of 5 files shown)


packages/common/src/models/User.ts
```diff
@@ -58,6 +58,7 @@ export type UserMetadata = {
   tikTokVerified?: boolean
   balance?: Nullable<StringWei>
   total_balance?: Nullable<StringWei>
+  usdc_balanc?: Nullable<StringWei>
```
dharit-tan, line 61: oops missing an e

schottra: Nope totally intentional!
😅

packages/common/src/models/Track.ts
```diff
@@ -74,12 +75,18 @@ export type PremiumConditionsSolNFTCollection = {
   externalLink: Nullable<string>
 }
 
+export type PremiumConditionsUSDCPurchase = {
+  price: StringWei
```
piazzatron, line 79: Worth discussing - I think we can have this be `StringUSDC` and pad it to the correct number of decimals on DN. We added extra padding to WAudio on DN because we combine the Eth + Sol AUDIO currencies together in the client, and thus need the same amount of precision. But we will never show USDC combined, so should be able to index it without extra padding. cc @rickyrombo

packages/common/src/services/audius-api-client/types.ts
```diff
@@ -26,6 +26,7 @@ export type APIUser = {
   artist_pick_track_id: Nullable<OpaqueID>
   blocknumber: number
   balance: string
+  usdc_balance: string
```
piazzatron, line 29: IIRC we decided that we wouldn't return `usdc_balance` from the API and just fetch it from chain

packages/common/src/models/USDCTransactions.ts
```diff
@@ -0,0 +1,79 @@
...
+      method: USDCTransactionMethod.SEND | USDCTransactionMethod.RECEIVE
+      date: string
+      change: StringUSDC
+      balance: StringUSDC
+      metadata?: Nullable<USDCPurchaseContentMetadata>
+    }
+  | {
+      signature: string
+      transactionType: USDCTransactionType.TRANSFER
+      method: USDCTransactionMethod.SEND | USDCTransactionMethod.RECEIVE
```
piazzatron, line 74: Don't think we will support receiving USDC (since we don't expose your USDC wallet anywhere), so should be able to cut this `RECEIVE`

packages/common/src/models/USDCTransactions.ts
```diff
@@ -0,0 +1,79 @@
+import { Nullable } from '../utils/typeUtils'
+
+import { StringUSDC } from './Wallet'
+
+export enum USDCTransactionType {
+  PURCHASE = 'PURCHASE',
```
piazzatron, line 6: nitty nit: maybe `PURCHASE_USDC`, so `PURCHASE_CONTENT` doesn't seem like a subset of this?

and then, could `USDCTransactionMetadataType` go away? I know we're modeling this off the old `Audio` tx history stuff

packages/common/src/models/User.ts
```diff
@@ -58,6 +58,7 @@ export type UserMetadata = {
   tikTokVerified?: boolean
   balance?: Nullable<StringWei>
   total_balance?: Nullable<StringWei>
+  usdc_balance?: Nullable<StringWei>
```
piazzatron, line 61: I think `StringUSDC`? Alternatively this could live in it's own slice (since only the current user will have this field populated), but no strong conviction either way there

piazzatron (approved):
Sweet!!!

smol nit: convention is to capitalize PR titles and issue number 🙂

packages/common/src/models/Track.ts
rickyrombo: I was wondering how many decimals USDC has in both Ethereum and Solana land so I did some poking around, and it seems like it's 6 decimals in both!

packages/common/src/models/USDCTransactions.ts
```diff
@@ -0,0 +1,79 @@
...
+  PURCHASE_USDC = 'PURCHASE_USDC',
+  PURCHASE_CONTENT = 'PURCHASE_CONTENT',
+  TRANSER = 'TRANSFER'
+}
+
+export type InAppUSDCPurchaseMetadata = {
+  discriminator: USDCTransactionMetadataType.PURCHASE_USDC
+  usd: string
+  usdc: StringUSDC
+  purchaseTransactionId: string
```
rickyrombo, line 30: fyi this could be redundant as the transaction might be able to directly go to the userbank (still probably need it though to associate on identity? I forget)

packages/common/src/models/USDCTransactions.ts
rickyrombo: Man I really overloaded "Buy" "Purchase" etc in the buy audio flow.. I could go either way - either keeping it "PURCHASE" to match the Audio one, or what Piazza is suggesting to make an improvement despite losing consistency

packages/common/src/models/USDCTransactions.ts
```diff
@@ -0,0 +1,79 @@
...
+  TRACK = 'TRACK'
+}
+
+export type USDCPurchaseContentMetadata = {
+  discriminator: USDCTransactionMetadataType.PURCHASE_CONTENT
+  amount: StringUSDC
+  senderUserId: string
+  receiverUserId: string
+  purchaseType: USDCContentPurchaseType
+  contentId: string
```
rickyrombo, line 43: Wonder if we want these all to be string ids or if we want numeric `ID`s? @piazzatron wdyt?

rickyrombo (approved):

packages/common/src/models/USDCTransactions.ts
piazzatron: Good catch - numeric IDs! Endpoint should return the encoded string ID but internally we always use the integer version

packages/common/src/models/Track.ts
schottra: Oh this makes sense. I remember now that AUDIO has different precision in the two environments, while USDC does not. I agree that we can/should just make this `StringUSDC`

packages/common/src/models/USDCTransactions.ts
schottra: This pattern came from audio transactions, and I was initially confused by it. It looks like we write two transactions each time Audio is sent (tipping), one for the sender and one for the receiver. I think having the `receive` item helps simplify the transactions list for any individual user because we select by `user_bank`, and this would correctly show the times you received USDC via a purchase.

packages/common/src/models/USDCTransactions.ts
schottra: I'm all for PURCHASE_USDC. I think having the separate types was mostly just convention/semantics. So I'll ditch the separate metadata type enum and just use the transaction type as the `discriminator`

packages/common/src/models/USDCTransactions.ts
schottra: Oh nice catch, thank you.

packages/common/src/models/USDCTransactions.ts
schottra: In my mind, I had imagined this was the id of a Solana or Stripe transaction that we wanted to keep track of because we weren't going to index it. But if this is going to be a duplicate of the `signature` column on `usdc_transactions`, then I don't think we need to store it separately.

packages/common/src/models/User.ts
schottra: Yeah I think this only needs to live here if we're planning on returning it with the User response from DN. If we're fetching the balance from chain directly, this can go.
I'll remove it for now until we implement that feature.

packages/common/src/services/audius-api-client/types.ts
schottra: Going to remove it for now. We can add it back if needed when implementing the balance tile.

packages/common/src/models/Wallet.ts
```diff
@@ -4,8 +4,10 @@ import { Brand } from 'utils/typeUtils'
 
 export type StringWei = Brand<string, 'stringWEI'>
 export type StringAudio = string
+export type StringUSDC = string
```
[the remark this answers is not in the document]
schottra: I updated this one to be branded. Since we use `StringAudio` in a lot of places, I don't want to change it for now in case there's som unknown side effect.

audius-infra:
Preview this change https://demo.audius.co/pay-1529-model-setup

PR #3653, opened 2023-06-26
[plat-1085] disable fav and repost on hidden tracks in play bar

sabrina-kiam:
### Description
Disable favorite and repost buttons in the play bar for hidden tracks. they should not be able to be liked/favorited, and there is no ui button for it on the page, but there is a way to do so via the play bar currently. this diff disables the buttons in the play bar

This change is only for tracks since you can't even view private playlists via links as of right now.

### How Has This Been Tested?

tested via ui, will attach video

https://github.com/AudiusProject/audius-client/assets/109105561/5556dfd4-8f29-4344-b4a5-3483c8eb8610


sabrina-kiam:
relevant slack thread - https://audius-internal.slack.com/archives/C02JBA52NKZ/p1687539668215279

PR #3656, opened 2023-06-27
Add Sentry error logging to Audius query C-2800

nicoback2:
### Description

Log caught API errors to Sentry

Not 100% sure this belongs here vs in the components using the queries/APIs, but I think this would be the most convenient.

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/common/src/audius-query/AudiusQueryContext.ts
+++ b/packages/common/src/audius-query/AudiusQueryContext.ts
@@ -1,4 +1,5 @@
 import { createContext } from 'react'
+
 import { AudiusSdk } from '@audius/sdk'
 
 import type { AudiusAPIClient } from 'services/audius-api-client'
--- a/packages/common/src/audius-query/AudiusQueryContext.ts
+++ b/packages/common/src/audius-query/AudiusQueryContext.ts
@@ -1,7 +1,7 @@
 import { createContext } from 'react'
 
 import { AudiusSdk } from '@audius/sdk'
-
+import { ReportToSentryArgs } from '../models'
 import type { AudiusAPIClient } from 'services/audius-api-client'
 import { AudiusBackend } from 'services/index'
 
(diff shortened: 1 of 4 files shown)


packages/common/src/audius-query/createApi.ts
```diff
@@ -296,6 +297,14 @@ const fetchData = async <Args, Data>(
 
     endpoint.onQuerySuccess?.(data, fetchArgs, { dispatch })
   } catch (e) {
+    context.reportToSentry({
+      error: e as Error,
+      level: ErrorLevel.Error,
+      additionalInfo: { fetchArgs, endpoint },
```
amendelsohn, line 303: ah good idea!

amendelsohn: Wondering if we should just restrict this to the call to `endpoint.fetch` or keep it around the whole block. This will capture errors in the before and after handlers which is probably good, and I don't think dispatching the actions should be too noisy so yeah let's go with this.

packages/common/src/audius-query/AudiusQueryContext.ts
```diff
@@ -1,14 +1,15 @@
...
-
+import { ReportToSentryArgs } from '../models'
 import type { AudiusAPIClient } from 'services/audius-api-client'
 import { AudiusBackend } from 'services/index'
 
 export type AudiusQueryContextType = {
   apiClient: AudiusAPIClient
   audiusSdk: () => Promise<AudiusSdk>
   audiusBackend: AudiusBackend
+  reportToSentry: (args: ReportToSentryArgs) => void
```
amendelsohn, line 12: Do we actually need this in the context if we're catching all errors in `CreateApi` anyway? It allows the endpoint's implementer to call report to sentry which is added utility anyway I suppose.

amendelsohn (approved):
This is a great change. And very clean!!

packages/common/src/audius-query/createApi.ts
nicoback2: SG!

packages/common/src/audius-query/AudiusQueryContext.ts
nicoback2: Ooh good point, but yeah I just added it here because I couldn't figure out a better way to report to Sentry since we're in common

audius-infra:
Preview this change https://demo.audius.co/nkang--sentry-aquery

PR #3660, opened 2023-06-27
[C-2787] Reregister device-token on app startup

dylanjeffers:
### Description

At app startup, reregister device token if notification permissions are enabled. This will prevent cases where a device id changes causing users to not receive notifications.
- Also improves some typing
- Also fixes issue with fast-image, there are some obscure cases where uri is undefined leading to app-ending bug

### DIFF
--- a/packages/mobile/src/components/image/FastImage.tsx
+++ b/packages/mobile/src/components/image/FastImage.tsx
@@ -15,12 +15,13 @@ export type ImageProps = Omit<FastImageProps, 'source'>
 export const FastImage = (props: FastImageProps) => {
   const { source, priority, ...other } = props
 
-  const imageSource =
-    typeof source === 'number'
-      ? source
-      : Array.isArray(source)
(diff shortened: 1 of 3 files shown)


packages/mobile/src/components/image/FastImage.tsx
```diff
@@ -15,12 +15,13 @@ export type ImageProps = Omit<FastImageProps, 'source'>
 export const FastImage = (props: FastImageProps) => {
   const { source, priority, ...other } = props
 
-  const imageSource =
-    typeof source === 'number'
-      ? source
-      : Array.isArray(source)
-      ? { uri: source[0].uri, priority }
-      : { uri: source.uri, priority }
+  const imageSource = !source
```
rickyrombo, line 18: nice, I think this fixes https://linear.app/audius/issue/PAY-1428/mobile-error-searching-for-ed-in-search-users-modal !

PR #3663, opened 2023-06-27
[PAY-1545] Updates to mobile gated content track tiles

dharit-tan:
### Description
- Remove uses of `useIsGatedContentEnabled` in all lineup tile related components
- Updates to gated content track tiles

### Dragons

### How Has This Been Tested?
Screenshots below were taken before I updated the font weight to medium, did not update the screenshots bc i am lazy :)

Local ios stage:
Simulator Screenshot - iPhone 14 Pro - 2023-06-27 at 18 28 20
Simulator Screenshot - iPhone 14 Pro - 2023-06-27 at 18 26 40
Simulator Screenshot - iPhone 14 Pro - 2023-06-28 at 17 32 17

Owner: 
Simulator Screenshot - iPhone 14 Pro - 2023-06-27 at 18 28 08

### Feature Flags ###

Not feature-flagged hence merging to feature branch.


packages/mobile/src/assets/images/iconLock.svg
schottra, line 1: Is this icon used anywhere else and does removing it break anything?
Seems like just an update to a smaller viewport?
Also wondering if I need to update this icon in web as well...

packages/mobile/src/components/lineup-tile/LineupTilePremiumContentTypeTag.tsx
```diff
@@ -0,0 +1,74 @@
...
+  SPECIAL_ACCESS = 'special access'
+}
+
+type LineupTilePremiumContentTypeTagProps = {
+  premiumConditions: Nullable<PremiumConditions>
+  doesUserHaveAccess?: boolean
+  isOwner: boolean
+}
+
+export const LineupTilePremiumContentTypeTag = ({
```
schottra, line 39: Mind adding a docstring to this describing what the component does?

packages/mobile/src/components/lineup-tile/LineupTilePremiumContentTypeTag.tsx
```diff
@@ -0,0 +1,74 @@
...
+        icon: IconSpecialAccess,
+        color: doesUserHaveAccess && !isOwner ? neutralLight4 : accentBlue,
+        text: messages.specialAccess
+      }
+    }
+  }, [accentBlue, doesUserHaveAccess, isOwner, neutralLight4])
+
+  const { icon: Icon, color, text } = premiumContentTypeMap[type]
+
+  return premiumConditions ? (
```
schottra, line 68: minor nit: Slightly less complicated if `premiumConditions` is required prop on this component and we skip rendering it entirely in the parent if we don't have `premiumConditions`.

packages/mobile/src/components/lineup-tile/LineupTilePremiumContentTypeTag.tsx
```diff
@@ -0,0 +1,74 @@
...
+    <View style={styles.root}>
+      <Icon fill={color} height={16} width={16} />
```
schottra, line 70: spacing units?

packages/mobile/src/components/lineup-tile/LineupTileStats.tsx
```diff
@@ -32,14 +41,12 @@ const formatPlayCount = (playCount?: number) => {
...
+const useStyles = makeStyles(({ spacing, palette }) => ({
   stats: {
     flexDirection: 'row',
-    flex: 0,
     alignItems: 'stretch',
-    paddingVertical: 2,
-    marginRight: 10,
-    height: 26
+    paddingVertical: spacing(2),
+    marginHorizontal: spacing(2.5)
```
schottra, line 49: I had thought the design had everything aligned to grid units. Why the 2.5 spacing here?

schottra (approved):
Looks good!

packages/mobile/src/assets/images/iconLock.svg
dharit-tan: I think the icon itself also looks slightly different. Probably also good to change in web!

packages/mobile/src/components/lineup-tile/LineupTilePremiumContentTypeTag.tsx
schottra: Ah, looking at this on web I could see a case where `isPremium` is true, but the conditions are empty. While that would be rare, we probably still want to render that case as "Special Access". So this is probably fine to leave as is

PR #3667, opened 2023-06-28
[C-2812] Use storageNodeSelector in libs, fix mobile images

dylanjeffers:
### Description

Uses latest sdk to leverage exported StorageNodeSelector in libs and sdk.
Uses this to fix mobile image fetches for v2 users


audius-infra:

rickyrombo (approved):
Looks great to me. Super minor nit, but maybe we should put our SDK singleton services under a nested `sdk` directory in services? I have a feeling we're going to make al ot of these 😢

audius-infra:

theoilie (approved):
many thanks!!

rickyrombo:
> > Looks great to me. Super minor nit, but maybe we should put our SDK singleton services under a nested `sdk` directory in services? I have a feeling we're going to make a lot of these 😢
> 
> Super down! though since its also used in libs, its kinda like a general service?

Yeah, they're still SDK services though (they come from SDK), they're just being used in libs right? Either way, it's super minor nit

sliptype (approved):

dylanjeffers:
> > > Looks great to me. Super minor nit, but maybe we should put our SDK singleton services under a nested `sdk` directory in services? I have a feeling we're going to make a lot of these 😢
> > 
> > 
> > Super down! though since its also used in libs, its kinda like a general service?
> 
> Yeah, they're still SDK services though (they come from SDK), they're just being used in libs right? Either way, it's super minor nit

oh yeah totally! updated

PR #3669, opened 2023-06-28
[PAY-1534] Allow Popup to be mounted inside a container

rickyrombo:
### Description

Attaching to the document body works for many popups, since they're in an area that isn't scrollable and we want the popup to not be cut off by the container in most cases. However, as soon as the body is scrollable and the popup shouldn't go outside its container, strange bugs arise:

In this first video, the popup doesn't scroll at all. This is because getScrollParent doesn't account for our custom scrollbar not using `overflow` to make the scrollbar appear.

https://github.com/AudiusProject/audius-client/assets/3690498/b183d845-4ecb-4e67-83c4-3b8c5d484a73

First step then, is to fix `getScrollParent` to detect our custom scrollbar implementation, easiest way is checking for `scrollbar-container` class. However, that results in the popup going beyond the bounds of the modal, which doesn't really make sense.

https://github.com/AudiusProject/audius-client/assets/3690498/42c99f72-b33f-4ce8-8b11-68aeb285b36b

This is because the popup isn't aware that it's meant to be contained inside the list, as it gets portal'ed all the way to the `document.body`. So instead, we need to portal it into the list so it gets masked when outside the bounds of the container thanks to `overflow: hidden`. This also requires recalculating the X and Y position to be relative to the mount element, and also makes the assumption that the mount element is the nearest parent with `position: relative` (in our case, the nearest parent with `position: relative` is actually the `scrollbar-container` a few ancestors up due to the scrollbar's nested divs, but it's effectively the same). 

https://github.com/AudiusProject/audius-client/assets/3690498/a8f2df93-74d9-4021-897e-7e4c45c67b05

Note that this means the `containerRef` and `mountRef` need to be the same! Otherwise, the popup wont get adjusted to appear within the mounted area, and if `overflow: hidden` is set, that means it won't be visible.

It's not the best implementation. I am _very open_ to alternative ideas/solutions/feedback. The easiest thing would be to dismiss on scroll to not worry about this but that feels sorta bad from a UX perspective.

### Dragons

Touches `Popup` which has a lot of usage throughout the product, although changes should be mostly isolated to new `mountRef` prop users.

Using `mountRef` makes some assumptions about `position: relative`

### How Has This Been Tested?

Tested:
- Navbar popup menu
- Notifs popup
- Chat overflow popup menu
- New Message modal chat user result overflow popup menu
- Playlist sidebar overflow popup menu
- New Playlist sidebar popup menu
- Upload track Add Artwork popup

### Feature Flags ###


packages/web/src/pages/MainContentContext.tsx
```diff
@@ -1,12 +1,14 @@
-import { createContext, memo, useRef, MutableRefObject } from 'react'
+import { createContext, memo, useRef, RefObject } from 'react'
 
-export const MainContentContext = createContext({
-  mainContentRef: {} as MutableRefObject<HTMLDivElement | undefined>
+export const MainContentContext = createContext<{
+  mainContentRef: RefObject<HTMLDivElement>
```
rickyrombo, line 4: fyi @raymondjacobson I updated this type, it feels more correct

packages/stems/src/components/Popup/types.ts
```diff
@@ -35,7 +35,7 @@ export type PopupProps = {
   /**
    * Children to render inside the Popup
    */
-  children: ReactChild
+  children: ReactNode
```
rickyrombo, line 38: `ReactChild` is deprecated

packages/stems/src/components/Popup/types.ts
```diff
@@ -18,7 +18,7 @@ export type PopupProps = {
   /**
    * A ref to the element whose position will be used to anchor the Popup
    */
-  anchorRef: MutableRefObject<HTMLElement | null>
+  anchorRef: RefObject<HTMLElement>
```
rickyrombo, line 21: Updated to RefObject as it's not modified internally. A `MutableRefObject<HTMLElement | null>` can be duck-typed as a `RefObject<HTMLElement>` so there's no breaking change here.

packages/stems/src/components/Popup/types.ts
```diff
@@ -52,7 +52,7 @@ export type PopupProps = {
    * to be the size of the container it belongs to. If the popup expands outside
    * the bounds of the container, it repositions itself.
    */
-  containerRef?: MutableRefObject<HTMLDivElement | undefined>
+  containerRef?: RefObject<HTMLElement>
```
rickyrombo, line 55: This one is a breaking change since it was `HTMLDivElement | undefined` before but is `HTMLDivElement | null` now

packages/stems/src/utils/scrollParent.ts
```diff
@@ -19,7 +19,9 @@ export const getScrollParent = (node: HTMLElement): Element | null => {
     style(_node, 'overflow') +
     style(_node, 'overflow-y') +
     style(_node, 'overflow-x')
-  const scroll = (_node: HTMLElement) => regex.test(overflow(_node))
+  const scroll = (_node: HTMLElement) =>
+    regex.test(overflow(_node)) ||
+    _node.classList.contains('scrollbar-container') // Perfect scrollbar does overflow: hidden
```
rickyrombo, line 24: without this, we weren't detecting scroll parents that were using `react-perfect-scrollbar`

packages/stems/src/components/Popup/types.ts
dylanjeffers: should this be ReactElement actually? i think this ensures its a single child and is reffable?

packages/stems/src/components/Popup/types.ts
dylanjeffers: i like the change to HtmlElement

dylanjeffers (approved):
wow lots of great looking changes here

packages/stems/src/components/Popup/types.ts
rickyrombo: I think `ReactNode` is the default for ComponentPropsWithChildren. See also https://stackoverflow.com/questions/58123398/when-to-use-jsx-element-vs-reactnode-vs-reactelement

dylanjeffers: word, yeah i guess we receive the containerRef as a prop, so its actual children can just be whatever

packages/stems/src/components/Popup/Popup.tsx
```diff
@@ -162,12 +163,13 @@ const getComputedOrigins = (
...
   left: number,
-  wrapperRect: DOMRect
+  wrapperRect: DOMRect,
+  containerRect: Pick<DOMRect, 'width' | 'height'>
 ): { adjustedTop: number; adjustedLeft: number } => {
   if (!wrapperRect) return { adjustedTop: 0, adjustedLeft: 0 }
 
-  const containerWidth = window.innerWidth - CONTAINER_INSET_PADDING
-  const containerHeight = window.innerHeight - CONTAINER_INSET_PADDING
+  const containerWidth = containerRect.width - CONTAINER_INSET_PADDING
```
rickyrombo, line 171: this is the one change that leaks a bit out to existing impls - we were originally always computing the adjusted position based on window, not the container, and only using the container to get the computed origins

I *think* this is more correct with the intended behavior? @raymondjacobson I think you added the container stuff, does this seem right? How would I test the usecase for the containerRef outside of my current one to make sure I didn't break anything?

PR #3673, opened 2023-06-29
[PAY-1535] Refresh chat messages on render and pull to refresh on chats list

rickyrombo:
### Description

Would appreciate more eyes and testing on this, it's rather large.

- On rendering a chat screen:
  - If it's the first load:
    - Get batches of messages until we have all the unread messages
  - Otherwise
    - Get all messages newer than what we have
- Adds pull to refresh on the chat list screen
- Updates reducer logic to take the earliest prev_cursor and latest next_cursor in both fetchMessages and fetchChats
- Updates messages batch size to 50 from 10

### Dragons

Known issue: Going into a chat that has new messages before a chat has been refreshed to acknowledge an unread count won't show the inline indicator of the unread count. This is because we cache the unread state on loading the chat screen prior to refreshing and finding those new messages.

Cursors: I think things should work roughly the same. I do want to be extra extra careful about the `hasMoreUnread` flag in the sagas, because an infinite loop would be devastating.

### How Has This Been Tested?

Tested on iOS Sim by turning off websockets manually in code.

### Feature Flags ###

### DIFF
--- a/packages/common/src/store/pages/chat/sagas.ts
+++ b/packages/common/src/store/pages/chat/sagas.ts
@@ -37,9 +37,11 @@ const {
   fetchUnreadMessagesCountFailed,
   goToChat,
   fetchChatIfNecessary,
+  fetchLatestChats,
   fetchMoreChats,
   fetchMoreChatsSucceeded,
   fetchMoreChatsFailed,
+  fetchLatestMessages,
   fetchMoreMessages,
   fetchMoreMessagesSucceeded,
   fetchMoreMessagesFailed,
@@ -74,6 +76,9 @@ const {
   getOtherChatUsers
 } = chatSelectors
 const { toast } = toastActions
+
+const CHAT_PAGE_SIZE = 30
+const MESSAGES_PAGE_SIZE = 50
 
 /**
  * Helper to dispatch actions for fetching chat users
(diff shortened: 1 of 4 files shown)


audius-infra:
Preview this change https://demo.audius.co/mjp-dms-refresh

packages/common/src/store/pages/chat/slice.ts
```diff
@@ -156,26 +156,61 @@ const slice = createSlice({
...
       }
-      chatsAdapter.addMany(state.chats, action.payload.data)
+      chatsAdapter.upsertMany(state.chats, data)
     },
     fetchMoreChatsFailed: (state) => {
       state.chats.status = Status.ERROR
     },
+    fetchLatestMessages: (state, action: PayloadAction<{ chatId: string }>) => {
+      // triggers saga
+      if (!state.messages[action.payload.chatId]) {
```
dharit-tan, line 205: nit: could clean this up - destructure `chatId` and is there a way to initialize state using `entityAdapter` fns?

packages/mobile/src/screens/chat-screen/ChatListScreen.tsx
```diff
@@ -162,6 +170,8 @@ export const ChatListScreen = () => {
               ))
           ) : (
             <FlatList
+              refreshing={chatsStatus === 'REFRESHING'}
```
dharit-tan, line 173: ty!

packages/common/src/store/pages/chat/slice.ts
rickyrombo: Yep it's initted below with entityAdapter, but agree we can destructure

packages/common/src/store/pages/chat/slice.ts
```diff
@@ -203,15 +238,21 @@ const slice = createSlice({
...
       }
 
-      // Update the summary to include the min of next_cursor/next_count and
-      // prev_cursor/prev_count.
+      // Update the summary to include the max of next_cursor and
+      // min of prev_cursor.
       const existingSummary = state.chats.entities[chatId]?.messagesSummary
       const summaryToUse = { ...summary, ...existingSummary }
-      if (summary.next_count < (existingSummary?.next_count ?? Infinity)) {
+      if (
```
dharit-tan, line 245: is the reason `count` no longer works here because the `existingSummary.next_count` might be out of date? And would that cause us to recalculateHasTail on the wrong index on L274?

dharit-tan (approved):

packages/common/src/store/pages/chat/slice.ts
dharit-tan: I think we're probably good just from comparing before/after logic. Way to test would be to see if tails randomly pop in/out when websocket reconnects and re-sends latest messages.

dharit-tan (approved):

packages/common/src/store/pages/chat/slice.ts
rickyrombo: When you first load the chat, you load from the most recent, so `next_count` is zero. When you refresh, you also load from the most recent, _but the most recent has changed_. The `next_count` is still 0, but the cursor changed. If we just looked at `next_count` then we won't update the cursor, and we'll keep refetching anything newer than that cursor over and over.

rickyrombo: tl;Dr `next_count` is always zero, but the cursor changes

packages/common/src/store/pages/chat/sagas.ts
```diff
@@ -134,6 +182,75 @@ function* doFetchMoreChats() {
...
+      data = data.concat(response.data)
+      hasMoreUnread =
+        // If we have no more messages with our after filter, we're done
+        (!!after && response.data.length > 0) ||
+        // If we don't have an after filter, eg on first messages load, then go until we get all the unreads
+        (!after &&
+          !!chat?.unread_message_count &&
+          chat.unread_message_count >
+            (response.summary?.next_count ?? 0) + data.length)
+      before = response.summary?.prev_cursor
```
piazzatron, line 223: Very possibly not getting this because it's late, but what happens here with the inclusion of both a before and after cursor in the request?

rickyrombo: It only fetches messages between before/after

piazzatron (approved):

PR #3674, opened 2023-06-29
[C-2808] Improve storageNodeSelector usage and perf

dylanjeffers:
### Description

- Ensures `getStorageNodeSelector` always returns a singleton
- Adds bootstrap nodes to ensure instant availability to select nodes
- Adds storageNodeSelector to app context to ensure image hooks are re-ran when it becomes active
- Ensures all image functions provide endpoints.

### Open question
How do we feel about the bootstrap nodes i selected? i essentially just pulled them down from current services


theoilie (approved):
looks great! just flagging the minor typecheck issue in CI

PR #3675, opened 2023-06-29
[C-2813] Catch collectibles runtime errors

dylanjeffers:
### Description

Fixes issue where some users collectibles don't render correctly, which leads to app-ending bug. Wraps collectibles-cards in an error boundary, similar to notifications, to ensure individual errors can be caught and handled without crashing the app.


sliptype (approved):
Great idea!

PR #3683, opened 2023-06-30
Finish implementation of typography CSS module

schottra:
### Description
This adds the remaining typography styles from the [Figma designs](https://www.figma.com/file/IFhreiwgRdt5kecYQPFjyO/Typography%3A-Base-%5BSTEM%5D?type=design&node-id=933-421&mode=dev). These should now be ready to use in web code whenever a design uses a named typography style.

I also added some comments at the top of the file to explain how the classes map to our typography system and show some examples for usage.

Between my last PR and this one, the `light` variant was changed to `weak`, so I updated the two places that variant was used for title and label font styles.

### Dragons
N/A

### How Has This Been Tested?
Verified locally

N/A

### Feature Flags ###
N/A

### DIFF
--- a/packages/web/src/components/typography/typography.module.css
+++ b/packages/web/src/components/typography/typography.module.css
@@ -17,91 +17,169 @@
  ```
 */
 
-/* Title Styles */
-.titleLarge {
-  font-size: var(--font-l);
+/* Display Styles */
+
+/* 96 / 120 */
+.diplayXLarge {
+  font-size: var(--unit-24);
+  font-weight: var(--font-bold);
+  line-height: calc(var(--unit) * 30);
+}
+/* 72 / 88 */
+.displayLarge {
+  font-size: var(--unit-18);
+  font-weight: var(--font-bold);
+  line-height: var(--unit-22);
+}
+/* 56 / 68 */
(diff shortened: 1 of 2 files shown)


packages/web/src/components/typography/typography.module.css
```diff
@@ -17,91 +17,169 @@
...
-.bodyStrong {
-  font-weight: var(--font-demi-bold);
-}
-
 /* Label Styles */
 
+/* 20 / 24 */
 .labelXLarge {
-  font-size: var(--font-xl);
+  font-size: var(--unit-5);
```
schottra, line 117: Note on this: I think we want the typography components/modules to replace our usage of the font size CSS variables long term? So I opted to change these so they derive from the grid units directly, instead of using the font size variables.

audius-infra:
Preview this change https://demo.audius.co/finish-typrography-constants

packages/web/src/components/typography/typography.module.css
dylanjeffers: Stellar job and makes sense using the units.

dylanjeffers (approved):
Stellar job and makes sense using units

PR #3684, opened 2023-06-30
[PAY-1505] USDC-gated mobile track tiles

dharit-tan:
### Description
Mobile track tile changes to support USDC-gated premium content. Feature-flagged.

### Dragons

`dogEarType` logic is getting hairy, gonna follow up with refactor of that stuff.

### How Has This Been Tested?

Local ios stag
Simulator Screenshot - iPhone 14 Pro - 2023-06-30 at 12 18 08
Simulator Screenshot - iPhone 14 Pro - 2023-06-30 at 12 15 22


audius-infra:
Preview this change https://demo.audius.co/rt-usdc-tile

packages/mobile/src/assets/images/iconCart.svg
schottra, line 1: white fill by default?

packages/mobile/src/components/details-tile/DetailsTile.tsx
```diff
@@ -266,10 +268,12 @@ export const DetailsTile = ({
 
     const dogEarType = showPremiumDogEar
       ? isOwner
-        ? premiumConditions.nft_collection
-          ? DogEarType.COLLECTIBLE_GATED
-          : DogEarType.SPECIAL_ACCESS
-        : DogEarType.LOCKED
+        ? isUSDCEnabled && premiumConditions.usdc_purchase
```
schottra, line 271: Was this supposed to set the shopping cart icon? It appears to be following through to the existing ternary if usdc is enabled and the premium condition has usdc?
Or am I misreading it?

packages/mobile/src/components/lineup-tile/LineupTile.tsx
```diff
@@ -68,15 +71,16 @@ export const LineupTile = ({
...
-  const dispatch = useDispatch()
 
   const showPremiumDogEar =
     premiumConditions &&
     (isOwner || !doesUserHaveAccess) &&
     !(showArtistPick && isArtistPick)
 
   const dogEarType = showPremiumDogEar
-    ? premiumConditions.nft_collection
+    ? isUSDCEnabled && premiumConditions.usdc_purchase
```
schottra, line 81: I assume you're following up with changes to use the helper function?

packages/mobile/src/utils/theme.ts
```diff
@@ -57,6 +57,7 @@ export const defaultTheme = {
   staticNeutralLight2: '#AAA7B8',
   staticNeutralLight8: '#F2F2F4',
   staticAccentGreenLight1: '#23AD1A',
+  staticSpecialLightGreen1: '#13C65A',
```
schottra, line 60: nit: In terms of the way our colors are defined, this isn't a static color (i.e. logically considered to never change across themes).
So I think we probably want to omit the `static` prefix here?
There's a broader question about how we expect to map color names in the design system with color names in code. My hope was that they would be as close of an exact match as possible. ('Default/Special/Light Green' -> `colors.specialLightGreen`)

cc @dylanjeffers in case he has any thoughts or suggestions

schottra (approved):
🥇

packages/mobile/src/components/lineup-tile/LineupTile.tsx
dharit-tan: yes

packages/mobile/src/utils/theme.ts
dharit-tan: makes sense

packages/mobile/src/assets/images/iconCart.svg
dharit-tan: oh wait sry i was wrong before - default fill is #000 which is actually black.

packages/mobile/src/components/details-tile/DetailsTile.tsx
dharit-tan: whoops that's why we need the helper fn lol

PR #3685, opened 2023-06-30
Migrate `getDogEarType` to common

dharit-tan:
### Description
`dogEarType` logic was getting hairy and also we were doing the same calculations in every component that used a `DogEar`. Moved to a util fn in `common` and updated all the callsites.

`getDogEarType` should now return null if there's not an exact match on the conditions, which also makes it possible to delete all the `showPremiumContent` vars too.

### How Has This Been Tested?

Local ios + web stage
There are probably some cases that I've missed but I think it's ok.
Screenshot 2023-06-30 at 1 09 25 PM
Simulator Screenshot - iPhone 14 Pro - 2023-06-30 at 13 10 02
Screenshot 2023-06-30 at 1 10 12 PM
Screenshot 2023-06-30 at 1 10 20 PM
Screenshot 2023-06-30 at 1 10 30 PM

Screenshot 2023-06-30 at 1 12 27 PM

Screenshot 2023-06-30 at 1 12 11 PM

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/rt-dog-ear-common

packages/common/src/utils/trackUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+
+type GetDogEarTypeArgs = {
+  doesUserHaveAccess?: boolean
+  isArtistPick?: boolean
+  isOwner?: boolean
+  isUnlisted?: boolean
+  premiumConditions?: Nullable<PremiumConditions>
+}
+
+export const getDogEarType = ({
```
dylanjeffers, line 14: Since we also have dog ear for collections, can we specify this to either handle both, or just say this is track related?

packages/mobile/src/components/details-tile/DetailsTile.tsx
```diff
@@ -263,23 +256,12 @@ export const DetailsTile = ({
   }, [onPressPlay])
 
   const renderDogEar = () => {
-    const showPremiumDogEar =
```
dylanjeffers, line 266: Really smart not having to duplicate this everywhere!

dylanjeffers (approved):
Excellent work!

packages/common/src/utils/trackUtils.ts
dharit-tan: I think this already handles both - correct me if I'm wrong but I grepped `DogEar` in `web` and I believe I've changed every instance. Only saw one collection-related instance on `CollectionPage` that was only used for `HIDDEN` so that case is covered.
Maybe I should just change the file name to `dogEarUtils.ts`.

schottra (approved):
Awesome.

I have a PR ready for mobile web, so I'll verify it there when I merge in this change.

PR #3686, opened 2023-06-30
[C-2799] Disallow protected m4p file uploads

amendelsohn:
### Description

These are drm protected files. Removing them from allowable uploads. 

### How was this tested?

Upload still working on iOS simulator.


dylanjeffers (approved):
boss, anything we need to do for mobile?

amendelsohn:
> boss, anything we need to do for mobile?

Looks like we don't check file type for mobile. We could move the whitelist to common and check against it there as well.

packages/common/src/utils/uploadConstants.ts
```diff
@@ -0,0 +1,47 @@
+export const ALLOWED_MAX_AUDIO_SIZE_BYTES = 250 * 1000 * 1000
+
+export const ALLOWED_AUDIO_FILE_EXTENSIONS = [
+  'mp2',
+  'mp3',
+  // mp4
+  'mp4',
+  'm4a',
```
amendelsohn, line 8: moved to common and removed `'m4p'`

packages/mobile/src/screens/upload-screen/utils/processTrackFile.ts
```diff
@@ -1,16 +1,32 @@
...
-
 export const processTrackFile = (
   trackFile: DocumentPickerResponse
 ): UploadTrack => {
-  const { name, size, fileCopyUri, uri } = trackFile
+  const { name, size, fileCopyUri, uri, type } = trackFile
   if (size && size > ALLOWED_MAX_AUDIO_SIZE_BYTES) {
     throw new Error('File too large')
   }
+  // Check file extension (heuristic for failure)
```
dylanjeffers, line 17: nice adds dude

PR #3687, opened 2023-06-30
[C-2465] Improve audius-query cache selection performance

dylanjeffers:
### Description

Improves query cache performance by checking denormalized result with `isEqual`. This is a temporary fix until we can update `denormalize` function to properly select from cache.


packages/common/src/audius-query/createApi.ts
```diff
@@ -143,7 +142,7 @@ const useQueryState = <Args, Data>(
   reducerPath: string,
   endpointName: string,
   endpoint: EndpointConfig<Args, Data>
-) => {
+): Nullable<PerKeyState<any> & { isInitialValue?: boolean }> => {
```
amendelsohn, line 145: does `PerKeyState<Data>` work? Maybe not cause it's like `PerKeyState<Normalized<Data>` really

packages/common/tsconfig.json
```diff
@@ -3,6 +3,7 @@
     "outDir": "dist",
     "baseUrl": ".",
     "paths": {
+      "audius-query": ["src/audius-query/index.ts"],
```
amendelsohn, line 6: what does this do?

amendelsohn (approved):
So glad this works!

audius-infra:
Preview this change https://demo.audius.co/dj-fix-audius-query-cache-performance

packages/common/src/audius-query/createApi.ts
dylanjeffers: Ah i see i can try that, the main thing was complaining about isInitialValue not being a port of PerStateKey

packages/common/tsconfig.json
dylanjeffers: makes absolute imports possible, which i kinda like with audius-query cause it's kinda like it's own package

amendelsohn: OH that's cool

PR #3689, opened 2023-06-30
[C-2478 C-2480] Add suggestedTracks

dylanjeffers:
### Description

Adds favorites and suggestedTracks api
Adds web-ui

image

TODO: copy coming from ux


audius-infra:
Preview this change https://demo.audius.co/dj-c-2465-suggested-tracks

packages/common/src/api/suggestedTracks.ts
```diff
@@ -0,0 +1,31 @@
...
+  const { data: favoritedTracks } = useGetFavoritedTrackList(
+    { currentUserId },
+    { disabled: !currentUserId }
+  )
+
+  const favoritedTracksSample = useMemo(() => {
+    return sampleSize(favoritedTracks, 5)
+  }, [favoritedTracks])
+
+  return useGetTracksByIds(
```
sliptype, line 22: 🔥

packages/web/src/pages/collection-page/components/desktop/SuggestedTracks.tsx
```diff
@@ -0,0 +1,87 @@
...
+    <Tile className={styles.root} elevation='mid'>
+      <div className={styles.heading}>
+        <h4 className={styles.title}>{messages.title}</h4>
+        <p className={styles.description}>{messages.description}</p>
+      </div>
+      <ul>
+        {divider}
+        {!suggestedTracks && status === Status.LOADING ? (
+          <LoadingSpinner className={styles.loading} />
+        ) : null}
```
sliptype, line 68: Is there an empty state (if the user doesn't have any favorited tracks)?

sliptype (approved):
Good stuff 🙌

packages/web/src/pages/collection-page/components/desktop/SuggestedTracks.tsx
dylanjeffers: yes in that case we should pull from trending which will be done in follow up

PR #3692, opened 2023-07-03
[PAY-1556] Update mobile web gated content experience for TrackTiles

schottra:
### Description
This brings the mobile web TrackTile into line with the latest mocks for gated content. It was a bit of an overhaul, but my goal was to clean this up and reuse some of the logic/styling/layout from the desktop experience. Long term, it would be nice to merge a lot of this code by switching to `<Tile />`.

* Moved shared messages and layout logic from desktop component so I could reuse them here
* Moved premium content elements around to match mocks
* `BottomButtons` component had a lot of inline function for rendering which weren't necessary. Cleaned that function up so it was doing less at render time (Felt mildly important as we render potentially dozens of TrackTiles at once)
* Cleaned up / refactored CSS for `TrackTile` and `BottomButtons` components. Replaced a lot of individual element margins with container padding and flex gaps.
* Fixed a subtle issue with showing dog ears on tracks while they were still loading, due to usage of `isBuffering` when we should be using `isLoading`
* Converted as many hard-coded pixel values to grid units as I could
* Cleaned up unused classes in TrackTile css module
* Updated crown icon with new SVG which is square
* Added "semantic" text name variants. The Harmony Figma (as well as some of our feature artboards) use semantic color names for some elements. These are designed to map to palette color names. In cases where designs specify a semantic name, I wanted devs to be able to use those names directly, instead of having to look up what palette color that name maps to.

### Dragons
Changes are mostly contained to mobile web code, so this should be relatively safe

### How Has This Been Tested?
Tested locally in Chrome mobile emulator

N/A

### Feature Flags ###
N/A

### Screenshots ###

Locked
image

---

Unlocked
image

---

Unlocked (Artist pick)
image

---

Artist View
image

---

Trending Page
image

---

Trending Page (Co-sign)
image


audius-infra:
Preview this change https://demo.audius.co/pay-1556-mobile-web-track-tile-updates

dylanjeffers (approved):
Wow massive lift, so many styles.

packages/web/src/components/track/trackTileMessages.ts
```diff
@@ -0,0 +1,13 @@
+import { pluralize } from '@audius/common'
+
+export const messages = {
```
piazzatron, line 3: Gooood looks pulling this out

piazzatron (commented):
Amazing campground cleanup! We should do a demo/lunch 'n learn around your CSS refactor approach here (using gap instead of individual margins, etc) :)

piazzatron (approved):

PR #3693, opened 2023-07-03
Fetch chats on app foreground

piazzatron:
### Description

There was a bug where our nice new `fetchLatestMessages` action/saga wouldn't fire if we backgrounded the app on a chat page and then foregrounded it onto the same chat page, because we wouldn't remount the component + trigger the useEffect that dispatches the action.

This change also listens to app state to trigger fetching new messages.

### Dragons

### How Has This Been Tested?

On device, turned off websockets, looked at logs. 

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/screens/chat-screen/ChatScreen.tsx
+++ b/packages/mobile/src/screens/chat-screen/ChatScreen.tsx
@@ -1,3 +1,4 @@
+import type { Dispatch } from 'react'
 import { useCallback, useEffect, useMemo, useRef } from 'react'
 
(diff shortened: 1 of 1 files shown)


dharit-tan (approved):
nice find!

rickyrombo:
@piazzatron Would using `useFocusEffect` have helped here?

piazzatron:
> @piazzatron Would using `useFocusEffect` have helped here?

I always forget about this. Yeah this sounds better, I'll switch it over

PR #3695, opened 2023-07-03
Fix track screen artist name font weight

dharit-tan:
### Description
Noticed the artist name was looking a little thin. Not sure when this changed but simple fix.

### Dragons

### How Has This Been Tested?

before:
Simulator Screenshot - iPhone 14 Pro - 2023-07-03 at 14 03 50

after:
Simulator Screenshot - iPhone 14 Pro - 2023-07-03 at 14 04 10

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/details-tile/DetailsTile.tsx
+++ b/packages/mobile/src/components/details-tile/DetailsTile.tsx
@@ -101,6 +101,7 @@ const useStyles = makeStyles(({ palette, spacing, typography }) => ({
   },
 
   artist: {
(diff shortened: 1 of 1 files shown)


dylanjeffers (changes requested):
yo thanks for doing this, can you actually update the text component to be the non-deprecated one, i think that comes with medium by default :)

you can also just do color='secondary' on the component props too, as well as `fontSize='large'` so can remove that whole style definition

dylanjeffers (approved):
ell ya

PR #3698, opened 2023-07-03
Convert GiantTrackTile to typescript

schottra:
### Description
Converts GiantTrackTile to typescript / function component pattern to make it easier to work with for upcoming gated content changes.

* Moved all propTypes to a type interface
* Removed defaultProps. The parent component uses a helper function to generate valid default values
* Moved one state variable (artworkLoaded) to a state hook
* Updated a callback name to be consistent with pattern for other callbacks
* Moved class helper functions into locally defined functions that read props from closure instead of class instance
* Updated some child components to have properly `Nullable` props for a few cases where the values they received from `GiantTrackTile` might actually be null. The underlying code is already null-safe, the props just didn't reflect that.
* Had to add an `any` cast for the overflow menu props, as TS is currently unable to correctly determine the track variant and it was becoming a hassle to try and refactor it so the checker would be happy. I don't love this, but since this component was already working and I didn't change how any of the values are passed/computed it seems like a safe cast for now.
* Fixed type for the `onExternalLinkClicked` handler. I'm not sure _why_ we do this, but that handler passes the entire event up the component stack, as opposed to the internal version which just passes a url 😅. Again, goal was to convert with minimal functional changes. 

### Dragons
There might be very subtle behaviors that stop working due to converting from class to functional component, or due to some of the minor changes I made.

### How Has This Been Tested?
Tested on web, compared a variety of track states to the same version on staging to make sure now large regressions or inconsistencies popped up

Will test as part of full QA pass for this feature branch

### Feature Flags ###
N/A


dharit-tan (approved):
wowza tysm!!

didn't really go thru with a fine-tooth comb, maybe you could point out where the logic changes were between the two files?

schottra:
> wowza tysm!!
> 
> didn't really go thru with a fine-tooth comb, maybe you could point out where the logic changes were between the two files?

Oh yep, I can try to add comments to the important bits. There aren't really any big logic changes.

packages/web/src/components/track/GiantTrackTile.tsx
```diff
@@ -0,0 +1,655 @@
...
+  repostButtonText: 'repost',
+  repostedButtonText: 'reposted',
+  unplayed: 'Unplayed',
+  timeLeft: 'left',
+  played: 'Played',
+  generatedWithAi: 'Generated With AI',
+  actionGroupLabel: 'collection actions'
+}
+
+export type GiantTrackTileProps = {
```
schottra, line 81: These were all converted from `GiantTrackTile.propTypes`. However, that type just used things like `object` and `string` for some props which have more explicit types (i.e. `Remix`, `PremiumConditions`, `CoverArtSizes`).  Pretty much anything in here that isn't a number/string/boolean is now more strongly typed.

packages/web/src/components/track/GiantTrackTile.tsx
```diff
@@ -0,0 +1,655 @@
...
+  released,
+  repostCount,
+  saveCount,
+  playing,
+  premiumConditions,
+  tags,
+  trackId,
+  trackTitle,
+  userId
+}: GiantTrackTileProps) => {
```
schottra, line 178: This was originally the `render` function, and didn't destructure as many props because the helper functions defined on the class would read them directly from `this.props`. Now since we are defining the helpers inline here, they are read from closure and we need to pull more items off of props.

packages/web/src/components/track/GiantTrackTile.tsx
```diff
@@ -0,0 +1,655 @@
...
+    [setArtworkLoading]
+  )
+  const isLongFormContent =
+    genre === Genre.PODCASTS || genre === Genre.AUDIOBOOKS
+  const isNewPodcastControlsEnabled = getFeatureEnabled(
+    FeatureFlags.PODCAST_CONTROL_UPDATES_ENABLED,
+    FeatureFlags.PODCAST_CONTROL_UPDATES_ENABLED_FALLBACK
+  )
+
+  const renderCardTitle = (className: string) => {
```
schottra, line 191: These were all converted from class methods. They all had their direct destructuring of `this.props` removed in favor of reading them from closure.

packages/web/src/components/track/GiantTrackTile.tsx
```diff
@@ -0,0 +1,655 @@
...
+      <div className={styles.topSection}>
+        <GiantArtwork
+          trackId={trackId}
+          coverArtSizes={coverArtSizes}
+          coSign={coSign}
+          callback={onArtworkLoad}
+        />
+        <div className={styles.infoSection}>
+          <div className={styles.infoSectionHeader}>
+            {renderCardTitle(cn(fadeIn))}
```
schottra, line 503: Had to add a `cn()` here because the helper function expects a simple string classname and `fadeIn` is an object with conditional classnames meant to be passed to the `classnames` function.

packages/web/src/components/track/GiantTrackTile.tsx
```diff
@@ -0,0 +1,655 @@
...
+            aria-label={messages.actionGroupLabel}
+          >
+            {renderShareButton()}
+            {renderMakePublicButton()}
+            {doesUserHaveAccess && renderRepostButton()}
+            {doesUserHaveAccess && renderFavoriteButton()}
+            <span>
+              {/* prop types for overflow menu don't work correctly
+              so we need to cast here */}
+              <Menu {...(overflowMenu as any)}>
```
schottra, line 561: Didn't want to go down a rabbit hole of trying to fix the props definition for Menu... 😢

packages/web/src/components/track/GiantTrackTile.tsx
```diff
@@ -0,0 +1,655 @@
...
+              textLabel={messages.generatedWithAi}
+            />
+          ) : null}
+          {badge ? (
+            <Badge className={styles.badgePlacement} textLabel={badge} />
+          ) : null}
+        </div>
+      </div>
+
+      {isPremium && premiumConditions ? (
```
schottra, line 592: Added a check here on premium conditions to avoid rendering this section altogether. Technically I think the child component is safe if it isn't passed. But in `TrackTile` we follow the pattern of not rendering the section if there are no premium conditions.

packages/web/src/components/track/GiantTrackTile.tsx
```diff
@@ -0,0 +1,655 @@
...
+              labelName='credit'
+              labelValue={credits}
+            />
+          ) : null}
+        </div>
+        {description ? (
+          <Linkify
+            options={{
+              attributes: {
+                onClick: (event: React.MouseEvent<HTMLAnchorElement>) => {
```
schottra, line 628: This handler needed a more specific type.

schottra (commented):
@dharit-tan Added some comments about the major things that changed in this file

packages/web/src/components/track/GiantTrackTileProgressInfo.tsx
```diff
@@ -59,7 +59,7 @@ export const GiantTrackTileProgressInfo = ({
         </div>
       )
     }
-  } else {
-    return <p className={styles.progressText}>{messages.unplayed}</p>
   }
+
+  return <p className={styles.progressText}>{messages.unplayed}</p>
```
schottra, line 64: This might seem weird. TS was complaining that this component couldn't be used in JSX because it might return `undefined`. Lo and behold, it might!

The way the conditionals were structured, there's a rare chance we would return nothing from this function (though it shouldn't be likely), which would be an error in React.

audius-infra:
Preview this change https://demo.audius.co/convert-giant-track-tile

PR #3704, opened 2023-07-05
[PAY-1573] Convert premium_content to discriminated union

dharit-tan:
### Description
Converts `premium_content` to a discriminated union. Updates usages in `usPremiumContent` and the premium content saga. Also updates setters in web and mobile to include a `type` field.

### Dragons
Note the `as unknown as any` in `reformat`. `premium_content` comes from the backend as the old `PremiumConditions` type (no `type` field), but we treat it as if it were the new type and therefore get type errors if we try to read `nft_collection`, `follower_user_id` directly. Open to ideas here.

### How Has This Been Tested?

Local ios stage + web stage. Tested both uploading + editing collectible-gated tracks on web + mobile.

### Feature Flags ###

### DIFF
--- a/packages/common/src/models/Track.ts
+++ b/packages/common/src/models/Track.ts
@@ -10,6 +10,7 @@ import { Repost } from './Repost'
 import { StemCategory } from './Stems'
 import { Timestamped } from './Timestamped'
 import { User, UserMetadata } from './User'
+import { StringUSDC } from './Wallet'
 
 type EpochTimeStamp = number
 
@@ -109,7 +110,10 @@ type PremiumConditionsTipGated = {
 
 type PremiumConditionsUSDCPurchase = {
   type: PremiumContentType.USDC_PURCHASE
-  usdc_purchase: PremiumConditionsUSDCPurchase
+  usdc_purchase: {
+    price: StringUSDC
+    slot: number
+  }
 }
 
 export type PremiumConditionsWithType =
--- a/packages/common/src/models/Track.ts
+++ b/packages/common/src/models/Track.ts
@@ -91,11 +91,15 @@ type PremiumConditionsBase = {
   usdc_purchase?: PremiumConditionsUSDCPurchase
 }
 
+// Must include undefined because during upload flow, the user could have selected
+// Collectible-gated option without specifying a collection yet. But afterweards
+// when we read PremiumConditions, nft_collection should always exist.
 type PremiumConditionsCollectibleGated = {
   type: PremiumContentType.COLLECTIBLE_GATED
   nft_collection:
     | PremiumConditionsEthNFTCollection
     | PremiumConditionsSolNFTCollection
+    | undefined
 }
 
 type PremiumConditionsFollowGated = {
--- a/packages/common/src/models/Track.ts
+++ b/packages/common/src/models/Track.ts
@@ -94,7 +94,7 @@ type PremiumConditionsBase = {
 // Must include undefined because during upload flow, the user could have selected
 // Collectible-gated option without specifying a collection yet. But afterweards
 // when we read PremiumConditions, nft_collection should always exist.
-type PremiumConditionsCollectibleGated = {
+type PremiumConditionsCollectibleGated = PremiumConditionsBase & {
   type: PremiumContentType.COLLECTIBLE_GATED
   nft_collection:
     | PremiumConditionsEthNFTCollection
@@ -102,17 +102,17 @@ type PremiumConditionsCollectibleGated = {
     | undefined
 }
 
-type PremiumConditionsFollowGated = {
+type PremiumConditionsFollowGated = PremiumConditionsBase & {
   type: PremiumContentType.FOLLOW_GATED
   follow_user_id: number
 }
 
-type PremiumConditionsTipGated = {
+type PremiumConditionsTipGated = PremiumConditionsBase & {
   type: PremiumContentType.TIP_GATED
   tip_user_id: number
 }
 
-type PremiumConditionsUSDCPurchase = {
+type PremiumConditionsUSDCPurchase = PremiumConditionsBase & {
   type: PremiumContentType.USDC_PURCHASE
   usdc_purchase: {
     price: StringUSDC
@@ -120,14 +120,11 @@ type PremiumConditionsUSDCPurchase = {
   }
 }
 
-export type PremiumConditionsWithType =
+export type PremiumConditions =
   | PremiumConditionsCollectibleGated
   | PremiumConditionsFollowGated
   | PremiumConditionsTipGated
   | PremiumConditionsUSDCPurchase
-
-export type PremiumConditions = PremiumConditionsBase &
-  PremiumConditionsWithType
 
 export type PremiumContentSignature = {
   data: string
--- a/packages/common/src/models/Track.ts
+++ b/packages/common/src/models/Track.ts
@@ -80,14 +80,33 @@ export type PremiumConditionsUSDCPurchase = {
   slot: number
 }
 
-export type PremiumConditions = {
-  nft_collection?:
-    | PremiumConditionsEthNFTCollection
-    | PremiumConditionsSolNFTCollection
-  follow_user_id?: number
-  tip_user_id?: number
-  usdc_purchase?: PremiumConditionsUSDCPurchase
+export enum PremiumContentType {
+  COLLECTIBLE_GATED = 'collectible gated',
+  FOLLOW_GATED = 'follower gated',
+  TIP_GATED = 'tip gated',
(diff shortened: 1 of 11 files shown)


audius-infra:
Preview this change https://demo.audius.co/rt-prem-content-type

audius-infra:
Preview this change https://demo.audius.co/rt-prem-content-type

packages/common/src/models/Track.ts
```diff
@@ -89,6 +91,40 @@ export type PremiumConditions = {
...
+  }
+}
+
+export type PremiumConditionsWithType =
+  | PremiumConditionsCollectibleGated
+  | PremiumConditionsFollowGated
+  | PremiumConditionsTipGated
+  | PremiumConditionsUSDCPurchase
+
+export type PremiumConditions = PremiumConditionsBase &
```
schottra, line 125: Does it not work if we do
```
type PremiumConditionsCollectibleGated = PremiumConditionsBase & {
  type: PremiumContentType.COLLECTIBLE_GATED
  nft_collection:
    | PremiumConditionsEthNFTCollection
    | PremiumConditionsSolNFTCollection
}
```
for each of the types and then just export a union of those types?
```
export type PremiumConditions =
  | PremiumConditionsCollectibleGated
  | PremiumConditionsFollowGated
  | PremiumConditionsTipGated
  | PremiumConditionsUSDCPurchase
```

My thinking was that it should never be possible to have a premium conditions object exist and _not_ have at least one of the conditions present. Seems like it would make some of the other logic easier.

Or is the issue that the response coming back from the server uses the same type and won't have the `type` field?

packages/web/src/common/store/cache/tracks/utils/reformat.ts
```diff
@@ -61,6 +66,31 @@ const setDefaultFolloweeSaves = <T extends TrackMetadata>(track: T) => {
   }
 }
 
+const setTypedPremiumConditions = <T extends TrackMetadata>(track: T) => {
+  const premium_conditions = track.premium_conditions as any
```
schottra, line 70: Yeah, I think since we're using the same type for the metadata in both the response and the reformatted version, we probably can't get around typecasting here.

This might be a place where we are supposed to use `unknown` instead of `any`, to indicate that the type isn't safe and we need to check it before using.

But otherwise, our only recourse would be to change the type used at the api layer to be different from `TrackMetadata` (I believe that would be in `retrieveTracks`). That seems _wayyyy_ beyond scope of what we're doing here and probably requires more planning.  So I think I'm okay with this for now.

cc @dylanjeffers @sliptype (We should consider if we want to separate the types for api responses vs. post-processing types)

packages/web/src/common/store/cache/tracks/utils/reformat.ts
```diff
@@ -61,6 +66,31 @@ const setDefaultFolloweeSaves = <T extends TrackMetadata>(track: T) => {
...
+  if (premium_conditions?.nft_collection != null) {
+    type = PremiumContentType.COLLECTIBLE_GATED
+  } else if (premium_conditions?.follow_user_id != null) {
+    type = PremiumContentType.FOLLOW_GATED
+  } else if (premium_conditions?.tip_user_id != null) {
+    type = PremiumContentType.TIP_GATED
+  } else if (premium_conditions?.usdc_purchase != null) {
+    type = PremiumContentType.USDC_PURCHASE
+  } else {
+    type = null
```
schottra, line 81: Oof, ideally we shouldn't need to do this.
I wonder if it's worth updating `PremiumConditionsBase` to a _different_ discriminated union where at least one of the nested properties is guaranteed to exist if the object exists at all? 

```
type PremiumConditionsBase = 
 | {
      nft_collection:
      | PremiumConditionsEthNFTCollection
      | PremiumConditionsSolNFTCollection 
  } 
  | { follow_user_id: number }
  | { tip_user_id: number }
  | { usdc_purchase: PremiumConditionsUSDCPurchase }
```
Essentially we're saying "if this conditions object exists at all, one and only one of the fields will have a value".

Is that a crazy idea? 😅

schottra (approved):
Thanks for digging in and doing this. I like us moving in a direction of strong, safe, clear types.

packages/common/src/store/premium-content/sagas.ts
```diff
@@ -233,8 +238,9 @@ function* handleSpecialAccessTrackSubscriptions(tracks: Track[]) {
     }
 
     const hasNoSignature = !premiumContentSignature
-    const isFollowGated = !!premiumConditions?.follow_user_id
-    const isTipGated = !!premiumConditions?.tip_user_id
+    const isFollowGated =
+      premiumConditions.type === PremiumContentType.FOLLOW_GATED
```
sddioulde, line 242: should this line and the one below it have the ? i.e. `premiumConditions?.type === PremiumContentType.FOLLOW_GATED` ?

packages/mobile/src/screens/edit-track-screen/components/CollectibleGatedAvailability.tsx
```diff
@@ -181,7 +181,12 @@ export const CollectibleGatedAvailability = ({
       setTrackAvailabilityFields(
         {
           is_premium: true,
-          premium_conditions: { nft_collection: selectedNFTCollection },
+          premium_conditions: selectedNFTCollection
```
sddioulde, line 184: hmm this might need to be
`premium_conditions: { type: PremiumContentType.COLLECTIBLE_GATED, nft_collection: selectedNFTCollection }`

packages/common/src/models/Track.ts
dharit-tan: think this works!

packages/web/src/common/store/cache/tracks/utils/reformat.ts
dharit-tan: I think this is the right idea but it's giving me a million type errors, the consumers aren't set up on those assumptions. Since we're trying to migrate away from the old world anyway I think this is probably fine? The `PremiumConditionsBase` needs to just match exactly what the old world had.

audius-infra:
Preview this change https://demo.audius.co/rt-prem-content-type

packages/web/src/common/store/cache/tracks/utils/reformat.ts
dharit-tan: Trying to think of a better way. Since `type` doesn't come from the backend and we're using `as any` it's hard to guarantee that at least one of the fields exist?

packages/common/src/models/Track.ts
```diff
@@ -89,6 +91,41 @@ export type PremiumConditions = {
   usdc_purchase?: PremiumConditionsUSDCPurchase
 }
 
+// Must include undefined because during upload flow, the user could have selected
+// Collectible-gated option without specifying a collection yet. But afterweards
+// when we read PremiumConditions, nft_collection should always exist.
+type PremiumConditionsCollectibleGated = PremiumConditionsBase & {
```
piazzatron, line 97: Hmm... feels sort of wrong to have the discrimination unions have optional types from the other cases. One could have a `USDC_PURCHASE` type with a possible defined `nft_collection` field, which half-way defeats the purpose of this discriminated union imo. Down to brainstorm a way to avoid this

packages/web/src/common/store/cache/tracks/utils/reformat.ts
piazzatron: My 2 cents: I don't mind the null check here since it's guaranteeing we get either a properly typed premium_conditions or none at all returned, due to the conditional

piazzatron (commented):
Thanks for taking this on, overall looking great and a big improvement from what we had before! Only real note is figuring out how to exclude the optional fields in the PremiumConditions union

packages/common/src/models/Track.ts
dharit-tan: This is only a migratory step - ideally we convert all consumers to check against `type` instead of existence of `nft_collection`, `follower_user_id` etc and then can come back and delete `PremiumConditionsBase`.
Or are you thinking there's a way to avoid this even in the intermediary state?

audius-infra:
Preview this change https://demo.audius.co/rt-prem-content-type

packages/common/src/store/premium-content/sagas.ts
```diff
@@ -544,8 +552,8 @@ function* updateSpecialAccessTracks(
     } = cachedTracks[id]
     const isGated =
       gate === 'follow'
-        ? premiumConditions?.follow_user_id
-        : premiumConditions?.tip_user_id
+        ? premiumConditions?.type === PremiumContentType.FOLLOW_GATED
```
dharit-tan, line 555: bug

PR #3709, opened 2023-07-06
[PAY-1566] Adds support for passing color directly to primary button

schottra:
### Description
This adds the ability to change the color of a primary button through a single property, with the appropriate hover and pressed states generated automatically. The allowable color names are restricted to a limited set defined in the base theme and exported as a string union. The component takes care of converting the color name to the appropriate css variable name and then the style is applied through css variable inheritance on the button.

For example, `<Button color='accentBlue' />` will result in the button using `var(--accent-blue)` as its background color.

The hover and pressed states have been updated to use overlays if the `color` prop is passed, with no overlay element if the color element is omitted.

The background color is set through a CSS variable `--button-color`, which defaults to `var(--primary)` if the variable has no value. 

Setting the color in the component is achieved by converting a camelCased color name into its corresponding CSS variable name and then setting the `--button-color` variable on the element via `style`, using the form `--button-color: var(--accent-blue)`. By using string substitution and keeping the `var`, as opposed to converting the color directly to a hex value, we maintain support for changing the colors via a theme.

So the end result is that if you pass `accentBlue` as the color, whatever you've defined at your root as `--accent-blue` will be used when applying the styles. Neat.

Also worth noting that I added a `shadows.css` module which contains the default set of shadows being used in modern versions of components. These can/should be used going forward to apply shadows from our design system to stems/web components.

### Dragons
Shouldn't be any issues as the styling behavior only changes when a `color` is passed, and that should only be in new code that opts in to using this method.

One note is that we previously decided to disable preservation of most CSS custom properties in our rollup config. I'm not entirely sure _why_ we did it. But to minimize surface impact, I chose to just add the button module to the short list of preserved modules instead of enabling preservation globally.

### How Has This Been Tested?
Tested locally in storybook and against a feature branch where I intend to use this functionality.

N/A

### Feature Flags
N/A

### Screenshots
Figma design
image

---

Button without interaction
image

---

Hovered (white overlay, 10% opacity)
image

---

Pressed (black overlay, 20% opacity)
image


packages/stems/src/components/Button/Button.module.css
```diff
@@ -95,19 +108,25 @@
   font-weight: var(--font-bold);
 }
 
-.primary .icon g path,
+.primary .icon path,
```
schottra, line 111: I think this was originally an oversight. Not that due to descendant selector behavior, this will still match elements which are nested under a `g`. But it was failing to correctly set the fill on some icons which do _not_ contain a `g` element.

packages/stems/src/components/Button/Button.module.css
```diff
@@ -81,12 +82,24 @@
...
 }
 
+.button .overlay {
+  background-color: var(--overlay-color, transparent);
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  opacity: var(--overlay-opacity, 0);
```
schottra, line 92: CSS variable inheritance is neat. If you tell an element with the `.overlay` class to consume a variable called `--overlay-opacity`, it will use the nearest value defined either on itself or its ancestors.

In the button hover/active classes below, we set values for this variable. And since those get applied to the parent `button` element, this element will inherit them and use it. 🎉

packages/stems/src/styles/types.ts
```diff
@@ -0,0 +1,20 @@
...
+ *   width: '10px',
+ *   '--custom-prop': 'var(--custom-var-value)
+ * }
+ *
+ * return <button style={style as CSSProperties} />
+ * ```
+ */
+export type CSSCustomProperties = CSSProperties & {
+  // Allow any CSS Custom Properties
+  [index: `--${string}`]: any
```
dylanjeffers, line 19: could it possibly just refer to `ColorValue` instead of all strings

packages/stems/src/assets/styles/shadows.css
```diff
@@ -0,0 +1,5 @@
+:root, root {
```
dylanjeffers, line 1: nice, i notcited we have this for the tile element in web, is this something we can reuse, or not the same for this?

dylanjeffers (approved):
seriously cool styling ninja stuff, great job

audius-infra:
Preview this change https://demo.audius.co/primary-button-color-overrides

packages/stems/src/assets/styles/shadows.css
schottra: Yeah, the idea is going forward you should be able to just do `var(--near)` for a shadow anywhere in stems or web and it'll pick this up.

packages/stems/src/styles/types.ts
```diff
@@ -0,0 +1,20 @@
+import { CSSProperties } from 'react'
+
+/** Type override to allow usage of custom properties in the
+ * `style' prop. We're using this instead of augmenting the csstype module
+ * because that currently breaks types for some third-party modules. Note that
```
sliptype, line 5: 👀 what does it break?

sliptype (approved):
This is so awesome! I can't help but think how much easier things would be in css-in-js but really great work making css modules do this in such a clean way!

audius-infra:
Preview this change https://demo.audius.co/primary-button-color-overrides

packages/stems/src/styles/types.ts
schottra: The normal strategy is to [augment the csstype module](https://github.com/frenic/csstype#what-should-i-do-when-i-get-type-errors) to allow custom properties. That type then gets inherited by any third-party module that derives from `CSS.Properties`. React's `CSSProperties` is included.

When I did that, it broke [this call to `useTransition`](https://github.com/AudiusProject/audius-client/blob/main/packages/stems/src/components/Modal/Modal.tsx#L194). The generic used for that hook expects the values of `from`, `to`, etc to be extensions of `CSSProperties`. However, it believed that `transform` wasn't a valid property. I explored for a while and couldn't find a module augmentation solution that worked without breaking some types that depended on `CSSProperties`. It seemed to me like the types weren't being merged? But even when I tried copying the `React.CSSProperties` definition and extending it to include custom properties, things still didn't work.

So! Deep deep type resolution problem that didn't feel worth the time down the rabbit hole.

sliptype: Interesting, surely there is a way to augment the type and make everything happy. But good call on manually defining the type for now, maybe we can figure it out in the future

PR #3710, opened 2023-07-07
Make embed not fetch metadata from CN

theoilie:
### Description
Content Node doesn't save metadata blobs anymore, so the embed player's CN fetching just fails ([example](https://audius.co/embed/secretPANDA007/collectibles)). This removes the failing requests. Collectibles still display. This only impacts their order. If we want to fix the order in the future, we'd need Discovery to index it (currently it only indexes `has_collectibles`).

### Dragons

### How Has This Been Tested?
`npm run start:prod` wouldn't work because packages couldn't be installed. I defaulted to manual inspection to make sure this looks safe. Errors (I tried building client+libs+stems and linking libs, but no luck):

`No matching version found for @audius/stems@1.5.31`

`Module not found: Error: Can't resolve './locale' in '/Users/theo/workspace/audius-client/packages/stems/node_modules/moment/src/lib/locale'`

`Module not found: Error: Can't resolve './bindings' in './node_modules/@audius/sdk/dist'`

### Feature Flags ###


michellebrier (changes requested):
discovery does have `collectibles` in the `cid_data` table. you can query with `v1/full/cid_data/:cid` - e.g.: `https://discoveryprovider3.audius.co/v1/full/cid_data/bagaaieratnwvphvjiib2bxahqbklnmtcb5ykfmtidintgr6x2bqvidmzh5nq`

so all you need to do here is change `fetchJsonFromCID` to query that endpoint on the selected discovery node for the cid rather than race query content nodes and you should have feature parity!

theoilie:
> discovery does have `collectibles` in the `cid_data` table, its just not actually indexed into the users table. you can query the metadata blob with `v1/full/cid_data/:cid` - e.g. this is tucker's most recent metadata: `https://discoveryprovider3.audius.co/v1/full/cid_data/bagaaieratnwvphvjiib2bxahqbklnmtcb5ykfmtidintgr6x2bqvidmzh5nq`
> 
> so all you need to do here is change `fetchJsonFromCID` to query that endpoint on the selected discovery node for the cid rather than race query content nodes and you should have feature parity!

great catch!

michellebrier (approved):
nice

PR #3711, opened 2023-07-07
[PAY-1573] Convert premium content type to union

dharit-tan:
### Description
Check the type of a `premiumConditions` by using the `in` operator instead of whether or not a field like `nft_collection` exists in `premiumConditions`. Also included some typeguards without which ternaries complain.

### Dragons

### How Has This Been Tested?

Tested by uploading/attempting to edit collectible + SA tracks on both web and mobile.

### Feature Flags ###


packages/common/src/hooks/usePremiumContent.ts
```diff
@@ -88,11 +98,15 @@ export const usePremiumContentAccessMap = (tracks: Partial<Track>[]) => {
...
-  } = premiumConditions ?? {}
+  const followUserId = isPremiumContentFollowGated(premiumConditions)
+    ? premiumConditions?.follow_user_id
+    : null
+  const tipUserId = isPremiumContentTipGated(premiumConditions)
+    ? premiumConditions?.tip_user_id
+    : null
+  const nftCollection = isPremiumContentCollectibleGated(premiumConditions)
+    ? premiumConditions?.nft_collection
+    : null
```
dharit-tan, line 110: ternaries here

dharit-tan, line 110: ternaries here

piazzatron (approved):
This looks great 🔥 Few small things around the typeguards but this generally looks super good

packages/common/src/hooks/usePremiumContent.ts
dharit-tan: still getting type errors even if I null check first. gonna stick with the type guard for now but let's chat about it tmr.

packages/mobile/src/screens/edit-track-screen/components/CollectibleGatedAvailability.tsx
```diff
@@ -253,18 +258,18 @@ export const CollectibleGatedAvailability = ({
             </Text>
             <IconCaretRight fill={neutralLight4} width={16} height={16} />
           </View>
-          {premiumConditions?.nft_collection && (
+          {isPremiumContentCollectibleGated(premiumConditions) && (
             <View>
               <Text style={styles.ownersOf}>{messages.ownersOf}</Text>
               <View style={styles.collection}>
-                {premiumConditions.nft_collection.imageUrl && (
+                {premiumConditions.nft_collection?.imageUrl && (
```
schottra, line 265: Is the `?` necessary here? I would expect that the type guard function narrows it to a type here where we know that `nft_collection` is defined.

packages/web/src/components/track-availability-modal/TrackAvailabilityModal.tsx
```diff
@@ -151,11 +152,11 @@ const TrackAvailabilityModal = ({
   const isInitiallySpecialAccess =
     !isUpload &&
     !!(
-      initialPremiumConditions?.follow_user_id ||
-      initialPremiumConditions?.tip_user_id
+      'follow_user_id' in (initialPremiumConditions ?? {}) ||
```
schottra, line 155: same question here regarding using the type guard functions

schottra (approved):
Generally looks good! Just think it might need a cleanup pass.

packages/common/src/store/premium-content/sagas.ts
```diff
@@ -233,8 +237,8 @@ function* handleSpecialAccessTrackSubscriptions(tracks: Track[]) {
     }
 
     const hasNoSignature = !premiumContentSignature
-    const isFollowGated = !!premiumConditions?.follow_user_id
-    const isTipGated = !!premiumConditions?.tip_user_id
+    const isFollowGated = 'follow_user_id' in premiumConditions
```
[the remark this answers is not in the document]
dharit-tan: Not necessarily but was thinking it's more readable/consistent to just use the type guards everywhere?

packages/mobile/src/components/details-tile/DetailsTilePremiumAccess.tsx
```diff
@@ -23,17 +24,10 @@ export const DetailsTilePremiumAccess = ({
...
-  const users = useSelector((state) =>
-    getUsers(state, {
-      ids: [followUserId, tipUserId].filter(removeNullable)
-    })
-  )
-  const followee = followUserId ? users[followUserId] : null
-  const tippedUser = tipUserId ? users[tipUserId] : null
   const shouldDisplay =
-    !!premiumConditions.nft_collection || followee || tippedUser
+    isPremiumContentCollectibleGated(premiumConditions) ||
```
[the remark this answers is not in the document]
dharit-tan: Unfortunately @saliou and I discovered that `nft_collection` specifically can be undefined, bc during upload flow user could set the track to collectible gated without specifying a collection yet. Kind of awkward because at every point afterwards all the fields should be defined.

packages/mobile/src/screens/edit-track-screen/components/CollectibleGatedAvailability.tsx
dharit-tan: Unfortunately @saliou and I discovered that `nft_collection` specifically can be undefined, bc during upload flow user could set the track to collectible gated without specifying a collection yet. Kind of awkward because at every point afterwards all the fields should be defined.

packages/mobile/src/components/details-tile/DetailsTilePremiumAccess.tsx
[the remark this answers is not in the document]
dharit-tan: ah good point. Probably good to keep it this way for now tho bc `usdc_purchase` is a valid option but I haven't written the logic to exclude that yet?

packages/mobile/src/screens/edit-track-screen/components/CollectibleGatedAvailability.tsx
dharit-tan: will include comment

packages/common/src/store/premium-content/sagas.ts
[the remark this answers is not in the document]
schottra: I guess I'm just wondering why we're using the new functions in some places, but not others?

fwiw, the `in` syntax here doesn't add anything that we weren't getting with the original code on this line (i.e. it doesn't tell the compiler that code beyond this is safe to assume that premiumConditions is a `PremiumConditionsTipGated`). So preference would be to use the checker that adds type information here.

I actually find the version with a quoted property name more irksome to read, but that's just personal preference :-p.

packages/mobile/src/screens/edit-track-screen/components/CollectibleGatedAvailability.tsx
schottra: Ouch. That does make this whole change a little less useful.

packages/common/src/store/premium-content/sagas.ts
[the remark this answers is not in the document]
dharit-tan: yep yep am converting all the `in` ones to the typegaurds rn.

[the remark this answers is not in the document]
dharit-tan: good point on the `in` not adding any type safety, didn't know that! and yeah agree on readability, part of why i figured better to just go with type guards everywhere.

packages/mobile/src/screens/edit-track-screen/components/CollectibleGatedAvailability.tsx
dharit-tan: yeah it was a bummer :( couldn't think of a way around it. still gonna be useful for usdc tho.

audius-infra:
Preview this change https://demo.audius.co/rt-prem-type-2

sddioulde (approved):
slightly more verbose but conditions look cleaner imo.

my only worry is the question marks e.g. `premiumConditions.nft_collection?.name` which used to be `premiumConditions.nft_collection.name`, meaning it is theoretically possible that some component not display text when it's supposed to be. We should be careful with it and hopefully find a way to avoid it in the future. not sure just how yet

packages/common/src/hooks/usePremiumContent.ts
piazzatron: I can live with the type guards!

PR #3717, opened 2023-07-07
[PAY-1513][PAY-1557] Update track details page with new gated content designs

schottra:
### Description
This brings the track details page on both web and mobile web into line with the updated gated content designs:
* Migrated FollowButton to TS
* Updated buttons in premium content section to render in `accentBlue` using new primary button color overrides
* Updated how dog ears are rendered
* Title text is no longer colored
* Reorganize premium content description section
* Fix some padding/spacing
* Use typography styles in a few places to simplify things
* Updated to use spacing variables in CSS where appropriate
* Fixed an issue in small widths where the second half of the updated description would always wrap immediately after the artist name in the unlocked case. Now it all flows inline.
* Fixed hidden tracks not showing dog ears on mobile web

### Dragons
None

### How Has This Been Tested?
Verified locally on Chrome in both desktop and mobile emulator modes. Tested against all premium content types in both owner and user scenarios, and also with locked/unlocking/unlocked states.

N/A

### Feature Flags ###
N/A

### Screenshots

**Locked**

image
image

---

image
image

---

image
image

---

**Unlocked**
image
image

---

image
image

---

image
image

---

**Owner**
image
image

---

**Hidden**
image
image


audius-infra:
Preview this change https://demo.audius.co/PAY-1513-update-track-page

packages/web/src/components/track/PremiumTrackSection.tsx
```diff
@@ -54,15 +53,15 @@ const messages = {
...
   specialAccess: 'SPECIAL ACCESS',
-  goToCollection: 'Go To Collection',
+  goToCollection: 'Open Collection',
   sendTip: 'Send Tip',
   followArtist: 'Follow Artist',
   period: '.',
   exclamationMark: '!',
   ownCollectibleGatedPrefix:
-    'Users can unlock access by linking a wallet containing a collectible from ',
+    'Users can unlock access by linking a wallet containing a collectible from',
```
dharit-tan, line 62: intentional to remove space?

packages/web/src/components/track/PremiumTrackSection.tsx
```diff
@@ -224,10 +243,11 @@ const LockedPremiumTrackSection = ({
     if (premiumConditions.nft_collection) {
       return (
         <Button
+          color='accentBlue'
```
dharit-tan, line 246: sry did not check out your other PR for passing colors down to stems components, just double checking that this is what we wanna do rather than passing `palette.accentBlue`?

dharit-tan (approved):
sweet!

packages/web/src/components/track/PremiumTrackSection.tsx
schottra: I don't think `palette` is a thing in web, since we use CSS variables for colors. This is about as close as we can get.
Though I think if we wanted to unify it on mobile, we would do a similar approach, but the allowable values will be any of the property names that exist on the `palette` object.

packages/web/src/components/track/PremiumTrackSection.tsx
schottra: Oops. In fact, no. I neglected to test the owner case for NFT collections and it definitely needs the space. Thanks for catching this.

PR #3718, opened 2023-07-07
Call the DN selection callback in AudiusBackend if using a cached DN

rickyrombo:
### Description

Our endless retry logic happened again. I so want to rework this AudiusAPIClient and AudiusBackend nonsense but in the meantime, just added this minor hack to allow cached DNs to work.

It calls the AudiusBackend DN change listener if the config for DN selector has an initialSelectedNode.

### Dragons

Relies on using a private member of DN selector, `config`, and as such requires a `@ts-ignore`.

### How Has This Been Tested?

### Feature Flags ###


dylanjeffers (approved):
the never ending bug :/ thanks for handling

PR #3722, opened 2023-07-10
[C-2785] Update collection screen focus effect to fetch lineup

Kyle-Shanks:
### Description
Small add to update the collection screen to fetch the lineup metadata whenever the screen is focused

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/web/src/common/store/pages/collection/sagas.js
+++ b/packages/web/src/common/store/pages/collection/sagas.js
@@ -27,7 +27,7 @@ const { fetchCollection, fetchCollectionSucceeded, fetchCollectionFailed } =
 
 function* watchFetchCollection() {
   yield takeLatest(collectionActions.FETCH_COLLECTION, function* (action) {
-    const { id: collectionId, permalink } = action
+    const { id: collectionId, permalink, fetchLineup } = action
     let retrievedCollections
     if (permalink) {
       retrievedCollections = yield call(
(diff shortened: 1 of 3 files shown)


packages/web/src/common/store/pages/collection/sagas.js
```diff
@@ -27,7 +27,7 @@ const { fetchCollection, fetchCollectionSucceeded, fetchCollectionFailed } =
 
 function* watchFetchCollection() {
   yield takeLatest(collectionActions.FETCH_COLLECTION, function* (action) {
-    const { id: collectionId, permalink } = action
+    const { id: collectionId, permalink, fetchLineup } = action
```
dylanjeffers, line 30: oh interesting, i see! okay hmm could we not have a useEffect that just calls fetchLineupMetada like we have with desktop? i honestly like what youve done here though which just ensures fechLineup gets called in addition to fetchCollection :)

dylanjeffers (approved):
awesome addition

audius-infra:
Preview this change https://demo.audius.co/kj-Update-collection-screen-focus

packages/web/src/common/store/pages/collection/sagas.js
Kyle-Shanks: for some reason there was a prefix error so I did it this way. controls the flow this way which i think is what fixes the error

PR #3723, opened 2023-07-10
[C-2809] Remove user from image hooks

dylanjeffers:
### Description

Now that we dont need user.creator_node_endpoint, a lot of the mobile image hooks dont need to pull in the user. this is great cause it simplifies the api and also reduces rerender count since we dont listen to user updates.


theoilie (approved):
🔥

PR #3724, opened 2023-07-10
[PAY-1559][PAY-1560] Adds USDC variants for track tiles on web/mobile web

schottra:
### Description
Adds the relevant variants for each state for track tiles in web and mobile web. This mostly amounted to adding logic to render things in green when the premium condition is usdc_purchase. Also included a little bit of logic to convert a wei USDC price string to a formatted dollar amount (rounds to nearest cent).

### Dragons
Formatting logic is a little basic at the moment. We will need to decide what our limits are for prices and make sure we correctly render all the edge cases.

### How Has This Been Tested?
Tested locally in Chrome (both web and emulated mobile)

N/A

### Feature Flags ###
N/A

### Screenshots

Screenshot 2023-07-10 at 5 39 50 PM

---

Screenshot 2023-07-10 at 5 39 59 PM

---

Screenshot 2023-07-10 at 6 18 26 PM

---

Screenshot 2023-07-10 at 5 37 19 PM

---

Screenshot 2023-07-10 at 5 37 27 PM


packages/web/src/components/track/LockedStatusBadge.tsx
```diff
@@ -5,14 +5,22 @@ import styles from './LockedStatusBadge.module.css'
 
 export type LockedStatusBadgeProps = {
   locked: boolean
+  variant: 'premium' | 'gated'
```
dharit-tan, line 8: naming a bit confusing since we use premium content and gated content somewhat interchangeably prior to usdc. would it make more sense to just use `usdc` and `gated`?

packages/web/src/components/track/desktop/TrackTile.tsx
```diff
@@ -153,6 +156,9 @@ const TrackTile = ({
   const isLongFormContent =
     genre === Genre.PODCASTS || genre === Genre.AUDIOBOOKS
 
+  const isPurchase =
+    premiumConditions && isPremiumContentUSDCPurchaseGated(premiumConditions)
```
dharit-tan, line 160: don't think we need the `premiumConditions` check here?

packages/web/src/components/track/desktop/BottomRow.tsx
```diff
@@ -101,21 +109,14 @@ export const BottomRow = ({
...
-            {messages.unlocking}
-          </div>
-        ) : (
-          <div className={styles.premiumContent}>
-            <IconLock />
-            {messages.locked}
-          </div>
-        )}
-        {!isLoading ? <div>{rightActions}</div> : null}
+        <PremiumConditionsPill
```
dharit-tan, line 115: nice

dharit-tan (approved):
lookin good

audius-infra:
Preview this change https://demo.audius.co/pay-1560-usdc-tiles

packages/web/src/components/track/PremiumConditionsPill.tsx
```diff
@@ -0,0 +1,47 @@
...
+  } else {
+    message = isPurchase
+      ? `$${formatUSDCWeiToUSDString(premiumConditions.usdc_purchase.price)}`
+      : messages.locked
+  }
+
+  const colorStyle = isPurchase ? styles.premiumContent : styles.gatedContent
+
+  return (
+    <div className={cn(styles.container, colorStyle)}>
```
rickyrombo, line 42: Is this not a button? I know it wasn't before but it seems like one now the way it looks...

packages/web/src/components/track/LockedStatusBadge.tsx
schottra: Thanks for context, @rickyrombo !
My thinking was indeed that we are making a shift to referring to purchasable content as "premium". But we do have a lot of lingering instances of "premium" referring to gated content in general.
This was my attempt to start using the new terminology.

Elsewhere in this PR I think I also mixed in the word "purchased" as an indication of usdc-gated content. So I at least need to pick one or the other going forward.

I think it probably deserves a discussion as a team on whether we want to actually make the change in code, or just leave a mapping between code->display strings and know that we refer to "premium" content internally as "purchasable" content.

packages/web/src/components/track/desktop/TrackTile.tsx
schottra: I would think so too! But it turns out on desktop, `premiumConditions` can be `undefined`, which isn't a valid input for the helper. I can, of course, update the helper to potentially accept undefined. Something about a function accepting undefined as a valid input has always bugged me :-)

packages/common/src/utils/wallet.ts
```diff
@@ -134,3 +133,26 @@ export const convertWeiToWAudio = (amount: BN) => {
...
+const BN_USDC_WEI = new BN('1000000')
+
+/** Formats a USDC wei string (full precision) to a fixed string suitable for
+display as a dollar amount. Note: WILL lose precision by rounding to nearest cent */
+export const formatUSDCWeiToUSDString = (amount: StringUSDC, precision = 2) => {
+  // Since we only need two digits of precision, we will multiply up by 100
+  // with BN, divide by $1 Wei, and then convert to JS number and divide back down
+  // before formatting to two decimal places.
+  const converted =
+    new BN(amount).muln(100).divRound(BN_USDC_WEI).toNumber() / 100
```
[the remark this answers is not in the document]
schottra: Yeah I figured we would discuss this more holistically with Product at the next sync up. Realistically, we won't allow setting sub-cent values through the UI, so this would never matter. But you could always write an update to chain from elsewhere. And we would just need to be clear about the policy in those cases.

FWIW, I think ceiling is probably the best approach, as it allows us to follow a policy where you pay _exactly_ the amount that the track costs down to the last Wei. That would make access-checking easy, since we don't have to ensure that rounding logic exists and matches in all the places.

packages/web/src/components/track/desktop/TrackTile.tsx
schottra: Updated my copy of the function to allow undefined. Will update the shared version when I merge the feature branch into this one.

packages/web/src/components/track/PremiumConditionsPill.tsx
schottra: It's not clickable, per se. You can click pretty much anywhere on the track tile to trigger the "how to unlock" modal. So this is acting more like a label than a button.

packages/common/src/utils/wallet.ts
```diff
@@ -134,3 +133,26 @@ export const convertWeiToWAudio = (amount: BN) => {
   const decimals = WEI_DECIMALS - SPL_DECIMALS
   return amount.div(new BN('1'.padEnd(decimals + 1, '0')))
 }
+
+/** USDC Utils */
+const BN_USDC_WEI = new BN('1000000')
+
+/** Formats a USDC wei string (full precision) to a fixed string suitable for
+display as a dollar amount. Note: WILL lose precision by rounding to nearest cent */
+export const formatUSDCWeiToUSDString = (amount: StringUSDC, precision = 2) => {
```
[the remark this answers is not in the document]
schottra: Oh yeah, that's easy. I think if we're doing rounding, we need to process it a little ahead of time.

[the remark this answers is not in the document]
schottra: I updated the code to add a rounding utility (since we will probably need that for some of the saga work) and then use the utility for formatting commas. Current behavior is to round up to the nearest cent and display that. We can discuss with Product and come back to change it later if needed.

audius-infra:
Preview this change https://demo.audius.co/pay-1560-usdc-tiles

packages/common/src/utils/wallet.ts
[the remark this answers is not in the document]
rickyrombo: Yeah despite what I said in standup let's double check with product

PR #3727, opened 2023-07-11
[PAY-1561][PAY-1558] Implement USDC variants on track details page

schottra:
### Description
Updates all of the code paths for web and mobile web track details and locked content modal to render the USDC variants for locked, unlocking, unlocked, and owner states.

* Added `Large` variant to Button
* Updated Secondary Button style to match new Harmony versions
  * It's only used in two places at the moment, so should be safe
* Updated Button stories to render all sizes to make things easier to compare
* Cleaned up GiantTrackTile and TrackHeader
  * Both now correctly show the Gated/Premium track section based on type/unlocked/owner status
  * TrackHeader got an overall to make the layout simpler and correctly order the sections (we had two copies of <PremiumTrackSection /> for no apparent reason)
* Added Preview button to both versions of the page. Will show if the user is owner or doesn't have access yet. Currently wired up to a console log

### Dragons
None

### How Has This Been Tested?
Tested in Chrome for desktop and emulated mobile browser

N/A

### Feature Flags ###
N/A

#### Locked

Screenshot 2023-07-12 at 7 50 20 PM
Screenshot 2023-07-12 at 7 52 42 PM

---

#### Unlocked

Screenshot 2023-07-12 at 7 52 11 PM
Screenshot 2023-07-12 at 7 52 59 PM

---

#### Owner
Screenshot 2023-07-12 at 7 54 51 PM
Screenshot 2023-07-12 at 7 55 01 PM

---

#### Button storybook changes

image

### DIFF
--- a/packages/stems/src/assets/styles/colors.css
+++ b/packages/stems/src/assets/styles/colors.css
@@ -64,6 +64,11 @@ root {
 
   --special-light-green: #13c65a;
 
+  /* Semantic text */
+  --text-default: var(--neutral);
+  --text-subdued: var(--neutral-light-4);
+  --text-disabled: var(--neutral-light-7);
+
   --primary-gradient: linear-gradient(135deg, #ee00ed 0%, #c514e0 100%);
   --secondary-gradient: linear-gradient(315deg, #6b0fb3 0%, #7e1bcc 100%);
   --page-header-gradient-color-1: #5b23e1;
--- a/packages/stems/src/components/Button/Button.module.css
+++ b/packages/stems/src/components/Button/Button.module.css
@@ -32,6 +32,25 @@
   height: var(--unit-8);
 }
 
+.button.large {
+  padding: var(--unit-5) var(--unit-6);
+  height: var(--unit-15);
+  gap: var(--unit-2);
+}
+
+.button.large .textLabel {
+  font-size: var(--font-xl);
+  font-weight: var(--font-bold);
+  letter-spacing: 0.25px;
+  line-height: var(--unit-5);
+  text-transform: uppercase;
+}
+
+.button.large .icon,
+.button.large .icon.left {
+  margin-right: 0;
+}
+
 .button.tiny {
   height: var(--unit-7);
   padding-left: var(--unit-4);
@@ -43,6 +62,10 @@
 }
 
 .button.small .icon svg {
+  width: var(--unit-4);
+}
+
+.button.tiny .icon svg {
   width: var(--unit-4);
 }
 
@@ -164,13 +187,13 @@
 /* Secondary Button */
 
 .secondary {
-  border: 1px solid var(--primary-dark-1);
+  border: 1px solid var(--neutral-light-5);
   border-radius: 4px;
   background: var(--white);
 }
 
 .secondary .textLabel {
-  color: var(--primary);
+  color: var(--text-default);
   font-size: var(--font-l);
   font-weight: var(--font-bold);
 }
@@ -187,28 +210,35 @@
 
 .secondary .icon path,
 .secondary .icon use {
-  fill: var(--primary);
+  fill: var(--text-default);
 }
 
 .secondary.includeHoverAnimations:hover:enabled {
   transform: var(--grow);
-  border: 1px solid var(--primary-dark-2);
-  background: var(--primary);
+  border-color: var(--primary);
 }
 
 .secondary.includeHoverAnimations:hover:enabled .textLabel {
-  color: var(--primary-secondary-text);
+  color: var(--primary);
 }
 
 .secondary.includeHoverAnimations:hover:enabled .icon path,
 .secondary.includeHoverAnimations:hover:enabled .icon use {
-  fill: var(--primary-secondary-text);
+  fill: var(--primary);
 }
 
 .secondary.includeHoverAnimations:active:enabled {
   transform: perspective(1px) scale3d(0.99, 0.99, 0.99);
-  border: 1px solid var(--primary-dark-2);
-  background: var(--primary-dark-2);
+  border-color: var(--primary-dark-2);
+}
+
+.secondary.includeHoverAnimations:active:enabled .textLabel {
+  color: var(--primary-dark-2)
+}
+
+.secondary.includeHoverAnimations:active:enabled .icon path,
+.secondary.includeHoverAnimations:active:enabled .icon use {
+  fill: var(--primary-dark-2);
 }
 
 /* Tertiary Button */
--- a/packages/stems/src/components/Button/Button.stories.tsx
+++ b/packages/stems/src/components/Button/Button.stories.tsx
@@ -1,6 +1,7 @@
 import { Story } from '@storybook/react'
 
 import * as Icons from 'components/Icons'
+import { ButtonSize } from 'index'
 
 import { Button } from './Button'
 import { ButtonProps, Type } from './types'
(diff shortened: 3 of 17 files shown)


audius-infra:
Preview this change https://demo.audius.co/pay-1561-usdc-track-page

packages/stems/src/assets/styles/colors.css
```diff
@@ -64,6 +64,11 @@ root {
 
   --special-light-green: #13c65a;
 
+  /* Semantic text */
```
rickyrombo, line 67: yay!

packages/stems/src/components/Button/Button.module.css
```diff
@@ -32,6 +32,25 @@
   height: var(--unit-8);
 }
 
+.button.large {
```
schottra, line 35: cc @dylanjeffers For the button updates. I tried to make the changes in a way that would only have a major effect if you used the new `LARGE` size. It's a reasonable starting point for getting buttons migrated over to the new harmony standards. Happy to chat more about how we could clean this up. But it's a much bigger audit to check all the places we use buttons 😅

audius-infra:
Preview this change https://demo.audius.co/pay-1561-usdc-track-page

packages/stems/src/components/Button/Button.module.css
```diff
@@ -46,6 +65,10 @@
   width: var(--unit-4);
 }
 
+.button.tiny .icon svg {
```
schottra, line 68: Noticed with the new stories rendering all sizes that the icons on the tiny variant are actually way too big. We only use it in one place (track edit modal)

packages/stems/src/components/Button/Button.stories.tsx
```diff
@@ -17,7 +18,12 @@ const baseProps: ButtonProps = {
 }
 
 const Template: Story<ButtonProps> = (args) => (
-  <Button {...baseProps} {...args} />
+  <div style={{ alignItems: 'center', display: 'flex', gap: '16px' }}>
+    <Button {...baseProps} size={ButtonSize.TINY} {...args} />
```
rickyrombo, line 22: thanks for adding these!

packages/web/src/components/track/GiantTrackTile.tsx
```diff
@@ -187,6 +188,11 @@ export const GiantTrackTile = ({
     FeatureFlags.PODCAST_CONTROL_UPDATES_ENABLED,
     FeatureFlags.PODCAST_CONTROL_UPDATES_ENABLED_FALLBACK
   )
+  const isPurchaseGated = isPremiumContentUSDCPurchaseGated(premiumConditions)
+  const showPreview = isOwner || (isPurchaseGated && !doesUserHaveAccess)
+  const onPreview = useCallback(() => {
+    console.log('Preview Clicked')
```
rickyrombo, line 194: super nit but maybe a TODO comment (bonus points if w/ linear item)?

packages/web/src/components/track/GiantTrackTile.tsx
```diff
@@ -345,26 +351,29 @@ export const GiantTrackTile = ({
...
+    if (!shouldShow) {
+      return null
+    }
     return (
-      shouldShow && (
-        <div className={styles.listens}>
-          {listenCount === 0 ? (
-            <span className={styles.firstListen}>
-              Be the first to listen to this track!
+      <div className={styles.listens}>
```
schottra, line 361: This looks weird because I pulled the `shouldShow` logic up to an early null return. Otherwise, this is nearly unchanged with the exception of hiding the "Be the first to listen" if you're the owner (will show 0 plays instead)

packages/web/src/pages/track-page/components/mobile/TrackHeader.module.css
```diff
@@ -54,8 +51,8 @@
   font-size: var(--font-l);
   font-weight: var(--font-medium);
   line-height: 24px;
+  justify-content: center;
   text-align: center;
-  margin-bottom: 16px;
```
schottra, line 58: The general approach for TrackHeader was to remove the margins and borders from all of the individual section. Instead, I'm using `gap` on the main container to space the sections evenly, and applying a shared "withDivider" class to do top borders/padding on each section.

packages/web/src/components/track/GiantTrackTile.tsx
```diff
@@ -526,12 +535,23 @@ export const GiantTrackTile = ({
           </div>
 
           <div className={cn(styles.playSection, fadeIn)}>
-            <PlayPauseButton
-              doesUserHaveAccess={doesUserHaveAccess}
-              playing={playing}
-              onPlay={onPlay}
-              trackId={trackId}
-            />
+            {doesUserHaveAccess ? (
```
rickyrombo, line 538: This logic is slightly confusing.. but I think I parsed it correctly

- Both normal play button and preview play button if user owns the track (shouldn't preview only appear for premium content?)
- Only the normal play button if user has access
- Only the preview play button if user does not have access _and_ it's purchase gated
- Nothing if the user doesn't have access and it's not purchase gated

It looks like we should always render the play/pause button but disabled if you don't have access for other gate types, so we might be missing a condition on this, maybe `doesUserHaveAccess || !isPurchasedGated`?

Also maybe we rename `doesUserHaveAccess` to `disabled` in PlayPauseButton and just pass through the disabled state directly, as it seems like that's all it's used for inside the PlayPauseButton component? wdyt?

Lastly minor nit but anything to make this logic easier to understand (comments or refactors/renames?) would probably be helpful in the future

packages/web/src/components/track/PremiumTrackSection.tsx
```diff
@@ -112,6 +122,11 @@ const LockedPremiumTrackSection = ({
     ? FollowSource.HOW_TO_UNLOCK_MODAL
     : FollowSource.HOW_TO_UNLOCK_TRACK_PAGE
   const account = useSelector(getAccountUser)
+  const isUSDCGated = isPremiumContentUSDCPurchaseGated(premiumConditions)
```
rickyrombo, line 125: nit: Feels like we should stick to either `isPurchaseGated` or `isUSDCGated`, seeing a mix of both in this pr

rickyrombo (commented):
Got through the first half, will look at teh second half in a bit but wanted to post what I had early

packages/web/src/components/track/GiantTrackTile.tsx
schottra: Yep, good idea, can do.

packages/web/src/components/track/GiantTrackTile.tsx
schottra: I agree that it's confusing. I think you got the analysis right. And based on that I made a mistake on one of the cases (showing a disabled play button instead of no button for non-purchase gated tracks).I updated it to:
* Calculate `showPlay` and `showPreview` explicitly outside of the rendering block. Each has a comment explaining what the conditions are
* Use `disabled` in the button components instead of passing `doesUserHaveAccess`, and then deriving the former from the latter as needed.

packages/web/src/components/track/PremiumTrackSection.tsx
schottra: Updated all instances to be `isUSDCPurchaseGated`

packages/web/src/components/track/GiantTrackTile.tsx
```diff
@@ -187,6 +188,18 @@ export const GiantTrackTile = ({
...
     FeatureFlags.PODCAST_CONTROL_UPDATES_ENABLED_FALLBACK
   )
+  const isUSDCPurchaseGated =
+    isPremiumContentUSDCPurchaseGated(premiumConditions)
+  // Preview button is shown for USDC-gated tracks if user does not have access
+  // or is the owner
+  const showPreview = isUSDCPurchaseGated && (isOwner || !doesUserHaveAccess)
+  // Play button is conditionally hidden for USDC-gated tracks when the user does not have access
+  const showPlay = isUSDCPurchaseGated ? doesUserHaveAccess : true
+
```
dharit-tan, line 198: nice comments i should add those in mobile

packages/web/src/pages/track-page/components/mobile/TrackHeader.module.css
dharit-tan: this feels so satisfying :)

dharit-tan (approved):
awesome work, nice job cleaning stuff up along the way!

packages/stems/src/components/Button/Button.module.css
schottra: cc @sliptype

PR #3728, opened 2023-07-12
[C-2547] Lineup pagination fixes

amendelsohn:
### Description

- Removes `containsDeleted` which was causing the bug in remix lineup [C-2547](https://linear.app/audius/issue/C-2547/[qa]-remix-lists-failing-to-load-completely)
- Fixes `hasMore` offset logic

### Dragons

Changing feed is scary

- `containsDeleted` was from https://github.com/AudiusProject/audius-client-old/pull/1043 which we believe is no longer relevant
- hasMore logic was definitely wrong, but maybe was for a case where an absolute limit was passed instead of a pageSize limit. Need to keep an eye on this in stage.

### DIFF
--- a/packages/common/src/models/Lineup.ts
+++ b/packages/common/src/models/Lineup.ts
@@ -38,7 +38,6 @@ export type LineupState<T> = {
   page: number
   isMetadataLoading: boolean
   dedupe?: boolean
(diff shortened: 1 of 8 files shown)


dylanjeffers (approved):
wonderful improvements

PR #3733, opened 2023-07-12
Update announcement notification style and hover behavior

dharit-tan:
### Description
Bonus: removed a few unused messages that were annoying during ctrl+f

### Dragons

Now if `NotificationTile` `disabled` prop is true, hover animation is also disabled. Originally it was only `active` styles that were disabled. Only other place that uses this is `TipReceivedNotification`, don't entirely understand the purpose of this but I tested against current stage and behavior seemed the same. cc @dylanjeffers .

### How Has This Been Tested?

Local web stage

https://github.com/AudiusProject/audius-client/assets/3893871/94973579-ac76-446f-a50b-d5e46cb77f2e

### Feature Flags ###

### DIFF
--- a/packages/web/src/components/notification/Notification/AnnouncementNotification.module.css
+++ b/packages/web/src/components/notification/Notification/AnnouncementNotification.module.css
@@ -8,7 +8,6 @@
   color: var(--secondary);
   font-size: var(--font-m);
   font-weight: var(--font-medium);
-  margin-left: var(--unit-1);
 }
 
 .readMore:hover {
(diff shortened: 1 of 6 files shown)


raymondjacobson (approved):

piazzatron (approved):
Nice! Two small things I'm noticing:
- The `Read More` text has some strange 1-unit left margin that looks wrong to me - can we check if this is up to design spec, and update if not?
- If the Panel isn't clickable, we probably shouldn't have a hover animation or have `cursor: pointer`

audius-infra:
Preview this change https://demo.audius.co/rt-long-desc

audius-infra:
Preview this change https://demo.audius.co/rt-long-desc

raymondjacobson (approved):
lgtm w/ piazza's comments on https://demo.audius.co/rt-long-desc/feed

dylanjeffers (approved):
noice

PR #3743, opened 2023-07-14
Get the first storage node for o-auth not all of them

amendelsohn:
### Description

Causing bug where the url was too large as we're serializing the entire storage node list into the request


amendelsohn:
Hold on. Missed some spots.

raymondjacobson (approved):

PR #3749, opened 2023-07-14
Use Text component in mobile DMs

dharit-tan:
### Description
Figured I'd go ahead and change all the `Text` usages in mobile chats to our `Text` component, after all the incorrect fonts we caught.

Also caught an erroneous `marginTop` that was causing incorrect layout in `ChatUserListItem`.

### Dragons

### How Has This Been Tested?

Local ios stage
Simulator Screenshot - iPhone 14 Pro - 2023-07-14 at 16 27 51

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/screens/chat-screen/ChatMessageSeparator.tsx
+++ b/packages/mobile/src/screens/chat-screen/ChatMessageSeparator.tsx
@@ -1,5 +1,6 @@
-import { View, Text } from 'react-native'
+import { View } from 'react-native'
 
+import { Text } from 'app/components/core'
 import { makeStyles } from 'app/styles'
 
 const useStyles = makeStyles(({ spacing, palette, typography }) => ({
@@ -17,9 +18,6 @@ const useStyles = makeStyles(({ spacing, palette, typography }) => ({
   },
   unreadTag: {
     color: palette.white,
-    fontSize: typography.fontSize.xxs,
-    fontFamily: typography.fontByWeight.bold,
-    textTransform: 'uppercase',
     letterSpacing: 0.5,
     backgroundColor: palette.neutralLight5,
     paddingHorizontal: spacing(2),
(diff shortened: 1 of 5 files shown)


packages/mobile/src/screens/chat-screen/ChatMessageSeparator.tsx
```diff
@@ -34,7 +32,14 @@ export const ChatMessageSeparator = ({ content }: { content: string }) => {
   return (
     <View style={styles.unreadTagContainer}>
       <View style={styles.unreadSeparator} />
-      <Text style={styles.unreadTag}>{content}</Text>
+      <Text
```
schottra, line 35: Is there a default color on Text if you don't specify one? I saw below that other usage included a `color` prop.

packages/mobile/src/screens/chat-screen/ChatUserListItem.tsx
```diff
@@ -258,7 +250,15 @@ export const ChatUserListItem = ({ userId }: ChatUserListItemProps) => {
                 )}
               </View>
               {user.does_follow_current_user && canCreateChat ? (
-                <Text style={styles.followsYouTag}>{messages.followsYou}</Text>
+                <Text
+                  fontSize='xxs'
+                  weight='heavy'
+                  color='neutralLight4'
```
schottra, line 256: Worth thinking about going forward if we should add the semantic variants of text color and use those instead (i.e. `textSubdued`)

schottra (approved):
Good cleanup!

packages/mobile/src/screens/chat-screen/ChatMessageSeparator.tsx
dharit-tan: yep it defaults to `neutral`.

PR #2956, opened 2023-03-02
[PAY-980] Treat premium content as deleted in lineups if feature flags disabled

rickyrombo:
### Description

Removes premium content from lineups / store entirely if the feature flag is disabled

### Dragons

*Is there anything the reviewer should be on the lookout for? Are there any dangerous changes?*

### How Has This Been Tested?

*Please describe the tests that you ran to verify your changes. Provide repro instructions & any configuration.*

Tested on web stage by looking at feed/profile pages. Need to double check mobile.

### How will this change be monitored?

*For features that are critical or could fail silently please describe the monitoring/alerting being added.*

### Feature Flags ###

*Are all new features properly feature flagged? Describe added feature flags.*

PR #3752, opened 2023-07-15
[PAY-1578] Hide USDC-gated tracks if the feature flag is disabled

schottra:
### Description
Using a similar strategy to #2956 , inserts a step into the lineup processing that treats USDC-gated tracks as "deleted" in all cases. This should take care of all environments since it's done at the generic lineup saga level.

### Dragons
You could still navigate directly to a USDC-gated track. But I think our goal for now is to not surface these tracks into feeds/trending/artist pages.

### How Has This Been Tested?
Tested locally on web/mobile web/ simulator

N/A

### Feature Flags ###
Uses the USDC_PURCHASES feature flag. On by default in development.

### DIFF
--- a/packages/web/src/common/store/lineup/sagas.js
+++ b/packages/web/src/common/store/lineup/sagas.js
@@ -12,7 +12,9 @@ import {
   queueActions,
   playerSelectors,
   queueSelectors,
-  getContext
+  getContext,
+  FeatureFlags,
+  isPremiumContentUSDCPurchaseGated
 } from '@audius/common'
 import {
   all,
@@ -29,6 +31,7 @@ import {
 
 import { getToQueue } from 'common/store/queue/sagas'
 import { isMobileWeb } from 'common/utils/isMobileWeb'
+import { getFeatureEnabled } from 'services/remote-config/featureFlagHelpers'
 
 const { getSource, getUid, getPositions } = queueSelectors
 const { getUid: getCurrentPlayerTrackUid, getPlaying } = playerSelectors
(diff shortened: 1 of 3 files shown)


packages/web/src/common/store/lineup/sagas.js
```diff
@@ -46,6 +49,10 @@ function* filterDeletes(tracksMetadata, removeDeleted) {
   const remoteConfig = yield getContext('remoteConfigInstance')
   yield call(remoteConfig.waitForRemoteConfig)
 
+  const isUSDCGatedContentEnabled = yield getFeatureEnabled(
```
schottra, line 52: Note: In the PR I based this on, there was a `waitForRead()` at this point. Looking into what `waitForRead` is waiting on (reachability and account), I don't think there's any additional need to wait on that condition here if the original lineup processing wasn't already waiting for it.
But if others can think of a reason why we should, I'm happy to add it.

dharit-tan (approved):
awesome!

packages/web/src/pages/upload-page/fields/availability/collectible-gated/CollectibleGatedFields.tsx
```diff
@@ -106,8 +107,11 @@ export const CollectibleGatedFields = ({
   // which makes the dropdown show the placeholder.
   // Otherwise, the default value is the nft collection which was previously selected,
   // which also includes the collection image.
-  const defaultCollectionName =
-    premiumConditionsValue?.nft_collection?.name ?? ''
+  const defaultCollectionName = isPremiumContentCollectibleGated(
```
schottra, line 110: @dharit-tan You were correct on the bad merge. There are some new reads of the gating conditions on main that needed to be updated to use the union type.
Starting to wonder if this is actually useful as it requires calling multiple type checker functions each time we're trying to switch on the gate type? 😬

audius-infra:
Preview this change https://demo.audius.co/pay-1578-hide-usdc-tracks

packages/web/src/pages/upload-page/fields/availability/collectible-gated/CollectibleGatedFields.tsx
dharit-tan: yeah sigh.. it's still gives more type safety tho right?

schottra: It does! And it's useful in the cases where you're checking for one specific condition before moving forward. I think it's clunky for the handful of cases where we are handling all of the types of premium content in one place.

PR #3761, opened 2023-07-17
[C-2858] Fix phantom signing

raymondjacobson:
### Description

When trying to connect a phantom wallet, we see this error message
```
"Caught error during handleConnectNewWallet:  TypeError: Cannot read private member from an object whose class did not declare it, resetting to initial state"
```
Presumably this is because of some change in a typescript dependency, not directly because of a phantom change.

I am unsure exactly how this change fixes it, but it seems to, probably because `this` context is necessary?

### Dragons

N/a

### How Has This Been Tested?

1. Ran client against stage
2. Connected phantom wallet

https://app.amplitude.com/analytics/audius/chart/1x2oshx

### Feature Flags ###

N/a

### DIFF
--- a/packages/web/src/store/token-dashboard/signMessage.ts
+++ b/packages/web/src/store/token-dashboard/signMessage.ts
@@ -2,7 +2,8 @@ import {
   accountSelectors,
   Chain,
   tokenDashboardPageActions,
-  tokenDashboardPageSelectors
+  tokenDashboardPageSelectors,
+  PhantomProvider
 } from '@audius/common'
 import { call, put, select } from 'typed-redux-saga'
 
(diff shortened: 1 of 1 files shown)


audius-infra:
Preview this change https://demo.audius.co/rj-c-2858

packages/web/src/store/token-dashboard/signMessage.ts
```diff
@@ -24,9 +29,9 @@ export function* signMessage(connection: WalletConnection) {
     case Chain.Sol: {
       const encodedMessage = new TextEncoder().encode(message)
       const signedResponse = yield* call(
-        connection.provider.signMessage,
-        encodedMessage,
-        'utf8'
+        solSign,
```
raymondjacobson, line 32: as @dharit-tan called out `[provider, provider.signMessage]` works here too, but it doesn't correctly infer the type of `provider`

dharit-tan (approved):

packages/web/src/store/token-dashboard/signMessage.ts
dylanjeffers: Great catch

PR #3763, opened 2023-07-18
[C-2853] Fix deletedCount calculation

amendelsohn:
### Description

We're relying on `deletedCount` to calculate `hasMore`.
For any request where limit was greater than the remaining entries on discovery, we'd set deleted count to the remainder.

### Dragons

Trying to find out if we rely on this for something. It was originally set in https://github.com/AudiusProject/audius-client/pull/1246

### How Has This Been Tested?

fixes profile loading on local web

### DIFF
--- a/packages/web/src/common/store/lineup/sagas.js
+++ b/packages/web/src/common/store/lineup/sagas.js
@@ -288,7 +288,10 @@ function* fetchLineupMetadatasAsync(
           return true
         })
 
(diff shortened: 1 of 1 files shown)


audius-infra:
Preview this change https://demo.audius.co/amendel-fix-profile-lineup

Kyle-Shanks (approved):
nice

dylanjeffers (approved):
such a great find, nice work

PR #3770, opened 2023-07-19
[plat-1111] add usdc purchase seller and buyer notifications

sabrina-kiam:
### Description
this adds both web and mobile usdc purchase seller and buyer notifications. apologies for the long diff - ended up requiring more changes than expected!

### Dragons

i dont think there are any dragons, but want to note that these notifs are behind the usdc purchases feature flag to prevent any accidental early rollout.

### How Has This Been Tested?

tested by running locally. i inserted a usdc_purchase row mocking a transaction between two users i have the accounts for, which triggered a usdc purchase notification row to be inserted on dn1's db. then i pointed my client at dn1 and was able to load the notification with the usdc purchase ff on in dev. 
Screenshot 2023-07-19 at 5 15 20 PM
Screenshot 2023-07-19 at 5 43 57 PM
Screenshot 2023-07-19 at 5 44 04 PM

### Feature Flags ###
everything is under the usdc purchases feature flag, which will block fetching the new notifications if it's disabled


audius-infra:
Preview this change https://demo.audius.co/sk-plat-1111

packages/mobile/src/screens/notifications-screen/Notifications/USDCPurchaseBuyerNotification.tsx
```diff
@@ -0,0 +1,98 @@
...
+}
+
+export const USDCPurchaseBuyerNotification = (
+  props: USDCPurchaseBuyerNotificationProps
+) => {
+  const { notification } = props
+  const navigation = useNotificationNavigation()
+  const track = useSelector((state) =>
+    getNotificationEntity(state, notification)
+  ) as Nullable<TrackEntity>
```
schottra, line 50: Would be nice if we could add typing directly to the selector in the future so we don't have to cast it :-(

packages/web/src/assets/img/iconCart.svg
schottra, line 1: Doesn't this file exist in stems already?
edit: I'm honestly not sure if we should be importing icons from stems or putting them into web directly.

cc @dylanjeffers on any plans around this with Harmony

packages/web/src/common/store/notifications/parseAndProcessNotifications.ts
```diff
@@ -144,6 +144,13 @@ export function* parseAndProcessNotifications(
       userIdsToFetch.push(notification.userId)
       trackIdsToFetch.push(notification.entityId)
     }
+    if (
+      type === NotificationType.USDCPurchaseBuyer ||
+      type === NotificationType.USDCPurchaseSeller
+    ) {
+      userIdsToFetch.push(...[...new Set(notification.userIds)])
```
schottra, line 151: This syntax threw me a bit and I had to look up `Set` to find the example of spreading it into an array. Is the goal to deduplicate the userIds?

packages/web/src/components/notification/Notification/components/icons.module.css
```diff
@@ -67,6 +67,15 @@
   margin-right: var(--unit-2);
 }
 
+.iconCart {
+  composes: icon;
+  height: 24px;
```
schottra, line 72: I think we're moving towards setting sizes in grid units (`var(--unit-6)`), and isn't the icon square? So inheriting the width of 30px is going to stretch this icon horizontally/

schottra (approved):
Just a few small questions, but looks good!

audius-infra:
Preview this change https://demo.audius.co/sk-plat-1111

packages/web/src/assets/img/iconCart.svg
sabrina-kiam: reed mentioned they're trying to move over stems too. we ended up doing this way because it was the only way we could get it to work at the time, but i think we can switch it back to stems. i'll leave a note in the notif icons file that we are trying to move to stems as well

packages/web/src/common/store/notifications/parseAndProcessNotifications.ts
sabrina-kiam: yeaaa i was following existing code in the same file that was de-duping userIds but agree it's confusing and there should be a better way to do this? maybe `...new Set(notification.userIds)` should work?

packages/web/src/components/notification/Notification/components/icons.module.css
sabrina-kiam: going to just leave this as composes:icon, i think we were just trying to follow what other icons were doing but looks fine just inheriting icon ?
Screenshot 2023-07-22 at 9 50 52 PM

audius-infra:
Preview this change https://demo.audius.co/sk-plat-1111

packages/web/src/common/store/notifications/parseAndProcessNotifications.ts
schottra: Tried it out locally and `push(...new Set(notification.userIds))` works just fine.
But was also thinking that we have multiple conditionals in this function that are all checking notifications and pushing ids. So across the various notifications we are looping here, we could end up with duplicates anyway.

If you wanted to take a cleanup opportunity, you could:
* Switch the arrays defined on line 49 to `Set`s
* Switch the `push` statements in all of these conditionals to `add`
* At the end, when we call the various retrieve functions, convert the Sets to arrays (`call(retrieveTracks, { trackIds: Array.from(trackIdsToFetch) }`)

audius-infra:
Preview this change https://demo.audius.co/sk-plat-1111

PR #3771, opened 2023-07-19
[C-2846] Suggested Tracks QA

dylanjeffers:
### Description

- [x]  add toast alert when suggested track is added
- [x]  Icon caret down vs icon down line
- [x]  Static height + skeletons for mobile
- [x]  Update drag to reorder to only have touch target on the drag icon
- [x]  See if we can improve lineup loading state after adding element
- [x]  Padding on the track list does not match the padding of the text above (i.e. tracks and modified and such)
- [x]  Fix artist name in add-some-tracks
- [x]  Filter out collectible gated content from trending


packages/mobile/src/components/track-list/TrackListItem.tsx
```diff
@@ -100,8 +105,8 @@ const useStyles = makeStyles(({ palette, spacing, typography }) => ({
   iconContainer: {
     marginLeft: spacing(2)
   },
-  icon: { height: 20, width: 20 },
-  removeIcon: { height: 24, width: 24 },
+  icon: { height: spacing(6), width: spacing(6) },
+  removeIcon: { height: spacing(6), width: spacing(6) },
```
Kyle-Shanks, line 109: the values for icon height and width changed, making sure thats intentional. also seems weird to use spacing for the height and width of things 🤔

dylanjeffers: yes it was intentional, and good point about using spacing, i just went with it to be consistent with other times ive seen it used for height and width....

Kyle-Shanks: ewww, sounds good

packages/mobile/src/components/track-list/TrackList.tsx
```diff
@@ -121,20 +130,36 @@ export const TrackList = ({
...
+    },
+    [onReorder]
+  )
+
   if (showSkeleton)
     return (
       <FlatList {...otherProps} data={data} renderItem={renderSkeletonTrack} />
     )
 
+  console.log({ scrollEnable })
```
Kyle-Shanks, line 151: log

packages/mobile/src/components/track-list/TrackListItem.tsx
dylanjeffers: lolll ill change it, please please sir i just found it on the ground

Kyle-Shanks (approved):

PR #3773, opened 2023-07-20
[C-2872] Fix image retries

dylanjeffers:
### Description
Updates fetchCID logic to retry storage nodes on failure.

### DIFF
--- a/packages/common/src/services/audius-backend/AudiusBackend.ts
+++ b/packages/common/src/services/audius-backend/AudiusBackend.ts
@@ -363,7 +363,7 @@ export const audiusBackend = ({
     }
   }
 
-  async function preloadImage(url: string) {
+  async function preloadImage(url: string): Promise<boolean> {
     if (!preloadImageTimer) {
       const batchSize =
         getRemoteVar(IntKeys.IMAGE_QUICK_FETCH_PERFORMANCE_BATCH_SIZE) ??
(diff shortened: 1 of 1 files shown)


packages/common/src/services/audius-backend/AudiusBackend.ts
```diff
@@ -387,35 +387,50 @@ export const audiusBackend = ({
...
-        preloadImageTimer.end(start)
-        resolve(false)
-      }, timeoutMs)
+      if (!response.ok) {
+        return false
+      }
 
-      // Avoid garbage collection by keeping a few images in an in-mem array
+      const blob = await response.blob()
+      const objectUrl = URL.createObjectURL(blob)
```
theoilie, line 412: should we remove this and just do `image.src = url` below?

theoilie (approved):
🔥

packages/common/src/services/audius-backend/AudiusBackend.ts
dylanjeffers: no cause then it would fetch again? or are you thinking since fetch already happened it wouldn't?

theoilie: ah nvm I think I was misunderstanding what this does here. this looks good!

PR #3775, opened 2023-07-20
[PAY-1618] Update DogEar rendering and fix spacing

schottra:
### Description
DogEars were rendering with a visible gap around the outside (the border of the container beneath them). The fix for this turned out to be a little trickier than expected, partially due to the way the component was implemented.

The original implementation was done with an 80x80px square linear gradient, rotated 45 degrees and then clipped by the parent container's border box to achieve the desired shape. While this works, it had two disadvantages:
* It requires `overflow:hidden` to be set on the parent. When using this overflow setting, no child is allowed to exceed the border box. And that means the DogEar cannot overlap the border, and the border will be visible around the outside.
* It was not in sync with the vectors being used in the Figma designs

I brought in the new SVG, which uses a path to draw the rounded edge and has the correct gradients built in. Overflow behavior was removed from the parent container, and logic was added to allow parents to position the `DogEar` with an offset to account for their own borders.

### Dragons
This did involve some changes to the `Tile` component on desktop, which is used as a base for components other than the TrackTile. But the changes are minimal and I did a verification pass to make sure none of them looked strange without the overflow behavior.

### How Has This Been Tested?
Tested locally in Chrome (web and mobile web emulator), on iOS Simulator and physical Android device.

Will verify as part of RC process and revert if necessary.

### Feature Flags ###
N/A

### Screenshots

**Web Track Tile**
Before
Screenshot 2023-07-20 at 10 57 34 AM

After
Screenshot 2023-07-20 at 10 50 16 AM

---

**Mobile Web Track Tile**
Before
Screenshot 2023-07-20 at 10 56 39 AM

After
Screenshot 2023-07-20 at 10 50 54 AM

--- 

**Mobile Track Tile**
Before
Screenshot 2023-07-20 at 10 59 00 AM

After
Screenshot 2023-07-20 at 10 06 02 AM

---

**Web Track Details**
Before
Screenshot 2023-07-20 at 10 57 44 AM

After
Screenshot 2023-07-20 at 11 04 06 AM

---

**Mobile Web Track Details**
Before
Screenshot 2023-07-20 at 10 57 02 AM

After
Screenshot 2023-07-20 at 11 01 04 AM

---

**Mobile Track Details**
Before
Screenshot 2023-07-20 at 10 58 46 AM

After
Screenshot 2023-07-20 at 10 52 32 AM

### DIFF
--- a/packages/mobile/src/components/core/DogEar.tsx
+++ b/packages/mobile/src/components/core/DogEar.tsx
@@ -1,8 +1,8 @@
 import { DogEarType } from '@audius/common'
 import type { ViewStyle } from 'react-native'
 import { View } from 'react-native'
-import LinearGradient from 'react-native-linear-gradient'
 
+import DogEarRectangle from 'app/assets/images/dogEarRectangle.svg'
 import IconCart from 'app/assets/images/iconCart.svg'
 import IconCollectible from 'app/assets/images/iconCollectible.svg'
 import IconHidden from 'app/assets/images/iconHidden.svg'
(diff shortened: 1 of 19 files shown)


packages/web/src/components/dog-ear/DogEar.module.css
```diff
@@ -1,61 +1,49 @@
-.artistPick {
```
schottra, line 1: I don't know why this class was named `artistPick` (probably the dogEar was originally just for that?), but it was clearly the "container" 😄

packages/web/src/components/tile/Tile.module.css
```diff
@@ -1,12 +1,18 @@
...
+  border: var(--border-width) solid var(--neutral-light-8);
   border-radius: var(--unit-2);
   background-color: var(--white);
   transition: all var(--expressive);
-  overflow: hidden;
+}
+
+.borderOffset {
+  position: absolute;
+  top: calc(-1 * var(--border-width));
```
schottra, line 14: Open to feedback on this pattern. My goal with this PR was to remove knowledge of the parent from DogEar. The place where we decide how far to offset a component to cover the border is the place that defines said border!

dharit-tan (approved):

audius-infra:
Preview this change https://demo.audius.co/pay-1618-fix-dog-ears

packages/mobile/src/components/core/DogEar.tsx
```diff
@@ -78,15 +90,14 @@ export const DogEar = (props: DogEarProps) => {
...
-      <LinearGradient
-        colors={colors}
-        style={[styles.container]}
-        start={{ x: 1, y: 1 }}
-        end={{ x: 0, y: 0 }}
-      />
-      <Icon fill={staticWhite} height={16} width={16} style={styles.icon} />
+    <View style={[styles.container, borderOffsetStyle, style]}>
+      <DogEarRectangle fill={colors[0]} style={styles.rectangle} />
+      <Icon width={16} height={16} fill={staticWhite} style={styles.icon} />
```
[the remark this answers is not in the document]
schottra: You are correct! I thought you could only use that function in `makeStyles` context and I was mistaken.

packages/mobile/src/components/core/DogEar.tsx
```diff
@@ -12,38 +12,50 @@ import { makeStyles } from 'app/styles'
...
+  rectangle: {
     position: 'absolute',
-    transform: [{ rotate: '45deg' }, { translateX: -68 }],
-    width: spacing(20),
-    height: spacing(20)
+    top: 0,
+    left: 0,
+    width: spacing(12),
+    height: spacing(12),
+    shadowColor: '#000',
```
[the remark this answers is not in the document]
schottra: I was following another pattern for setting shadow, and also this value matches what is shown in Figma. Why the preference for the named color?

packages/web/src/components/tile/Tile.module.css
schottra: If you set a class or inline style on any component using this class, and within that class/style set `--border-width: value`, this will pick up that value and use it. This is also inheriting the value from the nearest parent that has it set (in this case `.root`), so you can set the value on that instead.

If you were trying to do it from outside this module or the corresponding component, it might be a little trickier. But I think the idea is that this stays contained to only this module.

audius-infra:
Preview this change https://demo.audius.co/pay-1618-fix-dog-ears

PR #3777, opened 2023-07-20
[C-2866] Undo secondary button style changes

schottra:
### Description
I made changes to secondary buttons as part of our gated content updates. This broke some more subtle behavior of buttons which use Secondary as an "active" variant. For now, going to undo them to fix the regression. Will follow up later with a more backwards-compatible change.

### Dragons
Technically breaks the new "preview" button style slightly. But that's feature-flagged anyway. Will fix before release.

### How Has This Been Tested?
Manually tested locally

N/A

### Feature Flags ###
N/A

### DIFF
--- a/packages/stems/src/components/Button/Button.module.css
+++ b/packages/stems/src/components/Button/Button.module.css
@@ -187,13 +187,13 @@
 /* Secondary Button */
 
 .secondary {
-  border: 1px solid var(--neutral-light-5);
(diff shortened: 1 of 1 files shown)


dharit-tan (approved):

audius-infra:
Preview this change https://demo.audius.co/C-2866-fix-secondary-buttons

PR #3784, opened 2023-07-21
Change mobile drawer background to white

dharit-tan:
### Description
Drawer background should be `palette.white` instead of `palette.neutralLight10`.

### Dragons

Have not tested every single drawer, but it's a very slight color change so should be fine. The only problem I can think of is, if there was some content that was `palette.white` in order to stand out against the background, this would make that less visible. Again, very slight color change.

### How Has This Been Tested?

Local ios stage
Simulator Screenshot - iPhone 14 Pro - 2023-07-21 at 16 26 21
Simulator Screenshot - iPhone 14 Pro - 2023-07-21 at 16 26 48
Simulator Screenshot - iPhone 14 Pro - 2023-07-21 at 16 27 12

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/drawer/Drawer.tsx
+++ b/packages/mobile/src/components/drawer/Drawer.tsx
@@ -35,7 +35,7 @@ const OVERFLOW_FRICTION = 4
 
 export const useStyles = makeStyles(({ palette }) => ({
   drawer: {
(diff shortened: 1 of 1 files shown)


dylanjeffers (approved):
great

PR #3790, opened 2023-07-24
Fix mobile LockedStatusBadge padding

dharit-tan:
### Description
Realized vertical padding was 1px off.
Also setting a default for `variant` bc in USDC Purchase Drawer there is a case where the badge will only ever be locked and therefore background color will only ever be gray, so we don't care about which variant is used.

### Dragons

### How Has This Been Tested?

Screenshot 2023-07-24 at 1 35 09 PM

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/core/LockedStatusBadge.tsx
+++ b/packages/mobile/src/components/core/LockedStatusBadge.tsx
@@ -10,6 +10,7 @@ const useStyles = makeStyles(({ palette, spacing, typography }) => ({
   root: {
     backgroundColor: palette.accentBlue,
     paddingHorizontal: spacing(2),
+    paddingVertical: 1,
     borderRadius: spacing(10),
     justifyContent: 'center'
   },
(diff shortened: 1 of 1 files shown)


packages/mobile/src/components/core/LockedStatusBadge.tsx
```diff
@@ -10,6 +10,7 @@ const useStyles = makeStyles(({ palette, spacing, typography }) => ({
   root: {
     backgroundColor: palette.accentBlue,
     paddingHorizontal: spacing(2),
+    paddingVertical: 1,
```
schottra, line 13: If I'm reading the mocks correctly, this should be 2 pixels (or a half-unit)?
image

packages/mobile/src/components/core/LockedStatusBadge.tsx
```diff
@@ -23,13 +24,13 @@ const useStyles = makeStyles(({ palette, spacing, typography }) => ({
...
   locked: boolean
-  variant: 'purchase' | 'gated'
+  variant?: 'purchase' | 'gated'
 }
 
 /** Renders a small badge with locked or unlocked icon */
 export const LockedStatusBadge = ({
   locked,
-  variant
+  variant = 'gated'
```
schottra, line 33: Good idea

schottra (approved):

packages/mobile/src/components/core/LockedStatusBadge.tsx
dharit-tan: Yeah that confused me for a sec, but in the mock it also sets height to 16px which limits the padding to 1: icon is 14px + 1px on each side.

PR #3792, opened 2023-07-24
[C-2886] Improve cache performance

dylanjeffers:
### Description

- Removes `add` action and uses redux-thunk to check confirmer before calling `addSucceeded` action
- Improves/fixes `fast` cache option, using a smaller/configurable TTL. This will help prevent thrashing with updating a bunch of values over and over. ADD_SUCCESSFUL action time goes way down with this fix over an average session.
- Adds "simple" cache option to prevent a ton of rerenders due to cache subscriptions. Doesn't remove this entirely so we can potentially re-implement this. It would be helpful to have subscriptions which would could work in conjunction with TTL.


packages/common/src/services/remote-config/defaults.ts
```diff
@@ -31,7 +31,8 @@ export const remoteConfigIntDefaults: { [key in IntKeys]: number | null } = {
   [IntKeys.BUY_AUDIO_WALLET_POLL_MAX_RETRIES]: 120,
   [IntKeys.BUY_AUDIO_SLIPPAGE]: 3,
   [IntKeys.GATED_TRACK_POLL_INTERVAL_MS]: 1000,
-  [IntKeys.DISCOVERY_NOTIFICATIONS_GENESIS_UNIX_TIMESTAMP]: 0
+  [IntKeys.DISCOVERY_NOTIFICATIONS_GENESIS_UNIX_TIMESTAMP]: 0,
+  [IntKeys.CACHE_ENTRY_TTL]: 0
```
amendelsohn, line 35: is 0 a safe default? Would that invalidate everything immediately?

packages/common/src/store/cache/actions.ts
```diff
@@ -17,22 +20,64 @@ export const UNSUBSCRIBE_SUCCEEDED = 'CACHE/UNSUBSCRIBE_SUCCEEDED'
...
+      cache.entries,
+      Object.keys(confirmCalls).map((kindId) => getIdFromKindId(kindId))
+    )
+
+    const entriesToAdd = []
+    const entriesToSubscribe = []
+    entries.forEach((entry) => {
+      // If something is confirming and in the cache, we probably don't
+      // want to replace it (unless explicit) because we would lose client
+      // state, e.g. "has_current_user_reposted"
```
amendelsohn, line 58: will this prevent us from doing multiple local updates to the local cache, or is it ok because those would always go through update not add?

packages/common/src/store/cache/reducer.ts
```diff
@@ -199,18 +203,17 @@ const addEntries = (state: CacheState, entries: any[], replace?: boolean) => {
...
       // do nothing
     } else if (existing) {
-      let newMetadata = mergeWith(
+      const newMetadata = mergeWith(
         {},
         existing,
         entity.metadata,
         mergeCustomizer
       )
-      newMetadata = updateImageCache(existing, entity.metadata, newMetadata)
```
amendelsohn, line 213: safe to remove this?

packages/common/src/store/cache/reducer.ts
```diff
@@ -382,17 +401,8 @@ const actionsMap = {
...
-      entries: any
-      statuses: any
-      uids: any
-      subscribers: any
-      subscriptions: any
-      idsToPrune: any
-    },
-    action: { ids: any[] }
-  ) {
+  [REMOVE_SUCCEEDED](state: CacheState, action: { ids: any[] }) {
```
amendelsohn, line 404: love this `CacheState` change!

packages/mobile/src/screens/collection-screen/useFetchCollectionLineup.ts
```diff
@@ -106,7 +106,8 @@ export const useFetchCollectionLineup = (
         metadata: track
       }))
 
-      dispatch(cacheActions.add(Kind.TRACKS, cacheTracks, false, true))
+      // // @ts-expect-error
+      // dispatch(cacheActions.add(Kind.TRACKS, cacheTracks, false, true))
```
amendelsohn, line 110: can remove?

packages/mobile/src/screens/favorites-screen/useFavoritesLineup.ts
```diff
@@ -55,7 +55,8 @@ export const useFavoritesLineup = (fetchLineup: () => void) => {
         metadata: track
       }))
 
-      dispatch(cacheActions.add(Kind.TRACKS, cacheTracks, false, true))
+      // // @ts-expect-error
+      // dispatch(cacheActions.add(Kind.TRACKS, cacheTracks, false, true))
```
amendelsohn, line 59: can remove?

packages/mobile/src/store/offline-downloads/sagas/rehydrateOfflineDataSaga.ts
```diff
@@ -6,7 +6,7 @@ import type {
   UserMetadata
 } from '@audius/common'
 import { cacheActions, Kind, makeUid } from '@audius/common'
-import { call, select, put } from 'typed-redux-saga'
+import { call, put, select } from 'typed-redux-saga'
```
amendelsohn, line 9: 🔡 🎉

packages/web/src/common/store/cache/collections/createPlaylistSaga.ts
```diff
@@ -19,9 +19,10 @@ import {
   newCollectionMetadata,
   confirmerActions,
   confirmTransaction,
-  RequestConfirmationError
+  RequestConfirmationError,
+  put
```
amendelsohn, line 23: I worry this will get confusing and people will forget. Do we only need it when doing cache adds? What happens when we forget?

packages/web/src/common/store/cache/sagas.js
```diff
@@ -207,51 +207,6 @@ function* retrieveFromSourceThenCache({
...
-
-// Adds entries but first checks if they are confirming.
-// If they are, don't add or else we could be in an inconsistent state.
-function* watchAdd() {
-  yield takeEvery(cacheActions.ADD, function* (action) {
-    const { kind, entries, replace, persist } = action
-    yield call(add, kind, entries, replace, persist)
-  })
-}
-
```
amendelsohn, line 254: 🥳

packages/web/src/common/store/cache/sagas.js
```diff
@@ -327,27 +282,45 @@ function* watchRemove() {
...
-    remoteConfig.getFeatureEnabled,
+    getFeatureEnabled,
     FeatureFlags.SAFE_FAST_CACHE
   )
 
-  if (fastCache) {
-    yield put(cacheActions.setCacheType({ cacheType: 'fast' }))
+  let cacheType = 'normal'
+
+  if (fastCache || (isNativeMobile && fastMobileCache)) {
```
amendelsohn, line 301: do we want to be able to enable it for web but not mobile? seems like we can either do both or just mobile this way

amendelsohn (approved):
Wow, very excited to see this in action!!
Excellent job 🚀

packages/common/src/services/remote-config/defaults.ts
dylanjeffers: what's potentially really silly, is that this makes it falsy, so the DEFAULT value in the code gets used. BUT now that you have me thinking, we should just get rid of the constant and use this.

packages/common/src/store/cache/actions.ts
dylanjeffers: yeah exactly ^ (also this is migrated line-for-line by the original saga that did this)

packages/common/src/store/cache/reducer.ts
dylanjeffers: oh great find, whew, no not at all!

packages/mobile/src/screens/collection-screen/useFetchCollectionLineup.ts
dylanjeffers: oh great find as well, no

packages/mobile/src/screens/favorites-screen/useFavoritesLineup.ts
dylanjeffers: thanks, was testing something out when i removed them, bringing back

packages/mobile/src/store/offline-downloads/sagas/rehydrateOfflineDataSaga.ts
dylanjeffers: lmao

packages/web/src/common/store/cache/collections/createPlaylistSaga.ts
dylanjeffers: yeah it's super confusing. basically folks should literally not need to use it, it was just that redux-saga types don't know how to deal with dispatch allowing a ThunkAction. so i think will only be needed for the few times we call `put(add(...))`

packages/web/src/common/store/cache/sagas.js
dylanjeffers: thanks for the callout here, i realized we dont need a separate "fast" and fast-mobile" now that desktop issues are fixed. this was to allow us to roll forward with the fast cache on mobile cause desktop was acting up for a while there. but we are consistent now, so will remove.

packages/web/src/common/store/cache/collections/createPlaylistSaga.ts
amendelsohn: hmm ok I guess that's fine. Wondering if there's a place we can put a comment so someone dealing with the type error won't waste time on it

dylanjeffers: added a comment :)

audius-infra:
Preview this change https://demo.audius.co/dj-cache-improvements

PR #3791, opened 2023-07-24
[PAY-1587] Mobile USDC Purchase Drawer Skeleton

dharit-tan:
### Description
UI for purchasing USDC-gated tracks.

### Dragons
How do we feel about exporting the `PremiumContentType` enum in models?
Also cc @dylanjeffers wondering if there's a smarter way to do the colors in `TrackDetailsTile`. `Icon` takes a string, while `Text` takes a `keyof ThemeColors`.

### How Has This Been Tested?

Local ios stage
simulator_screenshot_3A17FEBC-5294-4CD5-A3F6-48A24367293E

### How will this change be monitored?

*For features that are critical or could fail silently please describe the monitoring/alerting being added.*

### Feature Flags ###

*Are all new features properly feature flagged? Describe added feature flags.*

PR #3793, opened 2023-07-24
[PAY-1587] Mobile USDC Purchase Drawer Skeleton

dharit-tan:
### Description
Re-trying this PR, had to close previous one due to git snafu: https://github.com/AudiusProject/audius-client/pull/3791

UI for purchasing USDC-gated tracks.

### Dragons

How do we feel about exporting the PremiumContentType enum in models?
Also cc @dylanjeffers wondering if there's a smarter way to do the colors in TrackDetailsTile. Icon takes a string, while Text takes a keyof ThemeColors.

### How Has This Been Tested?

Local ios stage
255672581-dde97e31-2354-45fe-9cd4-e8ed7ed16f51

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/premium-track-purchase-drawer/PremiumTrackPurchaseDrawer.tsx
+++ b/packages/mobile/src/components/premium-track-purchase-drawer/PremiumTrackPurchaseDrawer.tsx
@@ -1,0 +1,167 @@
+import { useCallback } from 'react'
+
+import {
+  cacheTracksSelectors,
+  formatUSDCWeiToUSDString,
+  isPremiumContentUSDCPurchaseGated
+} from '@audius/common'
+import { View } from 'react-native'
+import { useSelector } from 'react-redux'
+
+import IconCart from 'app/assets/images/iconCart.svg'
+import { LockedStatusBadge, Text, Button } from 'app/components/core'
+import { NativeDrawer } from 'app/components/drawer'
+import { useDrawer } from 'app/hooks/useDrawer'
+import { makeStyles, flexRowCentered } from 'app/styles'
+import { useThemeColors } from 'app/utils/theme'
+
+import { TrackDetailsTile } from '../track-details-tile'
+
+const { getTrack } = cacheTracksSelectors
+
+const PREMIUM_TRACK_PURCHASE_MODAL_NAME = 'PremiumTrackPurchase'
+
+const messages = {
+  title: 'Complete Purchase',
+  summary: 'Summary',
+  artistCut: 'Artist Cut',
+  audiusCut: 'Audius Cut',
+  alwaysZero: 'Always $0',
+  youPay: 'You Pay',
+  price: (price: string) => `$${price}`,
+  payToUnlock: 'Pay-To-Unlock',
+  disclaimer:
+    'By clicking on "Buy", you agree to our Terms of Use. Your purchase will be made in USDC via 3rd party payment provider. Additional payment provider fees may apply. Any remaining USDC balance in your Audius wallet will be applied to this transaction. Once your payment is confirmed, your premium content will be unlocked and available to stream.',
+  buy: (price: string) => `Buy $${price}`
+}
+
+const useStyles = makeStyles(({ spacing, typography, palette }) => ({
+  drawer: {
+    paddingVertical: spacing(6),
+    paddingHorizontal: spacing(4),
+    gap: spacing(6),
+    backgroundColor: palette.white
+  },
+  titleContainer: {
+    ...flexRowCentered(),
+    gap: spacing(2),
+    marginBottom: spacing(2),
+    alignSelf: 'center'
+  },
+  title: {
+    fontSize: typography.fontSize.xl,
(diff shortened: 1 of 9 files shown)


packages/mobile/src/components/track-details-tile/TrackDetailsTile.tsx
```diff
@@ -73,23 +83,70 @@ type TrackDetailsTileProps = {
...
     premiumConditions: track?.premium_conditions
   })
 
+  const type = isUSDCPurchaseGated
+    ? PremiumContentType.USDC_PURCHASE
+    : isCollectibleGated
+    ? PremiumContentType.COLLECTIBLE_GATED
+    : PremiumContentType.SPECIAL_ACCESS
+
+  const headerAttributes: Record<
```
dharit-tan, line 108: had to explicitly type this in order to get colorString as a keyof ThemeColors type. Open to suggestions!

packages/mobile/src/components/premium-track-purchase-drawer/PremiumTrackPurchaseDrawer.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const { trackId } = data
+  const track = useSelector((state) => getTrack(state, { id: trackId }))
+
+  const handleConfirmPress = useCallback(() => {
+    console.log('buy button pressed')
+  }, [])
+
+  const { premium_conditions: premiumConditions } = track ?? {}
+
+  if (!track || !isPremiumContentUSDCPurchaseGated(premiumConditions))
```
dharit-tan, line 109: Copying over convo from previous PR:
@schottra :
I would normally argue that this component should not even be rendered if the conditions here aren't true (prefer that components receive only valid data). But I would understand if that's a difficult thing to check at the layer where we render this component.
@dharit-tan :
yeah agree. i think the difficulty is since this is a drawer, all we pass to the drawer is a trackId and so in this component we don't know whether it's a USDC-gated track or if the track even exists in the cache. I suppose we could pass in the premiumConditions object to the drawer, but that seems to go against the "select everything at the leaf" mental model. Open to ideas here.

audius-infra:
Preview this change https://demo.audius.co/rt-usdc-drawer-2

dylanjeffers:
@dharit-tan it would be great for text to follow same pattern as icon... maybe can create a new prop for that so it doesn't affect the current color prop? Down to chat about best option with ya

packages/mobile/src/components/premium-track-purchase-drawer/PremiumTrackPurchaseDrawer.tsx
schottra: Ah interesting. I guess I'm hoping we don't launch the drawer at all if we don't have the right data available, even if we eventually only pass the trackId to the drawer.

So, in the case where some of these things conditions are for some reason false, we're going to see an empty drawer? Wonder if we should render some kind of error in this case instead of null...

Either way, probably okay to leave for now.

packages/mobile/src/components/track-details-tile/TrackDetailsTile.tsx
schottra: 💯  This is close to a pattern I regularly follow for mapping an enum or limited set of choices to a set of values/variants of a component. 

My only suggestion might be that if you want to make sure this always maps every enum value, don't use a `Record`, but a [mapped type](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html) :
```typescript
// defined outside the component
type HeaderAttributesMapping = { [k in PremiumContentType]: {
  message: string
  icon: ComponentType<SvgProps>
  color: string
  colorString: keyof ThemeColors
}}

// Used in the assignment
const headerAttributes: HeaderAttributesMapping = ...
```

Doing it that way _requires_ a value for every key in the `PremiumContentType` enum. So if we change that enum and don't change this definition, we will get a compiler error. It also has the benefit of the compiler _knowing_ that the value of headerAttributes will be defined, since all of the possible keys in this mapping have corresponding values.

schottra (approved):

dharit-tan:
> @dharit-tan it would be great for text to follow same pattern as icon... maybe can create a new prop for that so it doesn't affect the current color prop? Down to chat about best option with ya

makes sense! will take a look at this. gonna merge this PR for now and follow up with another one.

packages/mobile/src/components/premium-track-purchase-drawer/PremiumTrackPurchaseDrawer.tsx
dharit-tan: i hear you. the 1 place we open the drawer, we do guarantee that we have the correct data, so for now we're good.
packages/mobile/src/components/lineup-tile/LineupTileAccessStatus.tsx line 80.

packages/mobile/src/components/track-details-tile/TrackDetailsTile.tsx
dharit-tan: ooo nice ty!

PR #3794, opened 2023-07-25
[PAY-1651] Implements Harmony Buttons

schottra:
### Description
This adds the initial implementation of Harmony buttons, based on the new design system. Notable updates:
* Cleaned up size and color variant definitions. We now mostly just update some CSS variable values as the color variant class changes.
* Using `gap` for layout now, so no special classes for icons/text based on presence of each other
* Stories render disabled variants now
* Icon properties accept an icon component instead of a rendered element, allowing the button to control when the icon is actually rendered and apply classes directly to the SVG instead of a span above it.
* SVG paths will now use `currentColor` for the fill, so that they always match the text color
* Added a deprecation jsdoc comment to the original `Button` to remind devs to use the new component in new feature work.
* Renamed `iconCampFire.svg` -> `iconCampfire.svg` to match Figma
* Remove some unnecessary semantic color definitions which don't differ across themes (the variables they reference change)
* Added semantic colors for border/background since they are used in the new button

### Dragons
No changes to existing button component and this component is currently unused, so it should be a safe change.

### How Has This Been Tested?
Locally verified in Storybook

N/A

### Feature Flags ###
N/A

### Screenshots
Screenshot 2023-07-25 at 11 30 03 AM

---

Screenshot 2023-07-25 at 11 30 22 AM

---

Screenshot 2023-07-25 at 11 30 31 AM

---

https://github.com/AudiusProject/audius-client/assets/1815175/add85777-5191-4c71-8b6b-2eeed79c2b17

---

https://github.com/AudiusProject/audius-client/assets/1815175/1e259e73-8221-43d1-a46e-304a22b69478


audius-infra:
Preview this change https://demo.audius.co/pay-1651-harmony-buttons

packages/stems/src/assets/styles/colors.css
```diff
@@ -69,6 +69,16 @@ root {
   --text-subdued: var(--neutral-light-4);
   --text-disabled: var(--neutral-light-7);
 
+  /* Semantic Borders */
+  --border-default: var(--neutral-light-8);
```
dylanjeffers, line 73: wonderful

packages/stems/src/components/HarmonyButton/HarmonyButton.module.css
```diff
@@ -0,0 +1,220 @@
...
+  justify-content: center;
+  overflow: hidden;
+  position: relative;
+  text-align: center;
+  user-select: none;
+  white-space: nowrap;
+}
+
+.button:focus {
+  outline: none !important;
```
dylanjeffers, line 24: do we want to do a css reset on button? mabye `all: unset` on the root element`

packages/stems/src/components/HarmonyButton/HarmonyButton.module.css
```diff
@@ -0,0 +1,220 @@
...
+}
+
+/* Large */
+.large {
+  gap: var(--unit-2);
+  height: var(--unit-15);
+  padding: var(--unit-5) var(--unit-6);
+}
+
+.large .icon {
```
dylanjeffers, line 113: i think this is totally good, but wondering how you feel about specificity changes here? ive personally always tried to just have classname directly tied to the element, so i might do `largeIcon` class. Im no expert though so would love your insight here

packages/stems/src/components/HarmonyButton/HarmonyButton.module.css
```diff
@@ -0,0 +1,220 @@
...
+
+.primary:active {
+  --overlay-color: var(--static-black);
+  --overlay-opacity: 0.2;
+  box-shadow: none;
+}
+
+/* Secondary */
+.secondary {
+  --button-color: var(--neutral-light-5);
```
dylanjeffers, line 150: really cool with the css-vars defined in the className. do you mind giving a quick tldr on why we do this again?

packages/stems/src/components/HarmonyButton/HarmonyButton.tsx
```diff
@@ -0,0 +1,105 @@
...
+  [HarmonyButtonType.DESTRUCTIVE]: styles.destructive,
+  [HarmonyButtonType.GHOST]: styles.ghost
+}
+
+/**
+ * A common Button component. Includes a few variants and options to
+ * include and position icons.
+ */
+export const HarmonyButton = forwardRef<HTMLButtonElement, HarmonyButtonProps>(
+  function Button(
```
dylanjeffers, line 35: nice! maybe also call this HarmonyButton?

packages/stems/src/components/HarmonyButton/HarmonyButton.tsx
```diff
@@ -0,0 +1,105 @@
...
+    {
```
dylanjeffers, line 36: just since this will be our "core components" ill be extra nitty just to make sure we have alignment. so i personally prefer `function Component(props) {...}` and then destructure props in the body. i find this makes the top-level function definition a but easier to parse. if you feel strongly this way, all good as well, just curious :)

packages/stems/src/components/HarmonyButton/HarmonyButton.tsx
```diff
@@ -0,0 +1,105 @@
...
+      text,
+      variant = HarmonyButtonType.PRIMARY,
+      size = HarmonyButtonSize.DEFAULT,
+      leftIcon: LeftIconComponent,
+      rightIcon: RightIconComponent,
+      disabled,
+      widthToHideText,
+      minWidth,
+      className,
+      iconClassName,
```
dylanjeffers, line 47: so as a pattern going forward we want nested classnames to be indivudally named? in mobile and other libraries ive seen the `classes` pattern where it's an object of all the inner classnames we would want to open up.

packages/stems/src/components/HarmonyButton/HarmonyButton.module.css
dylanjeffers: and for example if we pass in iconClassName, doesnt specificity get a bit messed up when having to compete against `.default .icon`?

packages/stems/src/components/HarmonyButton/HarmonyButton.tsx
dylanjeffers: i forget but i think it also may prevent an extra nested body definition.

packages/stems/src/components/HarmonyButton/HarmonyButton.tsx
```diff
@@ -0,0 +1,105 @@
...
+          TYPE_STYLE_MAP[variant],
+          {
+            [styles.disabled]: disabled,
+            [styles.fullWidth]: fullWidth
+          },
+          className
+        )}
+        disabled={disabled}
+        ref={ref}
+        style={style as CSSProperties}
```
dylanjeffers, line 90: just wondering about `CSSCustomProperties` vs `CSSProperties`. anyway to avoid the cast?

packages/stems/src/components/HarmonyButton/HarmonyButton.tsx
```diff
@@ -0,0 +1,105 @@
...
+ * include and position icons.
+ */
+export const HarmonyButton = forwardRef<HTMLButtonElement, HarmonyButtonProps>(
+  function Button(
+    {
+      color,
+      text,
+      variant = HarmonyButtonType.PRIMARY,
+      size = HarmonyButtonSize.DEFAULT,
+      leftIcon: LeftIconComponent,
```
dylanjeffers, line 41: headsup mobile does `iconLeft`. im fine with either, but prob want consistency. if we go with leftIcon, can we make a note in mobile code on the prop definitions that we'd like to rename?

dylanjeffers (approved):
wonderful work, so simple and clean

packages/stems/src/components/HarmonyButton/HarmonyButton.module.css
schottra: Yeah good call. I was actually trying to use the least specificity possible, but I didn't really want to do separate classes for each piece (button, icon, text). I used a handy specificity calculator to check, and you are correct! The only way to make sure the class can be overridden is to use only one class, no descendants or combinators. Will update.

packages/stems/src/components/HarmonyButton/HarmonyButton.module.css
schottra: I had to look up how this works
>Specifies that all the element's properties should be changed to their inherited values if they inherit by default, or to their initial values if not.

I'm actually a bit confused on how it's useful. Could you give me an example scenario you're trying to account for?

packages/stems/src/components/HarmonyButton/HarmonyButton.module.css
schottra: Yep! There were two goals:
* Allowing override of the button color from JS by injecting a different value for the variable. (This is only used in primary buttons at the moment)
* Explicitly linking various properties logically to the button color. The idea is that you think about the button having a color, and then the style rules for each variant decide how the color is _applied_ to the button (ex. Secondary buttons apply the color to only the border, primary apply the color to the background).
* Defining how the color maps to the properties once, then just changing the color for hover/active states.

But I could see how this is arguably not any easier to follow than just setting the individual values per variant, per state. Open to simplifying it if you think it reads better.

packages/stems/src/components/HarmonyButton/HarmonyButton.tsx
schottra: I actually don't find either way more or less readable. Happy to destructure in the function body. I thought there was a style guide entry about doing it in the arguments, but I was unable to find it.

packages/stems/src/components/HarmonyButton/HarmonyButton.tsx
schottra: Oh I'd rather this be consistent. Can switch it to iconLeft. I think it actually works better because then `iconLeft` and `iconRight` sort next to each other lexographically.

packages/stems/src/components/HarmonyButton/HarmonyButton.tsx
schottra: I will say that with the `classes` pattern, it might be harder to write types such that you know what the valid class names are.
For example,
```
// Harder to write a props shape that will catch this typo!
<HarmonyButton classes={icn: styles.icon, text: styles.text} />
```

Not impossible, I could whip up a generic that makes it easier to define. Just wanted to point that out.

packages/stems/src/components/HarmonyButton/HarmonyButton.tsx
schottra: Oh right, I did work around this in a cleaner way elsewhere by just casting an object directly to CSSProperties.
The issue is that the built-in CSSProperties doesn't allow you to pass custom variables (`--button-color`)

dylanjeffers: gotchaa

packages/stems/src/components/HarmonyButton/HarmonyButton.tsx
dylanjeffers: yah totally agreed not having to pass styles in, so just removing it is great for now, and then we can all have a discussion about what pattern we want moving forward!! as far as classes type utility we added something somewhat basic in mobile that works well, so can prob reuse that if we go forward with classes

packages/stems/src/components/HarmonyButton/HarmonyButton.tsx
dylanjeffers: yeah its not official, just something i personally like and wanted to start a convo :)

packages/stems/src/components/HarmonyButton/HarmonyButton.module.css
dylanjeffers: thank you for all the context, this makes sense, esp with allowing it styled through js. lets continue with this pattern and see how folks get along with it 🔥

sliptype (approved):
Really awesome work! I think yall already covered everything I was going to point out. This is such a solid start to Harmony 🎉

audius-infra:
Preview this change https://demo.audius.co/pay-1651-harmony-buttons

PR #3795, opened 2023-07-25
[PAY-1650] Update play/preview buttons on track details to use HarmonyButton

schottra:
### Description
This reintroduces the correct styling for the Play/Pause button on the web and mobile web track details page. I had originally reverted the styling changes on `Button` because it caused some issues elsewhere. Now we have the  new `HarmonyButton` component and can use that instead.

### Dragons
None

### How Has This Been Tested?
Verified locally on Chrome

N/A

### Feature Flags ###
Display of the preview variant which uses the new styling is behind the USDC feature flag

### DIFF
--- a/packages/web/src/components/track/PlayPauseButton.tsx
+++ b/packages/web/src/components/track/PlayPauseButton.tsx
@@ -7,9 +7,9 @@ import {
   CommonState
 } from '@audius/common'
 import {
-  Button,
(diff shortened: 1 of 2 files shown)


dharit-tan (approved):

audius-infra:
Preview this change https://demo.audius.co/pay-1650-fix-button

PR #3796, opened 2023-07-25
[C-2890] Add first version of a typography component to web

Kyle-Shanks:
### Description
Testing out a first version of a Typography component for web.

Note: This does not use the css variables that were written by Randy bc this we mostly build out before that work was done

### Dragons

N/A

### How Has This Been Tested?

Manual testing

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/web/src/components/typography/Text.tsx
+++ b/packages/web/src/components/typography/Text.tsx
@@ -1,0 +1,72 @@
+import { CSSProperties, ReactNode } from 'react'
+
+import { getCurrentThemeColors, ThemeColor } from 'utils/theme/theme'
+
+import { fontWeightMap, variantInfoMap, variantStrengthMap } from './constants'
+import {
+  TextSizeInfo,
+  TextStrengthInfo,
+  TextVariant,
+  TextSize,
+  TextStrength
+} from './types'
+
+const defaultBodyInfo: TextSizeInfo & TextStrengthInfo = {
+  ...variantInfoMap.body.M!,
+  ...variantStrengthMap.body.default!
+}
+
+export type TextProps = {
+  className?: string
+  children?: ReactNode
+  variant?: TextVariant
+  size?: TextSize
+  strength?: TextStrength
+  color?: ThemeColor
+}
+
+export const Text = (props: TextProps) => {
+  const {
+    className,
+    children,
+    variant = 'body',
+    strength = 'default',
+    size = 'M',
+    color = '--neutral',
+    ...otherProps
+  } = props
+
+  const themeColors = getCurrentThemeColors()
+  const textColor = themeColors[color] ?? themeColors['--neutral']
+
+  const variantSizeInfo = variantInfoMap[variant][size] ?? {}
+  const variantStrengthInfo = variantStrengthMap[variant][strength] ?? {}
+
+  const {
+    tag: Tag,
+    fontSize,
+    lineHeight,
+    letterSpacing,
+    fontWeight,
+    textTransform
+  } = {
+    ...defaultBodyInfo,
+    ...variantSizeInfo,
+    ...variantStrengthInfo
+  }
+
+  const styleObject: CSSProperties = {
+    fontWeight: fontWeightMap[fontWeight].toString(),
+    fontSize,
+    lineHeight,
+    letterSpacing,
+    textTransform,
+    color: textColor
+  }
+
+  return (
+    <Tag className={className} style={styleObject} {...otherProps}>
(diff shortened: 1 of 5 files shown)


packages/web/src/components/typography/Text.tsx
```diff
@@ -0,0 +1,90 @@
...
+    ...variantSizeInfo,
+    ...variantStrengthInfo
+  }
+
+  const styleObject: CSSProperties = {
+    fontWeight: fontWeightMap[fontWeight].toString(),
+    fontSize,
+    lineHeight,
+    letterSpacing,
+    textTransform: isAllCaps ? 'uppercase' : 'inherit',
```
amendelsohn, line 81: is there a reason to use `isAllCaps` instead of `textTransform` as the key?

packages/web/src/components/typography/Text.tsx
```diff
@@ -0,0 +1,90 @@
...
+  TextSizeInfo,
+  TextStrengthInfo,
+  TextVariant,
+  TextSize,
+  TextStrength,
+  FontWeight
+} from './types'
+
+// TODO:
+// - Get theme colors in here
```
amendelsohn, line 18: still todo?

amendelsohn (approved):
Very nice! I like this approach.
Having props type checking the structured mappings in code is a win imo.

packages/web/src/components/typography/Text.tsx
Kyle-Shanks: nope 👍

packages/web/src/components/typography/Text.tsx
Kyle-Shanks: was on the fence about this, felt weird to type out the values for a css prop, but will try it out

audius-infra:
Preview this change https://demo.audius.co/kj-Add-typography-component-for-web

packages/web/src/components/typography/Text.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+const defaultBodyInfo: TextSizeInfo & TextStrengthInfo = {
+  tag: 'p',
+  fontSize: '16px',
+  lineHeight: '20px',
+  fontWeight: 'medium',
+  textTransform: 'inherit'
+}
+
+export type TextProps = {
```
dylanjeffers, line 36: this is wonderful

packages/web/src/components/typography/Text.tsx
```diff
@@ -0,0 +1,87 @@
...
+import {
+  TextSizeInfo,
+  TextStrengthInfo,
+  TextVariant,
+  TextSize,
+  TextStrength,
+  FontWeight
+} from './types'
+
+export const fontWeightMap: Record<FontWeight, number> = {
```
dylanjeffers, line 17: would it make sense to move these constants into the types file with some of the other definitions?

packages/web/src/components/typography/Text.tsx
```diff
@@ -0,0 +1,87 @@
...
+    letterSpacing,
+    fontWeight,
+    textTransform
+  } = {
+    ...defaultBodyInfo,
+    ...variantSizeInfo,
+    ...variantStrengthInfo
+  }
+
+  const styleObject: CSSProperties = {
```
dylanjeffers, line 73: as a first implementation, using styleObject is okay, but long term i do think it makes sense to use classNames so we can get better performance and readability when inspecting elements

packages/web/src/components/typography/Text.tsx
```diff
@@ -0,0 +1,87 @@
...
+  className?: string
+  children?: ReactNode
+  variant?: TextVariant
+  size?: TextSize
+  strength?: TextStrength
+  color?: ThemeColor
+}
+
+export const Text = ({
+  className,
```
dylanjeffers, line 46: a standard id like us to start using is to have the word `props` here instead of destructured, and then desturcture in the body. if you feel strongly not to do this, totally get it, but it's a stylistic preference that i like a lot.

dylanjeffers (approved):
really clean and this is going to change so much. lets also create and `index.ts` file and export Text and all the types so imports can be `import {Text} from 'components/typography`

packages/web/src/components/typography/Text.tsx
Kyle-Shanks: yea you're probably right here. using the styles like this will also make overwriting the styles more annoying. I'll update it in the future

PR #3797, opened 2023-07-25
Revert cacheActions.add thunk

dylanjeffers:
### Description

Undos cacheActions.add thunk update. This was a potential optimization to save a round-trip add -> add_successful call to improve cache performance. Noticing some strang race-condition degration with it, so undoing it for now, since the main win, using 'fast-cache' is done.


amendelsohn (approved):

audius-infra:
Preview this change https://demo.audius.co/dj-fix-cache

PR #3803, opened 2023-07-26
[Harmony] Add SelectablePill to Harmony PAY-1654

nicoback2:
### Description
This will be used for the filter buttons in the library
https://www.figma.com/file/hcVGqf35jfxueO7klqXlma/Harmony-Component-Library?node-id=796%3A5027&mode=dev

screencapture-localhost-6006-2023-07-26-09_35_35

### Dragons

### How Has This Been Tested?

### Feature Flags ###


packages/stems/src/components/SelectablePill/SelectablePill.tsx
```diff
@@ -0,0 +1,44 @@
...
+import cn from 'classnames'
+
+import styles from './SelectablePill.module.css'
+import { SelectablePillProps } from './types'
+
+export const SelectablePill = forwardRef<
+  HTMLButtonElement,
+  SelectablePillProps
+>(
+  ({
```
dylanjeffers, line 12: personally can we just call this props, so its `props: SelecteablePillProps` and then destructure in the body? i think it makes the layout a bit cleaner and consistent

dylanjeffers (commented):
A larger question we can address now or when we do the official harmony transition: i think the selectable pill is secretly a "radio-button" and for the best a11y, we should probably implement it as such, with a radio-group, input, label etc. curious thoughts here!

audius-infra:
Preview this change https://demo.audius.co/nkang--stems-selectable-pill

schottra:
> A larger question we can address now or when we do the official harmony transition: i think the selectable pill is secretly a "radio-button" and for the best a11y, we should probably implement it as such, with a radio-group, input, label etc. curious thoughts here!

Agree, but only if the values are mutually exclusive. Also, doesn't this represent one option in a radio group? So we would want a standard way in harmony to create a group of selectable pills and then wrap them in the correct markup.

If we allow more than one pill to be selected at a time, then they become checkboxes!

packages/stems/src/components/SelectablePill/SelectablePill.module.css
```diff
@@ -0,0 +1,67 @@
+.pill {
+  cursor: pointer;
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  color: var(--text-subdued);
+  user-select: none;
+  height: 24px;
```
schottra, line 8: Might be out of the loop, but likely these should also be using grid units?

packages/stems/src/components/SelectablePill/SelectablePill.module.css
```diff
@@ -0,0 +1,67 @@
...
+  box-shadow: none;
+}
+
+.large:active,
+.large.selected {
+  box-shadow: none;
+}
+
+.icon path {
+  fill: currentColor;
```
schottra, line 56: :100:

packages/stems/src/components/SelectablePill/SelectablePill.module.css
```diff
@@ -0,0 +1,67 @@
...
+}
+
+.icon {
+  margin-right: var(--unit-1);
```
schottra, line 60: Can I recommend using gap on the container instead of adding margins to icons?

packages/stems/src/components/SelectablePill/SelectablePill.tsx
```diff
@@ -0,0 +1,44 @@
...
+    size,
+    isSelected,
+    label,
+    icon: IconComponent,
+    onClick
+  }: SelectablePillProps) => {
+    const handleClick = () => {
+      if (!isSelected && onClick) {
```
schottra, line 20: Did you mean to not call the click handler here if the component isn't selected?

Also wondering if there was any discussion on what the desired _developer interface_ is for Harmony components. This feels like one of those nuances where we might want to introduce a convention and make all components work the same.
In this case, that convention is that SelectablePill is "controlled" and you have to pass the selection state in from above, and also handle the clicks there.

edit: Also wondering if we even need an intermediate handler here? I think Button just passes the `onClick` handler through to the native button.

nicoback2: oh wow I did not mean to do that, don't know what I was thinking lol. I think I copied some code over without thinking.
Yes SelectablePill should be controlled and we don't need an intermediate handler!

packages/stems/src/components/SelectablePill/SelectablePill.module.css
nicoback2: yess sg

packages/stems/src/components/SelectablePill/SelectablePill.module.css
nicoback2: yeah I thought about this, I wasn't sure if units should be used for things other than spacing (i.e. gaps, margins, padding) etc

nicoback2: @Kyle-Shanks @dylanjeffers @sliptype @amendelsohn what are your thoughts on using grid units for things like `height` and `width`?

sliptype: Ideally everything conforms to our grid so I say let's use grid units here too!

nicoback2:
> A larger question we can address now or when we do the official harmony transition: i think the selectable pill is secretly a "radio-button" and for the best a11y, we should probably implement it as such, with a radio-group, input, label etc. curious thoughts here!

ohh interesting, yeah i can def see it as either a radio button or a checkbox. id prefer to leave this as is for now to prevent growing scope in this pr but we should add SelectablePillGroup would render the pills and could control what HTML element they are

audius-infra:
Preview this change https://demo.audius.co/nkang--stems-selectable-pill

audius-infra:
Preview this change https://demo.audius.co/nkang--stems-selectable-pill

packages/stems/src/components/SelectablePill/HarmonySelectablePill.tsx
```diff
@@ -0,0 +1,37 @@
...
+        {
+          [styles.large]: size === 'large',
+          [styles.selected]: isSelected
+        },
+        className
+      )}
+      ref={ref}
+      {...restProps}
+    >
+      {IconComponent == null ? null : <IconComponent className={styles.icon} />}
```
dylanjeffers, line 33: ===

dylanjeffers (approved):
looks great, just one callout with the ===

packages/stems/src/components/SelectablePill/HarmonySelectablePill.tsx
schottra: I read this as intending to catch `null` or `undefined`. But I think the general pattern has been to just booleanize things? As there aren't really any values you can pass as a node which are falsey and we would actually _want_ to render.
`{IconComponent ? <IconComponent ... /> : null}`

dylanjeffers: ohh i see! yeah i prefer your last option (though @nicoback likes to have null come first, which i can get behind) what do you think @nicoback ?

PR #3805, opened 2023-07-26
Update typography component to use classnames

Kyle-Shanks:
### Description
Update the component to use the classnames that were already defined from before. makes things a little cleaner in the file.
Also needed to update the references to the classnames that were used in the code base.
This looks a little weird now, but will be replaced with the component in the future so this won't be an issue

### Dragons

slight risk of incorrect styles in the interim before the component is used. shouldn't be a big issue though

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A

### DIFF
--- a/packages/web/src/components/typography/constants.ts
+++ b/packages/web/src/components/typography/constants.ts
@@ -1,9 +1,4 @@
-import {
-  FontWeight,
-  TextVariant,
-  TextVariantSizeInfo,
-  TextVariantStrengthInfo
-} from './types'
+import { FontWeight, TextVariant, VariantSizeTagMap } from './types'
 
 export const fontWeightMap: Record<FontWeight, number> = {
   heavy: 900,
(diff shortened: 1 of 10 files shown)


audius-infra:
Preview this change https://demo.audius.co/kj-Update-typography-component-to-use-classnames

packages/web/src/components/typography/constants.ts
```diff
@@ -16,155 +11,36 @@ export const fontWeightMap: Record<FontWeight, number> = {
...
-      textTransform: 'uppercase'
-    },
-    XS: {
-      tag: 'label',
-      fontSize: '10px',
-      lineHeight: '12px',
-      letterSpacing: '0.5px',
-      textTransform: 'uppercase'
-    }
+    XLarge: 'label',
```
dylanjeffers, line 34: so just a headsup, "label" element has to be associated with an "input" element, just making sure that would actually be the case for all label types. otherwise span/p or whatever could just work

dylanjeffers (approved):
sick dude! and yeah in addition to using the text component for upload, would be great to also update cases that use the type styles :)

packages/web/src/components/typography/constants.ts
Kyle-Shanks: yea I assumed that's what these would be used for. can change later if we want to render something different 👍

PR #3815, opened 2023-07-28
[PAY-1660] Fix layout issues with TrackTile socials row with a lot of stats

schottra:
### Description
The stats row gets large if you happen to be following multiple people which reposted and favorited a track. The styling for the socials row includes a margin to make space for the plays. But it wasn't set up to restrict its content to the available space.

Also for small track tiles, we change the height of the row to 18px, which causes the plays count to be off-center since it is rendered absolutely. There is longer term cleanup needed on the socials row so that the content isn't sized larger than its container and centered. But it wasn't a safe change to make due to the multiple places we use the `Stats` component. This fixes the display issues for now, and we will address it more fully later.

### Dragons
N/A

### How Has This Been Tested?
Verified locally on Chrome

N/A

### Feature Flags ###
N/A

### Screenshots

**Before**
Screenshot 2023-07-28 at 2 35 56 PM

---

**After**
Screenshot 2023-07-28 at 2 36 37 PM

### DIFF
--- a/packages/web/src/components/track/desktop/TrackTile.module.css
+++ b/packages/web/src/components/track/desktop/TrackTile.module.css
@@ -254,9 +254,15 @@
   margin-right: var(--unit-20);
 }
 
+.small .bottomRight {
+  /* Match socialsRow to keep these in line */
+  height: 18px;
+}
+
 .socialInfo {
   display: inline-flex;
-  width: 100%;
+  flex: 1 1 100%;
+  min-width: 0;
   justify-content: flex-start;
 }
 
--- a/packages/web/src/components/track/desktop/ConnectedTrackTile.module.css
+++ b/packages/web/src/components/track/desktop/ConnectedTrackTile.module.css
@@ -81,7 +81,8 @@
 
 .socialInfo {
   display: inline-flex;
-  width: 100%;
+  flex: 1 1 100%;
+  min-width: 0;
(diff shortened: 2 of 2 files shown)


audius-infra:
Preview this change https://demo.audius.co/pay-1660-fix-stats-overlap

amendelsohn (approved):
Ah yeah that looks better 😁

amendelsohn:
I see that the reposts and favorites are granted different amounts of space. Is that by design? Maybe can fix the flex grow ratio somehow?

schottra:
> I see that the reposts and favorites are granted different amounts of space. Is that by design? Maybe can fix the flex grow ratio somehow?

I noticed this as well. The flex behavior is inside the `Stats` component, which I don't want to touch at the moment. I did a little fiddling with it and it's not straightforward to get them to behave equally due to the overflow behavior.

packages/web/src/components/track/desktop/TrackTile.module.css
```diff
@@ -254,9 +254,15 @@
   margin-right: var(--unit-20);
 }
 
+.small .bottomRight {
+  /* Match socialsRow to keep these in line */
+  height: 18px;
```
dharit-tan, line 259: var?

dharit-tan (approved):

PR #3818, opened 2023-07-28
[C-2742] Multi-track form pagination

amendelsohn:
### Description

Adds:
- Pagination for multiple tracks on the edit form
- Controlled fields to switch values
- Bugfixes for modal subform interactions with hide remix and availability

Out of scope:
- Layout & styling for navigation controls
- Floating track navigation sidebar
	- currently placeholder prev/next buttons

*Two paginated tracks on the same form:*
 

### Dragons

Make sure the `TabInput` changes don't break other usages elsewhere

### How Has This Been Tested?

local web

### Feature Flags ###

`UPLOAD_REDESIGN_ENABLED`


packages/web/src/components/data-entry/TagInput.js
```diff
@@ -38,6 +38,9 @@ class TagInput extends Component {
     if (this.state.typingMode && this.newTagInputRef.current) {
       this.newTagInputRef.current.focus()
     }
+    if (this.state.tags !== this.props.tags) {
```
dylanjeffers, line 41: is the performance okay with this? ive seen controlled components use the prop directly on the element to prevent needing a rerender

packages/web/src/components/form-fields/TagField.tsx
```diff
@@ -10,10 +10,11 @@ type TagFieldProps = Partial<TagInputProps> & {
 }
 export const TagField = (props: TagFieldProps) => {
   const { name, ...other } = props
-  const [field, , { setValue }] = useField(name)
+  const [field, , { setValue }] = useField<string>(name)
   const { value, ...otherField } = field
 
-  const tags = value.split(',').filter(removeNullable)
+  const tagList = (value ?? '').split(',').filter(removeNullable)
```
dylanjeffers, line 16: should we use-memo these?

packages/web/src/components/form-fields/TagField.tsx
```diff
@@ -22,7 +23,8 @@ export const TagField = (props: TagFieldProps) => {
 
   return (
     <TagInput
-      defaultTags={tags}
+      defaultTags={tagList}
```
dylanjeffers, line 26: since we are controlling the input, lets remove defaultTags

packages/web/src/components/form-fields/TextAreaField.tsx
```diff
@@ -8,9 +8,11 @@ type TextAreaFieldProps = TextAreaV2Props & {
...
 export const TextAreaField = (props: TextAreaFieldProps) => {
   const { name, ...other } = props
-  const [field, meta] = useField(name)
+  const [{ value, ...field }, meta] = useField(name)
 
   const hasError = Boolean(meta.touched && meta.error)
 
-  return <TextAreaV2 {...field} error={hasError} {...other} />
+  return (
+    <TextAreaV2 value={value ?? ''} {...field} error={hasError} {...other} />
```
dylanjeffers, line 16: lets not do this, instead ensure value is always controlled and at least empty string in the initialValues. lmk if that makes sense.

packages/web/src/pages/upload-page/components/EditPageNew.tsx
```diff
@@ -1,14 +1,18 @@
-import { useCallback, useMemo } from 'react'
+import { useCallback, useState, useMemo } from 'react'
 
-import { ExtendedTrackMetadata, Nullable } from '@audius/common'
-import { Button, ButtonType, IconArrow } from '@audius/stems'
+import { HarmonyButton, HarmonyButtonType, IconArrow } from '@audius/stems'
```
dylanjeffers, line 3: nice

packages/web/src/pages/upload-page/components/EditPageNew.tsx
```diff
@@ -47,53 +43,71 @@ const EditTrackSchema = Yup.object().shape({
...
-      tags: '',
-      licenseType: {
-        allowAttribution: null,
-        commercialUse: null,
-        derivativeWorks: null
-      }
+      trackMetadatas: tracks.map((track) => ({
+        ...track.metadata,
+        artwork: null,
+        description: '',
```
dylanjeffers, line 51: so for example i added '' for description to ensure we dont get the "controlled/uncontrolled" warning in the console.

packages/web/src/pages/upload-page/components/TrackModalArray.tsx
```diff
@@ -6,14 +6,19 @@ import { TrackAvailabilityModalForm } from '../forms/TrackAvailabilityModalForm'
...
+export const TrackModalArray = (props: TrackModalArrayProps) => {
+  const { index } = props
   return (
     <div className={styles.root}>
-      <ReleaseDateModalForm />
-      <RemixModalForm />
-      <SourceFilesModalForm />
-      <TrackAvailabilityModalForm />
-      <AttributionModalForm />
+      <ReleaseDateModalForm index={index} />
```
dylanjeffers, line 17: cool how easy that was to update

audius-infra:
Preview this change https://demo.audius.co/amendel-upload-multi-track-2

packages/web/src/pages/upload-page/fields/MultiTrackSidebar.tsx
```diff
@@ -0,0 +1,31 @@
...
+import layoutStyles from 'components/layout/layout.module.css'
+
+type MultiTrackSidebarProps = {
+  index: number
+  limit: number
+  setIndex: (index: number) => void
+}
+
+export const MultiTrackSidebar = (props: MultiTrackSidebarProps) => {
+  const { index, setIndex, limit } = props
```
dylanjeffers, line 13: just something to consider long term, if we can try to understand what the best api would be for this element. do we use `onChange` as the prop instead of setIndex? or maybe `onNext` `onPrevious`? basically trying to be as clear as we can with what the expected usage is for a component, and minimize the number of actions a component can allow. this is all more philosophical, so curious your thoughts.

packages/web/src/pages/upload-page/fields/MultiTrackSidebar.tsx
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <div className={cn(layoutStyles.col, layoutStyles.gap2)}>
+      Track {index + 1} of {limit}
```
dylanjeffers, line 16: probably messages object?

packages/web/src/pages/upload-page/fields/TrackMetadataFields.tsx
```diff
@@ -24,16 +26,24 @@ const messages = {
...
 
-export const TrackMetadataFields = () => {
+type TrackMetadataFieldsProps = {
+  /** Whether or not the preview is playing. */
+  playing: boolean
+  index: number
+}
+
+export const TrackMetadataFields = (props: TrackMetadataFieldsProps) => {
+  const { index } = props
```
dylanjeffers, line 36: would we want to put index in some context so we dont have to pass a bunch of stuff down? not convinced that is a good idea, but i wonder if establishing a multi-track hierarchy via context some other mech would be compelling?

dylanjeffers: maybe a `useSomethingField` that has context of index and can handle it all for us? that way we dont have to pass index around and call `getTrackFieldName` for every field

packages/web/src/pages/upload-page/forms/AttributionModalForm.tsx
```diff
@@ -91,28 +91,42 @@ const derivativeWorksValues = [
 type AttributionFormValues = {
   [IS_AI_ATTRIBUTED]: boolean
   [AI_USER_ID]?: number
+  [ISRC]: string
```
dylanjeffers, line 94: nice

packages/web/src/pages/upload-page/forms/AttributionModalForm.tsx
```diff
@@ -262,6 +276,8 @@ const AttributionModalFields = () => {
             <SegmentedControl
               defaultSelected={allowAttribution}
```
dylanjeffers, line 277: so typically should remove "defaultValue" if using a controlled value. not sure if aware but `defaultValue` is a way to set a default for an uncontrolled component.

packages/web/src/pages/upload-page/forms/RemixModalForm.tsx
```diff
@@ -95,11 +101,11 @@ export const RemixModalForm = () => {
     set(
       initialValues,
       IS_REMIX,
-      !!remixOfValue?.tracks.some((track) => !!track)
+      !!remixOfValue?.tracks.some((track: number) => !!track)
```
dylanjeffers, line 104: can we type the track in the useField instantation?

dylanjeffers (approved):
wow great job threading through this feature. i think required item to update is removing defaultValue, and would also love a discussion about how to incorporate index in the upload form

packages/web/src/pages/upload-page/fields/MultiTrackSidebar.tsx
amendelsohn: my thinking is setIndex or an onChange will be correct because we select by index in the sidebar, and eventually the next/prev buttons move to the bottom of the form

packages/web/src/pages/upload-page/fields/TrackMetadataFields.tsx
amendelsohn: I thought about context, but field hooks is another interesting idea... I'll think about it

packages/web/src/pages/upload-page/forms/AttributionModalForm.tsx
amendelsohn: yeah makes sense. A bunch of these were "required" but I bet they function fine when passing a value

audius-infra:
Preview this change https://demo.audius.co/amendel-upload-multi-track-2

PR #3822, opened 2023-07-31
Use audius-query in USDC Purchase Drawer

dharit-tan:
### Description
Use audius-query to fetch track data. Turns out I didn't need `status` anyway as the track should always exist (since we only open this drawer from a track tile) but might as well merge this!

### How Has This Been Tested?

Local ios stage

### Feature Flags ###


amendelsohn (approved):
Awesome!

PR #3828, opened 2023-08-01
[PAY-1687] Web: Share tracks, playlists, and albums via Direct Message

rickyrombo:
### Description

Adds support for sharing track tiles, playlist tiles, and album tiles via direct messages. Also adds embed to share modal.

https://github.com/AudiusProject/audius-client/assets/3690498/998b99a8-2a98-4e5d-9e99-ce687f7f1209

### Dragons

### How Has This Been Tested?

Tested locally against stage

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/mjp-dms-share-track

packages/web/src/components/share-modal/ShareModal.tsx
```diff
@@ -48,6 +52,19 @@ export const ShareModal = () => {
   const isOwner =
     content?.type === 'track' && account?.user_id === content.artist.user_id
 
+  const handleShareToDirectMessage = useCallback(async () => {
+    if (!content) return
+    dispatch(setVisibility({ modal: 'CreateChat', visible: true }))
+    dispatch(
+      setCreateChatModalState({
+        // Just care about the link
+        presetMessage: (await getTwitterShareText(content, false)).link,
```
dharit-tan, line 63: Will this include "#audius" in the string? is that weird?

packages/web/src/pages/chat-page/components/ChatComposer.tsx
```diff
@@ -86,11 +90,12 @@ export const ChatComposer = (props: ChatComposerProps) => {
 
   // Set focus and clear on new chat selected
   useEffect(() => {
-    if (chatId) {
+    if (chatId !== chatIdRef.current) {
```
dharit-tan, line 93: just confirming you've tested this case, seems like we might be resetting `chatIdRef` to current `chatId` on every load so this would always be false? Assuming you want it to be true when we navigate here from the "share via DM" flow.

dharit-tan (approved):
would be nice to get a screencap of this in action!

packages/web/src/pages/chat-page/components/ChatComposer.tsx
rickyrombo: Want it to be false from initial navigation, but true when you click on the sidebar to change the chat. That way this prevents this effect from clearing the default preset message on first render

packages/web/src/components/share-modal/ShareModal.tsx
rickyrombo: nope, this is just the link part of it!

rickyrombo:
@dharit-tan added screencap!

piazzatron (approved):
Code looks great!! A few visual things I noticed:

- Header doesn't appear centered, icon spacing doesn't match figma (nor does color/capitalization, but not sure what 'version' of modal titles we're on here)
Screen Shot 2023-08-01 at 5 48 07 PM

- Hover state for these buttons is very large (maybe this is a harmony thing, and we punt for now):


Screen Shot 2023-08-01 at 5 48 43 PM

- Spacing above and below "spread the word" CTA doesn't match figma (less space in figma)

- Modal transitions: the compose modal pops up below the share modal, before the share model closes, which feels pretty jarring to me. Any we to ensure the share model closes before the compose modal appears?

Then one last UX nit, which we shouldn't do if is tricky – when we drop into the chats page from this share flow, would be nice if the text input automatically had focus so you can just press enter to send your track

rickyrombo:
@piazzatron :
1. Modal Header: Gonna punt as I think we'll want to do a full modal pass to get all modals to match that style (this has been a mismatch between figma and production for a while, so I assume safe to ignore)
2. Button hover: will leave as-is and wait for design feedback
3. Spacing above and below: Good catch, fixed
4. Modal transitions: There's no easy clean way to do this afaik. Simplest thing would be to put the Share modal _under_ the chat create modal? I can play with it in a follow up but I don't think we should block on this. I think modals generally need a bit of an overhaul to support a modal stack, I'm not in love with my "cancel" hack
5. Autofocus: Will see if I can address separately

audius-infra:
Preview this change https://demo.audius.co/mjp-dms-share-track

PR #3830, opened 2023-08-01
[C-2917] Don't show supporters section if all are deactivated users (web)

nicoback2:
### Description
Before:
If all supporters are deactivated users, the "top supporters" box on user profile shows up but is empty

After:
Web: If all supporters are deactivated users, the "top supporters" box does not show up on user profile
Mobile: Still same as before :( Due to the mobile component's logic (which uses `user.supporters_count` to determine whether to render), it'd be more complex/risky to fix it there - figured it wasn't worth it now since this is a pretty edge-casey bug.

Notes:
The optimal fix would be to exclude deactivated users from the /supporters response, but that is much more time consuming/risky due to the way that query is set up than the quick fix here. Should revisit in the future though. Ticket here: https://linear.app/audius/issue/PAY-1683/[mobile]-empty-top-supporterssupportings-list-on-user-profile-if-all

After:
Screenshot 2023-07-31 at 9 19 14 PM

Before:
Screenshot 2023-08-01 at 9 18 37 AM

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/screens/profile-screen/ProfileHeader/TopSupporters.tsx
+++ b/packages/mobile/src/screens/profile-screen/ProfileHeader/TopSupporters.tsx
@@ -85,7 +85,6 @@ const useSelectTopSupporters = (userId: number) =>
       const topSupporters = topSupporterIds
         .map((id) => supporterUsers[id])
         .filter(removeNullable)
-        .filter((user) => !user.is_deactivated)
       return topSupporters
     },
     [userId]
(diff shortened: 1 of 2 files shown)


audius-infra:
Preview this change https://demo.audius.co/nkang--deactivated-supporters

sddioulde (commented):
this makes sense for this particular part but there will be many other places to filter i.e. where we do ranking logic, where we show num supporters, etc.
I think we should prioritize and do a backend fix here. wdyt?

nicoback2:
@sddioulde yeah that works! i can close this then. Here is the ticket for backend fix https://linear.app/audius/issue/PAY-1683/%5Bmobile%5D-empty-top-supporterssupportings-list-on-user-profile-if-all

PR #3831, opened 2023-08-01
[C-2918] Don't show supporting section on web if all supportings are deactivated users (web)

nicoback2:
### Description
Companion PR: https://github.com/AudiusProject/audius-client/pull/3830

Before:
If all supportings are deactivated users, the "Supporting" sectionon user profile shows up but the content is empty

After:
Web: If all supportings are deactivated users, the "Supporting" section does not show up on user profile
Mobile: Still same as before :( Due to the mobile component's logic (which uses `user.supporting_count` to determine whether to render), it'd be more complex/risky to fix it there - figured it wasn't worth it now since this is a pretty edge-casey bug.

Notes:
The optimal fix would be to exclude deactivated users from the /supporters response, but that is much more time consuming/risky due to the way that query is set up than the quick fix here. Should revisit in the future though.

### Dragons

### How Has This Been Tested?

### Feature Flags ###

### DIFF
--- a/packages/common/src/hooks/useRankedSupporters.ts
+++ b/packages/common/src/hooks/useRankedSupporters.ts
@@ -1,8 +1,11 @@
 import { ID } from '../models/Identifiers'
+import { Supporting } from '../models/Tipping'
+import { cacheUsersSelectors } from '../store/cache'
 import { tippingSelectors } from '../store/tipping'
 import { stringWeiToBN } from '../utils'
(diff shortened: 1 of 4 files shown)


sddioulde (commented):
thx for these changes, but similar comment to https://github.com/AudiusProject/audius-client/pull/3830#pullrequestreview-1557457167 :)

PR #3832, opened 2023-08-01
Fix tag input

amendelsohn:
### Description

Fixes a regression in TagInput that caused tags to be undefined

### DIFF
--- a/packages/web/src/components/data-entry/TagInput.js
+++ b/packages/web/src/components/data-entry/TagInput.js
@@ -34,9 +34,15 @@ class TagInput extends Component {
 
   newTagInputRef = createRef()
 
-  componentDidUpdate = () => {
+  componentDidUpdate = (prevProps, prevState) => {
     if (this.state.typingMode && this.newTagInputRef.current) {
       this.newTagInputRef.current.focus()
+    }
+    if (
+      prevState.tags === this.state.tags &&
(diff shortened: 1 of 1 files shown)


Kyle-Shanks (approved):

packages/web/src/components/data-entry/TagInput.js
```diff
@@ -38,9 +38,6 @@ class TagInput extends Component {
     if (this.state.typingMode && this.newTagInputRef.current) {
       this.newTagInputRef.current.focus()
     }
-    if (this.state.tags !== this.props.tags) {
```
raymondjacobson, line 41: I assume this was here to allow for the component to be "controlled" externally. I imagine how we're using it is different now, but this might be unsafe?

raymondjacobson (approved):

packages/web/src/components/data-entry/TagInput.js
amendelsohn: The only controlled usage is in the new upload form. I just added another round of changes that puts back the controlled behavior without breaking the old usages.

Also okay to just rip it out for now and revisit later.

raymondjacobson: sweet! thanks for clarification

PR #3846, opened 2023-08-02
Add npm run clean script

dharit-tan:
### Description
I find myself wanting to blast all `node_modules` and build folders frequently, this would make it easier.

### Dragons

* is it sketch to have `rm -rf` in a script?
* Could not figure out where the build folder for `web` lives.
* This won't do everything that an `npm run clean` in `mobile` will but that's ok.

### How Has This Been Tested?

Ran locally followed by `npm i`.

### Feature Flags ###


dylanjeffers:
maybe another option is to add a "clean" method to each sub package so it's not all or nothing?

raymondjacobson (approved):
web is

build, build-staging, and build-production

dharit-tan:
> maybe another option is to add a "clean" method to each sub package so it's not all or nothing?

I guess I run into it more with mobile, bc I want to blast mobile, common, and then also have to blast web sometimes due to web common.

rickyrombo:
I also echo doing `rm -rf node_modules & rm -rf packages/*/node_modules` rather than any `cd`s

dharit-tan:
> Personally I would expect `clean` as only meaning build artifacts, not node_modules, so maybe better to have separate `npm run clean` and `npm run clean:modules` or something? The former should probably call upon the mobile `npm run clean` script rather than manually doing `rm -rf`s imo? can do `react-native clean-project-auto` to avoid interactive element: https://www.npmjs.com/package/react-native-clean-project

Good call - the `auto` doesn't include the brew update so that'll work. I like this idea.

package.json
```diff
@@ -9,6 +9,7 @@
     "install-hooks": "mookme init --only-hook --skip-types-selection",
     "lint": "lerna run lint",
     "lint:fix": "lerna run lint:fix",
+    "clean": "rm -rf node_modules && cd packages/web && rm -rf node_modules && cd ../mobile && rm -rf node_modules && rm -rf android/build && rm -rf ios/build && cd ../common && rm -rf node_modules && cd ../stems && rm -rf node_modules && cd ..",
```
sddioulde, line 12: Should the end be "cd ../.."

audius-infra:

rickyrombo (approved):

package.json
rickyrombo: yeah probably unnecessary? tbh not sure whether npm commands change your actual terminal's working dir

dylanjeffers (approved):
dope!

PR #3850, opened 2023-08-03
Fix send audio flow

dharit-tan:
### Description
Fixes send audio flow - we had changed `solanaWeb3Manager` mint type but in this file, `audiusLibs` is `any` so we didn't catch it!

### Dragons

### How Has This Been Tested?

Tested locally against prod

### Feature Flags ###


nicoback2 (approved):

piazzatron (approved):

rickyrombo:
We should probably follow up by replacing this with a libs method? I don't think we should reach into libs internals like this generally

PR #3851, opened 2023-08-03
Add DirectMessages Banner and Update All Banners

rickyrombo:
### Description

Kinda snowballed into a total banner redesign, against my better judgement decided to push through

- Converts all the banners to TypeScript
- Creates `CallToActionBanner` as a generic component that the others use to create banners with actions. Unifies styles to match Figma. Only one that doesn't consume it is the `FanburstBanner`
- Moves as much show/hide logic for each banner to be handled by itself rather than `App.js` (the only one lacking that is the `UpdateAppBanner` because it make sense to have that logic in `App.js` and limits the scope of this change
- (risky part of the change) Modifies the structure of the web app by moving the banners container outside of the main app div, and having it push content down naturally rather than tracking `--banner-margin` throughout our layout
  - This involved making things that were `position: fixed` => `position: absolute`, which works because on desktop web we don't actually scroll the main page, but instead scroll the using the underlying "main content" overflow
  - This made it unnecessary to track a `DesktopHeaderContext `, as instead, we can simply portal the gutter to the parent of the scroll container which will be (1) inside the app div, so moved down by banners if necessary, and (2) outside the main content div, so it won't scroll
  - This caused headaches on mobile web when banners are showing, but I think I mitigated those effectively and we don't have any banners showing on mobile web ever anyway
- Improves desktop web banner experience - previously we were hiding the close icon

And finally:
- Adds new banner announcing DMs that appears when logged in on desktop web (and hide the mobile app download banner on desktop web when logged in)

### Dragons

- This changes how banners appear on mobile web (they no longer push content down)
- This changes how a couple banners look so that they all look the same (Metamask error and Update App banners both have bold instead of medium font weight, all banner text now has 16px padding above/below
- This plays with how things are laid out on web, mobile web, and desktop web. This impacted things like the scroll gutter/headers, the app wide notice, the public landing page, scroll experiences etc. Will post screenshots of everything before/after shortly

### How Has This Been Tested?

Tested mobile web, desktop web, and desktop app all look ok with and without banners, and made sure notices showed up appropriately. Pictures to come

# Desktop Web
UpdateAppBanner is not possible

### New
Screenshot 2023-08-03 at 3 45 27 PM

### Before
Screenshot 2023-08-03 at 3 43 37 PM

### After
Screenshot 2023-08-03 at 3 44 55 PM

### Before
Screenshot 2023-08-03 at 3 46 44 PM

### After
Screenshot 2023-08-03 at 3 47 13 PM

### Before
Screenshot 2023-08-03 at 4 09 30 PM

### After
Screenshot 2023-08-03 at 4 15 48 PM

# Mobile Web

## Web3ErrorBanner
I don't think there's any chance of this happening, but this is the only banner that's even possible on mobile web.

### Before
Screenshot 2023-08-03 at 3 51 29 PM

### After
Screenshot 2023-08-03 at 3 52 17 PM

### Before
Screenshot 2023-08-03 at 4 32 23 PM

### After
Screenshot 2023-08-03 at 4 32 11 PM

# Desktop App

### New
Screenshot 2023-08-03 at 3 59 31 PM

### Before
Screenshot 2023-08-03 at 4 06 22 PM

### After
Screenshot 2023-08-03 at 4 03 13 PM

### Before
Screenshot 2023-08-03 at 4 05 37 PM

### After
Screenshot 2023-08-03 at 4 03 32 PM

# Misc
## Testing notice still works
Screenshot 2023-08-03 at 4 20 12 PM


audius-infra:
Preview this change https://demo.audius.co/mjp-dms-banner

packages/web/src/components/banner/DirectMessagesBanner.tsx
```diff
@@ -0,0 +1,51 @@
...
+const messages = {
+  text: 'Direct Messaging Now Available!',
+  pill: 'New'
+}
+
+const { getHasAccount } = accountSelectors
+
+const DIRECT_MESSAGES_BANNER_LOCAL_STORAGE_KEY = 'dismissDirectMessagesBanner'
+
+export const DirectMessagesBanner = () => {
```
piazzatron, line 22: Let's add analytics for this banner; would like to see how many times it's seen as well as how many ppl click through/close it

packages/web/src/components/banner/DirectMessagesBanner.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const dispatch = useDispatch()
+  const hasAccount = useSelector(getHasAccount)
+  const [isVisible, setIsVisible] = useState(
+    !window.localStorage.getItem(DIRECT_MESSAGES_BANNER_LOCAL_STORAGE_KEY) &&
+      ((getClient() === Client.DESKTOP && hasAccount) ||
```
piazzatron, line 27: If you're logged out on electron, we still show?

packages/web/src/components/banner/DirectMessagesBanner.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const dispatch = useDispatch()
+  const hasAccount = useSelector(getHasAccount)
+  const [isVisible, setIsVisible] = useState(
+    !window.localStorage.getItem(DIRECT_MESSAGES_BANNER_LOCAL_STORAGE_KEY) &&
+      getClient() === Client.DESKTOP &&
+      hasAccount
+  )
+
+  const handleAccept = useCallback(() => {
+    dispatch(pushRoute(CHATS_PAGE))
```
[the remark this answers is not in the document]
piazzatron: Also wonder if we should dismiss the banner here - question for design

packages/web/src/components/banner/FanburstBanner.tsx
```diff
@@ -12,27 +11,29 @@ const messages = {
...
     <Banner
-      {...props}
-      className={cn(styles.banner, { [styles.isMobile]: props.isMobile })}
+      isMobile={isMobile}
+      onClose={onClose}
+      className={cn(styles.banner, { [styles.isMobile]: isMobile })}
     >
       <div className={styles.text}>
         <span>
-          {props.isMboile ? messages.mobileHome : messages.home}
```
piazzatron, line 23: `isMboile` 😆

packages/web/src/components/banner/DownloadAppBanner.tsx
```diff
@@ -0,0 +1,47 @@
...
+  text: 'Download the Audius App',
+  pill: 'New'
+}
+export const DownloadAppBanner = () => {
+  const dispatch = useDispatch()
+  const hasAccount = useSelector(getHasAccount)
+  const [isVisible, setIsVisible] = useState(
+    !window.localStorage.getItem(MOBILE_BANNER_LOCAL_STORAGE_KEY) &&
+      getClient() === Client.DESKTOP &&
+      !hasAccount
```
piazzatron, line 26: let's document the logged in vs logged out banner viz behavior between this and the DMs banner somewhere 🙏

packages/web/src/components/banner/AppBannerWrapper.tsx
```diff
@@ -0,0 +1,15 @@
+import cn from 'classnames'
+
+import { isMobile } from 'utils/clientUtil'
+
+import styles from './AppBannerWrapper.module.css'
+
+export const AppBannerWrapper = ({
```
piazzatron, line 7: Since we don't have any banners on mobile web, do we need this component? Could we just put these styles on the base banner CTA? I wonder if we shouldn't support this at all

packages/web/src/pages/App.js
```diff
@@ -466,52 +439,36 @@ class App extends Component {
...
-              isElectron={client === Client.ELECTRON}
-              onClose={this.dismissWeb3ErrorBanner}
-            />
-          ) : null}
+      <div className={styles.root}>
+        <AppBannerWrapper>
+          <DownloadAppBanner />
+          <DirectMessagesBanner />
+          <Web3ErrorBanner />
+          {/* Other banners' logic is self-contained, but since this one uses the IPC
```
piazzatron, line 451: What happens if this banner is shown while the DMs banner is shown?

packages/web/src/pages/App.js
```diff
@@ -466,52 +439,36 @@ class App extends Component {
...
+            and can result in either required or optional updates, keeping the visibility 
+            controlled from this parent for now */}
           {showWebUpdateBanner ? (
             <UpdateAppBanner
               onAccept={this.acceptWebUpdate}
               onClose={this.dismissUpdateWebAppBanner}
             />
           ) : null}
+        </AppBannerWrapper>
+        {this.props.isChatEnabled ? <ChatListener /> : null}
```
piazzatron, line 461: Should the DMs banner also care about isChatEnabled? Or maybe we just lose this here?

packages/web/src/index.css
```diff
@@ -159,8 +159,7 @@ body {
 
 html,
 body,
-#root,
-#root > div {
```
rickyrombo, line 163: Note: This change means `App`, `ToastContext`, `SomethingWrong`, `RequiresUpdate` and any other divs directly descending from root (please tell me if there's more, I probably missed one) need to _set these styles themselves_, but I ultimately think this is best left to be "non-magical"

packages/web/src/components/banner/FanburstBanner.tsx
```diff
@@ -12,27 +11,27 @@ const messages = {
...
```
rickyrombo, line 23: the joy of JavaScript 😢

piazzatron (approved):
Generally looks great - have a few specific questions. Definitely want another pair of eyes to QA this – I did poke around, and didn't notice anything out of the ordinary.

Small things:
- Pill differs slightly from Figma (I'm guessing this is a Harmony thing, and can ignore)
- Slightly different spacing I think between icon + text vs figma
Screen Shot 2023-08-03 at 7 12 16 PM

- If you start logged out and then log in, we still show the app download banner. Very minor

packages/web/src/components/banner/DirectMessagesBanner.tsx
```diff
@@ -0,0 +1,51 @@
...
+const { getHasAccount } = accountSelectors
+
+const DIRECT_MESSAGES_BANNER_LOCAL_STORAGE_KEY = 'dismissDirectMessagesBanner'
+
+export const DirectMessagesBanner = () => {
+  const dispatch = useDispatch()
+  const hasAccount = useSelector(getHasAccount)
+  const [isVisible, setIsVisible] = useState(
+    !window.localStorage.getItem(DIRECT_MESSAGES_BANNER_LOCAL_STORAGE_KEY) &&
+      getClient() === Client.DESKTOP &&
```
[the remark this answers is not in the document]
piazzatron: outdated comment

packages/web/src/components/banner/DirectMessagesBanner.tsx
rickyrombo: yeah good catch, was thinking it would take you to signup but it doesn't

packages/web/src/components/banner/AppBannerWrapper.tsx
rickyrombo: Yeah this is a gross leftover remnant from prototyping/testing, kinda want to leave it so at least we have the option of banners on mobile web? The _real_ goal was making it completely separate (ie not take children) and have the banner list self contained inside of it, and have it decide which to show but that proved to hard to separate from the required update logic cleanly

packages/web/src/pages/App.js
rickyrombo: They all stack in front of each other - whichever is the last one goes on top, and then they dismiss down to the lowest one.

We also have the option of making them stack vertically, though. I chose to keep the existing behavior

packages/web/src/pages/App.js
rickyrombo: Very good catch - I don't think we need a DMs kill switch, maybe a follow up pr to strip out the checks for `isChatEnabled` is in order

audius-infra:
Preview this change https://demo.audius.co/mjp-dms-banner

packages/web/src/components/banner/DownloadAppBanner.tsx
```diff
@@ -0,0 +1,47 @@
...
+  )
+
+  const handleAccept = useCallback(() => {
+    dispatch(setAppModalCTAVisibility({ isOpen: true }))
+  }, [dispatch])
```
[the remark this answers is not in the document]
rickyrombo: ie it should close on accept, right?

packages/web/src/components/banner/DirectMessagesBanner.tsx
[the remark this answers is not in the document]
rickyrombo: yea good design qs

audius-infra:
Preview this change https://demo.audius.co/mjp-dms-banner

packages/web/src/components/banner/DownloadAppBanner.tsx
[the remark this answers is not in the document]
piazzatron: yeah. fwiw was less sure about it closing (sounds like design wants that), more that we should add the key to local storage so the next session we don't show it again if you've clicked thru

packages/web/src/components/banner/AppBannerWrapper.tsx
piazzatron: Yeah makes sense. Don't mind leaving it in

packages/web/src/pages/App.js
piazzatron: Ooh, interesting. I imagine we probably only ever want to show one banner at a time, stacking them is weird - was the current behavior to stack?

PR #3857, opened 2023-08-08
PLAT-1171: discovery relay flag

alecsavvy:
### Description
adds optimizely flags to control if a user of the client will use their selected discovery node or identity for relay operations

### Dragons
I'm not sure client side if waiting for optimizely initialization is supposed to be done in the init sdk function. Feels a little odd. Is there a better way to do this?

### How Has This Been Tested?

### Feature Flags ###
Two optimizely flags:
`discovery_relay` - user specific feature flag to enable discovery relay, this is good for a subset of users to try it out before the entire network starts using it
`USE_DISCOVERY_RELAY` - network wide feature flag, if this is set to true and the previously mentioned flag is off this one will override.


audius-infra:
Preview this change https://demo.audius.co/as/plat-1171-discovery-relay-flag

dylanjeffers (approved):
looks great, heads up we are using .info instead of .log for things that you want to have appear in the console, also can you revert the embed package-lock?

alecsavvy:
closing post monorepo + libs change

PR #3864, opened 2023-08-09
Add library albums and playlists audius-query hook + migrate collection reformat util; bump SDK PAY-1679

nicoback2:
### Description
-Add audius-query hook for getting library albums
-Migrate collection `reformat` util to common
-Some type + ResponseAdapter fixing

Note this is going into main, not feature branch. See feature branch PR here: 
https://github.com/AudiusProject/audius-client/pull/3865

BEFORE MERGING:
-Merge protocol PR 
-Generate SDK
-Bump SDK ver in client

### Dragons

### How Has This Been Tested?

### Feature Flags ###


packages/common/src/api/library.ts
```diff
@@ -0,0 +1,74 @@
...
+            audiusBackendInstance: audiusBackend,
+            omitUser: false
+          })
+        )
+        return albums
+      },
+      options: {
+        kind: Kind.COLLECTIONS,
+        schemaKey: 'collections',
+        idListArgKey: 'playlist_id'
```
amendelsohn, line 68: We don't need `idListArgKey` here.

`idListArgKey` is only useful if you have a hook argument containing the ids of the results you expect. It allows cache hits before making the initial fetch.

Example:
`fetchCollections(collection_ids: ID[])`
`idArgKey: 'collection_ids'`
usage: `useFetchCollections({ collection_ids: ['123', '456'] })` now if collections 123 and 456 are found in the cache, we skip the initial fetch.

However, for this case, we don't know upfront what ids we are going to need, so we always have to fetch and construct the entity list based on the response.

packages/common/src/api/library.ts
```diff
@@ -0,0 +1,74 @@
...
+          .filter(removeNullable) as UserCollectionMetadata[]
+        const albums = albumsMetadata.map((am) =>
+          reformatCollection({
+            collection: am,
+            audiusBackendInstance: audiusBackend,
+            omitUser: false
+          })
+        )
+        return albums
+      },
```
amendelsohn, line 64: Amazing! This is a perfect use of audius-query 🎉

packages/common/src/api/library.ts
```diff
@@ -0,0 +1,74 @@
...
+            sortDirection,
+            type: 'all',
+            encodedDataMessage: data,
+            encodedDataSignature: signature
+          })
+        const albumsMetadata = rawAlbums
+          .map((r: APIActivityV2) => makeActivity(r))
+          .filter(removeNullable) as UserCollectionMetadata[]
+        const albums = albumsMetadata.map((am) =>
+          reformatCollection({
```
amendelsohn, line 57: Nice. This makes me wonder if it's worth building in the reformatters for each entity type? Do you know if we always want to transform the data for every instance of collection?

amendelsohn (approved):

packages/web/src/common/store/cache/collections/utils/processAndCacheCollections.ts
```diff
@@ -40,7 +40,7 @@ export function* processAndCacheCollections(
   yield addTracksFromCollections(collections)
 
   let reformattedCollections = collections.map((c) =>
-    reformat(c, audiusBackendInstance)
+    reformatCollection({ collection: c, audiusBackendInstance })
```
amendelsohn, line 43: Oh wait, I see. The cache already does this, so by calling `addEntries` in audius-query, we already do it.

Have you tried it without calling this in the endpoint's fetch implementation? It probably helps the types match up, but I think the data should still be there if you don't. We should be saving the results to cache which internally calls `reformatCollection` then selecting the cache data as the return from the hook.

packages/common/src/api/library.ts
```diff
@@ -0,0 +1,74 @@
...
+          userId,
+          offset,
+          limit,
+          query = '',
+          sortMethod = 'added_date',
+          sortDirection = 'desc'
+        } = args
+        const sdk = await audiusSdk()
+        const { data, signature } =
+          await audiusBackend.signDiscoveryNodeRequest()
```
schottra, line 39: Do we have to do this for every sdk request? Feels like it should be baked into the sdk :-/

nicoback2: yeah 100%, it's a follow up task for the SDK https://linear.app/audius/issue/C-2928/enable-sdk-reads-for-authenticated-routes-[follow-on]

packages/common/src/api/library.ts
nicoback2: ahh thanks for the clarification

audius-infra:

audius-infra:

audius-infra:
Preview this change https://demo.audius.co/nkang--albums-query

PR #3866, opened 2023-08-09
Preserve CIDs for track and collection cover arts

rickyrombo:
### Description

Before this PR, edits to tracks and playlists would result in the `cover_image_sizes` prop on the entity being set to the local URL of the file (eg. blob:// or file://). This was so that we could load the image optimistically on mobile. However, the downside was that the local URL would also get written on a subsequent update to the track and break the cover art.

This changes makes it so that we never touch the `cover_art_sizes` field, even on mobile, and updates mobile to check for overrides in `_cover_art_sizes` instead for optimistic updates.

### DIFF
--- a/packages/mobile/src/components/image/CollectionImage.tsx
+++ b/packages/mobile/src/components/image/CollectionImage.tsx
@@ -24,7 +24,11 @@ type UseCollectionImageOptions = {
   collection: Nullable<
     Pick<
       Collection,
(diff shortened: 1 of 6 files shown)


audius-infra:
Preview this change https://demo.audius.co/mjp-fix-track-images

dylanjeffers (approved):
nice, probably just a matter of ensuring the types work?

PR #3874, opened 2023-08-11
[PAY-1701] Fix "Share to DMs" to work through InboxUnavailableModal

rickyrombo:
### Description

Might be useful to look at commit by commit

- Migrates `CreateChatModal` to use the `createModal` helper, making this easier
- Migrates `InboxUnavailableModal` to use `createModal` helper, making this easier
- Updates `SearchUserModal` to not take a `modalName` etc and just accept `isOpen`
- Refactors the tipping integration to be less coupled and take in a pair of "callback" actions `onSuccess` and `onSuccessConfirmed` which are used to navigate the user to the chat and create the chat respectively in the InboxUnavailable flow.

### Dragons

### How Has This Been Tested?

Tested E2E on web against stage. Mobile has a slightly different impl, looking at that next and will try to unify

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/mjp-dms-share-2

packages/web/src/components/search-users-modal/SearchUsersModal.tsx
```diff
@@ -104,10 +99,9 @@ export const SearchUsersModal = (props: SearchUsersModalProps) => {
   )
 
   const handleClose = useCallback(() => {
-    setIsVisible(false)
     setQuery('')
-    onClose?.()
-  }, [setIsVisible, onClose])
+    onClose()
```
raymondjacobson, line 111: can this still be undefined?

packages/web/src/components/inbox-unavailable-modal/InboxUnavailableModal.tsx
```diff
@@ -60,7 +65,11 @@ const actionToContent = (
     case ChatPermissionAction.TIP:
       return {
         content: messages.tipContent(
-          <UserNameAndBadges user={user} onNavigateAway={onClose} />
+          user ? (
+            <UserNameAndBadges user={user} onNavigateAway={onClose} />
+          ) : (
+            'this user'
```
raymondjacobson, line 74: should be in a `messages`?

packages/web/src/components/inbox-unavailable-modal/InboxUnavailableModal.tsx
```diff
@@ -83,55 +92,50 @@ const actionToContent = (
...
   const { callToAction } = useSelector((state) =>
-    getCanCreateChat(state, { userId: user.user_id })
+    getCanCreateChat(state, { userId: user?.user_id })
   )
   const hasAction =
     callToAction === ChatPermissionAction.TIP ||
     callToAction === ChatPermissionAction.UNBLOCK
 
   const handleClick = useCallback(() => {
+    if (!user) return
```
raymondjacobson, line 111: can this ever happen? maybe worth a log

packages/web/src/pages/chat-page/components/ChatHeader.tsx
```diff
@@ -53,7 +58,7 @@ export const ChatHeader = forwardRef<HTMLDivElement, ChatHeaderProps>(
     ref
   ) => {
     const dispatch = useDispatch()
-    const [, setCreateChatVisible] = useModalState('CreateChat')
+    const { onOpen: openCreateChatModal } = useCreateChatModal()
```
raymondjacobson, line 61: it's a little weird that this is called `onOpen` versus `open`

packages/web/src/common/store/tipping/sagas.ts
```diff
@@ -385,15 +391,12 @@ function* sendTipAsync() {
       yield* put(
         fetchPermissions({ userIds: [sender.user_id, recipient.user_id] })
       )
+      if (onSuccessConfirmedAction) {
```
raymondjacobson, line 394: wtf this is wild

raymondjacobson (approved):
🙆‍♂️

packages/web/src/components/inbox-unavailable-modal/InboxUnavailableModal.tsx
rickyrombo: techincally user is undefined when the modal is closed, but should ever get this far no - adding a console error

packages/web/src/components/search-users-modal/SearchUsersModal.tsx
rickyrombo: no

packages/web/src/pages/chat-page/components/ChatHeader.tsx
rickyrombo: I feel the same way about `onClose`, should I change them to `open` and `close`?

raymondjacobson: ye i think so, but ok if in sep pr for me

packages/common/src/store/tipping/slice.ts
```diff
@@ -104,6 +104,8 @@ const slice = createSlice({
         user: User | null
         source: TipSource
         trackId?: ID
+        onSuccessActions?: Action[]
```
nicoback2, line 107: Might be worth a comment to clarify the difference between these / what they mean ^ v since it is not clear from the names

packages/common/src/store/ui/inbox-unavailable-modal/index.ts
```diff
@@ -0,0 +1,26 @@
+import { Action } from '@reduxjs/toolkit'
+
+import { User } from 'models/User'
+
+import { createModal } from '../modals/createModal'
+
+export type InboxUnavailableModalState = {
+  user?: User
```
nicoback2, line 8: does the whole user object need to be passed (guess I'll see further down in the PR but just commenting as I go hehe)?

audius-infra:
Preview this change https://demo.audius.co/mjp-dms-share-2

packages/common/src/store/tipping/slice.ts
nicoback2: Also 2 qs:
-why have these be actions and not just normal functions? (in which an action(s) is dispatcheD)
-why have one be an array and one be just one action?

nicoback2 (approved):
Nicee LGTM! no need to make changes before merging but left some comments

packages/common/src/store/tipping/slice.ts
rickyrombo: All very good points/concerns!
1. `onSuccessAction` is an action that fires when the tip is sent, `onSuccessConfirmedAction` is an action that fires when the tip is indexed by DN. I'll add comments!
2. It's best practice to not put functions in redux, because they're non-serializable. Although it did cause me great pain to work around that 😢 . We can definitely do it, it technically does work, but it breaks time traveling and possibly other things like rehydration etc.
3. Because I'm lazy and there's no usecases for the other one to be more than one action (yet) 😅, otherwise they'd both be arrays

packages/common/src/store/ui/inbox-unavailable-modal/index.ts
rickyrombo: ooh very good point, we should probably just pass ID and fetch the user from the cache when needed

packages/common/src/store/tipping/slice.ts
rickyrombo: more on (2) here https://redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state

audius-infra:
Preview this change https://demo.audius.co/mjp-dms-share-2

PR #3877, opened 2023-08-12
[PAY-1702] Use existing chats as default user list when sharing to DMs

rickyrombo:
### Description

Use the users existing chat list instead of their followers as the default user list when opening DMs via the Share dialog

### Dragons

### How Has This Been Tested?

Tested mobile and web

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/mjp-dms-default-to-chats

packages/common/src/store/pages/chat/selectors.ts
```diff
@@ -254,6 +256,18 @@ export const getCanCreateChat = createSelector(
       }
     }
 
+    // Check for existing chat and short circuit if exists
+    // Note: this only works if the respective chat has been fetched already
```
rickyrombo, line 260: This prevents the new user list from claiming you can't create a chat with someone, since you're not "creating" a chat if you already have one

packages/common/src/hooks/chats/useChatsUserList.ts
```diff
@@ -0,0 +1,31 @@
...
+} from 'store/pages/chat/selectors'
+import { decodeHashId } from 'utils/hashIds'
+import { removeNullable } from 'utils/typeUtils'
+
+export const useChatsUserList = () => {
+  const currentUserId = useSelector(getUserId)
+  const chats = useSelector(getChats)
+  const hasMoreChats = useSelector(getHasMoreChats)
+  const chatsStatus = useSelector(getChatsStatus)
+  const chatUserListIds = chats
```
schottra, line 18: Any use in memoizing this and/or using the fancy selector helper we wrote to track all of these dependencies and only recompute on change? Or do we only run it once?

packages/web/src/pages/chat-page/components/CreateChatModal.tsx
```diff
@@ -48,10 +50,14 @@ export const CreateChatModal = () => {
 
   const loadMore = useCallback(() => {
     if (currentUser) {
-      dispatch(followersUserListActions.setFollowers(currentUser?.user_id))
-      dispatch(userListActions.loadMore(FOLLOWERS_USER_LIST_TAG))
+      if (defaultUserList === 'chats') {
+        dispatch(fetchMoreChats())
```
dharit-tan, line 54: nit: feels like there's potentially a cleaner way to do this, but i think this is fine since we're probably not gonna extend this functionality beyond these two cases.

dharit-tan (approved):

packages/common/src/store/pages/chat/selectors.ts
```diff
@@ -254,13 +256,24 @@ export const getCanCreateChat = createSelector(
       }
     }
 
+    // Check for existing chat, since unblocked users with existing chats
```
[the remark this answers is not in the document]
rickyrombo: No, I intentionally said unblocked because blocked users still can't message

packages/common/src/hooks/chats/useChatsUserList.ts
rickyrombo: Great call-out, will look into reselect for this. I don't think the other helper works here because it operates on individual chats but let me double check

packages/web/src/pages/chat-page/components/CreateChatModal.tsx
rickyrombo: Yeah you're right, there's likely a cleaner way. But yeah we can cross that bridge later

packages/common/src/store/pages/chat/selectors.ts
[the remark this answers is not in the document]
dharit-tan: sry still confused - if the user is blocked/blocker, wouldn't they get filtered out on line 273/274? And on 275 aren't we checking permissions (not blocker/blockee)?
are you referring to the case where someone was blocked and then unblocked? bc i thought this was referring to the case where user A changed their perms, but if they had a previously existing chat with user B, B would still be able to message A.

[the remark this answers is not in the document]
rickyrombo: Yep that's exactly correct! Maybe my wording here is awkward? I'm trying to say that we check for existing chats in the case where the user is not blocked by/is not blocking the other user, since they don't need permission if they already have a chat (and aren't blocked by or blocking the other user)

PR #3878, opened 2023-08-12
[PAY-1701] Fix "Share to DMs" on mobile to go through InboxUnavailable modal

rickyrombo:
### Description

Follow up for #3874 but for mobile

### Dragons

### How Has This Been Tested?

iOS sim

### Feature Flags ###

### DIFF
--- a/packages/mobile/src/components/inbox-unavailable-drawer/InboxUnavailableDrawer.tsx
+++ b/packages/mobile/src/components/inbox-unavailable-drawer/InboxUnavailableDrawer.tsx
@@ -9,7 +9,8 @@ import {
   CHAT_BLOG_POST_URL,
   accountSelectors,
   makeChatId,
-  useInboxUnavailableModal
+  useInboxUnavailableModal,
+  cacheUsersSelectors
 } from '@audius/common'
 import { View } from 'react-native'
 import { useDispatch, useSelector } from 'react-redux'
@@ -108,8 +109,10 @@ const DrawerContent = ({ data, onClose }: DrawerContentProps) => {
   const dispatch = useDispatch()
   const navigation = useNavigation()
 
-  const { user, presetMessage } = data
-  const userId = user?.user_id
+  const { userId, presetMessage } = data
+  const user = useSelector((state) =>
+    cacheUsersSelectors.getUser(state, { id: userId })
+  )
   const { callToAction } = useSelector((state) =>
     getCanCreateChat(state, { userId })
   )
@@ -151,10 +154,12 @@ const DrawerContent = ({ data, onClose }: DrawerContentProps) => {
             presetMessage
           })
         ],
-        onSuccessConfirmedAction: chatActions.createChat({
-          userIds: [user.user_id],
-          skipNavigation: true
-        })
+        onSuccessConfirmedActions: [
+          chatActions.createChat({
+            userIds: [user.user_id],
+            skipNavigation: true
+          })
+        ]
       })
     )
     navigation.navigate('TipArtist')
--- a/packages/mobile/src/components/inbox-unavailable-drawer/InboxUnavailableDrawer.tsx
+++ b/packages/mobile/src/components/inbox-unavailable-drawer/InboxUnavailableDrawer.tsx
@@ -7,8 +7,11 @@ import {
   tippingActions,
   cacheUsersSelectors,
   ChatPermissionAction,
-  CHAT_BLOG_POST_URL
+  CHAT_BLOG_POST_URL,
+  accountSelectors,
+  makeChatId
 } from '@audius/common'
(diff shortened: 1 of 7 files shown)


packages/mobile/src/components/inbox-unavailable-drawer/InboxUnavailableDrawer.tsx
```diff
@@ -102,48 +100,71 @@ const useStyles = makeStyles(({ spacing, typography, palette }) => ({
...
-    )
-  }, [dispatch])
+  const currentUserId = useSelector(accountSelectors.getUserId)
 
   const handleUnblockPress = useCallback(() => {
-    dispatch(unblockUser({ userId }))
-    if (shouldOpenChat && canCreateChat) {
-      dispatch(createChat({ userIds: [userId] }))
+    if (!userId) {
+      console.error(
```
schottra, line 123: Just double-checking this is sufficient handling of the error. Does the user see anything? Will we get a message in sentry or some other type of analytics. It reads as a silent failure as is.

packages/mobile/src/components/inbox-unavailable-drawer/InboxUnavailableDrawer.tsx
```diff
@@ -102,48 +100,71 @@ const useStyles = makeStyles(({ spacing, typography, palette }) => ({
...
-    closeDrawer()
-  }, [closeDrawer, onPressLearnMore])
+    onClose()
+  }, [onClose, onPressLearnMore])
 
   const handleTipPress = useCallback(() => {
-    dispatch(beginTip({ user, source: 'inboxUnavailableModal' }))
+    if (!currentUserId || !user) {
+      console.error(
+        'Unexpected undefined currentUserId or user when starting tip'
```
schottra, line 142: Same question here about error propagation.

packages/mobile/src/screens/tip-artist-screen/TipSentScreen.tsx
```diff
@@ -84,20 +83,20 @@ export const TipSentScreen = () => {
...
-      source === 'inboxUnavailableModal' &&
-      account?.user_id &&
-      recipient?.user_id
-    ) {
-      dispatch(
-        chatActions.goToChat({
-          chatId: makeChatId([account.user_id, recipient.user_id])
-        })
-      )
+    if (onSuccessActions && account?.user_id && recipient?.user_id) {
```
schottra, line 86: This is an interesting pattern. Are there any implications of order of actions, failure of an action blocking dispatch of the rest, etc?

schottra (approved):

packages/mobile/src/screens/profile-screen/MessageLockedButton.tsx
```diff
@@ -24,24 +23,17 @@ const useStyles = makeStyles(({ palette, spacing }) => ({
 }))
 
 type MessageLockedButtonProps = {
-  profile: Pick<User, 'user_id'>
+  userId: ID
```
dharit-tan, line 26: ty for cleanup

packages/mobile/src/components/inbox-unavailable-drawer/InboxUnavailableDrawer.tsx
rickyrombo: Yeah this is a good q. Sentry does capture our `console.error` messages, but I was torn about whether to throw here and crash the app, or just log an error, or do nothing. In theory, this should never happen - the drawer should always be opened with a userId - but typecheck makes some of these checks necessary and it felt like I should at least log them if they do occur.

packages/mobile/src/screens/tip-artist-screen/TipSentScreen.tsx
rickyrombo: Yeah if an action fails/crashes the app, the next one won't be processed. I'm not in love with this pattern, it was the best idea I had though - open to alternatives!

dharit-tan (approved):
nice work, lots of plumbing!

packages/mobile/src/components/inbox-unavailable-drawer/InboxUnavailableDrawer.tsx
schottra: Makes sense. I just wanted to make sure that we have actionable data if this _does_ happen, as it would probably be in a context where something else has gone wrong in an unexpected way.

> but typecheck makes some of these checks necessary

This is a pattern worth thinking about. There's a way to design where values are _not_ typed as potentially null or undefined. That would essentially mean that we throw at a much lower layer (validating values as they come across the wire), and we don't even render a component that would consume a value in cases where the value doesn't exist.

It's kind of a fundamental shift, though. And not a trivial amount of effort, so I'm not pushing it 😅 . This is a reasonable approach given that we can't change the type to be required without doing a lot of other work.

PR #3882, opened 2023-08-14
Client uses cids in requests to CN for images

michellebrier:
### Description
***WIP, needs more testing***
Per changes in this [PR](https://github.com/AudiusProject/audius-protocol/pull/5783), when querying for an image variant from the client, use the variant's cid (ingested from `*_cids` fields in API user/track/playlist responses) to query CNs. So the query is `${cid's preferred storage node}/content/${cid}` rather than `${non-specific storage node}/content/${upload id}/${size}`. This reduces redirects and image latency.

### Dragons

### How Has This Been Tested?
Tested against stage node with API changes, used network developer tools panel to verify that the image variant's cid was queried directly from its preferred node with no redirects.
image

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/mbrier/con-820/request-cids

packages/mobile/src/components/image/TrackImage.tsx
```diff
@@ -62,7 +71,8 @@ export const useTrackImage = ({ track, size }: UseTrackImageOptions) => {
     cid,
     size,
```
dylanjeffers, line 72: so just to double check, the size param is mostly irrelevant for this new path since we set the cid = the cover_art_cids[size] right?

packages/mobile/src/components/image/UserCoverImage.tsx
```diff
@@ -25,16 +25,28 @@ const interpolateImageTranslate = (animatedValue: Animated.Value) =>
...
+    | 'updatedCoverPhoto'
+  >
 >
 
 export const useUserCoverImage = (user: CoverImageUser) => {
-  const cid = user ? user.cover_photo_sizes || user.cover_photo : null
+  let cid = null
+  if (user) {
+    cid = user.cover_photo_cids
+      ? user.cover_photo_cids[WidthSizes.SIZE_640]
```
dylanjeffers, line 41: is there ever a chance that we dont have the cid for the specific size? ie should we check that cover_photo_cids exist and that the size specific one exists, otherwise fallback to cover_photo_sizes?

dylanjeffers (approved):
EXCELLENT work, this makes a ton of sense. main question is about missing sizes if that's something reasonable wed want to catch? Also looks like mobile init failed. lmk if you need help with that

dylanjeffers (commented):
Also another thought, did you test photo updates? i think for instance when a user updates their cover_photo, we use the "OVERRIDE" section of _cover_art_sizes to use a local image. it appears with this new setup that wont be respected?

packages/mobile/src/components/image/TrackImage.tsx
michellebrier: yes!

packages/mobile/src/components/image/UserCoverImage.tsx
michellebrier: good point - will add that in

dylanjeffers: Sweet if that is the case one thing we can do is pass the cid map as a new arg instead of the "useRoute" param and we can check inside the function since we have access to the size?

michellebrier: added in https://github.com/AudiusProject/audius-client/pull/3882/commits/d2f7677fb435ba48037ebd2fef41a1dc8c06b460!

audius-infra:
Preview this change https://demo.audius.co/mbrier/con-820/request-cids

audius-infra:
Preview this change https://demo.audius.co/mbrier/con-820/request-cids

audius-infra:
Preview this change https://demo.audius.co/mbrier/con-820/request-cids

audius-infra:
Preview this change https://demo.audius.co/mbrier/con-820/request-cids

audius-infra:
Preview this change https://demo.audius.co/mbrier/con-820/request-cids

audius-infra:
Preview this change https://demo.audius.co/mbrier/con-820/request-cids

audius-infra:
Preview this change https://demo.audius.co/mbrier/con-820/request-cids

PR #3883, opened 2023-08-14
[PAY-1707] Implements usage of existing balance during content purchases

schottra:
### Description
* Added logic to modal flow for detecting the current balance and doing the math to determine the leftover amount needed
* Added logic to saga to get the difference between price and balance to pass to the stripe modal flow

### Dragons
None

### How Has This Been Tested?
Locally verified in Chrome

N/A

### Feature Flags ###
Experience is behind the usdc feature flag

### Screenshots

**Balance covers entire purchase**
Screenshot 2023-08-14 at 10 49 11 AM

---

**Balance covers all but one cent**
Screenshot 2023-08-14 at 10 49 30 AM

---

**Balance is less than a cent**
Screenshot 2023-08-14 at 10 58 41 AM


packages/common/src/services/audius-backend/solana.ts
```diff
@@ -106,6 +106,39 @@ function isCreateUserBankIfNeededError(
...
+  const ethAddress =
+    sourceEthAddress ?? audiusLibs.Account!.getCurrentUser()?.wallet
+
+  if (!ethAddress) {
+    throw new Error(
+      `getUserbankAccountInfo: unexpected error getting eth address`
+    )
+  }
+
+  if (!audiusLibs.solanaWeb3Manager!.doesUserbankExist({ ethAddress, mint })) {
```
dharit-tan, line 127: Doesn't this call the same fns as we're doing below? Can we just check it ourselves before returning to save on the solana RPC calls?

audius-infra:
Preview this change https://demo.audius.co/pay-1707-use-existing-balance

dharit-tan (approved):

audius-infra:
Preview this change https://demo.audius.co/pay-1707-use-existing-balance

PR #3884, opened 2023-08-14
Change nullish check to falsey check for collection track times

Kyle-Shanks:
### Description
Small change to fix how we process falling back to the time from metadata_time of a track.
Was causing an issue when a track had a metadata_time of 0

### Dragons

N/A

### How Has This Been Tested?

N/A

N/A

### Feature Flags ###

N/A


sliptype (approved):

audius-infra:
Preview this change https://demo.audius.co/kj-Fix-time-check-for-collection-tracks

PR #3885, opened 2023-08-14
[PAY-1588] Use existing balance in purchase flow on mobile

dharit-tan:
### Description
Moved some of the util fns in `web` over to `common` and used them to calculate `existingBalance`/`amountDue` for the `PurchaseSummaryTable`.

### Dragons

### How Has This Been Tested?

Confirmed web behavior is correct.

Simulator Screenshot - iPhone 14 Pro - 2023-08-14 at 16 05 58
Simulator Screenshot - iPhone 14 Pro - 2023-08-14 at 16 05 32
Simulator Screenshot - iPhone 14 Pro - 2023-08-14 at 16 37 04

### Feature Flags ###

### DIFF
--- a/packages/common/src/store/purchase-content/utils.ts
+++ b/packages/common/src/store/purchase-content/utils.ts
@@ -15,15 +15,17 @@ export const isContentPurchaseInProgress = (stage: PurchaseContentStage) => {
   ].includes(stage)
 }
 
-export const getPurchaseSummaryValues = (
-  price: number,
-  currentBalance: BNUSDC = zeroBalance()
-): {
+type PurchaseSummaryValues = {
   amountDue: number
   existingBalance: number | undefined
   basePrice: number
   artistCut: number
-} => {
+}
+
+export const getPurchaseSummaryValues = (
+  price: number,
+  currentBalance: BNUSDC = zeroBalance()
+): PurchaseSummaryValues => {
   let amountDue = price
   let existingBalance
   const priceBN = new BN(price).mul(BN_USDC_CENT_WEI)
--- a/packages/common/src/store/purchase-content/utils.ts
+++ b/packages/common/src/store/purchase-content/utils.ts
@@ -1,4 +1,11 @@
+import BN from 'bn.js'
+
+import { BNUSDC } from 'models/Wallet'
+import { BN_USDC_CENT_WEI, ceilingBNUSDCToNearestCent } from 'utils/wallet'
+
 import { PurchaseContentStage } from './types'
+
+export const zeroBalance = () => new BN(0) as BNUSDC
 
 export const isContentPurchaseInProgress = (stage: PurchaseContentStage) => {
   return [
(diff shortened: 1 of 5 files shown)


packages/common/src/store/purchase-content/utils.ts
```diff
@@ -1,9 +1,48 @@
...
+    // is between cents so that we aren't advertising *lower* than what the user
+    // will have to pay.
+    const diff = priceBN.sub(currentBalance)
+    amountDue = ceilingBNUSDCToNearestCent(diff as BNUSDC)
+      .div(BN_USDC_CENT_WEI)
+      .toNumber()
+    existingBalance = price - amountDue
+  }
+
+  return { amountDue, existingBalance, basePrice: price, artistCut: price }
```
schottra, line 47: Might want to actually define this type if it's not going to live with the components?
I'm fine if we just copy the shape of PurchaseSummaryProps here.

packages/mobile/src/components/premium-track-purchase-drawer/PremiumTrackPurchaseDrawer.tsx
```diff
@@ -126,61 +141,73 @@ export const PremiumTrackPurchaseDrawer = () => {
...
       drawerName={PREMIUM_TRACK_PURCHASE_MODAL_NAME}
       onClosed={handleClosed}
     >
-      <View style={styles.drawer}>
-        <View style={styles.titleContainer}>
-          <IconCart fill={neutralLight2} />
-          <Text style={styles.title}>{messages.title}</Text>
+      {isLoading ? (
+        <View style={styles.spinnerContainer}>
+          <LoadingSpinner />
```
schottra, line 155: fwiw, I did not check with design on what the loading pattern should be for this flow.
It's possible we should be using shimmer states instead.

packages/web/src/components/premium-content-purchase-modal/components/PurchaseDetailsPage.tsx
```diff
@@ -83,7 +49,7 @@ export type PurchaseDetailsPageProps = {
 }
 
 export const PurchaseDetailsPage = ({
-  currentBalance = zeroBalance(),
+  currentBalance,
```
schottra, line 52: Curious why this got removed? I think this was to prevent problems with users who don't have a user bank yet.

schottra (approved):

packages/mobile/src/components/premium-track-purchase-drawer/PremiumTrackPurchaseDrawer.tsx
dharit-tan: good point, i don't anticipate this happening often but we can surface with design

packages/web/src/components/premium-content-purchase-modal/components/PurchaseDetailsPage.tsx
dharit-tan: i moved it as a default arg in the `getPurchaseSummaryValues` fn. buttt i can see an argument for keeping it here instead, makes more sense conceptually. wdyt?

schottra: Oh I missed that. Probably fine, just might complicate things if we need the current balance for any other reason inside this component.

PR #3897, opened 2023-08-17
[PAY-1720] Implements PlainButton

schottra:
### Description
This adds a full implementation of the harmony PlainButton component, which was originally partially implemented as a variant of HarmonyButton.

* Moved non-styling functionality into a private `BaseButton` component.
* Updated HarmonyButton to layer its styling on top of the new base component
* Added `HarmonyPlainButton` component as a derived component implementing the new button type

### Dragons
Should be none

### How Has This Been Tested?
Locally verified in Storybook

N/A

### Feature Flags ###
N/A

### Screenshots

Default Variant
Kapture 2023-08-17 at 13 24 24

---

Subdued Variant
Kapture 2023-08-17 at 13 25 34

---

Inverted Variant
Kapture 2023-08-17 at 13 26 15

### DIFF
--- a/packages/stems/src/components/HarmonyButton/BaseButton.module.css
+++ b/packages/stems/src/components/HarmonyButton/BaseButton.module.css
@@ -1,0 +1,43 @@
+/* ===Base Styles=== */
+.button {
+  align-items: center;
+  box-sizing: border-box;
+  cursor: pointer;
+  display: inline-flex;
+  flex-shrink: 0;
+  justify-content: center;
+  overflow: hidden;
+  position: relative;
+  text-align: center;
+  user-select: none;
+  white-space: nowrap;
+}
+
+.button:focus {
+  outline: none !important;
+}
+
+/* Only add hover styles on devices which support it */
+@media (hover: hover) {
+  .button:not(.disabled):hover {
+    transition: all var(--hover);
+    transform: scale(1.04);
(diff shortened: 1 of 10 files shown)


audius-infra:
Preview this change https://demo.audius.co/pay-1720-plain-button

packages/stems/src/components/HarmonyButton/BaseButton.module.css
```diff
@@ -0,0 +1,43 @@
+/* ===Base Styles=== */
+.button {
```
dylanjeffers, line 2: absolutely LOVE the base-button approach

dylanjeffers (approved):
LOVE this refactor. So what im hearing is that PlainButton is too different to be a variant of HarmonyButton? Code-wise everything looks perfect

schottra:
> LOVE this refactor. So what im hearing is that PlainButton is too different to be a variant of HarmonyButton? Code-wise everything looks perfect

@dylanjeffers It could conceivably be done. But with the way the variants work for each, it would probably require 3 new variants of button, and then some messy style overrides for sizing. It made more sense to make it a separate button. It behaves just like a `Button` in terms of functionality and takes all the same base props. 

I would expect LinkButton to also follow this pattern, deriving from BaseButton and implementing its style variants separately.

PR #3901, opened 2023-08-18
[C-2970] Add Link, Improve Text

dylanjeffers:
### Description

Adds generic Link component which will be used across the app to improve seo
Improves Text component 
- Adds `as` prop to customize underlying rendered element.
- Updates color prop to use the same color config as Icon and HarmonyButton

### DIFF
--- a/packages/stems/src/styles/colors.ts
+++ b/packages/stems/src/styles/colors.ts
@@ -38,3 +38,4 @@ export type ColorValue =
   | 'accentPurple'
   | 'accentBlue'
   | 'specialLightGreen'
+  | 'darkmodeStaticWhite'
--- a/packages/web/src/components/typography/Text.tsx
+++ b/packages/web/src/components/typography/Text.tsx
@@ -1,40 +1,49 @@
-import { CSSProperties, ElementType, ReactNode } from 'react'
+import { ComponentProps, ElementType, ReactNode } from 'react'
 
+import {
+  ColorValue,
+  CSSCustomProperties,
+  toCSSVariableName
+} from '@audius/stems'
 import cn from 'classnames'
 import { camelCase } from 'lodash'
-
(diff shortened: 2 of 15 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-c-2970-add-link

packages/stems/src/styles/colors.ts
```diff
@@ -38,3 +38,4 @@ export type ColorValue =
   | 'accentPurple'
   | 'accentBlue'
   | 'specialLightGreen'
+  | 'darkmodeStaticWhite'
```
raymondjacobson, line 41: staticWhite was supposed to be the same in light/dark, should it be in here?

raymondjacobson (approved):

packages/stems/src/styles/colors.ts
dylanjeffers: staticWhite should also be in here eventually, this add was for compatibility with the new banner color add

packages/web/src/components/typography/Text.tsx
```diff
@@ -1,40 +1,49 @@
...
 
-  const Tag: ElementType = variantTagMap[variant][size] ?? 'p'
+  const Tag: ElementType = as ?? variantTagMap[variant][size] ?? 'p'
 
-  const themeColors = getCurrentThemeColors()
-  const textColor = themeColors[color] ?? themeColors['--neutral']
-  const styleObject: CSSProperties = {
-    color: textColor
+  const styleObject: CSSCustomProperties = {
+    '--text-color': `var(${toCSSVariableName(color)})`
```
Kyle-Shanks, line 46: we should probably fall back to inherit here in case a color is not passed

Kyle-Shanks (approved):
Nice

PR #3902, opened 2023-08-18
[C-2971] Add Avatar

dylanjeffers:
### Description

Adds basic Avatar component, which will be used to improve seo by properly wrapping user images in an anchor tag with proper accessibility

Also adds improved useUserProfilePicture hook which only needs user_id. kinda lame name but eventually we can migrate to this

### DIFF
--- a/packages/web/src/components/avatar/Avatar.tsx
+++ b/packages/web/src/components/avatar/Avatar.tsx
@@ -8,7 +8,7 @@ import {
 import { Link } from 'react-router-dom'
 
 import DynamicImage from 'components/dynamic-image/DynamicImage'
-import { useUserProfilePicture2 } from 'hooks/useUserProfilePicture'
+import { useProfilePicture } from 'hooks/useUserProfilePicture'
 import { useSelector } from 'utils/reducer'
 import { SIGN_IN_PAGE, profilePage } from 'utils/route'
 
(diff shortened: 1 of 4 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-avatar

packages/web/src/components/avatar/Avatar.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { SIGN_IN_PAGE, profilePage } from 'utils/route'
+
+import styles from './Avatar.module.css'
+
+const { getAccountUser } = accountSelectors
+
+const { getUser } = cacheUsersSelectors
+
+const messages = {
+  goTo: 'Go to',
```
raymondjacobson, line 22: is `go to` std verbiage?

packages/web/src/hooks/useUserProfilePicture.ts
```diff
@@ -28,6 +32,27 @@ export const useUserProfilePicture = (
   })
 }
 
+export const useUserProfilePicture2 = (
```
raymondjacobson, line 35: maybe a more descriptive name or a comment suggesting the different args

raymondjacobson (approved):

packages/web/src/components/avatar/Avatar.tsx
dylanjeffers: From chat gpt perspective yeah, but can dig into whether there is better verbiage

packages/web/src/hooks/useUserProfilePicture.ts
dylanjeffers: > maybe a more descriptive name or a comment suggesting the different args

Yeah agreed, may also deprecate the old one

PR #3904, opened 2023-08-18
[PAY-1628] Navigate to track after purchase

schottra:
### Description
This adds logic to detect a completed purchase in the modal and navigate to the track (if not already on the page) behind the modal.

Some other cleanup:
* Added logic to close the LockedContentModal when opening the purchase modal. This is to prevent it hanging around behind the purchase modal and being visible after the purchase is complete
* Removed the page structure in the purchase modal. We don't need a loading page and it had the effect of flashing on the screen briefly before we show the actual content

### Dragons
N/A

### How Has This Been Tested?
Verified locally on Chrome

Will test as part of overall feature validation

### Feature Flags ###
Behind usdc purchase feature flag

### Screenshots

https://github.com/AudiusProject/audius-client/assets/1815175/e5f8948f-ec3e-491f-8255-cb457784cae2

### DIFF
--- a/packages/web/src/components/premium-content-purchase-modal/components/PurchaseDetailsPage.tsx
+++ b/packages/web/src/components/premium-content-purchase-modal/components/PurchaseDetailsPage.tsx
@@ -1,4 +1,4 @@
-import { useCallback } from 'react'
+import { useCallback, useEffect } from 'react'
 
 import {
   BNUSDC,
@@ -21,6 +21,7 @@ import {
   IconCheck,
   IconError
 } from '@audius/stems'
+import { push as pushRoute } from 'connected-react-router'
 import { useDispatch, useSelector } from 'react-redux'
 
 import { Icon } from 'components/Icon'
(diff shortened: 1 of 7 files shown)


audius-infra:
Preview this change https://demo.audius.co/pay-1628-navigate-to-track

packages/web/src/components/premium-content-purchase-modal/components/PurchaseDetailsPage.tsx
```diff
@@ -85,6 +98,8 @@ export const PurchaseDetailsPage = ({
     )
   }, [isUnlocking, dispatch, track.track_id])
 
+  useNavigateOnSuccess(track, stage)
```
rickyrombo, line 101: super nit picky and no need to change, but I'm curious why you decided to split the hook out rather than using `useEffect` inline?

rickyrombo (approved):

dharit-tan (approved):

packages/web/src/components/premium-content-purchase-modal/components/PurchaseDetailsPage.tsx
schottra: I think I have an undocumented preference for components to not have _too much_ hook logic at the top and this seemed like a nice way to clearly indicate what the hook is doing and keep the component itself simple.

PR #3905, opened 2023-08-18
[C-2969] Fix related artist images not loading

amendelsohn:
### Description

Since audius-query adds items to cache, we need to perform the same processing we do at other cache ingress points. Here, we were missing the `getUserImages` in `reformatUser` which sets the `_profile_picture_sizes` field in the UserMetadata.

Also fixed all the circular imports in `@audius/common`

image

### DIFF
--- a/packages/common/src/audius-query/createApi.ts
+++ b/packages/common/src/audius-query/createApi.ts
@@ -279,7 +279,8 @@ const fetchData = async <Args, Data>(
       data = result
 
       // Format entities before adding to cache
-      entities[Kind.USERS] = (entities[Kind.USERS] ?? []).map(
+      entities[Kind.USERS] = mapValues(
+        entities[Kind.USERS] ?? [],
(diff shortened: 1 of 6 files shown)


raymondjacobson (approved):
not the best reviewer, but it seems reasonable. approving in case needed review b/c blocking

audius-infra:
Preview this change https://demo.audius.co/amendel-related-artist-photos

packages/common/src/audius-query/createApi.ts
```diff
@@ -274,6 +277,12 @@ const fetchData = async <Args, Data>(
         apiResponseSchema
       )
       data = result
+
+      // Format entities before adding to cache
+      entities[Kind.USERS] = mapValues(
```
dylanjeffers, line 282: nice!

dylanjeffers (approved):
great adds

PR #3907, opened 2023-08-19
[C-2972] Fix feed, trending track-page seo

dylanjeffers:
### Description

Fixes seo for feed, trending, track-page, and any page with track tiles on desktop and mobile

### TODO
We should probably not hide the anchor tags for desktop track tiles, mobile correctly renders them behind a skeleton

### DIFF
--- a/packages/web/src/components/track/GiantTrackTile.tsx
+++ b/packages/web/src/components/track/GiantTrackTile.tsx
@@ -34,6 +34,7 @@ import { ReactComponent as IconRobot } from 'assets/img/robot.svg'
 import { ArtistPopover } from 'components/artist/ArtistPopover'
 import DownloadButtons from 'components/download-buttons/DownloadButtons'
 import { EntityActionButton } from 'components/entity-page/EntityActionButton'
+import { Link } from 'components/link'
 import LoadingSpinner from 'components/loading-spinner/LoadingSpinner'
 import Menu from 'components/menu/Menu'
 import RepostFavoritesStats from 'components/repost-favorites-stats/RepostFavoritesStats'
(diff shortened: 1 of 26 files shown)


packages/web/src/components/track/GiantTrackTile.tsx
```diff
@@ -523,16 +524,23 @@ export const GiantTrackTile = ({
             </div>
             <div className={styles.artistWrapper}>
               <div className={cn(fadeIn)}>
-                <span>By</span>
+                <span>By </span>
```
raymondjacobson, line 527: why was this space necessary? is it possible this will cause other hover issues, etc? probably fine but worth thinking about a little which maybe you have

packages/web/src/components/track/mobile/ConnectedPlaylistTile.tsx
```diff
@@ -167,14 +166,6 @@ const ConnectedPlaylistTile = ({
     [goToRoute, getRoute]
   )
 
-  const goToArtistPage = useCallback(
-    (e: MouseEvent<HTMLElement>) => {
-      e.stopPropagation()
```
raymondjacobson, line 172: ok to lose stop propagating?

packages/web/src/components/track/desktop/ConnectedTrackTile.module.css
```diff
@@ -48,35 +48,8 @@
...
-  color: inherit;
-  text-decoration: inherit;
-}
-
-.userName {
-  display: inline-flex;
-  font-weight: var(--font-medium);
-  align-items: center;
-  /**
-   * Some webkit browsers (safari) have some trouble rendering this specific
```
raymondjacobson, line 63: i imagine losing this ok, but we check safari just in case?

raymondjacobson (approved):
Absolutely heroic

audius-infra:
Preview this change https://demo.audius.co/dj-c-2972-feed-trending-track-page

packages/web/src/components/track/mobile/ConnectedPlaylistTile.tsx
dylanjeffers: yes cause the Link component stops prop by default, confirmed clicking the links didnt start playing a track

packages/web/src/components/track/GiantTrackTile.tsx
dylanjeffers: there was a paddingLeft on the link to simulate a space, so removed that and added space here :)

packages/web/src/components/track/desktop/ConnectedTrackTile.module.css
dylanjeffers: just confirmed we chillin, good call

PR #3908, opened 2023-08-19
[PAY-1706] Merge modalsWithState with modals in common store

rickyrombo:
### Description

- Unify the store structure for modal states
- Move modal store stuff (at least the ones using `createModal`) to be nested inside the `modals` folder
- Make the modals slice default to objects instead of just a simple `boolean | 'closing'` and update usages

### Dragons

### How Has This Been Tested?

Tested both kinds of modals still function as expected

### Feature Flags ###


audius-infra:
Preview this change https://demo.audius.co/mjp-modals-slice

packages/common/src/store/ui/modals/reducers.ts
```diff
@@ -0,0 +1,37 @@
...
+import { BaseModalState } from './createModal'
+import { inboxUnavailableModalReducer } from './inbox-unavailable-modal'
+import { leavingAudiusModalReducer } from './leaving-audius-modal'
+import parentReducer, { initialState } from './parentSlice'
+import { Modals, ModalsState } from './types'
+
+/**
+ * Create a bunch of reducers that do nothing, so that the state is maintained and not lost through the child reducers
+ */
+const noOpReducers = Object.keys(initialState).reduce((prev, curr) => {
```
schottra, line 13: I _think_ I understand why this is needed, but it seems a little messy? Is this something that goes away if/when all the modals move to the new strategy?

packages/common/src/store/ui/modals/reducers.ts
```diff
@@ -0,0 +1,37 @@
...
+  CreateChatModal: createChatModalReducer,
+  InboxUnavailableModal: inboxUnavailableModalReducer,
+  LeavingAudiusModal: leavingAudiusModalReducer
+})
+
+/**
+ * Return a reducer that processes child slices, then parent slice.
+ * This maintains backwards compatibility between modals created without createModal
+ */
+export const rootModalReducer = (state: ModalsState, action: Action) => {
```
schottra, line 34: love it

schottra (approved):
Nice work!

packages/common/src/store/ui/modals/reducers.ts
rickyrombo: It's not my favorite either. But yeah it's basically stubbing out "child" slice reducers as no-op passthroughs, although somewhat hackily using the initial state to determine what keys to create values for. If we had child slices for each modal, this could go away, but I don't see that happening. One other method would be to explicitly make these reducers rather than using the loop - not sure if that helps it feel less messy or more though

PR #3910, opened 2023-08-21
[C-2911] Update new select page of the upload flow

Kyle-Shanks:
### Description
Updated the components and styles for the new select page of the upload flow

### Dragons

N/A

### How Has This Been Tested?

Manually tested

N/A

### Feature Flags ###

N/A


packages/web/src/components/upload/TrackPreview.module.css
```diff
@@ -58,19 +59,77 @@
...
+
 .remove:hover g path {
   fill: var(--neutral-light-4);
 }
+
 .remove:active {
   transform: scale(0.95);
 }
+
+.trackPreviewNew {
```
dylanjeffers, line 83: would we want to split up the css files too?

packages/web/src/components/upload/TrackPreviewNew.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const {
+    displayIndex = false,
+    index,
+    fileType = 'audio/mp3',
+    trackTitle = 'Untitled',
+    fileSize,
+    onRemove
+  } = props
+
+  const imgSrc = useMemo(() => fileTypeIcon(fileType), [fileType])
```
dylanjeffers, line 53: since this is a simple if/else branch function, its more performant to just let the computation happen. Typically only when the computation itself is expensive, or what's being returned is not a primitive value, like an array or object, is useMemo necessary

dylanjeffers: can we also name this better, like fileTypeSrc? also we can probably just put this inline in the img component, too

packages/web/src/pages/upload-page/components/TracksPreview.module.css
```diff
@@ -59,3 +59,42 @@
 .tabSlider {
   margin-bottom: 15px;
 }
+
+.containerNew {
```
dylanjeffers, line 63: same here, but whatever you think is easiest to reason about, whether these are in a new file so its easy to remove the old code eventually.

packages/web/src/pages/upload-page/components/TracksPreviewNew.tsx
```diff
@@ -0,0 +1,105 @@
...
+        className={cn(styles.tracksContainerNew, {
+          [styles.shortScroll]:
+            props.uploadType !== UploadType.INDIVIDUAL_TRACKS
+        })}
+      >
+        {tracks.map((track, i) => (
+          <TrackPreviewNew
+            index={i}
+            displayIndex={tracks.length > 1}
+            key={track.metadata.title + i}
```
dylanjeffers, line 81: is there a way to set the key without needing the index?

packages/web/src/pages/upload-page/components/TracksPreviewNew.tsx
```diff
@@ -0,0 +1,105 @@
...
+            fileType={track.file.type}
+            fileSize={track.file.size}
+            onRemove={() => onRemove(i)}
+          />
+        ))}
+      </Scrollbar>
+      <div className={cn(styles.infoContainerNew, styles.footerContainerNew)}>
+        <Text size='small'>
+          {tracks.length === 1
+            ? `${tracks.length} Track Added`
```
dylanjeffers, line 92: messages obj for "track/s added"?

packages/web/src/pages/upload-page/components/TracksPreviewNew.tsx
```diff
@@ -0,0 +1,105 @@
...
+            : `${tracks.length} Tracks Added`}
+        </Text>
+        <HarmonyButton
+          variant={HarmonyButtonType.PRIMARY}
+          text={messages.continue}
+          name='continue'
```
dylanjeffers, line 98: do we need a name?

dylanjeffers (approved):
really clean and nice dude

packages/web/src/pages/upload-page/components/TracksPreview.module.css
Kyle-Shanks: yea i've been updating things with the thought that we would remove old stuff later

packages/web/src/pages/upload-page/components/TracksPreviewNew.tsx
Kyle-Shanks: we don't have a track id yet and this is how we were doing it before so i figure this should work

packages/web/src/pages/upload-page/components/TracksPreviewNew.tsx
Kyle-Shanks: i didn't think so, but it was on the old button so I didn't want to remove anything just in case

audius-infra:
Preview this change https://demo.audius.co/kj-Update-new-select-page

PR #3916, opened 2023-08-21
[PAY-1741] Add routes for transactional pages

schottra:
### Description
Just adding routes for now, will be used in follow-up PRs

### Dragons
None, these routes are currently not handled in the router.

### How Has This Been Tested?
Locally verified, 404s as expected

N/A

### Feature Flags ###
N/A

### DIFF
--- a/packages/web/src/utils/route.ts
+++ b/packages/web/src/utils/route.ts
@@ -73,6 +73,9 @@ export const CHECK_PAGE = '/check'
 export const DEACTIVATE_PAGE = '/deactivate'
 export const CHATS_PAGE = '/messages'
 export const CHAT_PAGE = '/messages/:id?'
+export const PURCHASES_PAGE = '/purchases'
+export const SALES_PAGE = '/sales'
+export const WITHDRAWALS_PAGE = '/withdrawals'
 
 // Param routes.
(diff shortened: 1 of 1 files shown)


dharit-tan (approved):

raymondjacobson (approved):
nice. there should also be a spot where we prevent handles from being signed up with. lmk if you need a hand finding the code pointer

schottra:
> nice. there should also be a spot where we prevent handles from being signed up with. lmk if you need a hand finding the code pointer

@raymondjacobson Yep, we use the `orderedRoutes` list as part of the larger [`restrictedRoutes` list](https://github.com/AudiusProject/audius-client/blob/b0c2e4adf4530d09157da8db8ca6c72bd07a993a/packages/web/src/utils/restrictedHandles.js#L6), which is used to[ validate handles at signup ](https://github.com/AudiusProject/audius-client/blob/1277a41e698cc23a47cc095b44668c6ee0db5898/packages/web/src/common/store/pages/signon/sagas.js#L187-L188).

So this should already be...handled.

audius-infra:
Preview this change https://demo.audius.co/pay-1715-reserved-routes

rickyrombo (approved):

PR #3920, opened 2023-08-22
[PAY-1742] Remove useMetaMask on invalid account

raymondjacobson:
### Description

Drop `useMetaMask` local storage item when account fetch fails on web.
Address issue where a user may accidentally get halfway through the MetaMask flow, bail out, and then brick themselves from future sign up attempts.

### How Has This Been Tested?

1. Start MetaMask sign up flow
2. Verify useMetaMask set in local storage
3. Exit sign up, refresh, watch it get removed

### DIFF
--- a/packages/web/src/common/store/account/sagas.js
+++ b/packages/web/src/common/store/account/sagas.js
@@ -186,6 +186,10 @@ export function* fetchAccountAsync({ isSignUp = false }) {
         reason: 'ACCOUNT_NOT_FOUND'
       })
     )
+    if (!isNativeMobile) {
+      const localStorage = yield getContext('localStorage')
+      yield call([localStorage, 'removeItem'], 'useMetaMask')
+    }
     return
   }
   if (account.is_deactivated) {
--- a/packages/web/src/hooks/useTabs/useTabs.tsx
+++ b/packages/web/src/hooks/useTabs/useTabs.tsx
@@ -755,7 +755,6 @@ const BodyContainer = memo(
     dimensionsAreDirty,
     didSetDimensions
   }: BodyContainerProps) => {
-    console.log('hmm', elements, activeIndex)
     // Get a ref to the element to use for calculating height
     const {
       containerWidth,
--- a/packages/web/src/components/banner/Web3ErrorBanner.tsx
+++ b/packages/web/src/components/banner/Web3ErrorBanner.tsx
@@ -15,7 +15,7 @@ const messages = {
 }
 
 const META_MASK_SETUP_URL =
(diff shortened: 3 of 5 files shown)


audius-infra:
Preview this change https://demo.audius.co/rj-pay-1742

packages/web/src/common/store/account/sagas.js
```diff
@@ -186,6 +186,10 @@ export function* fetchAccountAsync({ isSignUp = false }) {
         reason: 'ACCOUNT_NOT_FOUND'
       })
     )
+    if (!isNativeMobile) {
```
schottra, line 189: I know @rickyrombo was seeing some Sentry errors with localStorage recently on Safari mobile. Just wondering if it's safe to _always_ clear this regardless of whether we've actually set it and if there were any learnings from that issue we should apply here to fail safely if localStorage isn't accessible?

packages/web/src/hooks/useTabs/useTabs.tsx
```diff
@@ -755,7 +755,6 @@ const BodyContainer = memo(
     dimensionsAreDirty,
     didSetDimensions
   }: BodyContainerProps) => {
-    console.log('hmm', elements, activeIndex)
```
schottra, line 758: 😄

schottra (approved):
lgtm, just the one question

packages/web/src/common/store/account/sagas.js
raymondjacobson: Tested on ios safari. it seems ok. i thought about getting & then removing if it exists, but it's just strictly more work because the implementation of remove per spec is 

https://html.spec.whatwg.org/multipage/webstorage.html

---- 

The removeItem(key) method steps are:

If [this](https://webidl.spec.whatwg.org/#this)'s [map](https://html.spec.whatwg.org/multipage/webstorage.html#concept-storage-map)[key] does not [exist](https://infra.spec.whatwg.org/#map-exists), then return null.

Set oldValue to [this](https://webidl.spec.whatwg.org/#this)'s [map](https://html.spec.whatwg.org/multipage/webstorage.html#concept-storage-map)[key].

[Remove](https://infra.spec.whatwg.org/#map-remove) [this](https://webidl.spec.whatwg.org/#this)'s [map](https://html.spec.whatwg.org/multipage/webstorage.html#concept-storage-map)[key].

[Reorder](https://html.spec.whatwg.org/multipage/webstorage.html#concept-storage-reorder) [this](https://webidl.spec.whatwg.org/#this).

[Broadcast](https://html.spec.whatwg.org/multipage/webstorage.html#concept-storage-broadcast) [this](https://webidl.spec.whatwg.org/#this) with key, oldValue, and null.

PR #3928, opened 2023-08-22
Migrate withdraw USDC saga to web common

dharit-tan:
### Description
Migrates the withdraw USDC sagas to web common, so we can use the jupiter service that's already there and avoid having to migrate it to common.

### Dragons

### How Has This Been Tested?

Dispatching actions via redux extension.

### Feature Flags ###


packages/common/src/services/audius-backend/solana.ts
```diff
@@ -43,6 +43,25 @@ export const isValidSolDestinationAddress = async (
   }
 }
 
+export const isSolAddress = async (
```
rickyrombo, line 46: nit: `isSolWallet`?

rickyrombo (approved):

PR #3930, opened 2023-08-23
Add lint check for console.log and remove some bad ones

raymondjacobson:
### Description

Add lint rule to index.js and update console.log instances to something more relevant.

### How Has This Been Tested?

`npm run verify`


audius-infra:

sliptype (approved):
Nice clean up! Prettier won't automatically remove console.logs, for example when debugging right?

packages/mobile/src/services/sdk/audius-sdk.ts
```diff
@@ -31,16 +31,12 @@ const initSdk = async () => {
 
 export const audiusSdk = async () => {
   if (inProgress) {
-    console.log('SDK in progress...')
```
rickyrombo, line 34: thank

rickyrombo (approved):

raymondjacobson:
> can you remove the locks?

done

packages/web/src/services/sentry.ts
```diff
@@ -30,6 +30,13 @@ export const initializeSentry = () => {
     normalizeDepth: 5,
     maxBreadcrumbs: MAX_BREADCRUMBS,
     beforeBreadcrumb: (breadCrumb, hint) => {
+      // filter out info and debug logs
+      if (
+        (breadCrumb.level === 'info' || breadCrumb.level === 'debug') &&
```
rickyrombo, line 35: Do we want to filter out _info_ from breadcrumbs?

raymondjacobson: i think so. @dylanjeffers and i were talking about this. info and debug have the same semantic meaning to us internally and there are no real rules of the trade (they show up in console identically...). i think we may just want to only have either debug or info, but left that for future.

audius-infra:

packages/web/src/services/sentry.ts
rickyrombo: My brain thinks of them differently:
- Error: Highest concern/severity
- Warn: Something unexpected, things might be ok though
- Info: Useful for debugging prod issues, not noisy
- Debug: Useful when debugging locally, more permissive to be noisy

I guess my assumption is mainly "debug" => dev, "info" => prod

Maybe I'm unique and other people default to "info" if "log" isn't available, and this methodology would be difficult to get folks to adopt.

I could see a nice use-case of turning debug logs to info logs if we start seeing bugs in prod that we cant repro locally, so that we can debug the issue based on breadcrumbs

audius-infra:

audius-infra:
Preview this change https://demo.audius.co/rj-bye-console-log

dylanjeffers (approved):
noice

PR #3932, opened 2023-08-23
[PAY-1727] USDC Withdrawals saga pt. 1

dharit-tan:
### Description
Part 1 takes us through swapping USDC for sol on the sender's side, in the case of a fresh account with no USDC ATA.

Also updates Jupiter service to handle ExactOut swapmode.

Scope creep: added a libs and solana web service and updated `BuyAudio` and `WithdrawUSDC` services to use those.

There are a few TODOs
- Handling existing sol balances
- Handling funding of root solana account

### How Has This Been Tested?

Manually inspected instructions/logs to verify behavior against the POC script.
Did not test against relay bc relay won't accept it yet.

### Feature Flags ###

This flow will only be launchable from USDC UI which should all be feature-flagged.


packages/web/src/services/audius-backend/WithdrawUSDC.ts
```diff
@@ -0,0 +1,224 @@
...
+  swapUSDCForSol(
+    desiredSolAmount,
+    solanaRootAccount.publicKey,
+    usdcUserBank,
+    solanaUSDCATA,
+    feePayerPubkey
+  )
+}
+
+export const getOrCreateDestinationATA = async (
```
schottra, line 193: Do you mind adding jsdocs to exported functions so it's easier to understand them when using in calling code?

packages/web/src/services/audius-backend/WithdrawUSDC.ts
```diff
@@ -0,0 +1,224 @@
...
+} from '@solana/spl-token'
+import { LAMPORTS_PER_SOL, PublicKey, Transaction } from '@solana/web3.js'
+
+import { JupiterSingleton } from 'services/audius-backend/Jupiter'
+import { waitForLibsInit } from 'services/audius-backend/eagerLoadUtils'
+
+import { getRootSolanaAccount, getSolanaConnection } from './BuyAudio'
+
+// @ts-ignore
+const libs = (): AudiusLibs => window.audiusLibs
```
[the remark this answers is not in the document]
schottra: Second thought:

I'm wondering if we should discuss a plan for what our direction is for libs/backend stuff. I only mention it because this file differs from the pattern in `solana.ts` where a backend instance is passed in, each function that needs lib is written in dependency injection style, and functions wait on libs directly. In this file, we're reading it directly off the window instead of dependency-injecting the backend instance.
I can see pros and cons to both ways. And I think ultimately we're trying to get to a place where we can have most code use a well-typed libs instance that doesn't require using overrides (`!`) and ignores in various places to get around the fact that we can't be sure the instance is correctly initialized 😞 .

One additional argument for the DI route, though it's a bit tedious, is that we aren't directly bound to reading something off the window. Which would make things slightly more testable if/when we get to that point.

Anyway, a lot of text to say that maybe if we aren't going to actually use a backend instance here, we should move it out of the `audius-backend` subfolder and up to its own folder in `services`. 

Thoughts?
cc @rickyrombo

packages/web/src/services/audius-backend/WithdrawUSDC.ts
```diff
@@ -0,0 +1,224 @@
...
+const createUSDCATAInstruction = async (
+  solanaUSDCATA: PublicKey,
+  solanaRootAccount: PublicKey,
+  feePayer: PublicKey
+) => {
+  const createSolanaUSDCAccountInstr =
+    Token.createAssociatedTokenAccountInstruction(
+      ASSOCIATED_TOKEN_PROGRAM_ID, // associatedProgramId
+      TOKEN_PROGRAM_ID, // programId
+      libs().solanaWeb3Manager!.mints.usdc, // mint
```
schottra, line 61: I don't know why but this feels a little odd to me.
Technically, wouldn't this require waiting on libs?

packages/web/src/services/audius-backend/WithdrawUSDC.ts
```diff
@@ -0,0 +1,224 @@
...
+  return closeSolanaUSDCAccountInstr
+}
+
+// Swaps USDC for SOL, taking USDC from the user's USDC user bank and
+// depositing SOL into the user's root solana account.
+const swapUSDCForSol = async (
+  amount: number,
+  solanaRootAccountPubkey: PublicKey,
+  usdcUserBank: PublicKey,
+  solanaUSDCATA: PublicKey,
```
schottra, line 90: Minor nit: How do you feel about expanding ATA out to AssociatedTokenAccount where it makes sense in this file to make this easier to follow for folks coming in later who aren't as familiar with the concepts.

schottra (approved):
Overall looks good. Might want another pair of eyes since I'm only moderately familiar with how all of this works.

audius-infra:
Preview this change https://demo.audius.co/rt-usdc-saga

packages/web/src/services/audius-backend/WithdrawUSDC.ts
```diff
@@ -0,0 +1,224 @@
...
+}
+
+const createAndFundDestinationATA = async (
+  destinationPubkey: PublicKey,
+  feePayerPubkey: PublicKey
+) => {
+  // TODO: factor in existing sol balance
+  // TODO: might have to pay rent for root sol account, see BuyAudio.ts
+  const desiredSolAmount =
+    (await getRentAndTransferTransactionFee(destinationPubkey)) /
```
rickyrombo, line 230: nit: I think calling/reusing `getRent` and `getTransactionFee` and summing them here makes more sense than wrapping in another fn

packages/web/src/services/audius-backend/WithdrawUSDC.ts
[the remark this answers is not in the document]
rickyrombo: good call, I think most of these should probably live under a `solana` folder instead

packages/web/src/services/audius-backend/WithdrawUSDC.ts
```diff
@@ -0,0 +1,224 @@
...
+    feePayerPubkey
+  )
+
+  const instructions = [
+    createATAInstruction,
+    ...transferInstructions,
+    ...swapInstructions,
+    closeATAInstruction
+  ]
+  return await libs().solanaWeb3Manager!.transactionHandler.handleTransaction({
```
rickyrombo, line 195: super nit: maybe this fn can return the transaction, and the saga can be responsible for sending it (with or without libs/transaction handler?)

audius-infra:
Preview this change https://demo.audius.co/rt-usdc-saga

audius-infra:

audius-infra:
Preview this change https://demo.audius.co/rt-usdc-saga

rickyrombo (approved):

audius-infra:

audius-infra:
Preview this change https://demo.audius.co/rt-usdc-saga

PR #3937, opened 2023-08-23
[C-2968] Fix private collection action buttons

dylanjeffers:
### Description

Fixes issue where viewers of a private collection can repost/favorite/share.


audius-infra:
Preview this change https://demo.audius.co/dj-c-2968-private-collection

Kyle-Shanks (approved):
I thought that if you disabled the button, the tooltip would not appear, but if you tested this and it works, then awesome!

dylanjeffers:
> I thought that if you disabled the button, the tooltip would not appear, but if you tested this and it works, then awesome!

So it had to with missing the pointer-events style: 
image
now that that's there, disabled buttons can still show tooltips!

PR #3939, opened 2023-08-24
Update pull_request_template.md

raymondjacobson:
### Description

We don't use the other fields ever... Maybe we will have higher quality with two. This is the template we use for the protocol repo.

### How Has This Been Tested?

_Please describe the tests that you ran to verify your changes. Provide repro instructions & any configuration._

I used it here

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,5 +1,8 @@
 ### Description
 
+
 ### How Has This Been Tested?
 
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,18 +1,5 @@
 ### Description
 
-### Dragons
```
amendelsohn, line 3: I use dragons pretty frequently. I would prefer to keep a section for callouts to the reviewer

rickyrombo (approved):

PR #3942, opened 2023-08-24
[C-2987] Add UserGeneratedText

dylanjeffers:
### Description

Adds UserGeneratedText using Linkify, and ensuring text is consistently formatted.
Improves a bunch of other components to make this work.

TODO: There is one more linkify instance in messages @rickyrombo leaving that for you to make sure it's done properly. happy to try my hand though

### How Has This Been Tested?

Checked every usage multiple times

### DIFF
--- a/packages/web/src/components/user-generated-text/UserGeneratedText.module.css
+++ b/packages/web/src/components/user-generated-text/UserGeneratedText.module.css
@@ -1,0 +1,4 @@
+.root {
+  white-space: pre-line;
+  /* user-select: none; */
+}
--- a/packages/web/src/components/user-generated-text/UserGeneratedText.tsx
+++ b/packages/web/src/components/user-generated-text/UserGeneratedText.tsx
@@ -1,6 +1,10 @@
 import { forwardRef, useMemo, MouseEvent } from 'react'
 
-import { getPathFromAudiusUrl, isAudiusUrl } from '@audius/common'
+import {
+  getPathFromAudiusUrl,
+  isAudiusUrl,
+  squashNewLines
+} from '@audius/common'
 import cn from 'classnames'
 import Linkify from 'linkify-react'
 import { IntermediateRepresentation, Opts } from 'linkifyjs'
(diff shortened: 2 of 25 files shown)


audius-infra:
Preview this change https://demo.audius.co/dj-c-2987-user-generated-text

packages/web/src/components/user-generated-text/UserGeneratedText.module.css
```diff
@@ -0,0 +1,4 @@
+.root {
+  white-space: pre-line;
+  /* user-select: none; */
```
rickyrombo, line 3: nit: comment

packages/web/src/components/user-generated-text/UserGeneratedText.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  const children =
+    typeof childrenProp === 'string'
+      ? squashNewLines(childrenProp)
+      : childrenProp
+
+  return (
+    <Linkify
+      options={options}
+      // @ts-expect-error
```
rickyrombo, line 80: why is this needed?

rickyrombo (approved):
This is awesome, killer work! Love that this has more deletions than additions <3

I can make the DMs use this in a follow up, do you mind making a ticket?

packages/web/src/components/user-generated-text/UserGeneratedText.module.css
dylanjeffers: nice catch

packages/web/src/components/user-generated-text/UserGeneratedText.tsx
dylanjeffers: linkify's "as" isn't as generic as it should be basically :/

rickyrombo: sad :( they should have used asChild :joy:

dylanjeffers: after doing this pr, i have seen the light with asChild though, i think for harmony we may want to go with that instead

PR #3943, opened 2023-08-25
Use does_current_user_subscribe API field

michellebrier:
### Description
Discovery v0 API returns a `does_current_user_subscribe` field now, just like it has `does_current_user_follow`. Use this to render the subscribe button on prof pages rather than issuing a separate network req.

### How Has This Been Tested?
pointed client to stage, verified via debugger the API field was being used. subscribed/unsubscribed + navigated away and back to verify caching


dylanjeffers (approved):
nice! also lmk when all DNs are updated cause we can do some further simplification with subscriptions!

audius-infra:
Preview this change https://demo.audius.co/mbrier/con-835/client-curr-user-subscribed

PR #3946, opened 2023-08-25
[PAY-1748][PAY-1731][PAY-1729][PAY-1730] DMs link fixes

rickyrombo:
### Description

- Fix overflows in link previews
- Use `UserGeneratedText` in DMs
- Add `LeavingAudius` modal for all external links not just DMs (sans social links)

### How Has This Been Tested?

Tested bio links, social links, and DM links

### Screenshots

Non-english, non-hyphenated:
image

English, hyphenated:
image

### DIFF
--- a/packages/mobile/src/components/core/Hyperlink.tsx
+++ b/packages/mobile/src/components/core/Hyperlink.tsx
@@ -52,7 +52,6 @@ export const Hyperlink = (props: HyperlinkProps) => {
     source,
     styles: stylesProp,
     style,
-    warnExternal,
     ...other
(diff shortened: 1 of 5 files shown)


dylanjeffers (approved):
not much more i can say, looks perfect. do we need the www duplicate for the allowlist?

rickyrombo:
> do we need the www duplicate for the allowlist?

@dylanjeffers probably not, but better safe than sorry 🤷 lmk if there's other sites I should add. can also make this an optimizely config if necessary

PR #3948, opened 2023-08-26
[PAY-1733] Remove Gated Prompt Modal

rickyrombo:
### Description

Removes dead code


raymondjacobson (approved):
number go up! (or down in this case i guess)

PR #3949, opened 2023-08-26
Fix Share to IG/T/S sticker consistency [C-3009]

nicoback2:
### Description
Bug: share to X sticker (square with track/artist info) would not show up sometimes

This was a doozy

- Keep Share Drawer open while story being generated
  - Since the share drawer would close right after the button was clicked, the sticker component would occasionally unmount  before the capture was taken 
  -  Note: I tried to move the sticker component out to both the App level and the on progress drawer, but this did not work after hours of trying for various reasons 💀. Can try again another time - I think the drawer thing is an OK solution for now. 

- Fixing this issue now allows us to compute dominant colors and take the screenshot concurrently for perf win

### How Has This Been Tested?
On Android device

### Screenshots

https://github.com/AudiusProject/audius-client/assets/36916764/bef2bfe1-7fe0-4724-8cfd-0bfe2ce9b01d

### DIFF
--- a/packages/mobile/src/components/share-drawer/useShareToStory.tsx
+++ b/packages/mobile/src/components/share-drawer/useShareToStory.tsx
@@ -293,19 +293,14 @@ export const useShareToStory = ({
         toggleProgressDrawer(true, platform)
         dispatch(setProgress(10))
 
-        // Step 1/2: Render and take a screenshot of the sticker, calculate the dominant colors of the cover art:
+        // Step 1: Render and take a screenshot of the sticker
         let stickerUri: string | undefined
+
+        // Step 2: Calculate the dominant colors of the cover art:
         let rawDominantColorsResult: Color[] | string | undefined // `getDominantRgb` returns a string containing a single default color if it couldn't find dominant colors
-        let dominantColorHex1: string
-        let dominantColorHex2: string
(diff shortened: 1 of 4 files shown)


packages/mobile/src/components/share-drawer/useShareToStory.tsx
```diff
@@ -291,30 +291,36 @@ export const useShareToStory = ({
...
-        const storyVideoPath = path.join(
-          RNFS.TemporaryDirectoryPath,
-          `storyVideo-${uuid()}.mp4`
-        )
-        const audioStartOffsetConfig =
-          content.track.duration && content.track.duration >= 20
-            ? '-ss 10 '
-            : ''
-
+        // Step 1/2: Render and take a screenshot of the sticker, calculate the dominant colors of the cover art:
```
raymondjacobson, line 296: maybe this could just be step 1 and step 2 still. this confused me for a sec

raymondjacobson (approved):
wow!!

packages/mobile/src/components/share-drawer/useShareToStory.tsx
nicoback2: oo ya for sure

PR #3960, opened 2023-08-28
[C-3023] Fix collectible visibility

dylanjeffers:
### Description

Fixes issue where user's collectibles incorrectly show up even though a wallet has been removed. Ultimately the issue has to do with the cached user, and us preferring stale values over new ones.

### DIFF
--- a/packages/common/src/store/cache/reducer.ts
+++ b/packages/common/src/store/cache/reducer.ts
@@ -87,7 +87,9 @@ const unwrapEntry = (entry: { metadata: any }) => {
 const forceUpdateKeys = new Set([
   'field_visibility',
   'followee_reposts',
-  'followee_saves'
+  'followee_saves',
+  'associated_wallets',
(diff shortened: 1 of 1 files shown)


packages/common/src/store/cache/reducer.ts
```diff
@@ -87,7 +87,9 @@ const unwrapEntry = (entry: { metadata: any }) => {
 const forceUpdateKeys = new Set([
   'field_visibility',
   'followee_reposts',
-  'followee_saves'
+  'followee_saves',
+  'associated_wallets',
```
nicoback2, line 91: what's this for?

nicoback2 (approved):

packages/common/src/store/cache/reducer.ts
dylanjeffers: i refactored those two over to this one cause the intent is the same, which is to always prefer the incoming value

PR #3961, opened 2023-08-29
[C-3025] Fix race-condition with stem/remix fetch

dylanjeffers:
### Description

Fixes issue where user is unable to play tracks coming from search page. This is due to a race condition where we fetch a tracks remixes and stems before the track itself, causing cases where remix values return and update the track entry before it even exists.

By moving these fetches to after the track is fetched, like we do with `fetchTrackBySlugAndHandle`, it works as expected.

### How Has This Been Tested?

searching and going to https://audius.co/russdiemon/freed-up-107829 works on mobile.


michellebrier (approved):
nice

PR #3966, opened 2023-08-29
[C-3026] Close now-playing when sharing to dm

dylanjeffers:
### Description

Fixes issue where sharing to DM from the now-playing drawer does not close the drawer, leading to a funky user state.

### How Has This Been Tested?

Tested on local ios

### Screenshots

https://github.com/AudiusProject/audius-client/assets/8230000/975dbf00-db2f-4ad6-a44a-89733518f6c0


rickyrombo (approved):
Great catch!

PR #3967, opened 2023-08-29
Fix jupiter Buffer is not defined

sliptype:
### Description
Importing Jupiter in the new workspaces setup throws a `Buffer is not defined` error at runtime. This seems to be bc Jupiter expects global `Buffer` to exist when it is imported, but bc of the workspaces change, Buffer doesn't exist until `index.tsx` is run. 

We really only need the string value of the enum so we can avoid importing Jupiter for now. The real fix will be to make webpack ProvidePlugin properly put Buffer into the global namespace


amendelsohn (approved):

rickyrombo (approved):
Thanks for this! Strange one

PR #3968, opened 2023-08-29
[PAY-1590] Hook up content previews for USDC content

schottra:
### Description
This PR hooks up track previews for web and mobile web.

I added logic to the player saga to detect attempts to play tracks to which the user does not have access, switching to a preview instead if its available. This can be overridden by passing a boolean flag to the action. The previewing state is stored in the slice for retrieval later, which is helpful for determining the source of playback (full access of preview). I opted _not_ to use a new UID for this as it felt like it would have gotten quite complicated attempting to combine multiple sources (previewing from a lineup vs playing the same track from a lineup?). 

The TrackPageProvider is the only place that uses the boolean override, since we need to allow the track owner to preview the track.

The track page also makes use of the previewing state flag to determine which button shows "pause" when we're playing a preview. And we also make use of the previewing flag in the queue saga to force a reset if you switch from preview to full access.

### How Has This Been Tested?
Tested locally in Chrome (desktop and mobile web emulation)

### Screenshots

Playing Preview for non-owner

https://github.com/AudiusProject/audius-client/assets/1815175/95cd5ebb-ee1a-408c-ac49-ace4555dc211

---

Playing Preview for non-owner from lineups

https://github.com/AudiusProject/audius-client/assets/1815175/7a87bf0f-2c56-4bf7-85f8-0ffbc928b1e1

---

Playing preview for owner

https://github.com/AudiusProject/audius-client/assets/1815175/03f844f8-0252-46ae-8112-1c515376d139

### DIFF
--- a/packages/web/src/common/store/player/sagas.ts
+++ b/packages/web/src/common/store/player/sagas.ts
@@ -66,11 +66,12 @@ const { getIsReachable } = reachabilitySelectors
 const PLAYER_SUBSCRIBER_NAME = 'PLAYER'
 const RECORD_LISTEN_SECONDS = 1
 const RECORD_LISTEN_INTERVAL = 1000
+const PREVIEW_LENGTH_SECONDS = 15
 
 export function* watchPlay() {
   const getFeatureEnabled = yield* getContext('getFeatureEnabled')
   yield* takeLatest(play.type, function* (action: ReturnType<typeof play>) {
-    const { uid, trackId, onEnd } = action.payload ?? {}
+    const { uid, trackId, isPreview, onEnd } = action.payload ?? {}
 
     const audioPlayer = yield* getContext('audioPlayer')
     const isNativeMobile = yield getContext('isNativeMobile')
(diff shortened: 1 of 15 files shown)


audius-infra:
Preview this change https://demo.audius.co/pay-1590-hook-up-previews

packages/web/src/common/store/player/sagas.ts
```diff
@@ -113,6 +114,19 @@ export function* watchPlay() {
...
+
+      let trackDuration = track.duration
+
+      if (isPreview) {
+        // Add preview query string and calculate preview duration for use later
+        const previewStartSeconds = track.preview_start_seconds || 0
+        queryParams.preview = true
+        trackDuration = Math.min(
+          track.duration - previewStartSeconds,
+          PREVIEW_LENGTH_SECONDS
```
rickyrombo, line 126: nice, good looks

packages/web/src/common/store/queue/sagas.ts
```diff
@@ -181,13 +187,16 @@ export function* watchPlay() {
...
+        (uid !== playerUid || !!isPreview !== playerIsPreviewing)
       if (noTrackPlaying || trackIsDifferent || trackIsSameButDifferentUid) {
         yield* put(
           playerActions.play({
             uid,
+            isPreview,
             trackId: playActionTrack.track_id,
-            onEnd: next
+            // Don't auto-advance after previews
+            onEnd: isPreview ? playerActions.stop : next
```
rickyrombo, line 199: Are you sure that's what we want? In lineups, that would make streaming stop, no?

rickyrombo (approved):
This is awesome, great work

packages/web/src/common/store/queue/sagas.ts
schottra: I can follow up with design and double check. Generally I think we need to clarify the subtleties of how previewable tracks interact with lineups and queu.

PR #3969, opened 2023-08-29
Fix embed ci with workspaces

sliptype:
### Description

Update embed CI flows to work in new workspaces set up


dylanjeffers (approved):
nice find

PR #3975, opened 2023-08-30
Add category to library query endpoint + more useAllPaginated fixes

nicoback2:
### Description
-Add category param to library audius-query
-Some more fixes for useAllPaginatedQuery
-- See comments

### DIFF
--- a/packages/common/src/audius-query/hooks/usePaginatedQuery.ts
+++ b/packages/common/src/audius-query/hooks/usePaginatedQuery.ts
@@ -95,30 +95,30 @@ export const useAllPaginatedQuery = <
   )
 
   const notError = result.status !== Status.ERROR
-  const notStillLoadingCurrentPage =
-    !loadingMore && !(result.status === Status.LOADING)
+  const stillLoadingCurrentPage =
+    loadingMore || result.status === Status.LOADING
   const notStarted = result.status === Status.IDLE && allData.length === 0
   const hasNotFetched = !result.data && result.status !== Status.SUCCESS
   const fetchedFullPreviousPage = result.data?.length === pageSize
 
   const hasMore =
     notError &&
-    notStillLoadingCurrentPage &&
+    !stillLoadingCurrentPage &&
     (notStarted || hasNotFetched || fetchedFullPreviousPage)
 
   const loadMore = useCallback(() => {
-    if (!notStillLoadingCurrentPage) {
+    if (stillLoadingCurrentPage) {
       return
     }
     setLoadingMore(true)
     setPage(page + 1)
-  }, [notStillLoadingCurrentPage, page])
+  }, [stillLoadingCurrentPage, page])
   return {
     ...result,
     // TODO: add another status for reloading
     status: allData?.length > 0 ? Status.SUCCESS : status,
     data: allData,
-    isLoadingMore: !notStillLoadingCurrentPage,
+    isLoadingMore: stillLoadingCurrentPage,
     loadMore,
     hasMore
   }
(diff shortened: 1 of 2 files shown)


packages/common/src/audius-query/hooks/usePaginatedQuery.ts
```diff
@@ -59,55 +59,61 @@ export const useAllPaginatedQuery = <
...
-    ...baseArgs,
-    limit: pageSize,
-    offset: page * pageSize
-  } as ArgsType
-  const result = useQueryHook(args, queryHookOptions)
 
   useCustomCompareEffect(
     () => {
       setAllData([])
+      setPage(0)
```
nicoback2, line 66: Make sure we fetch from beginning if params are changed

packages/common/src/audius-query/hooks/usePaginatedQuery.ts
```diff
@@ -59,55 +59,61 @@ export const useAllPaginatedQuery = <
...
+  const notStillLoadingCurrentPage =
+    !loadingMore && !(result.status === Status.LOADING)
   const notStarted = result.status === Status.IDLE && allData.length === 0
   const hasNotFetched = !result.data && result.status !== Status.SUCCESS
   const fetchedFullPreviousPage = result.data?.length === pageSize
 
   const hasMore =
-    notError && (notStarted || hasNotFetched || fetchedFullPreviousPage)
+    notError &&
+    notStillLoadingCurrentPage &&
```
nicoback2, line 101: Need this check here to prevent calling loadMore in the wrong order of operations on base params change

audius-infra:
Preview this change https://demo.audius.co/nkang--paginated-query-again

packages/common/src/audius-query/hooks/usePaginatedQuery.ts
```diff
@@ -60,57 +60,65 @@ export const useAllPaginatedQuery = <
...
+    (notStarted || hasNotFetched || fetchedFullPreviousPage)
 
   const loadMore = useCallback(() => {
-    if (loadingMore) {
+    if (!notStillLoadingCurrentPage) {
```
amendelsohn, line 110: if we're going to negate this here, I think I'd rather have the var be `isStillLoadingCurrentPage` and use `!` when we want the negative case. Avoids the double negative

amendelsohn: would it be better to check `hasMore` here since it includes the loadingCurrentPage check?

packages/common/src/audius-query/hooks/usePaginatedQuery.ts
```diff
@@ -60,57 +60,65 @@ export const useAllPaginatedQuery = <
...
     setPage(page + 1)
-  }, [loadingMore, page])
-
+  }, [notStillLoadingCurrentPage, page])
   return {
     ...result,
     // TODO: add another status for reloading
     status: allData?.length > 0 ? Status.SUCCESS : status,
     data: allData,
+    isLoadingMore: !notStillLoadingCurrentPage,
```
amendelsohn, line 121: yeah let's just reverse the variable

amendelsohn (approved):
Awesome, thank you for the fixes!!

amendelsohn:
@schottra since I know you're working with `useAllPaginatedQuery` as well. Maybe check if this conflicts or causes/fixes any bugs with your current changes

dylanjeffers:
@nicoback any chance this could get merged soon? would like to use the useAllPaginated fixes for something :)

nicoback2:
> @nicoback any chance this could get merged soon? would like to use the useAllPaginated fixes for something :)

oh yeah can get on this in an hour or so!

packages/common/src/audius-query/hooks/usePaginatedQuery.ts
nicoback2: i prefer the loading more since it's clearer but i think hasmore would probably work too!

amendelsohn: I'm happy either way

PR #3976, opened 2023-08-30
Fix text color for DMs

rickyrombo:
### Description

Fixes bug where non-links weren't correctly colored.

### BEFORE
image
image

### AFTER
image
image

### DIFF
--- a/packages/web/src/pages/chat-page/components/ChatMessageListItem.module.css
+++ b/packages/web/src/pages/chat-page/components/ChatMessageListItem.module.css
@@ -20,7 +20,6 @@
   display: inline-block;
   max-width: 448px;
   filter: var(--drop-shadow-near);
-  color: var(--static-netural);
   transition: background-color var(--expressive), filter var(--expressive);
 }
 
@@ -40,7 +39,6 @@
 }
 
 .isAuthor .bubble {
-  color: var(--static-white);
   --bubble-color: var(--secondary-light-2);
 }
 
--- a/packages/web/src/pages/chat-page/components/ChatMessageListItem.tsx
+++ b/packages/web/src/pages/chat-page/components/ChatMessageListItem.tsx
@@ -207,9 +207,12 @@ export const ChatMessageListItem = (props: ChatMessageListItemProps) => {
             />
           ) : null}
           {!hideMessage ? (
-            <div className={styles.text}>
-              <UserGeneratedText>{message.message}</UserGeneratedText>
-            </div>
+            <UserGeneratedText
+              className={styles.text}
+              color={isAuthor ? 'staticWhite' : 'neutral'}
+            >
+              {message.message}
+            </UserGeneratedText>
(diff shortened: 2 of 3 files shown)


packages/web/src/pages/chat-page/components/ChatMessageListItem.module.css
```diff
@@ -20,7 +20,6 @@
   display: inline-block;
   max-width: 448px;
   filter: var(--drop-shadow-near);
-  color: var(--static-netural);
```
rickyrombo, line 23: turns out this was a typo anyway

packages/web/src/pages/chat-page/components/ChatMessageListItem.tsx
```diff
@@ -207,9 +207,12 @@ export const ChatMessageListItem = (props: ChatMessageListItemProps) => {
             />
           ) : null}
           {!hideMessage ? (
-            <div className={styles.text}>
-              <UserGeneratedText>{message.message}</UserGeneratedText>
-            </div>
+            <UserGeneratedText
+              className={styles.text}
+              color={isAuthor ? 'staticWhite' : 'neutral'}
```
rickyrombo, line 212: lowkey liked it better when it was using CSS

dylanjeffers (approved):
nice! if it was css then you'd have to have logic for classnames, which imo is more work than just passing a color prop?

rickyrombo:
> nice! if it was css then you'd have to have logic for classnames, which imo is more work than just passing a color prop?

@dylanjeffers but here the color comes as a direct result of a class on the parent. Having one check for `isAuthor` at the root level and passing that class and then styling the bubble, text, links, tail, etc all based on that is much easier than having to check `isAuthor` in every subcomponent's props for various styles. Eg. I can style the bubble and use as many `Text` components as I want without having to manually style their color each time.

I would have preferred the `Text` component not automatically set the color and default to the default stylesheet, so that it's easily overridden by CSS

dylanjeffers:
> > nice! if it was css then you'd have to have logic for classnames, which imo is more work than just passing a color prop?
> 
> 
> 
> @dylanjeffers but here the color comes as a direct result of a class on the parent. Having one check for `isAuthor` at the root level and applying the class `.isAuthor`, and then styling the bubble, text, links, tail, etc all based on that is much easier than having to check `isAuthor` in every subcomponent's props for various styles. Eg. I can style the bubble and use as many `Text` components as I want without having to manually style their color each time.
> 
> 
> 
> I would have preferred the `Text` component not automatically set the color and default to the default stylesheet, so that it's easily overridden by CSS

Ah gotcha, I can see how that would be nice, to style inner complements with css. Id like to check with folks if that's actually what our preference should be, cause it kinda breaks encapsulation a bit. It sounds like what you need is to use the "inherit" variant on text, in that cause it'll just be whatever the upper parent tells it to be!

PR #3977, opened 2023-08-30
[C-2921] Update subgenre value passed in the genre field

Kyle-Shanks:
### Description
Update the subgenre value that is passed to remove the 'Electronic - ' in the beginning

### How Has This Been Tested?
Manually tested

### Screenshots
Screenshot 2023-08-30 at 2 12 34 PM


packages/common/src/store/upload/selectors.ts
Kyle-Shanks, line 1: this was another update that i spotted, but not related to the ticket

PR #3978, opened 2023-08-30
Fix TextField input spacing

dylanjeffers:
### Description

Fixes issue where "demiBold" fontFamily leads to input text settling too high in it's container. this issue only exists on ios

### Screenshots

Before: 

image

After:
image


Kyle-Shanks (approved):
ezpz

PR #3981, opened 2023-08-30
Fix validation on access and sale modal

amendelsohn:
### Description

Fixes release-blocking issue with access and sale modal.

Fixes
- Form was not able to be submitted when a validation error was present in other hidden fields
- Cleans form values when swapping between availability types

### How Has This Been Tested?
local web

### DIFF
--- a/packages/web/src/pages/upload-page/fields/AccessAndSaleField.tsx
+++ b/packages/web/src/pages/upload-page/fields/AccessAndSaleField.tsx
@@ -23,7 +23,7 @@ import {
   RadioButtonGroup
 } from '@audius/stems'
 import cn from 'classnames'
-import { useField } from 'formik'
+import { useField, useFormikContext } from 'formik'
 import { get, isEmpty, set } from 'lodash'
 import { useSelector } from 'react-redux'
 import { z } from 'zod'
@@ -515,6 +515,10 @@ export const AccessAndSaleMenuFields = (props: AccesAndSaleMenuFieldsProps) => {
       setfieldVisibilityValue
     ]
   )
+
+  const { values, errors } = useFormikContext()
+
+  console.log({ values, errors })
 
   return (
     <div className={cn(layoutStyles.col, layoutStyles.gap4)}>
--- a/packages/web/src/pages/upload-page/fields/AccessAndSaleField.tsx
+++ b/packages/web/src/pages/upload-page/fields/AccessAndSaleField.tsx
@@ -114,7 +114,7 @@ const messages = {
     preview: {
       tooEarly: 'Preview must start during the track.',
       tooLate:
-        'Preview must start at lest 30 seconds before the end of the track.'
+        'Preview must start at least 30 seconds before the end of the track.'
     }
   }
 }
@@ -519,10 +519,10 @@ export const AccessAndSaleMenuFields = (props: AccesAndSaleMenuFieldsProps) => {
             isPremiumContentCollectibleGated(premiumConditionsValue)
           )
             break
-          setPremiumConditionsValue(null)
+          setPremiumConditionsValue({})
           break
         case TrackAvailabilityType.HIDDEN:
-          setPremiumConditionsValue(null)
+          setPremiumConditionsValue({})
           if (!fieldVisibilityValue) break
           setfieldVisibilityValue({
             ...fieldVisibilityValue,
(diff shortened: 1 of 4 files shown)


packages/web/src/pages/upload-page/fields/AccessAndSaleField.tsx
```diff
@@ -140,26 +140,47 @@ export type AccessAndSaleFormValues = {
...
+      },
+      { message: messages.errors.preview.tooEarly, path: [PREVIEW] }
+    )
+    .refine(
+      (values) => {
+        const formValues = values as AccessAndSaleFormValues
+        if (isPremiumContentUSDCPurchaseGated(formValues[PREMIUM_CONDITIONS])) {
+          return (
+            formValues[PREVIEW] === undefined ||
+            (formValues[PREVIEW] >= 0 && formValues[PREVIEW] < trackLength - 15)
```
rickyrombo, line 177: fwiw it turns out track previews are implemented on the content end as 30s, so maybe easiest to update the UI side? Either way, maybe a comment/constant for the 15 here?

rickyrombo (approved):
Noticed on stage I was unable to change access from gated to public and have it stick? Is that related?

packages/web/src/pages/upload-page/fields/AccessAndSaleField.tsx
amendelsohn: ah great call. I'll make it 30 before the end

audius-infra:
Preview this change https://demo.audius.co/amendel-access-sale-regression-fix

amendelsohn:
> Noticed on stage I was unable to change access from gated to public and have it stick? Is that related?

Digging into this... looks like we're calling `confirmEditTrack` with `premium_conditions: null` and `is_premium: false`. Wondering if the `null` is getting dropped during the entity merge not replacing the premium conditions?

rickyrombo:
> > Noticed on stage I was unable to change access from gated to public and have it stick? Is that related?
> 
> Digging into this... looks like we're calling `confirmEditTrack` with `premium_conditions: null` and `is_premium: false`. Wondering if the `null` is getting dropped during the entity merge not replacing the premium conditions?
> 
> When the track reloads later, it has `premium_conditions: { follower_id: 12345 }` but `is_premium: false`, and the dog ear is visible but not the "Special Access" label

That sounds like what's happening. https://github.com/AudiusProject/audius-protocol/blob/7b3120fd1da44563bd5530698ac91ab6c49a2594/discovery-provider/src/tasks/entity_manager/entities/track.py#L220-L226

Maybe we should send an empty object for premium conditions?

packages/web/src/pages/upload-page/fields/AccessAndSaleField.tsx
```diff
@@ -109,12 +109,12 @@ const messages = {
...
     price: {
       tooLow: 'Price must be at least $0.99.',
-      tooHigh: 'Price must be less than $9.99.'
+      tooHigh: 'Price must be less than $9999.99.'
     },
     preview: {
       tooEarly: 'Preview must start during the track.',
       tooLate:
-        'Preview must start at lest 15 seconds before the end of the track.'
+        'Preview must start at lest 30 seconds before the end of the track.'
```
dylanjeffers, line 117: Least*

packages/web/src/pages/upload-page/fields/AccessAndSaleField.tsx
```diff
@@ -109,12 +109,12 @@ const messages = {
   errors: {
     price: {
       tooLow: 'Price must be at least $0.99.',
-      tooHigh: 'Price must be less than $9.99.'
+      tooHigh: 'Price must be less than $9999.99.'
```
dylanjeffers, line 112: Lol

dylanjeffers (approved):
Awesome. Now just checking the usdc field is hidden in prod anyway right?

audius-infra:
Preview this change https://demo.audius.co/amendel-access-sale-regression-fix

amendelsohn:
> Maybe we should send an empty object for premium conditions?

Oof sent `premium_conditions: {}` but still getting old `premium_conditions: { follower_id: 123 }` from discovery after confirmation.

amendelsohn:
> Awesome. Now just checking the usdc field is hidden in prod anyway right?

Correct, though the other bug affects prod track edit

PR #3983, opened 2023-08-30
Revert premium_conditions empty object change

amendelsohn:
### Description

Revert change to use `{}` instead of `null` for empty `premium_conditions`

### How Has This Been Tested?
local web


rickyrombo (approved):
Thanks for trying this

PR #3986, opened 2023-08-31
[PAY-1591][PAY-1775] Implements content previews on native mobile

schottra:
### Description
This PR adds plumbing to the mobile component layer to pass the preview flag in the same way as web.

In addition, I updated the data model of our Queue slice to support holding the preview flag in the `Queueable` model. 

Playback on mobile does not actually respond to any of the common player actions. The implementation for `audioPlayer` is stubbed out. So the logic we have in player sagas for reading the incoming action's `isPreview` flag and calculating the correct URL/duration have no effect.

Instead, mobile uses its own logic in a component `Audio.tsx`, where it watches for changes to the queue redux slice and does similar logic to what occurs in the player sagas. Additionally, the Now Playing components on mobile do not derive duration from the player the way that web does. Instead, they read duration from the track. So the logic needed to be updated to determine if we're currently playing the preview version of the track (which _is_ read from the player state). In the preview case, we now use a shared common function to determine the duration of the preview stream.

There is also logic in the `Audio` component to watch for auto-advance from the native player and sync changes back to the player state. So I needed to make sure that all the player actions set the proper values for `previewing` to correctly drop out of the previewing state when we finish a preview track.

It's not the cleanest implementation. But I think a major refactor would be in order to get it much better.

### How Has This Been Tested?
Locally tested on iOS simulator.


audius-infra:
Preview this change https://demo.audius.co/pay-1591-mobile-previews

packages/mobile/src/screens/track-screen/TrackScreenDetailsTile.tsx
```diff
@@ -289,20 +290,37 @@ export const TrackScreenDetailsTile = ({
...
+      uid,
+      isPlayingId,
+      dispatch,
+      isPlaying,
+      isPreviewing,
+      isLineupLoading
+    ]
+  )
+
+  const handlePressPlay = useCallback(() => play({ isPreview: false }), [play])
```
dharit-tan, line 319: if you set up the default args above in line 294, do you need to explicitly pass `isPreview: false` here?

dharit-tan (approved):
lgtm!

packages/mobile/src/screens/track-screen/TrackScreenDetailsTile.tsx
schottra: Oh nope. Clearly I went through a couple iterations on this 😅

PR #3988, opened 2023-08-31
[C-3022] Fix broken multi-track index after deleting

amendelsohn:
#### The problem:
When deleting a track other than the selected track, the form got into a bad state where the selected index could be the deleted track. This was happening because the onClick for the row to select the track as active was also being activated when clicking the delete icon.

#### The fix:
Stop propagation on the onClick for remove track so we don't also set the selected index to the track to be deleted.

#### Also:
- removes the deleted track from the tracks list
- fix bug in selected index calculation after deletion

### How Has This Been Tested?
local web

### DIFF
--- a/packages/web/src/pages/upload-page/fields/MultiTrackSidebar.tsx
+++ b/packages/web/src/pages/upload-page/fields/MultiTrackSidebar.tsx
@@ -1,4 +1,4 @@
-import { useCallback } from 'react'
+import { MouseEvent, useCallback } from 'react'
 
 import { imageBlank as placeholderArt } from '@audius/common'
 import {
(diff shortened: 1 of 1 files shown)


packages/web/src/pages/upload-page/fields/MultiTrackSidebar.tsx
```diff
@@ -166,11 +171,11 @@ const TrackRow = (props: TrackRowProps) => {
             </Text>
           </div>
           {values.trackMetadatas.length > 1 ? (
-            <div className={styles.iconRemove}>
-              <IconTrash
-                fill='--default'
-                onClick={() => handleRemoveTrack(index)}
-              />
+            <div
```
dylanjeffers, line 174: Nice!

dylanjeffers (approved):
Great find. Long term that should be a button not a div but all good

audius-infra:
Preview this change https://demo.audius.co/amendel-upload-delete-tracks-fix

PR #3989, opened 2023-08-31
[C-3014] Add prompt on leaving upload form in progress

amendelsohn:
### Description
Adds a warning before navigating away if the form has any changed values.
Does not trigger on form submission or on the screen's back button.

*Note: Currently, this only blocks navigation within the app. Refreshing the page or navigating away via the address bar does not trip the condition.*

### Screenshots
Screenshot 2023-09-01 at 3 24 50 PM

### DIFF
--- a/packages/web/src/components/navigation-prompt/NavigationPrompt.tsx
+++ b/packages/web/src/components/navigation-prompt/NavigationPrompt.tsx
@@ -1,11 +1,21 @@
 import React, { useEffect, useState } from 'react'
 
-import { Modal, ModalContent, ModalHeader, ModalTitle } from '@audius/stems'
+import {
+  HarmonyButton,
+  HarmonyButtonType,
+  Modal,
+  ModalContent,
+  ModalHeader
+} from '@audius/stems'
+import cn from 'classnames'
 import { Location } from 'history'
 import { Prompt } from 'react-router-dom'
 
-import ConfirmationBox from 'components/confirmation-box/ConfirmationBox'
+import layoutStyles from 'components/layout/layout.module.css'
+import { Text } from 'components/typography'
 import { useNavigateToPage } from 'hooks/useNavigateToPage'
+
+import styles from './NavigationPrompt.module.css'
 
 interface Props {
   when?: boolean | undefined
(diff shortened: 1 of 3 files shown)


dylanjeffers (commented):
Cool! I think we'd want to run this by UX my guess is we should use the prompt we use when user tries to refresh the page when a block confirmation is still in flight

amendelsohn:
> Cool! I think we'd want to run this by UX my guess is we should use the prompt we use when user tries to refresh the page when a block confirmation is still in flight

Ah, true. I'll see if I can use a version of that instead. Thanks!

audius-infra:
Preview this change https://demo.audius.co/amendel-upload-prompt

dylanjeffers (commented):
@amendelsohn also can we just see what ux is thinking for this interaction in general? Like should we use the modal or the alert or something else?

audius-infra:
Preview this change https://demo.audius.co/amendel-upload-prompt

packages/web/src/components/navigation-prompt/NavigationPrompt.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { Location } from 'history'
+import { Prompt } from 'react-router-dom'
+
+import layoutStyles from 'components/layout/layout.module.css'
+import { Text } from 'components/typography'
+import { useNavigateToPage } from 'hooks/useNavigateToPage'
+
+import styles from './NavigationPrompt.module.css'
+
+interface Props {
```
dylanjeffers, line 20: Type and I usually call it componentnameprops

packages/web/src/pages/upload-page/forms/EditTrackForm.tsx
```diff
@@ -107,11 +114,12 @@ export const EditTrackForm = (props: EditTrackFormProps) => {
 }
 
 const TrackEditForm = (props: FormikProps<TrackEditFormValues>) => {
-  const { values } = props
+  const { values, dirty } = props
   const isMultiTrack = values.trackMetadatas.length > 1
 
   return (
     <Form>
+      <NavigationPrompt when={dirty} messages={messages.navigationPrompt} />
```
dylanjeffers, line 122: Sickkkkk

dylanjeffers (approved):
Very cool great work

PR #3991, opened 2023-08-31
[C-3028] make the usdc inputs white

amendelsohn:
### Description

Make the inputs white so they don't look disabled

### Screenshots

image

### DIFF
--- a/packages/web/src/pages/upload-page/fields/availability/UsdcPurchaseFields.module.css
+++ b/packages/web/src/pages/upload-page/fields/availability/UsdcPurchaseFields.module.css
@@ -7,5 +7,5 @@
 }
 
 .inputRoot {
-  background: var(--background-white)
+  background: var(--background-white);
 }
--- a/packages/web/src/pages/upload-page/fields/availability/UsdcPurchaseFields.module.css
+++ b/packages/web/src/pages/upload-page/fields/availability/UsdcPurchaseFields.module.css
@@ -5,3 +5,7 @@
   border: 1px solid var(--border-strong);
   background: var(--background-surface-1);
 }
+
+.input > div:first-child {
+  background: var(--background-white)
+}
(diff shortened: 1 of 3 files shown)


packages/web/src/pages/upload-page/fields/availability/UsdcPurchaseFields.module.css
```diff
@@ -5,3 +5,7 @@
   border: 1px solid var(--border-strong);
   background: var(--background-surface-1);
 }
+
+.input > div:first-child {
```
amendelsohn, line 9: Is this against our css guidelines? The `InputV2` className prop is a level up from this. Any better ideas on how to make this happen?

Kyle-Shanks: i feel like the background should be white by default in the component right? it was probably not set bc its normally used on a white background, but i can't thing of a case where we would want it to have a transparent background.

Kyle-Shanks: alternatively, we could add a prop like inputClassName or something so that we can just pass the class to the right spot

amendelsohn: InputV2 has `background-color: var(--neutral-light-10);` and I didn't want to break that, but yeah could just pass another prop

Kyle-Shanks: ooh thats weird. sounds good 👍

amendelsohn: done

Kyle-Shanks (approved):
nice

PR #3995, opened 2023-08-31
Fix misc embed deploy issues

raymondjacobson:
### Description

* Issues with redirecting to old /embed api
* Issues with iframe restrictions

### How Has This Been Tested?

on prod via wrangler build


dylanjeffers (approved):
awesome! so it's the case that any change we make to embed, we should build and commit the changes?

raymondjacobson:
> awesome! so it's the case that any change we make to embed, we should build and commit the changes?

not necessarily needed, but it's a good idea. deploys should all happen from CI going fwd

PR #3997, opened 2023-08-31
Fix CI (partially)

sliptype:
### Description

Partially fix CI breakages due to npm workspaces. Still a couple outstanding issues including building storybook and the codepush steps


.circleci/src/commands/@web-commands.yml
```diff
@@ -93,6 +93,12 @@ web-distribute:
     - checkout
     - attach_workspace:
         at: ./
+    - run:
+        name: Install dependencies
+        command: rm -rf node_modules && SKIP_POD_INSTALL=true npm ci
```
dylanjeffers, line 98: just curious, we couldn't use the cached version from the init phase?

.circleci/src/jobs/@stems-jobs.yml
```diff
@@ -23,7 +23,7 @@ stems-init:
     - persist_to_workspace:
         root: ./
         paths:
-          - packages/node_modules
+          - node_modules
```
dylanjeffers, line 26: nice

audius-infra:

packages/mobile/android/app/build.gradle
```diff
@@ -58,7 +58,7 @@ project.ext.envConfigFiles = [
 ]
 
 apply from: "../../../../node_modules/react-native-code-push/android/codepush.gradle"
-apply from: "../../node_modules/@sentry/react-native/sentry.gradle"
+apply from: "../../../../node_modules/@sentry/react-native/sentry.gradle"
```
dylanjeffers, line 61: nice

scripts/postinstall.sh
```diff
@@ -1,11 +1,15 @@
...
-# Symlink react-native into the mobile package bc npm doesn't
-# support nohoist
-cd packages/mobile/node_modules
-ln -s ../../../node_modules/react-native react-native
-cd ../ios
-if command -v pod >/dev/null; then
-  pod install
-fi
+
+if [[ -z "${SKIP_POD_INSTALL}" ]]; then
```
dylanjeffers, line 6: sneaky, makes sense tho

dylanjeffers (approved):
the lords work

.circleci/src/commands/@web-commands.yml
sliptype: Actually a very crazy issue that I don't quite understand. Electron builder always fails doing symlinks from the node_gyp artifacts with a `Read Only File System` error. This is stuff that just exists in the node_modules folder. Only those node_gyp artifacts (like `python3` for example) throw this error when you try to modify them. But, making a copy of the file and renaming it to python3 fixes everything. Reinstalling node modules has the same effect. Maybe it has something to do with node_gyp compiling dependencies for one platform in the CI init step and then trying to consume them on a different platform (the mac os circle ci image). But idk how the change in dependency structure would have caused that

audius-infra:
Preview this change https://demo.audius.co/sk-ci-fixes

.circleci/src/commands/@web-commands.yml
dylanjeffers: I see what you mean about maybe compiling for different platforms but to your point it's weird that it worked before. Well something to note for monorepo

PR #3999, opened 2023-09-01
[PAY-1734] Implement skeleton withdraw USDC modal

raymondjacobson:
### Description

Implement skeleton UI. Does not connect to store state beyond pagination and has some visual/layout issues still (mostly just inputs).
UI all hidden under withdraw in the USDCCard on /dashboard, which is gated by the USDC flag.

### How Has This Been Tested?

```
npm run web:stage
localhost:3001/dashboard
```

### Screenshots

image
image
image
image


dharit-tan (approved):
incomplete but we'll both keep working on it!

audius-infra:
Preview this change https://demo.audius.co/rj-pay-1734

PR #4000, opened 2023-09-01
[C-3017] Reset the new upload flow state when the upload button in the side nav is clicked

Kyle-Shanks:
### Description
Add a new selector for shouldReset and add effect in the new upload flow to listen for it and reset the state

### How Has This Been Tested?
Manually tested

### Screenshots
[reset upload flow image here]


dylanjeffers (approved):
great changes!

PR #4002, opened 2023-09-01
[PAY-1791] Withdraw USDC - transfer to dest account

dharit-tan:
### Description
Final step of withdraw USDC flow - transfer to a destination account.
Works even if destination account has not been funded (but swap for sol flow doesn't work yet - solana wallet needs to be prefunded)

### How Has This Been Tested?
Local web stage
Example txs:
https://solscan.io/tx/3q1SofTyHpD1YzGGy9B2h63jCCpSFRxzcJyTaMGU1E3wvdbKzAkSutqj7v5kTATP8vdjqshZbND6JZg3eBi9V9Bc
https://solscan.io/tx/46uB41vDt3tFWUXnooyeyfsB4FCPHWVcmW3tqpbJyVJCj3qRvR5cPiknpnuKRjNUYEBf19S982pYcSdXzXj67kBv

### Screenshots
Screenshot 2023-09-01 at 2 36 33 PM

### DIFF
--- a/packages/web/src/store/application/ui/buy-audio/sagas.ts
+++ b/packages/web/src/store/application/ui/buy-audio/sagas.ts
@@ -65,7 +65,8 @@ import { audiusBackendInstance } from 'services/audius-backend/audius-backend-in
 import {
   getRootAccountRentExemptionMinimum,
   getRootSolanaAccount,
-  getSolanaConnection
+  getSolanaConnection,
+  getTransferTransactionFee
 } from 'services/solana/solana'
 import { reportToSentry } from 'store/errors/reportToSentry'
 import { waitForWrite } from 'utils/sagaHelpers'
@@ -280,7 +281,13 @@ function* getSwapFees({ route }: { route: RouteInfo }) {
   const feesCache = yield* select(getFeesCache)
   const rootAccount = yield* call(getRootSolanaAccount)
 
-  const rootAccountMinBalance = yield* call(getRootAccountRentExemptionMinimum)
+  const transferFee = yield* call(
+    getTransferTransactionFee,
+    rootAccount.publicKey
+  )
+  // Allows for 3 transaction fees
+  const rootAccountMinBalance =
+    (yield* call(getRootAccountRentExemptionMinimum)) + transferFee * 3
 
   const associatedAccountCreationFees = yield* call(
     getAssociatedAccountCreationFees,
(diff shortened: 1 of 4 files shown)


audius-infra:
Preview this change https://demo.audius.co/rt-transfer

packages/web/src/store/application/ui/buy-audio/sagas.ts
```diff
@@ -280,7 +281,13 @@ function* getSwapFees({ route }: { route: RouteInfo }) {
...
   const rootAccount = yield* call(getRootSolanaAccount)
 
-  const rootAccountMinBalance = yield* call(getRootAccountRentExemptionMinimum)
+  const transferFee = yield* call(
+    getTransferTransactionFee,
+    rootAccount.publicKey
+  )
+  // Allows for 3 transaction fees
+  const rootAccountMinBalance =
+    (yield* call(getRootAccountRentExemptionMinimum)) + transferFee * 3
```
raymondjacobson, line 290: should 3 should be a const at the top of this file?

packages/web/src/store/application/ui/withdraw-usdc/sagas.ts
```diff
@@ -179,7 +184,9 @@ function* doWithdrawUSDC({ payload }: ReturnType<typeof beginWithdrawUSDC>) {
         }
         console.debug(`Withdraw USDC - swap successful: ${swapRes}`)
 
-        // Then create and fund the destination associated token account.
+        // Then create and fund the destination associated token account
+        // using funds from the root solana account that we just swapped for.
+        // TODO: use existing funds if possible
```
raymondjacobson, line 189: let's track this in a ticket somewhere

packages/web/src/store/application/ui/withdraw-usdc/sagas.ts
```diff
@@ -203,8 +210,52 @@ function* doWithdrawUSDC({ payload }: ReturnType<typeof beginWithdrawUSDC>) {
         )
       }
     }
-    // TODO: transfer USDC to destination associated token account
+    // TODO: math.min(amount, balance)
```
raymondjacobson, line 213: do we need to have this check here? ideally UI will handle this

packages/web/src/store/application/ui/withdraw-usdc/sagas.ts
```diff
@@ -203,8 +210,52 @@ function* doWithdrawUSDC({ payload }: ReturnType<typeof beginWithdrawUSDC>) {
...
+    let destinationTokenAccount = destinationAddress
```
raymondjacobson, line 214: code suggestion

```
if (!libs.solanaWeb3Manager) {
  throw new Error('No libs solanaWeb3Manager defined)
}
```

and you can remove the `!` elsewhere

raymondjacobson (approved):
wonderful

packages/web/src/store/application/ui/withdraw-usdc/sagas.ts
dharit-tan: I think we do bc there might be slightly less USDC after swapping for fees. will open ticket tho.

packages/web/src/store/application/ui/withdraw-usdc/sagas.ts
dharit-tan: lovely ty

PR #4007, opened 2023-09-01
Handle local library changes (e.g. favoriting, reposting) PAY-1714

nicoback2:
### Description
- Make sure tracks and collections that are favorited or reposted are added to the library
- Make sure tracks and collections that are un-favorited or un-reposted are removed from the library

For the collections piece of this, I originally tried to use `useUpdateQueryData` from the audius-query library API, but it's not meant/built to be used in a saga. Plus, going that route would get quite tricky and bug prone when/if we add different sorting options to the collections tabs.

So I landed on keeping the locally added/removed collection IDs in the Redux store and using those along with the fetched library collections (from API) to calculate the right collections to show in the library.

### DIFF
--- a/packages/common/src/store/pages/saved-page/reducer.ts
+++ b/packages/common/src/store/pages/saved-page/reducer.ts
@@ -1,28 +1,21 @@
-// @ts-nocheck
-// TODO(nkang) - convert to TS
 import { ID } from 'models/Identifiers'
 import { asLineup } from 'store/lineup/reducer'
 import {
+  ADD_LOCAL_COLLECTION,
+  ADD_LOCAL_TRACK,
+  END_FETCHING,
+  FETCH_MORE_SAVES,
+  FETCH_MORE_SAVES_FAILED,
+  FETCH_MORE_SAVES_SUCCEEDED,
   FETCH_SAVES,
+  FETCH_SAVES_FAILED,
   FETCH_SAVES_REQUESTED,
   FETCH_SAVES_SUCCEEDED,
-  FETCH_SAVES_FAILED,
-  FETCH_MORE_SAVES,
-  FETCH_MORE_SAVES_SUCCEEDED,
-  FETCH_MORE_SAVES_FAILED,
-  ADD_LOCAL_TRACK_REPOST,
-  REMOVE_LOCAL_TRACK_REPOST,
-  ADD_LOCAL_TRACK_PURCHASE,
-  ADD_LOCAL_COLLECTION_FAVORITE,
-  ADD_LOCAL_COLLECTION_REPOST,
-  REMOVE_LOCAL_COLLECTION_FAVORITE,
-  REMOVE_LOCAL_COLLECTION_REPOST,
-  ADD_LOCAL_TRACK_FAVORITE,
-  REMOVE_LOCAL_TRACK_FAVORITE,
-  END_FETCHING,
-  SET_SELECTED_CATEGORY,
+  INIT_COLLECTIONS_CATEGORY_FROM_LOCAL_STORAGE,
   INIT_TRACKS_CATEGORY_FROM_LOCAL_STORAGE,
-  INIT_COLLECTIONS_CATEGORY_FROM_LOCAL_STORAGE
+  REMOVE_LOCAL_COLLECTION,
+  REMOVE_LOCAL_TRACK,
+  SET_SELECTED_CATEGORY
 } from 'store/pages/saved-page/actions'
 import tracksReducer, {
   initialState as initialLineupState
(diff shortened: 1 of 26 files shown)


audius-infra:

packages/common/src/store/pages/saved-page/reducer.ts
```diff
@@ -1,5 +1,6 @@
 // @ts-nocheck
 // TODO(nkang) - convert to TS
```
schottra, line 2: Seeing as this is your todo, figured I should ask if it's actually done already? This file seems to be in TS... :-)

packages/common/src/store/pages/saved-page/reducer.ts
```diff
@@ -21,19 +29,106 @@ import { signOut } from 'store/sign-out/slice'
...
+  localAlbumFavorites: [],
+  localAlbumPurchases: [],
+  localAlbumReposts: [],
+  localRemovedAlbumFavorites: [],
+  localRemovedAlbumReposts: [],
+  localPlaylistFavorites: [],
+  localPlaylistReposts: [],
+  localPlaylistPurchases: [],
+  localRemovedPlaylistFavorites: [],
+  localRemovedPlaylistResposts: [],
```
schottra, line 48: typo :-)

packages/common/src/store/pages/saved-page/reducer.ts
```diff
@@ -21,19 +29,106 @@ import { signOut } from 'store/sign-out/slice'
...
   hasReachedEnd: false,
   fetchingMore: false,
   tracks: initialLineupState,
   selectedCategory: LibraryCategory.Favorite
 } as SavedPageState
 
+/** Utility to get the name of key in which locally added or removed collections are stored in SavedPageState.
+ * For example, playlists that are favorited/unfavorited locally go into `localPlaylistFavorites`/`localPlaylistRemovedFavorites`.
+ */
+const getLocalCollectionStateKeys = ({
```
schottra, line 60: Just a thought: If all of these follow similar patterns, you could just structure this as nested objects based on that pattern, i.e. `state.local[kind][category][mutation]` where
 * `kind` is Track, Playlist, Album and an enum
 * `category` is your existing category
 * `mutation` is 'added' or 'removed'

That would still allow you to strictly define the type in a way that doesn't require building a string for indexing.

This could also let you pare down the actions to a single set where you pass the kind/category/mutation values instead of an action for each permutation of all the values together.

edit: I can see below why having it all flattened makes the merging logic easier (spread the state and modify just the set of add/remove values for the kind/category you are changing)

packages/common/src/store/pages/saved-page/reducer.ts
```diff
@@ -21,19 +29,106 @@ import { signOut } from 'store/sign-out/slice'
...
+}
+
+const removeCollectionLocally = ({
+  state,
+  action,
+  type
+}: {
+  state: SavedPageState
+  action: { collectionId: ID; isAlbum: boolean }
+  type: Omit<LibraryCategoryType, 'All'>
```
schottra, line 95: Love it

packages/common/src/store/pages/saved-page/reducer.ts
```diff
@@ -21,19 +29,106 @@ import { signOut } from 'store/sign-out/slice'
...
+  state: SavedPageState
+  action: { collectionId: ID; isAlbum: boolean }
+  type: 'repost' | 'favorite'
+}) => {
+  const { additionKey, removalKey } = getLocalCollectionStateKeys({
+    isAlbum: action.isAlbum,
+    type
+  })
+  return {
+    ...state,
```
schottra, line 124: eek, really wish this was using `createSlice`. It feels error-prone to make sure we're correctly spreading state in all of these actions.

packages/common/src/store/pages/saved-page/selectors.ts
```diff
@@ -2,20 +2,149 @@ import { CommonState } from 'store/commonStore'
...
+  state.pages.savedPage.localRemovedPlaylistResposts
+
+/** Get the tracks in currently selected category that have been added to the library in current session */
+export const getSelectedCategoryLocalAdds = (state: CommonState) => {
+  const selectedCategory = getSelectedCategory(state)
+  const localFavorites = getLocalTrackFavorites(state)
+  const localPurchases = getLocalTrackPurchases(state)
+  const localReposts = getLocalTrackReposts(state)
+  let localLibraryAdditions
+  if (selectedCategory === LibraryCategory.Favorite) {
```
schottra, line 53: Could be a switch?

packages/common/src/store/pages/saved-page/selectors.ts
```diff
@@ -2,20 +2,149 @@ import { CommonState } from 'store/commonStore'
...
+    return localFavorites
+  } else if (selectedCategory === LibraryCategory.Purchase) {
+    return localPurchases
+  } else if (selectedCategory === LibraryCategory.Repost) {
+    return localReposts
+  } else {
+    // Category = ALL
+    return Array.from(
+      new Set([...localReposts, ...localFavorites, ...localPurchases])
```
schottra, line 110: Love the usage of Set here. Just confirming I understand correctly based on the docs for `Set`. It's going to use strict equal checking to compare items. So this would only filter out items that are references to the exact same object in memory. Is that the behavior we're going for?

I _think_ it works since collection ids are actually just `numbers` and the values we are putting in these arrays are just the IDs.

packages/common/src/store/pages/saved-page/types.ts
```diff
@@ -21,9 +21,24 @@ export function isLibraryCategory(value: string): value is LibraryCategoryType {
   return Object.values(LibraryCategory).includes(value as LibraryCategoryType)
 }
 export interface SavedPageState {
-  localSaves: { [id: number]: UID }
+  localTrackFavorites: { [id: number]: UID }
+  localTrackReposts: { [id: number]: UID }
+  localTrackPurchases: { [id: number]: UID }
```
schottra, line 26: It's a tad confusing that tracks use an object mapping from IDs-> UIDs but I suppose this pattern was already here and you're getting it functional again?

schottra (approved):
That's a lot of redux state! But I don't know that trying to pare it down makes it easier to follow. The explicit keys might be simpler.

Looks good!

audius-infra:

audius-infra:
Preview this change https://demo.audius.co/nkang--local-state-library

audius-infra:

audius-infra:
Preview this change https://demo.audius.co/nkang--local-state-library

audius-infra:

audius-infra:
Preview this change https://demo.audius.co/nkang--local-state-library

packages/common/src/store/pages/saved-page/reducer.ts
nicoback2: ah i like this idea! will ponder

packages/common/src/store/pages/saved-page/reducer.ts
nicoback2: 100%

audius-infra:
Preview this change https://demo.audius.co/nkang--local-state-library

packages/common/src/store/pages/saved-page/types.ts
nicoback2: Yeah exactly, prob worth another revisit in the future

nicoback2: I'm not entirely sure that the UID thing is necessary (prob is but wasn't obvious)

packages/common/src/store/pages/saved-page/selectors.ts
schottra: Oh no, set is totally fine. I'm all about not adding lodash methods for things that are easy to do with built-ins :-)

audius-infra:

audius-infra:
Preview this change https://demo.audius.co/nkang--local-state-library

PR #4010, opened 2023-09-02
[C-3035] Stale stats for offline tracks

dylanjeffers:
### Description

Fixes issue where downloaded tracks have stale stats when user is online. This is because the downloaded track is populated in the cache, and then doesn't update until cache TTL. Added update where any track, collection, or user that enters the cache from offline data has a `local` flag on it so the cache knows to flush it next time it's requested. We do the same thing for local account user.


amendelsohn (approved):
Very cool!

audius-infra:
Preview this change https://demo.audius.co/dj-c-3035-stale-stats

PR #4014, opened 2023-09-05
Offline mode drawer PAY-1713

nicoback2:
### Description
Designs: https://www.figma.com/file/ClKssErI6tIMmZFwQifp9f/%F0%9F%92%B0-USDC-%2B-Payments%3A-Feature?type=design&node-id=1078-48999&mode=design&t=azQUY0z4AurR5HWo-0

offline

removeoffline


nicoback2:
Note - re-used the drawer header from Share to Story drawer instead of following the Figma for now. Going to ask design if we're going to standardize the header soon.

nicoback2:
Note - re-used the drawer header from Share to Story drawer instead of following the Figma for now. Going to ask design if we're going to standardize the header soon.

packages/mobile/src/components/drawers/OfflineListeningDrawer.tsx
```diff
@@ -0,0 +1,217 @@
...
+  const dispatch = useDispatch()
+  const { data, onClose } = useDrawer('OfflineListening')
+  const { isFavoritesMarkedForDownload, onSaveChanges } = data
+
+  const [isFavoritesOn, setIsFavoritesOn] = useState(
+    isFavoritesMarkedForDownload
+  )
+
+  const handleSaveChanges = useCallback(() => {
+    if (isFavoritesMarkedForDownload) {
```
dylanjeffers, line 129: could maybe collapse the if else's into something a bit more explicit:
`if (isFavoritesMarkedForDownload && !isFavoritesOn) {...} else if (!isFavoritesMarkedForDownload && isFavoritesOn) {...} ? main idea being to make clear we do stuff when the value has changed

packages/mobile/src/components/drawers/OfflineListeningDrawer.tsx
```diff
@@ -0,0 +1,217 @@
...
+            style={styles.titleIcon}
+            fill={neutralLight2}
+            height={20}
+            width={24}
+          />
+          <View>
+            <Text
+              weight='heavy'
+              color='neutralLight2'
+              fontSize={'xl'}
```
dylanjeffers, line 173: as much as we can, prefer text string so `fontSize='xl'`

packages/mobile/src/components/drawers/OfflineListeningDrawer.tsx
```diff
@@ -0,0 +1,217 @@
...
+    <NativeDrawer drawerName='OfflineListening'>
+      <View style={styles.container}>
+        <View style={styles.title}>
+          <IconDownload
+            style={styles.titleIcon}
+            fill={neutralLight2}
+            height={20}
+            width={24}
+          />
+          <View>
```
dylanjeffers, line 169: do we need this extra view?

packages/mobile/src/components/drawers/OfflineListeningDrawer.tsx
```diff
@@ -0,0 +1,217 @@
...
+  },
+  descriptionText: {
+    textAlign: 'center',
+    lineHeight: typography.fontSize.large * 1.3
+  },
+  descriptionContainer: {
+    width: '100%'
+  },
+  titleIcon: {
+    position: 'relative',
```
dylanjeffers, line 49: not sure of exact styling, but seems a bit weird that we are positioning the icon like this instead of the icon and title just being in a flex box?

packages/mobile/src/components/drawers/OfflineListeningDrawer.tsx
```diff
@@ -0,0 +1,217 @@
...
+  },
+  titleText: {
+    textTransform: 'uppercase',
+    marginTop: spacing(4)
+  },
+  descriptionText: {
+    textAlign: 'center',
+    lineHeight: typography.fontSize.large * 1.3
+  },
+  descriptionContainer: {
```
dylanjeffers, line 45: do we need this view? my guess is that text align: center with just the text would go the full width of its container?

packages/mobile/src/screens/favorites-screen/FavoritesDownloadSection.tsx
```diff
@@ -1,27 +1,55 @@
...
+        data: {
+          isFavoritesMarkedForDownload: switchValue,
+          onSaveChanges: setSwitchValue
+        }
+      })
+    )
+  }, [dispatch, switchValue])
+
+  return (
+    <TouchableOpacity onPress={handlePressStatusIndicator}>
```
dylanjeffers, line 40: it might not work, but i wonder if we could use IconButton for this? the nice thing with that is it comes in with some built-in hitSlop to make it easier to press.

dylanjeffers (approved):
excellent job! it's looking really clean

packages/mobile/src/screens/favorites-screen/FavoritesDownloadSection.tsx
nicoback2: sadly we cannot because the "loading" indicator state is a rive component not an icon component. can add in a hitslop here if we run into issues pressing the button in qa!

packages/mobile/src/components/drawers/OfflineListeningDrawer.tsx
nicoback2: removed!

packages/mobile/src/screens/favorites-screen/FavoritesDownloadSection.tsx
dylanjeffers: nice yeah makes sense, and sounds like a plan :)

packages/mobile/src/components/drawers/OfflineListeningDrawer.tsx
nicoback2: removed ty

audius-infra:

audius-infra:
Preview this change https://demo.audius.co/nkang--offline-mode-drawer

PR #4019, opened 2023-09-05
[C-2536] Fix track name cutoff

dylanjeffers:
### Description

Fixes issue where track names were cutoff weirdly in mobile. This actually has to do with `\n` in the track name. Regular html converts that into a space, whereas react-native treats it as a line break. The fix for us was to update the Text component to convert `\n` into spaces

### How Has This Been Tested?

All areas where the track name could appear were tested: search results, track page, track tile

### DIFF
--- a/packages/mobile/src/components/core/Text.tsx
+++ b/packages/mobile/src/components/core/Text.tsx
@@ -41,6 +41,7 @@ export const Text = (props: TextProps) => {
     weight,
     fontSize: fontSizeProp,
     textTransform,
+    children: childrenProp,
     ...other
   } = props
   const variant = variantProp ?? 'body'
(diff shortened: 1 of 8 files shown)


packages/mobile/src/components/core/Text.tsx
```diff
@@ -85,5 +86,14 @@ export const Text = (props: TextProps) => {
     ]
   )
 
-  return <RNText style={[styles.root, customStyles, style]} {...other} />
+  const children =
+    typeof childrenProp === 'string'
+      ? childrenProp.replace('\n', ' ')
```
amendelsohn, line 91: Neat! How should we be doing line breaks in our own messages? Do you always do two separate messages spaced using flexbox or is there another way?

amendelsohn (approved):
Nice! Interesting bug

packages/mobile/src/components/core/Text.tsx
dylanjeffers: its a good question, actually we should probably have a prop here to disable this, since i see some areas of the mobile code where we do explicit line breaks. Ill fix those entries. Do you have an idea on if it should be default on vs off?

amendelsohn: I think default on is good to prevent the bug reoccurring. As long as it's not too much work to find the existing newline usages, I would go with that.

amendelsohn: Better to require desired newline cases to be explicit about it so we never do it by accident.

PR #4020, opened 2023-09-05
Install pip for android release

sliptype:
### Description

Fix android release CI. This broke bc we are using a new circle ci image


dylanjeffers (approved):
youre the best

PR #4028, opened 2023-09-06
Empty states for different library category filters PAY-1797

nicoback2:
### Description

Screenshot 2023-09-05 at 4 13 16 PM
Screenshot 2023-09-05 at 4 58 38 PM

### DIFF
--- a/packages/web/src/pages/saved-page/components/desktop/AlbumsTabPage.tsx
+++ b/packages/web/src/pages/saved-page/components/desktop/AlbumsTabPage.tsx
@@ -3,7 +3,8 @@ import { useMemo } from 'react'
 import {
   LibraryCategory,
   statusIsNotFinalized,
-  savedPageSelectors
+  savedPageSelectors,
+  CommonState
 } from '@audius/common'
 import { useSelector } from 'react-redux'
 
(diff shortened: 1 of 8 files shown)


packages/web/src/pages/saved-page/components/desktop/AlbumsTabPage.tsx
```diff
@@ -25,6 +33,16 @@ export const AlbumsTabPage = () => {
     fetchMore,
     collections: albums
   } = useCollectionsData('album')
+  const selectedCategory = useSelector(getSelectedCategory)
```
dylanjeffers, line 36: If selected category is not used elsewhere, you could maybe just do it all in the selector

dylanjeffers (approved):
Wow a lot of states to handle, nice

packages/web/src/pages/saved-page/components/emptyStateMessages.ts
```diff
@@ -0,0 +1,18 @@
+export const emptyStateMessages = {
```
sliptype, line 1: Is there a reason emptyStateMessages is separate for desktop but defined in the messages object for mobile? Not a big deal but we could potentially be more consistent here

sliptype (approved):

packages/web/src/pages/saved-page/components/emptyStateMessages.ts
nicoback2: yeah it's because in web it's shared by mobile web and desktop components so it was getting super duplicative! can move to common if you think that's better, but i thought that might be overkill

audius-infra:

audius-infra:
Preview this change https://demo.audius.co/nkang--empty-states-library

PR #4030, opened 2023-09-06
Add guard to prevent attempted read of undefined queue items

schottra:
### Description
`getToQueue` has code paths that can return `undefined`, so we need to guard the check against `uid` for the new preview functionality

### How Has This Been Tested?
I am unable to reproduce the original crash, but will do coordinated testing with somebody who can.

### DIFF
--- a/packages/web/src/common/store/lineup/sagas.js
+++ b/packages/web/src/common/store/lineup/sagas.js
@@ -373,7 +373,7 @@ function* play(lineupActions, lineupSelector, prefix, action) {
           const queueable = yield call(getToQueue, lineup.prefix, e)
           // If the entry is the one we're playing, set isPreview to incoming
           // value
-          if (queueable.uid === action.uid) {
+          if (queueable?.uid === action.uid) {
             queueable.isPreview = isPreview
           }
           return queueable


audius-infra:
Preview this change https://demo.audius.co/fix-queueable-crash

sliptype (approved):
Confirmed working!

dylanjeffers (approved):
Works but did we figure out which element is failing us?

schottra:
> Nice! So which queue items fail this?

@dylanjeffers My testing accounts don't reproduce this, so I'm not actually sure. But it would be any items that cause the `getToQueue` function to return undefined. That would be anything without `Kind.TRACKS` as entity type, or if for some reason a collection was being returned in the favorites list for tracks?
Otherwise, I'm not entirely sure how the other entity types get set. But the next few lines of code read to me like a way to filter out the bad entries when attempting to queue a lineup.

dylanjeffers:
> > Nice! So which queue items fail this?
> 
> 
> 
> @dylanjeffers My testing accounts don't reproduce this, so I'm not actually sure. But it would be any items that cause the `getToQueue` function to return undefined. That would be anything without `Kind.TRACKS` as entity type, or if for some reason a collection was being returned in the favorites list for tracks?
> 
> Otherwise, I'm not entirely sure how the other entity types get set. But the next few lines of code read to me like a way to filter out the bad entries when attempting to queue a lineup.

Gotcha! Let me try this out real quick and see what item it is

PR #4034, opened 2023-09-06
[C-2548] Fix favorites table saved date filter

dylanjeffers:
### Description

Fixes silly bug where we were not setting the correct query-param when filtering a track table based on "dateAdded" and "dateSaved"


audius-infra:
Preview this change https://demo.audius.co/dj-c-2548-fix-favorites-saved-date

dylanjeffers:
> oooo subtle. I think this is something we could catch in the future by using the exported types from SDK as the allowable values.
> 
> [We did that for the purchases page](https://github.com/AudiusProject/audius-client/blob/main/packages/web/src/pages/purchases-and-sales/PurchasesPage.tsx#L50)

wait that is so sick... let me look into doing this rn

PR #4036, opened 2023-09-07
Fix Jupiter caching bad route info

dharit-tan:
### Description
The two calls to `computeRoutes` were happening close enough together that Jupiter was caching the response, but for some reason the routes were empty in the cached response. Force-fetching fixes the bug.

Also includes some small cleanup.

### How Has This Been Tested?
Local web stage


rickyrombo (approved):
Interesting... caching would be great if it worked since we're actively trying to use the same route we're quoted for initially

audius-infra:
Preview this change https://demo.audius.co/rt-fixes

PR #4039, opened 2023-09-07
[PAY-1725] Display proper amount in track sold notification

sddioulde:
### Description

Needed to use diff formatting function since amount comes back as usdc wei.

### How Has This Been Tested?

Web and mobile on stage.

### Screenshots

before:
image

after:
Screenshot 2023-09-07 at 12 41 18 PM


audius-infra:
Preview this change https://demo.audius.co/sd-fix-usdc-notif

PR #4040, opened 2023-09-07
[PAY-1811] Fix default/empty profile picture in now playing bar/dashboard

raymondjacobson:
### Description

* Dashboard didn't have background color set
* Now playing bar setting was removing the background on `moveBehind` in dynamic image. When we switch images in DynamicImage, remove the background color prop first before setting it to what we want.

Probably should refactor this whole beast a bit at some point, but these changes at least are consistent w/ current patterns.

### How Has This Been Tested?

Locally vs. staging -- played a few diff tracks on trending & tested /dashboard

### Screenshots

image

image


audius-infra:
Preview this change https://demo.audius.co/rj-pay-1811

schottra (approved):

PR #4042, opened 2023-09-07
[PAY-1763][PAY-1810] Wire up links to Purchases/Sales pages

schottra:
### Description
This implements the links from the Artist Dashboard and Settings pages to the purchases/sales/withdrawals page.
There will be a follow up PR to implement the withdrawals page after this one is merged.

Also a quick detour to fix a crash when navigating away from the artist dashboard. A bad type causes us to attempt to yield on a channel and then call close on the result, instead of just closing the channel.

### How Has This Been Tested?
Tested locally against staging

### Screenshots
Screenshot 2023-09-07 at 4 05 23 PM

### DIFF
--- a/packages/web/src/pages/artist-dashboard-page/components/USDCCard.tsx
+++ b/packages/web/src/pages/artist-dashboard-page/components/USDCCard.tsx
@@ -35,8 +35,6 @@ const messages = {
   withdrawalHistory: 'Withdrawal History'
 }
 
-const MediumWithdrawIcon = () => <Icon icon={IconWithdraw} size='medium' />
-
 export const USDCCard = ({ balance }: { balance: BNUSDC }) => {
   const goToRoute = useGoToRoute()
   const { onOpen: openWithdrawUSDCModal } = useWithdrawUSDCModal()
@@ -100,7 +98,7 @@ export const USDCCard = ({ balance }: { balance: BNUSDC }) => {
             variant={HarmonyButtonType.SECONDARY}
             text={messages.withdraw}
             fullWidth
-            iconLeft={MediumWithdrawIcon}
+            iconLeft={IconWithdraw}
             onClick={() =>
               openWithdrawUSDCModal({
                 page: WithdrawUSDCModalPages.ENTER_TRANSFER_DETAILS
--- a/packages/web/src/pages/artist-dashboard-page/components/USDCCard.tsx
+++ b/packages/web/src/pages/artist-dashboard-page/components/USDCCard.tsx
@@ -21,6 +21,8 @@ import BN from 'bn.js'
 
 import { Icon } from 'components/Icon'
 import { Text } from 'components/typography'
+import { useGoToRoute } from 'hooks/useGoToRoute'
+import { SALES_PAGE } from 'utils/route'
 
 import styles from './USDCCard.module.css'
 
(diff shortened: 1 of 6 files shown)


packages/web/src/pages/artist-dashboard-page/store/sagas.ts
```diff
@@ -167,11 +168,11 @@ function* pollForBalance() {
...
     IntKeys.DASHBOARD_WALLET_BALANCE_POLLING_FREQ_MS
   )
-  const chan = yield* call(doEvery, pollingFreq || 1000, function* () {
+  const chan = (yield* call(doEvery, pollingFreq || 1000, function* () {
     yield* put(getBalance())
-  })
+  })) as unknown as EventChannel<any>
   yield* take(dashboardActions.reset.type)
-  ;(yield* chan).close()
+  chan.close()
```
schottra, line 175: cc @raymondjacobson

packages/web/src/pages/artist-dashboard-page/components/USDCCard.tsx
```diff
@@ -33,7 +35,10 @@ const messages = {
   withdrawalHistory: 'Withdrawal History'
 }
 
+const MediumWithdrawIcon = () => <Icon icon={IconWithdraw} size='medium' />
```
rickyrombo, line 38: nit: I personally would prefer this inlined

packages/web/src/pages/artist-dashboard-page/store/sagas.ts
raymondjacobson: switching things to ts is scary! glad it's fixed now though. ty so much

raymondjacobson (approved):
nice!

rickyrombo (approved):

packages/web/src/pages/artist-dashboard-page/store/sagas.ts
schottra: I debated whether we should even be doing this in a saga. I think it would make sense to use a hook that mounts/unmounts with the components and polls while the component is active. But I was going for minimal change to fix the issue since it's unrelated to this PR

packages/web/src/pages/artist-dashboard-page/components/USDCCard.tsx
schottra: Worth discussion. I generally avoid redefining function props on every render if I can help it, unless I'm really sure it won't result in anything below here re-rendering.

I assume that if you change the value of this prop each time, the button under the hood is going to be mounting a new instance of it every time as well and that might result in React doing updates we don't really need to do.

But I could probably test this first and make sure it's not smart enough to figure out that the resulting tree is the same and not do the extra work.

rickyrombo: Oh does it have to be a function prop? I didn't realize that - thought we could pass in the Icon component directly. Nevermind then, carry on!

schottra: You are correct, normally we _should_ just pass in the component directly and let the button render an instance of it with the appropriate styling.
And as it turns out, the button _does_ do the right thing in this case. I had thought this was originally an override to fix the icon being the wrong size due to the source icon file not being the right scale.
But it looks correct to me if I just pass the icon in directly. So I will remove this intermediate component.

Thanks for asking questions!

audius-infra:
Preview this change https://demo.audius.co/pay-1763-links-to-transactions

PR #4043, opened 2023-09-07
[PAY-1753] Implements Withdrawals Page

schottra:
### Description
This change adds the Withdrawals page, api, and table. It follows the implementations of sales/purchases fairly closely, but uses a different API and column definitions.

There is some repeated code, but it's likely not enough to be worth trying to make the pages/tables generic across all three cases.

This is behind the USDC feature flag.
The metadata for usdc transactions currently returns as empty. So we don't show the wallet address yet. But that should be resolved automatically once the API starts returning values.

### How Has This Been Tested?
Tested locally against staging


packages/common/src/audius-query/createApi.ts
```diff
@@ -272,7 +272,7 @@ const fetchData = async <Args, Data>(
     endpoint.onQueryStarted?.(fetchArgs, { dispatch })
 
     const apiData = await endpoint.fetch(fetchArgs, context)
-    if (!apiData) {
+    if (apiData == null) {
```
schottra, line 275: This line was breaking fetch endpoints which return falsey values, most notably `count` endpoints when they return 0.

packages/common/src/models/USDCTransactions.ts
```diff
@@ -38,46 +27,12 @@ export type USDCPurchaseDetails = {
...
+export type USDCTransactionDetails = {
+  signature: string
+  transactionType: USDCTransactionType
+  method: USDCTransactionMethod
+  transactionDate: string
+  change: StringUSDC
+  balance: StringUSDC
+  metadata?: Nullable<object> | string
 }
-
```
schottra, line 55: These types were originally added when the tech spec document was done, but they weren't in use yet. The actual types coming back from the API are a little different. And we don't really need the discriminated union.

audius-infra:
Preview this change https://demo.audius.co/pay-1753-withdrawals-page

audius-infra:
Preview this change https://demo.audius.co/pay-1753-withdrawals-page

packages/common/src/models/USDCTransactions.ts
```diff
@@ -38,46 +27,12 @@ export type USDCPurchaseDetails = {
...
-  amount: StringUSDC
-  destination: string
+export type USDCTransactionDetails = {
+  signature: string
+  transactionType: USDCTransactionType
+  method: USDCTransactionMethod
+  transactionDate: string
+  change: StringUSDC
+  balance: StringUSDC
+  metadata?: Nullable<object> | string
```
rickyrombo, line 37: fwiw this is probably always going to be a string

packages/common/src/utils/wallet.ts
```diff
@@ -151,11 +151,13 @@ export const ceilingBNUSDCToNearestCent = (value: BNUSDC): BNUSDC => {
 /** Formats a USDC wei string (full precision) to a fixed string suitable for
 display as a dollar amount. Note: will lose precision by rounding _up_ to nearest cent */
 export const formatUSDCWeiToUSDString = (amount: StringUSDC, precision = 2) => {
+  // remove negative sign if present.
+  const amountPos = amount.replace('-', '')
```
rickyrombo, line 155: This was a nice little hack for the AUDIO tx history, but it feels ripe for bugs in other potential (future?) usages... we should probably revisit both of these at some point.

packages/web/src/pages/purchases-and-sales/components/NoTransactionsContent.tsx
```diff
@@ -9,21 +9,21 @@ import { Icon } from 'components/Icon'
...
-export type NoPurchasesContentProps = {
+export type NoTransactionsContentProps = {
   headerText: string
   bodyText: string
   ctaText: string
   onCTAClicked: () => void
 }
 
-export const NoPurchasesContent = ({
+export const NoTransactionsContent = ({
```
rickyrombo, line 21: All that's missing is an `icon` prop and I think this could be a super general `NoRows` or `EmptyTable` component :eyes:

rickyrombo (approved):
Awesome!

PR #4044, opened 2023-09-07
[C-3024] Fix solana dapp store ci

dylanjeffers:
### Description

Now that we are node 18 everywhere, we can drop the special nvmrc and nvm install for dapp-store, which should simplify and fix the process.


sliptype (approved):
Fire

audius-infra:

PR #4052, opened 2023-09-08
[PAY-1814] Store USDC balances in wallet slice

raymondjacobson:
### Description

Store USDC balances in common slice so that it stays fresh when updated from misc. usage of `useUSDCBalance`

Issue:
1. User with balance 100 withdraws 5
2. Withdrawal modal UI shows 95, but dashboard page shows 100 still

I think I like this pattern more than polling for balance refresh every 1s.

### How Has This Been Tested?

Locally vs. staging. Tested via withdrawal flow.

### DIFF
--- a/packages/common/src/hooks/useUSDCBalance.ts
+++ b/packages/common/src/hooks/useUSDCBalance.ts
@@ -1,19 +1,35 @@
-import { useEffect, useState } from 'react'
+import { useCallback, useEffect, useState } from 'react'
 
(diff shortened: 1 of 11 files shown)


rickyrombo (approved):
Good call, super easily to do with the hook!

audius-infra:
Preview this change https://demo.audius.co/rj-pay-1814

dharit-tan (approved):
it works!

PR #4054, opened 2023-09-08
[PAY-1794] Math.min(withdrawalAmount, latestBalance) after using USDC to pay tx fee

raymondjacobson:


audius-infra:
Preview this change https://demo.audius.co/rj-pay-1794

dharit-tan (approved):
sick ty

PR #4055, opened 2023-09-08
[C-3060] Fix mobile outbound link experience

dylanjeffers:
### Description

- Fixes issue where the touch targets for confirming/cancelling an outbound link in the LeavingAudiusDrawer are way too small
- Fixes issue where allowListed external links still open the LeavingAudiusDrawer
- Fixes issues with drawer header styles.


packages/common/src/utils/linking.ts
```diff
@@ -0,0 +1,18 @@
+export const externalLinkAllowList = new Set([
+  'facebook.com',
+  'instagram.com',
+  'tiktok.com',
+  'twitter.com',
+  'x.com',
+  'blog.audius.co'
+])
+
+export const isAllowedExternalLink = (link: string) => {
```
dylanjeffers, line 10: lmk if folks can think of better name. "known" possibly?

audius-infra:
Preview this change https://demo.audius.co/dj-c-3060-fix-outbound-link-experienc

packages/mobile/src/components/action-drawer/ActionDrawer.tsx
```diff
@@ -4,11 +4,13 @@ import { useCallback } from 'react'
 import type { Modals } from '@audius/common'
 import type { TextStyle, ViewStyle } from 'react-native'
 import { TouchableHighlight, View } from 'react-native'
+import type { SetOptional } from 'type-fest'
```
amendelsohn, line 7: ooh cool

packages/mobile/src/components/core/Text.tsx
```diff
@@ -68,7 +68,13 @@ export const Text = (props: TextProps) => {
...
         marginTop:
           weight === 'demiBold' && Platform.OS === 'ios'
-            ? spacing(fontSize === 'large' ? 1 : fontSize === 'small' ? 0.5 : 0)
+            ? spacing(
+                fontSize && ['xl', 'large'].includes(fontSize)
+                  ? 1
+                  : fontSize === 'small'
+                  ? 0.5
+                  : 0
+              )
```
amendelsohn, line 77: size -> spacing map maybe better

packages/mobile/src/components/feed-filter-drawer/FeedFilterDrawer.tsx
```diff
@@ -74,7 +61,11 @@ export const FeedFilterDrawer = () => {
   return (
     <ActionDrawer
       modalName={MODAL_NAME}
-      renderTitle={() => <Text style={styles.title}>{messages.title}</Text>}
+      title={
+        <Text color='neutral' textTransform='none'>
```
amendelsohn, line 65: this is about to get wrapped in another text inside DrawerHeader, right?

amendelsohn: should we take a different prop when you are passing a component vs plain text?

amendelsohn (approved):

packages/common/src/utils/linking.ts
```diff
@@ -0,0 +1,18 @@
...
+  try {
+    let hostname = new URL(link).hostname
```
rickyrombo, line 12: Does react-native support `URL.hostname` now? Nice!

packages/mobile/src/screens/chat-screen/LinkPreview.tsx
```diff
@@ -101,7 +102,7 @@ export const LinkPreview = ({
   const { onOpen: openLeavingAudiusModal } = useLeavingAudiusModal()
 
   const handlePress = useCallback(() => {
-    if (isAudiusUrl(href)) {
+    if (isAudiusUrl(href) || isAllowedExternalLink(href)) {
```
rickyrombo, line 105: thanks!

packages/mobile/src/components/core/Hyperlink.tsx
```diff
@@ -98,10 +102,10 @@ export const Hyperlink = (props: HyperlinkProps) => {
 
   const handlePress = useCallback(
     (url) => {
-      if (!isAudiusUrl(url)) {
-        openLeavingAudiusModal({ link: url })
-      } else {
+      if (isAudiusUrl(url) || isAllowedExternalLink(url)) {
```
rickyrombo, line 105: Thanks!

packages/mobile/src/components/drawer/Drawer.tsx
```diff
@@ -111,7 +111,7 @@ export type DrawerProps = {
   /**
    * Header title if this drawer has a header
    */
-  title?: string
+  title?: ReactNode
```
rickyrombo, line 114: <3

rickyrombo (approved):

packages/mobile/src/components/core/Text.tsx
dylanjeffers: completely agree

packages/mobile/src/components/feed-filter-drawer/FeedFilterDrawer.tsx
dylanjeffers: i could def see that, i was thinking if you want to update the title text, you can just pass a text component ready to go, and it's okay that it's wrapped. really only applies to one drawer and ideally we should all have the same title. can see the idea of having a "renderTitle" thing that goes all the way down though.

packages/common/src/utils/linking.ts
dylanjeffers: yeah it does thank heavens

audius-infra:
Preview this change https://demo.audius.co/dj-c-3060-fix-outbound-link-experienc

packages/mobile/src/components/feed-filter-drawer/FeedFilterDrawer.tsx
amendelsohn: yeah either way is okay I think. As long as double wrapping the text uses the inner one I guess

PR #4060, opened 2023-09-08
Fixes solana-dapp-store ci with new node

dylanjeffers:
### Description

Fixes a few more issues with solana-dapp-store ci
- Removes redundant checkout
- Fixes corepack enable, it needs to be installed to local directory otherwise complains about permissions
- Also need to symlink react-native like other android builds.
- Removes nvm references since we are all on same version of node 18


audius-infra:
Preview this change https://demo.audius.co/dj-fix-dapp-store-ci

sliptype (approved):

PR #4066, opened 2023-09-09
[PAY-1833] Increase trending playlists fetch limit

sddioulde:
### Description

For some reason, limit is 3 and we are not getting enough playlists back, maybe due to some bug in the lineup.
Setting the limit to 10 so we at least get enough playlists back from first fetch for now.

### How Has This Been Tested?

local web dapp vs prod


raymondjacobson (approved):

dylanjeffers (approved):
nice!

sddioulde:
> nice! will this make it take longer to fetch cause playlist data is a lot?

won't make a noticeable diff

PR #4067, opened 2023-09-09
[PAY-1822][PAY-1824][PAY-1825] Address withdrawal QA

raymondjacobson:
### Description

* Match designs
* Add error state to confirmation and error page of modal
* Add min balance

### How Has This Been Tested?

npm run web:stage

### Screenshots

image

image

image

image

### DIFF
--- a/packages/web/src/components/withdraw-usdc-modal/WithdrawUSDCModal.module.css
+++ b/packages/web/src/components/withdraw-usdc-modal/WithdrawUSDCModal.module.css
@@ -1,5 +1,5 @@
 .modal {
-  width: 672px;
+  width: 720px;
 }
 
 .title {
--- a/packages/web/src/components/withdraw-usdc-modal/WithdrawUSDCModal.module.css
+++ b/packages/web/src/components/withdraw-usdc-modal/WithdrawUSDCModal.module.css
@@ -1,7 +1,3 @@
-.modal {
-  width: 720px;
-}
-
 .title {
   align-items: center;
   display: flex;
--- a/packages/web/src/components/withdraw-usdc-modal/WithdrawUSDCModal.tsx
+++ b/packages/web/src/components/withdraw-usdc-modal/WithdrawUSDCModal.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useState } from 'react'
+import { useCallback, useEffect, useState } from 'react'
 
 import {
   SolanaWalletAddress,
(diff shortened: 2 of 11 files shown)


audius-infra:
Preview this change https://demo.audius.co/rj-withdrawal-qa

packages/web/src/components/withdraw-usdc-modal/WithdrawUSDCModal.module.css
```diff
@@ -1,5 +1,5 @@
 .modal {
-  width: 672px;
+  width: 720px;
```
rickyrombo, line 2: `size='medium'` prop should do this!

packages/web/src/components/withdraw-usdc-modal/WithdrawUSDCModal.tsx
```diff
@@ -65,6 +76,16 @@ export const WithdrawUSDCModal = () => {
   const balanceNumberCents = formatUSDCWeiToFloorCentsNumber(
     (balance ?? new BN(0)) as BNUSDC
   )
+  const withdrawalStatus = useSelector(getWithdrawStatus)
+
+  const [priorBalanceCents, setPriorBalanceCents] =
+    useState<Nullable<number>>(null)
+
+  useEffect(() => {
```
rickyrombo, line 84: do we need an effect here? can we just have the default for `priorBalanceCents` be `balanceNumberCents` instead of null? (And maybe even use a ref?)

packages/web/src/components/withdraw-usdc-modal/components/Error.tsx
```diff
@@ -0,0 +1,81 @@
...
+        <Text size='xSmall' strength='default' color='accentRed'>
+          {messages.error}
+        </Text>
+        <Text
+          className={styles.tryAgain}
+          size='xSmall'
+          strength='default'
+          color='secondary'
+          onClick={handleTryAgain}
+        >
```
rickyrombo, line 75: should this be an anchor or a button or something more accessible?

rickyrombo (approved):
Nice!

packages/web/src/components/withdraw-usdc-modal/WithdrawUSDCModal.module.css
raymondjacobson: ty! back ONGRID

packages/web/src/components/withdraw-usdc-modal/WithdrawUSDCModal.tsx
raymondjacobson: balanceNumberCents won't necessarily be defined correctly initially (or it isn't a requirement to), so there does need to be some kinda setter thing for it.

